PR #87, opened 2014-01-08
postgresql support

srgpqt:
A few notes:
1. The current schema cannot be loaded by postgres because the story_cache column of the stories has a 2GB limit, but postgresql has a 1GB -1 limit for text columns.
2. I renovated Keystore a little bit. Removed the redundant methods, and added a generic way to put/increment/update values in it, and modified the traffic counting code to take advantage of it (now makes 3 to 4 queries instead of 5). I left the specialized Mysql and Sqlite put/increment code intact, but perhaps we could remove it.
3. Sphinx indexing works with the last commit in this set, but I'm not sure this is the correct approach. I saw one of your earlier commits about "cast(upvotes as signed)" but this doesn't work with postgres -- I've put "integer" instead of "signed", hoping (but unsure) that this would be sufficient, as integer type is supposed to be signed... Feedback needed.

### DIFF
--- a/app/indices/comment_index.rb
+++ b/app/indices/comment_index.rb
@@ -3,7 +3,7 @@ ThinkingSphinx::Index.define :comment, :with => :active_record do
   indexes comment
   indexes user.username, :as => :author
 
-  has "(cast(upvotes as signed) - cast(downvotes as signed))",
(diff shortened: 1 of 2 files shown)


ckib16:
@srgpqt and @jcs - Thanks so much for doing this Postgresql work!

PR #169, opened 2014-08-30
Add confirm box when extraneous components are detected in title.

walle:
I've gotten a couple of submissions moderated because of this, so I thought I would try and fix it.
Not sure if this is up to code standards or so. And it only detects a couple of formats, but might be something to build on anyway.

Tell me what you think.


jcs:
8771afc5f53ac is helping out a lot, so this probably won't be needed.  thanks though.

walle:
Looks like a better solution.

PR #184, opened 2014-12-14
Update to Rails v4.1.8

srgpqt:
Rails 4.1.8 is quite mature now. With 4.2.0 coming out soon, it's a good time to update.
* Updated the thinking-sphinx gem also, for compatibility with Rails 4.1.x (tested and seems to work fine).
* No other gems were updated, aside from rails, thinking-sphinx and their dependencies.
* I've updated the config files to closely match the new Rails defaults. Should behave the same as before.
* The new Rails-recommended way to set secret_key_base is via the SECRET_KEY_BASE environment variable read by secrets.yml, but setting it in config/initializers/secret_token.rb as per the Lobsters readme still works fine, as a fallback.
* Slightly simplified readme instructions for creating initial user and tag.


jcs:
Thanks.

The upgrade went mostly okay except that all existing requests from logged-in users triggered exceptions:

```
A JSON::ParserError occurred in stories#show:

  795: unexpected token at '^D^HI"^Osession_id^F:^FEFI"%845c299312ddf4e72...^F;'
  app/controllers/application_controller.rb:11:in `authenticate_user'
```

I had to generate a new session key to invalidate all existing sessions and things seem okay now.

PR #186, opened 2015-01-12
Give textareas a default background colour.

jordigh:
My default background colour for textareas is inherited from my OS
colours, which is a dark grey. Thus setting a foreground colour
without setting a background colour makes it impossible for me to type
in textareas. This patch should fix that.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -118,6 +118,7 @@ button,
 select,
 textarea {
 	color: #555;
+	background-color: white;
 	line-height: 1.2em;
 	padding: 3px 5px;
 }
@@ -142,6 +143,7 @@ input:focus,
 textarea:focus {
 	border-color: #888;
 	color: #303030;
+	background-color: white;
 	outline: 0;
 }
 textarea:disabled {


app/assets/stylesheets/application.css
```diff
@@ -118,6 +118,7 @@ button,
 select,
 textarea {
 	color: #555;
+	background-color: white;
```
jcs: This also applies to `input`, `button`, and `select`.  You should put this in the block right under this which is just for `textarea`.

jordigh:
Oh, I see. Well, input fields should also affected, but I suppose button and select shouldn't have white backgrounds.

What's the intended background colour for button and select? Doesn't this background colour normally vary across browsers and operating systems?

jcs:
Buttons are already colored specially.  Select doesn't really matter but I guess we can force it to white.

jordigh:
Ah, I see again. Well, that just leaves buttons, but their colour is set further below by another rule, so it's already getting the correct colour. I did notice that I failed to set the colour on focussed textareas and input fields, so I've amended my commit accordingly.

I think your concerns are addressed?

jordigh:
Kind regards and my gratitude for your prompt reply.

Issue #189, opened 2015-02-09
provide comment tree lines


jcs:
On some long threads (especially on mobile) it can be hard to immediately tell which parent a child comment is in response to, so some very light colored tree lines might be helpful (as on the [user tree](https://lobste.rs/u)).

daGrevis:
Maybe it's time to reconsider https://lobste.rs/s/m5dlun/comment_folding.

jcs:
That could be added later, but it really has nothing to do with this.

bentrevor:
Looks like this can be accomplished by just adding a `border-left` to `ol.comments`.  Something like `border-left: 1px dotted #ccc;` if you want it to look like Reddit.

PR #201, opened 2015-04-25
Fixes #181. Add comment folding as decribed in: https://lobste.rs/s/m5dl...

### LINKED ISSUE #181: Option to show top-level comments only
[This thread](https://lobste.rs/s/5ovrc2/what_s_your_most_controversial_technical_opinion) makes me really wish for the ability to only show top-level comments, or at least distinguish top-level comments somehow so yo can quickly identify them while scanning down the list of comments. I'm usually more interested in the top-level comments than the discussions hanging off from each of them.

BennyHallett:
Added comment folding feature as decribed in https://lobste.rs/s/m5dlun/comment_folding

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -270,6 +270,24 @@ $(document).ready(function() {
     return Lobsters.moderateStory(this);
   }),
 
+  $(document).on("click", "a.comment_folder", function() {
+    $(this).addClass("comment_unfolder").removeClass("comment_folder");
+    $(this).text("unfold");
+    var comment = $(this).closest(".comment");
+    comment.nextAll(".comments").each(function() {
+      $(this).addClass("collapsed");
+    });
+  });
+
+  $(document).on("click", "a.comment_unfolder", function() {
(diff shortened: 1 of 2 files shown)


BennyHallett:
The JS isn't firing if you're not logged in, sorry. I'll fix that up and commit a fix asap.

daGrevis:
(y)

jcs:
There is no JS for logged out users, so the fold link should probably just not be there if there is no user.

Though if #189 were implemented, I wonder how useful folding would even be.  There are rarely huge threads that are hard to scroll through like on Reddit, but with vertical thread lines, it would make it much easier to see which comments are replying to which in a heavily nested comment thread.

BennyHallett:
Thanks @jcs , I've stopped displaying the fold link when there is no user.

I see your point about the thread lines and having smaller threads. I still think folding is useful though, even with comment counts as low as 20 or so as in this thread: https://lobste.rs/s/kgdrdj/my_switch_to_openbsd_first_impressions

I like the ability to close off threads as I read them, or choose to ignore them based on the root comment.

BennyHallett:
@jsc is there anything you need me to do to have this PR merged in?

PR #210, opened 2015-07-15
more flexible url regexp validation

VitalySalnikov:
like railscasts #301 but with \A and \z ruby features
https://github.com/railscasts/301-extracting-a-ruby-gem/blob/master/url_formatter/lib/url_formatter.rb

for example with old regexp I couldnot submit a story with url like HTTPS://lobste.rs/, but now I can.


jcs:
`\A^` doesn't make sense, but I committed a fix to allow uppercase characters to validate.

PR #214, opened 2015-08-02
Update the right header links

zg:
It'd be more consistent if the username linked to your profile and a new link titled "Settings" linked to settings.


jcs:
I think it's fairly common for clicking on your own username to go to a preferences/settings page.

There is also not a lot of room in the header for any more links.  I'd actually prefer to get rid of some.

zg:
What about icons as header links?

PR #224, opened 2015-09-23
Implement comment folding in pure css

ABorgna:
Allows guests and users with javascript disabled to fold comments.


daGrevis:
Woah, this is amazing! It totally fixes [the issue](https://github.com/jcs/lobsters/issues/223) I had yesterday.

jcs:
Thanks!

PR #227, opened 2015-10-03
Indicate if a story has a description

pushcx:
To encourage folks to write story descriptions on links, let's mark stories with descriptions so that folks know to click through.

The pilcrow is a little visually heavy, but it seemed better than an arbitrary symbol.

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -26,6 +26,9 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
           "original submitter" %>]
       <% end %>
     </span>
+    <% if story.markeddown_description.present? %>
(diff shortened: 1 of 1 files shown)


jcs:
````
ActionView::Template::Error (undefined method `markdown_description' for #<Story:0x007fcfa3cc5bd0>):
````

I think you mean `markeddown_description`.

jcs:
just for posterity:

    [11/04 10:34] <jcs> yeah i don't like that character but i'm not opposed to showing something in some way when there is text
    [11/04 10:34] <jcs> and i don't know what else to show so i have not updated the pr
    [11/04 10:35] <jcs> maybe just another tag-like thing? i don't know
    [11/04 10:35] <pushcx> Ha! I also don't like that character, but failed to think of something more appropriate.

PR #233, opened 2015-10-26
Add comment tree lines

AndrewBrinker:
This commit closes issue #189, adding a line to the left of each comment to
assist with identifying which comments are nested where in a larger
conversation.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -370,6 +370,7 @@ ol.comments {
 	margin-bottom: 0em;
 	padding-left: 6px;
 	line-height: 1.35em;
+	border-left: 1px dotted #bbb;
 }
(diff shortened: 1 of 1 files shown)


jcs:
I have something similar to that in a local tree but it's not accurate.

Here is what yours looks like: https://i.imgur.com/SXYiWft.png
Here's a mockup of what I would like: https://i.imgur.com/677PsUV.png

Notably the line goes all the way up along the left to the parent vote count, and the collapse links are moved onto the line.  For the last child node of each branch, the line stops.

Sort of like Windows Explorer or whatever else uses this kind of tree:

http://i.imgur.com/cZ1YDaC.png

AndrewBrinker:
You're right. Your version looks better.

PR #235, opened 2015-11-06
Add tooltip for the ︙ marking description_present

roryokane:
When I saw that only some of the stories had the ‘︙’ symbol, I was mystified. It looked like a drag handle, but didn’t react to anything. I only figured out its purpose by using the developer tools and seeing that it had the class `description_present`.

This `title` attribute allows the user to see the meaning of the symbol, by hovering over it.


jcs:
Committed something similar

PR #240, opened 2015-11-13
Update byline for non-authored submissions.

apg:
(I couldn't find any indication that this was previously discussed and didn't feel like the effort involved was too great to avoid working on it before discussion, so... here's a PR. Close at will)

I love the recent addition of the 'authored by' checkbox, but unfortunately, the distinction between 'authored by' and 'by' is too subtle to be picked up without careful attention (at least for me). I tend to see 'by' as implicitly including 'authored ', because I don't tend to read the whole byline, just the '... by &lt;username&gt;.' Naturally, this leads to confusion on my part.

This PR makes a subtle change to the byline of stories. Instead of attempting to classify the type of attribution with a modifier (e.g. 'authored', or potentially 'submitted'), it avoids the use of 'by' all together for non-authored submissions, choosing 'via' instead. 'via' may not be the correct choice, but, to me, implies 'discovery' not 'authorship.' and is easily separated visually from 'by'.

Comments remain unchanged, of course, as they are always authored by, and attributed to the commenter.

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -69,11 +69,11 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
                 srcset="<%= ms.user.avatar_url(16) %> 1x,
                 <%= ms.user.avatar_url(32) %> 2x" class="avatar"></a>
             <% end %>
-
             <% if story.user_is_author? %>
-              authored
+              authored by
+            <% else %>
+              via
             <% end %>
-            by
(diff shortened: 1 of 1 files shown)


apg:
Thanks @jcs!

On Fri, Nov 13, 2015 at 1:52 PM, joshua stein <notifications@github.com>
wrote:

> Merged #240 <https://github.com/jcs/lobsters/pull/240>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jcs/lobsters/pull/240#event-464092716>.
>



-- 
http://www.apgwoz.com

PR #252, opened 2015-12-16
Missing word in duplicate detection headline

holsta:


jcs:
This should actually just be using the `time_ago_in_words_label` helper which includes the "ago" part.

holsta:
Ah, Ruby's not my thing.

PR #255, opened 2015-12-20
elasticsearch support

sandys:
hi,
I have added elasticsearch support to lobsters (as you recommended I do !). It works generally well  - the weights dont match Sphinx, but then the ranking is different anyway.

Is this something you will consider merging into the mainline ? I ask because I have tried to maintain clean separation of the elasticsearch code so that your main sphinx code is not affected in any way. But I'm not sure how you would want me to structure the code. I dont know if it is possible to make the code such that the search infrastructure is choosable using config files.

if you would like me to structure my code in a way that would make it acceptable for you to merge, I'll be happy to do that.
if this is not something you'd consider for merge, please do let me know.

thanks!!
-sandeep


jcs:
There are a ton of changes here that have nothing to do with Elastic Search.  You need to separate those out.

sandys:
sorry about the delay - have been having poor health issues. I'll work on it this week.

PR #256, opened 2016-01-05
mail_new_activity: QP-encode subject

kyrias:
There's no reason to force the subject to be ASCII and mangle many UTF-8
titles when we can just use QP-encoding in the subject like we do in the
body.

This is especially noticable in the IRC channel since the bot mockturtle
uses the mailing list emails to get new stories.

Signed-off-by: Johannes Löthberg <johannes@kyriasis.com>

### DIFF
--- a/script/mail_new_activity
+++ b/script/mail_new_activity
@@ -8,19 +8,6 @@ require APP_PATH
 Rails.application.require_environment!
 
 class String
-  def force_to_ascii
-    # fixup some "smart" quotes and things instead of forcing them to "?" chars
-    gsub("\u201C", '"').
-    gsub("\u201D", '"').
-    gsub("\u2018", "'").
-    gsub("\u2019", "'").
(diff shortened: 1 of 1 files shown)


kyrias:
So, I just noticed that you mentioned a "weird wrapping issue" when adding QP support to the body, (though the body doesn't actually need QP, but that's a separate topic). It's not really weird at all, it's just doing QP properly, which requires a newline every 76 char, and it adds a newline to the end of the message.

This would have been all fine and dandy if the script had been using a proper mailing library which had support for wrapping the header properly, but since it doesn't, and I'm not comfortable enough in Ruby to make it so, I think I'll leave that to someone else feeling up to the task eventually.

As a work-around, me and @alanpost discussed this on IRC and figured that just adding the non-butchered title to the body of the email as well would be simplest. Do you have any comments on this solution, @jcs?

jcs:
28a6faab9

PR #262, opened 2016-02-11
Prettier messaging

crertel:
This PR moves the message creation box up above the message history box, so you don't have to scroll so far in order to get to it. It probably really would prefer just to be a lightbox.

This PR also reverses the order of the messages displayed so that the most recent ones show up on top first--this could probably be done in the controller, but fastest thing was just to reverse it in the view.

Thoughts?

### DIFF
--- a/app/views/messages/index.html.erb
+++ b/app/views/messages/index.html.erb
@@ -8,20 +8,6 @@
 </script>
 
 <div class="box wide">
-  <div class="legend" style="float: right;">
-    <% if @direction == :out %>
-      <a href="/messages">View Received</a>
-    <% else %>
-      <a href="/messages/sent">View Sent</a>
-    <% end %>
-  </div>
-  <div class="legend">
-    Private Messages
-    <% if @direction == :out %>
-      Sent
-    <% end %>
-  </div>
-  
   <div class="legend">
     Compose Message
   </div>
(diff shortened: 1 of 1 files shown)


jcs:
This is kind of done on purpose to get people to delete old messages.

crertel:
Why not just setup a task to remove messages older than 30 days?

crertel:
Also, is that why the "my thread history" is only like a page long?

jcs:
I don't want to delete anyone's important messages, but I suppose a time limit could be imposed.

By "my thread history" do you mean /threads?  That just needs pagination added to it.

crertel:
So, do you want me to kill this, or improve it?

A small part of this that could be worth saving is just the compose message box on top of the page, and leave the sorting alone for now until we come up with a better way of handling the messages.

jcs:
Yeah I'll take a version that just moves the compose box for now.

crertel:
There, that little snip should keep the messaging behavior the same for you!

crertel:
@jcs ?

jcs:
Please roll up the commits into one so I can merge it

crertel:
You mean just roll uip into single commit via rebase?

jcs:
yep

PR #268, opened 2016-03-05
Add script to ease setup for contributors

derekprior:
Some of the steps in the README are not necessary if all you want to do
is contribute back to Lobsters, rather than run your own instance.
Additionally, the steps that are necessary can be scripted without too
much hassle. `bin/setup` will produce a runnable local version of the
application (without Sphinx).

I also updated the README to break up the instructions for running a
local version of Lobsters and configuring your own version for
deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,75 +16,49 @@ site unless you are contributing code that will also benefit [lobste.rs](https:/
 Please see the [CONTRIBUTING](https://github.com/jcs/lobsters/blob/master/CONTRIBUTING.md)
 file.
 
-####Initial setup
+####Development setup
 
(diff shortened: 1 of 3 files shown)


jcs:
Wouldn't this still require manually creating a mysql database and its user?

derekprior:
`db:reset` includes `db:drop` (if necessary) and `db:setup`. `db:setup` includes `db:schema:load` and `db:seed`.

jcs:
Oh, I didn't realize `db:create` will prompt you for the root mysql password.

If you want to resolve the merge conflicts I will merge this.

PR #269, opened 2016-03-05
Support Ruby 2.2.0 and 2.3.0

derekprior:
Judging from the test suite, the only issue in supporting Ruby 2.2.0+
was with the `mysql2` gem. Updating this to the latest 0.3.x allows
Lobsters to run with newer Rubies.

We can't update to the absolute latest `mysql2` gem without first
updating Rails, which is why I used the pessimistic version constraint
for `mysql2`.

This PR also removes the test for username linking functionality which was disabled in 0f498d3. The corresponding test was not removed, leaving the suite failing. We can restore the test when the functionality is returned.

Finally, `htmlentities` was updated to remove a warning about a duplicate key in one of its hashes, which is a new warning in later versions of Ruby, I believe.


derekprior:
@jcs I should explain the series of PRs incoming...

I'm developing a workshop for RailsConf on upgrading apps to Rails 5. We're considering using Lobsters as our demo app, so I'm kicking the tires a bit  and making some changes to make the project easier to setup and run for attendees. I figured that while I was doing this, I would contribute the changes back upstream. If you'd rather not have these changes, let me know and I'll keep them on my own fork.

Thanks!

PR #280, opened 2016-04-05
Fixes #272. Add runSelect2 method to call after previewStory.

### LINKED ISSUE #272: select2 box breaks on new story preview
.

imwally:
The select2 code failed to run after previewing a story
submission. This fix creates a new method that can be called upon
completion of the previewStory load() call.


jcs:
thanks!

PR #287, opened 2016-04-16
Add sample crontab

pushcx:


jcs:
I would rather just put the examples in the README, since the file is not actually used by anything.

PR #290, opened 2016-04-30
Add database creation command to README.md

nickhould:
With the current instructions in the README, when running `rake db:schema:load` the following error message is displayed: `ActiveRecord::NoDatabaseError: Unknown database 'lobsters_dev'Run `$ bin/rake db:create db:migrate` to create your database`.

Prior to running `rake db:schema:load`, the user needs to run  `rake db:create`.


crertel:
Um...you might want to keep this down to just changes to the readme.

This appears to have a bunch of trees, an upgrade to Ruby 2.3, and a bunch of other changes.

PR #294, opened 2016-05-18
I18n support for layout/application.html view

chaica:
I18n support for layout/application.html view

- app/views/layouts/application.html.erb : i18n support for layout/application.html view
- config/locales/en.yml : 18n english strings for layout/application.html
- config/locales/fr.yml : 18n french strings for layout/application.html

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -50,17 +50,17 @@
         title="<%= Rails.application.name %> (Current traffic: <%= @traffic.to_i %>)"></a>
 
         <% links = {
-          "/" => @cur_url == "/" ? Rails.application.name : "Home",
-          "/recent" => "Recent",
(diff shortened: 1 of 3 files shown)


chaica:
up about this PR for the i18n of lobste.rs :)

chaica:
up about this PR for the start of the week ;)

jcs:
@chaica what is the preferred way to handle pluralized words?  Is there something that makes the `pluralize` helper work automatically or does there have to be two entries in the locale file with the logic in the view to pick one or the other?

chaica:
@jcs 

The official I18n Rails documentation gives [an explanation about pluralize](http://guides.rubyonrails.org/i18n.html#pluralization) and [the following howto](https://robots.thoughtbot.com/pluralizing-rails-i18n-translations) is quite straight-forward imho.

PR #312, opened 2016-07-07
Search bar instead of redirecting it to another page

polveenomials:


padi:
@polveenomials hurrah for your first ever PR!

I'll be including a few comments after this one. :)

app/controllers/hats_controller.rb
```diff
@@ -8,8 +8,8 @@ def build_request
 
     @hat_request = HatRequest.new
   end
-
-  def index
+
```
padi: Remove unnecessary whitespaces here and the line below this.

config/routes.rb
```diff
@@ -46,8 +46,7 @@
 
     get "/t/:tag" => "home#tagged", :as => "tag", :format => /html|rss|json/
     get "/t/:tag/page/:page" => "home#tagged"
-
-    get "/search" => "search#index"
+    get"/search" => "search#index"
```
padi: This doesn't add anything new except changing the spacing in between routes.

padi:
In general, follow the notes here https://github.com/jcs/lobsters/blob/master/CONTRIBUTING.md#making-changes

There's also a note about commit messages in that link, but here's an article explaining the rationale in detail: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Issue #197, opened 2015-03-30
Hide comments from specific users


wickedchicken:
Hi,

Much in the same way stories can be hidden, I'd like for comments by certain users to be hidden as well. Ideally this could show up as <comment by a hidden user> which could be clicked and expanded if desired.

The reason I am requesting this is that there are certain commenters whom I would like to avoid, but don't think should be outright banned from the site. A private block seems like a happy solution.

Thanks for lobste.rs, it's great!

jcs:
That would break threading, so is it supposed to hide the entire thread under a blocked comment?

kyrias:
You could possibly replace the comment with an empty dummy-comment, having the user listed as ignored.

bconway:
+1. I've found an increasing number of users that inject their politics into every post they make and find myself avoiding discussions entirely.

wickedchicken:
The UI I was imagining would be as follows (using ⋅⋅⋅ to signify indents, markdown doesn't seem to like spaces):

user1 2 hours ago | link | reply
An insightful comment.

⋅⋅⋅ [hidden] 16 minutes ago | unhide

⋅⋅⋅⋅⋅⋅ user1 4 minutes ago | link | reply
⋅⋅⋅⋅⋅⋅⋅⋅⋅Another insightful comment responding to the hidden one.

pushcx:
I wrote a greasemonkey script for myself a while ago

```javascript
// ==UserScript==
// @name        Atheist Lobsters
// @namespace   http://lobste.rs/u/pushcx
// @description separation of church and sanity
// @include     https://lobste.rs/*
// @version     1
// @grant       none
// ==/UserScript==

domain = 'example.com';
username = 'example';

// hide stories
[].slice.call(document.querySelectorAll('.story a.domain[href="/search?q=domain:' + domain + '&order=newest"]')).forEach(function (el) { el.closest(".story").remove() });

// hide comments
[].slice.call(document.querySelectorAll('.comment a[href="/u/' + username + '"]')).forEach(function (el) { el.closest(".comment").remove() });
```

crertel:
@wickedchicken @bconway have you considered just...ignoring...those posts? Also, you might want to link to some examples to show what you mean.

Adding additional echochamber features is probably not a great idea.

bconway:
@crertel Thanks for the tip. Since I left the site over a year ago, it's probably safe for me to unsubscribe from this conversation.

wickedchicken:
[...]

PR #329, opened 2016-11-25
Hide comments from specific users [issue #197]

andela-cdaniel:
This PR implements the feature requested in #197

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem "unicorn"
 gem "mysql2", ">= 0.3.14"
 
 # uncomment to use PostgreSQL
-# gem "pg"
+gem "pg"
 #
 # NOTE: If you use PostgreSQL, you must still leave enabled the above mysql2
 # gem for Sphinx full text search to function.
@@ -37,4 +37,5 @@ group :test, :development do
   gem "machinist"
   gem "sqlite3"
   gem "faker"
+  gem "capybara"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem "unicorn"
 gem "mysql2", ">= 0.3.14"
 
 # uncomment to use PostgreSQL
-gem "pg"
+# gem "pg"
 #
 # NOTE: If you use PostgreSQL, you must still leave enabled the above mysql2
 # gem for Sphinx full text search to function.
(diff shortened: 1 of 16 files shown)


crertel:
Some questions:

* Why did you add a dependency on `capybara` when that wasn't there before, when the current tests seem to not use it?
* Why did you add a dependency on `pg` that wasn't there before, and why is the PG extension now a hard requirement, when the current codebase uses MySQL?
* Why did you add a dependency on `addressable` that wasn't there before?
* Why did you remove the limits on the text columns (and what does this mean for people submitting maliciously large data)?
* Why did you increase the limit on the vote column in the vote table?

Good testing though! Just maybe try to stick with the idioms already present in the codebase.

andela-cdaniel:
> Why did you add a dependency on pg that wasn't there before, and why is the PG extension now a hard requirement, when the current codebase uses MySQL?

An oversight on my part (I'm most familiar with postgres and I used it instead of mysql).  I'll update the pr and remove the dependency on the postgres gem

> Why did you add a dependency on addressable that wasn't there before?

It's a runtime dependency for `Capybara` (added by the capybara gem)

> Why did you remove the limits on the text columns (and what does this mean for people submitting maliciously large data)?
> Why did you increase the limit on the vote column in the vote table?

This I got to look into as I only added the migration to create one new table (blocked users) and I am unclear as to why this would cause the limit constraints to be removed.

crertel:
Thank you for answering my questions. :)

pushcx:
> Why did you increase the limit on the vote column in the vote table?

Rails creates `schema.rb` by introspecting on the database. If PostgresSQL and MySQL differ in their understanding of a migration or implementation of a table, you'll get churn in the file.

The best way to avoid it is to `git add -p` after running a migration. To fix it, rollback the migration, `git add db/schema.rb`, re-run the migration, then `git add -p` only the deliberate changes (which may require manual editing). Then commit and `git checkout db/schema.rb` to toss the spurious changes.

andela-cdaniel:
@crertel, @pushcx, I've updated the PR (removed the `pg` gem and retain the limits enforced on particular columns)

danielcompton:
After squashing these commits, the rebase was trivial. What's needed next to move this PR forward?

jcs:
I would like some community discussion on Lobsters on whether this type of feature is wanted.

danielcompton:
I've made a meta post about this here: https://lobste.rs/s/debnsq/hiding_comments_by_specific_users

jcs:
Sorry, but there didn't seem to be a lot of concensus around wanting this.

Issue #273, opened 2016-03-21
Ctrl+Enter should act like clicking the "Post" button


soc:
It would be amazing if Ctrl+Enter would work as an alternative for clicking the "Post" button when commenting.

GitHub does it, and it is amazing!

PR #337, opened 2017-01-19
Submit comment with ctrl+enter or cmd+enter [issue #273]

josecanhelp:
Resolves issue #273

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -484,4 +484,23 @@ $(document).ready(function() {
     $("#story_guidelines").toggle();
     return false;
   });
+
+  $(document).on("focus", "textarea#comment", function() {
+    var trigger = false;
+
+    $(".comment_form_container").on("keydown", function(e) {
+      // left cmd - 91 / right cmd - 93 / left ctrl - 17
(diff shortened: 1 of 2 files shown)


jcs:
Can this be moved to `application.js.erb`?

josecanhelp:
@jcs Absolutely. I'll work on that.

josecanhelp:
@jcs Yes, much cleaner in application.js.erb. Thanks for the recommendation. 👍

app/assets/javascripts/application.js.erb
```diff
@@ -484,4 +484,23 @@ $(document).ready(function() {
     $("#story_guidelines").toggle();
     return false;
   });
+
+  $(document).on("focus", "textarea#comment", function() {
+    var trigger = false;
+
+    $(".comment_form_container").on("keydown", function(e) {
```
jcs: I think this installs a duplicate event handler every time the field is focused, which would cause this to run more than once if the user types text, unfocuses the field, then focuses it again.

josecanhelp: This is a fair concern. I apologize for not considering it. After some research, I came up with a cleaner method.

Issue #344, opened 2017-02-26
make password change require current password


jcs:
https://lobste.rs/s/1cyltz/two_factor_authentication_now_available/comments/rxgla0#c_rxgla0

PR #345, opened 2017-02-27
Require current password on password change [issue #344]

nickfrostatx:


jcs:
I was hoping to have this just when changing the password, not changing any of the other settings.

nickfrostatx:
If the password field is left blank, then the current password is ignored and the settings are updated as before.

But if you try to change your password and, say, the "About" box, then it will check the current password.  If the current password is incorrect, it will make no updates.

On an unsuccessful password change attempt, would it be better to still update the rest of the settings, or nothing at all?

PR #353, opened 2017-03-23
Show @2x images for commenter avatars

### LINKED ISSUE #352: Retina images for commenter avatars
Retina avatars are used for the byline of the submitter, but not for comments on a post. It looks like [app/views/comments/_comment.html.erb#L40-L43](https://github.com/jcs/lobsters/blob/310a0bf4a2f6f6554c0c5d9881aadef3d382d269/app/views/comments/_comment.html.erb#L40-L43) needs to be updated to match [app/views/stories/_listdetail.html.erb#L66-L71](https://github.com/jcs/lobsters/blob/310a0bf4a2f6f6554c0c5d9881aadef3d382d269/app/views/stories/_listdetail.html.erb#L66-L71), or perhaps factor the avatar logic out separately so it can be reused everywhere?

danielcompton:
Fixes #352. I'm not a Rails developer, so I wasn't sure how best to factor out the avatar code. Also, I'm not sure if this requires test/better commit message/e.t.c. as it is a small change. Happy to update with whatever you need here.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -39,7 +39,9 @@ class="comment <%= comment.current_vote ? (comment.current_vote[:vote] == 1 ?
 
       <% if (@user && @user.show_avatars?) || !@user %>
         <a href="/u/<%= comment.user.username %>"><img
-          src="<%= comment.user.avatar_url(16) %>" class="avatar"></a>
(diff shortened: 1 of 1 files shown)


jcs:
Thanks, I implemented this in f4e09205ce4c8b842e59161c92b381d16a66ba35

PR #354, opened 2017-04-12
Use CommonMarker

### LINKED ISSUE #288: fix remaining markdown issues
- tildes for strikethrough don't work, because we have to pass `:strict` (see cbf14021ac06).
- `@username` expansion doesn't work, because of security issues (see 0f498d3ee3).

kivikakk:
Fixes #288. We add CommonMark compatibility, strikethrough, `@username` expansion (with a kinda awkward AST walk, but it's simple if verbose), and with any luck general malleability going forward.

### DIFF
--- a/extras/markdowner.rb
+++ b/extras/markdowner.rb
@@ -44,7 +44,7 @@ class Markdowner
 
   def self.postprocess_text_node(node)
     while node
-      return unless node.string_content =~ /\B(@[\w\-]+)/
+      return unless node.string_content =~ /\B(@#{User::VALID_USERNAME})/
       before, user, after = $`, $1, $'
 
       node.string_content = before
--- a/extras/markdowner.rb
+++ b/extras/markdowner.rb
@@ -27,7 +27,7 @@ class Markdowner
 
     # make links have rel=nofollow
     ng.css("a").each do |h|
-      h[:rel] = "nofollow" unless URI.parse(h[:href]).host.nil?
+      h[:rel] = "nofollow" unless (URI.parse(h[:href]).host.nil? rescue false)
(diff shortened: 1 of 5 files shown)


extras/markdowner.rb
```diff
@@ -7,49 +7,71 @@ def self.to_html(text, opts = {})
...
     end
 
+    if !opts[:allow_images]
+      ng.css("img").remove
+    end
+
     # make links have rel=nofollow
     ng.css("a").each do |h|
-      h[:rel] = "nofollow"
+      h[:rel] = "nofollow" unless h[:href].starts_with?("/")
```
jcs: A link like `[test](//jcs.org/)` would fail that, but still need `nofollow`.

extras/markdowner.rb
```diff
@@ -7,49 +7,71 @@ def self.to_html(text, opts = {})
...
+    node.each do |child|
+      walk_text_nodes(child, &block)
     end
-end
+  end
 
-    ng.at_css("body").inner_html
+  def self.postprocess_text_node(node)
+    while node
+      return unless node.string_content =~ /\B(@[\w\-]+)/
```
jcs: It might be worth moving the username validation regex in `User` out to a constant and checking against that.

extras/markdowner.rb
kivikakk: Good catch. Now checking for a nil host on URI parse.

extras/markdowner.rb
kivikakk: :+1:

jcs:
Thanks a lot for taking care of this!

PR #367, opened 2017-06-10
Upgrade rails 5 1

HParker:
Rails 4.2 only receives severe security fixes now. It makes sense to upgrade if possible.

I ran though the major things that came up to get the tests to pass and what I found when clicking around.

In addition, I ran `rails app:update` to get the new scripts that come with 5.1 I think it would be totally fine to do this upgrade without them if you like, so I left all that in a different commit so you can exclude it if you like.


HParker:
I realize this might create a bit of extra work for a maintainer to test and assure that I didn't miss any cases.

If you like, I am happy to contribute tests or any other confidence measures we want to put in place.

jcs:
#yolo

PR #378, opened 2017-07-12
Add tableheader for invite deletion

danielgustafsson:
Commit b592aae9 allowed the _Delete_ option for non-moderators, but didn't add the `<th></th>` to extend the grey tableheader bar over the new option. This adds an empty `<th>` cell like the one over _Send Invitation_.

### DIFF
--- a/app/views/invitations/index.html.erb
+++ b/app/views/invitations/index.html.erb
@@ -15,9 +15,7 @@
     <th width="20%">Name</th>
     <th width="40%">Memo</th>
     <th width="15%"></th>
-    <% if @user.is_moderator? %>
-      <th></th>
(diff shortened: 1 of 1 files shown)


jcs:
I think the proper fix is to just remove the `if` conditional around the existing `<th>`

danielgustafsson:
> On 12 Jul 2017, at 21:25, joshua stein <notifications@github.com> wrote:
> 
> I think the proper fix is to just remove the if conditional around the existing <th>

*facepalm* .. you are of course absolutely right; for some reason I completely missed that. Force-pushed a rebased branch with that fix unless you’ve already fixed it.

PR #397, opened 2017-09-20
Update delete, reply, edit links to be tabable

lord:
Currently the delete, reply, and edit links on comments are skipped over when
using the tab key to flip through the various links. Adding an href allows tab
to navigate to these links, and `javascript:void(0)` prevents the page from
scrolling when these links are clicked.

Thanks again for the awesome site! Glad to make any edits to this you deem necessary, just thought I'd submit a quick patch since I noticed this the other day when using the site without a mouse.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -64,24 +64,24 @@ class="comment <%= comment.current_vote ? (comment.current_vote[:vote] == 1 ?
 
         <% if comment.is_editable_by_user?(@user) %>
           |
-          <a class="comment_editor">edit</a>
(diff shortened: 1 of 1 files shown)


jcs:
`javascript:void(0)` links are no good.  The proper way to fix this would be to set [`tabindex=0`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) on each of those `a`'s.

> tabindex="0" means that the element should be focusable in sequential keyboard navigation, but its order is defined by the document's source order.

lord:
Gotcha. PR updated!

PR #400, opened 2017-09-25
Comment tree lines

talklittle:
ref #189 

Based on mockup https://i.imgur.com/677PsUV.png from #233 

Current result (tested on Firefox / Ubuntu desktop only):

**Updated with narrower indentation**

lobsters-comment-tree-lines-2

[Previous screenshot](https://user-images.githubusercontent.com/87961/30788320-38a662bc-a14f-11e7-9b79-ca3d7cf80448.png)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -376,7 +376,7 @@ ol.comments {
 	padding-left: 6px;
 }
 ol.comments1 {
-	margin-left: 0;
+	margin-left: -20px;
 	padding-left: 0;
 }
 ol.comments.comments1 {
@@ -463,6 +463,7 @@ ol.stories li.story div.story_liner {
 	clear: both;
 	padding-top: 0.4em;
 	padding-bottom: 0.4em;
+	padding-left: 20px;
 }
 
 .comment a {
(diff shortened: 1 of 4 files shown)


xcombelle:
the arrow space is too large at first sight. half of the space should be good

colindean:
What if there was a subtle color difference in each level of the thread?

image

Each top comment gets a different color than the thread before or after it, maybe choosing out of seven or eight colors. Subthreads get a variation on that color or a different color altogether so that it's possible to tell when a subthread has ended based on the color, too.

talklittle:
@xcombelle Good point, I narrowed the indentation back to the original width. See updated screenshot.

@colindean Copying [my comment](https://lobste.rs/s/rmlnke/add_comment_tree_lines_lobste_rs#c_ois2a8) here for completeness

> If we agree that color-coding is an improvement, there's still a debate whether the colors should be per top-level comment (vertical rainbow), or the color should indicate the depth of the comment (horizontal rainbow).
>
>So I think it's best to do that in a future pull request after some more thought and experimentation.

colindean:
Definitely, merge it and then iterate!

jcs:
Thanks for tackling this.  The mobile/responsive view seems broken though.

https://imgur.com/a/xVj8R

talklittle:
@jcs Thanks for catching that. I've pushed a commit to hide the tree lines on mobile. I think the addition of the top-level folder buttons and tree lines makes the comments section too cramped horizontally.

jcs:
I backed this out locally, there's still some bugs to work out.

In [this shot](https://i.imgur.com/C5tCrfO.png) you can see at the bottom, the vertical line doesn't continue through the three posts under angersock's.  Same in [this one](https://i.imgur.com/Rv4noeo.png) missing a line between tedu and SeanTAllen.  Also, the top-level collapse box no longer lines up under the '24' at the very top.  It would be nice to preserve that alignment if possible.

PR #401, opened 2017-09-26
Fix comment tree lines

talklittle:
Fixes bugs pointed out in https://github.com/jcs/lobsters/pull/400#issuecomment-332069284


jcs:
Thanks, I manually squashed these into your previous commit.

PR #403, opened 2017-09-26
Github-style markdown previewer tab

AlessandroMinali:
A view cleanup and Github-style markdown previewer tab proof of concept.

View can be further cleaned up(split and logic removed) and this new tab can be generalized for comments as well, if this PR makes sense.

https://github.com/jcs/lobsters/issues/101


pushcx:
Hey @AlessandroMinali, thanks for starting on this. I just had a chance to dig in, I see you started with previewing stories.

Why did you break out the story errors, titles, tags, and suggestions to partials rather than break the story display out into a partial for stories/show view and the preview tab on the edit form to use?

AlessandroMinali:
The reason for creating partials (errors, titles, tags) there was to do the minimal work to separate logic out of the view, in this case whether or not we are suggesting and I think passing locales as a flag is bad practice (https://github.com/lobsters/lobsters/search?utf8=✓&q=suggesting&type=). In my opinion it is much clearer now what a suggestion involves vs a story since they aren't jumbled together. (https://github.com/lobsters/lobsters/pull/403/files#diff-86163f97337184101a51544ed9ce5f98)

The preview tab can be made it a partial pretty easily since it would just be 4 lines. (https://github.com/lobsters/lobsters/pull/403/files#diff-114e72f0d9783597d0e4c07d7cb7183f)

There is no work here concerning `stories#show` at least not intentionally, so I'm not to sure what you are trying to say 😅

pushcx:
In that case, I'd prefer a boolean for suggestion mode to splitting the code out into several partials.

Also, minor style issue, please use `"` for html attributes instead of `'`.

I look forward to landing this when it's working on comments and we've worked out the little details.

PR #412, opened 2017-10-04
Update .travis.yml

hmadison:
Sets up services needed for test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,7 @@
 language: ruby
 services:
   - mysql
+env:
+ - DATABASE_URL=mysql2://root:@localhost/lobsters_test
 before_script:
   - ./bin/rails db:setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: ruby
 services:
   - mysql
 env:
- - DATABASE_URL=mysql2://root:@localhost/lobsters_test
+  - DATABASE_URL=mysql2://root:@localhost/lobsters_test
 before_script:
-  - ./bin/rails db:setup
+  - ./bin/rails db:create
+  - ./bin/rails db:schema:load
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,6 @@
 ---
 language: ruby
+services:
+  - mysql
+before_script:
+  - ./bin/rails db:setup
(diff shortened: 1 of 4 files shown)


talklittle:
Awesome! Maybe add the badge to the README while you're at it?

    [Build Status](https://travis-ci.org/lobsters/lobsters)

spec/models/user_spec.rb
```diff
@@ -6,7 +6,7 @@
     expect { User.make!(:username => "") }.to raise_error
     expect { User.make!(:username => "*") }.to raise_error
 
-    User.make!(:username => "test")
+    User.make!(:username => "test") unless User.find_by(username: 'test')
```
pushcx: Why is this needed? Is this papering over a test failing to clean the database?

pushcx:
Thanks, this is great to have on PRs.

spec/models/user_spec.rb
hmadison: This fixed a failure on two runs:
* https://travis-ci.org/lobsters/lobsters/builds/283332809?utm_source=github_status&utm_medium=notification
* https://travis-ci.org/lobsters/lobsters/builds/283330930?utm_source=github_status&utm_medium=notification

We have transnational fixtures turned on, so we should be clearing the database after each run.

pushcx: Uh... that sure looks like either test data failed to get cleaned or some general pre-test hook is creating duplicate data. There's something wrong here, this `unless` shouldn't exist. I'm sorry I didn't think of this before hitting merge.

hmadison: Ill dig at this more locally and try to see whats causing the issue.

pushcx:
Reopening so I can't forget about the `unless` again. @hmadison, please reuse this PR if you can land a fix on your `travis--conf` branch. Hopefully GitHub does the right thing.

.travis.yml
```diff
@@ -1,2 +1,9 @@
 ---
 language: ruby
+services:
+  - mysql
+env:
+ - DATABASE_URL=mysql2://root:@localhost/lobsters_test
```
talklittle: nitpick: indentation :)

hmadison:
@pushcx The issue was the seeds.rb file had a definition for the test user which was getting loaded in before the test run.

pushcx:
Great, thank you for running that down!

PR #413, opened 2017-10-05
story: check for duplicate URL on blur

### LINKED ISSUE #149: check for duplicate article immediately
This is really a usability bug, not a logic bug, but it's my biggest frustration right now.

When posting a new article, it should check for a duplicate url immediately after you paste it into the url field. That would save me a lot of effort coming up with tags and typing a short summary, only to find out at the last stage that someone else has already posted the link.

talklittle:
fixes #149


app/controllers/stories_controller.rb
```diff
@@ -329,6 +329,22 @@ def unsave
...
   end
 
+  def check_url_dupe
+    @story = Story.new(story_params)
+
+    if @story.url.present?
+      @story.check_url
+
+      # ignore other types of errors (e.g., invalid URL format)
+      if @story.already_posted_story.blank?
```
pushcx: Instead of extracting all of `check_url` from `validate`, extract just the part that sets `already_posted_story` and maybe adds the error about it being submitted recently. Leave the regexp that checks format and adds that error.

I don't want this piece of code to have to know about all the other possible errors on URL and be maintained in sync with the model.

app/views/stories/_formerrors.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="form_errors_header">
```
pushcx: Please name this file `_form_errors` rather than run the words together.

pushcx:
Great to have this. It might be nice to use `slideDown` or something to make this a little less jarring, but I'd rather merge this now than wait.

PR #415, opened 2017-10-06
story: return nil if no similar url

talklittle:
allows use of Ruby 2.3 Safe Navigation Operator (&.)


talklittle:
fixed in master as cca4f06807296c49bbdac866ffe2fc9078032b4e

PR #416, opened 2017-10-07
Allow creating and editing tags for admin users

masongup:
I'm trying to start my own site, and was bothered by needing to login to a MySQL console to create and update tags, so I did a quick and basic UI for it, limited to admin users. I saw JCS's complaint about it on the post for passing the torch, so I thought I'd PR it for you guys.


app/controllers/tags_controller.rb
```diff
@@ -10,4 +12,27 @@ def index
...
     end
   end
+
+  def new
+    @tag = Tag.new
+    @title = "Create Tag"
+  end
+
+  def create
+    tag = Tag.create!(params.require(:tag).permit(:tag, :description, :privileged, :is_media, :inactive, :hotness_mod))
```
pushcx: Please break out the params filtering to a method for this and `update` to share.

pushcx: Tag needs to validate the non-empty presence of the `tag` field, the `tags` table should be migrated to remove the `""` default, and this action and `update` need to check that validations passed. I'm not really OK with the console letting you break things in such easily prevented ways, and a web interface definitely shouldn't allow it.

app/views/layouts/application.html.erb
```diff
@@ -88,6 +88,9 @@
         <% if lookup_context.template_exists?("footer", "layouts", true) %>
           <%= render :partial => "layouts/footer" %>
         <% else %>
+          <% if @user && @user.is_admin? %>
```
pushcx: Please remove this and just note the link in `README.md`. Tags are edited so infrequently that it's not worth the space. It can get a link on the admin dashboard when that gets built.

app/views/tags/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      <%= f.label :tag, 'Name:', class: 'required' %>
+      <%= f.text_field :tag  %>
+    </div>
+
+    <div class="boxline">
+      <%= f.label :description, 'Description:' %>
+      <%= f.text_field :description %>
+    </div>
+
+    <div class="boxline">
```
pushcx: These three booleans could use a little descriptive help text about what they do.

app/views/tags/index.html.erb
```diff
@@ -3,10 +3,16 @@
...
+  <% if is_admin %>
+    <div>
+      <%= link_to 'Create New Tag', new_tag_path %>
+    </div>
+  <% end %>
   <% max_size = @tags.map{|t| t.stories_count }.max %>
   <% @tags.each do |tag| %>
     <% mod = (max_size.to_f / ((tag.stories_count.to_f == 0) ? 1 : tag.stories_count.to_f )) %>
-    <%= link_to tag.tag, tag_path(tag), :class => tag.css_class,
+    <%= link_to tag.tag, is_admin ? edit_tag_path(tag.id) : tag_path(tag), :class => tag.css_class,
```
pushcx: Add an edit link, don't replace the link to the per-tag view.

app/views/tags/new.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="box wide">
+  <div class="legend">
+    Tag Create
+  </div>
+
+  <%= form_for @tag, url: tags_path, method: :post do |f| %>
```
pushcx: Share a partial between the new and edit views.

config/routes.rb
```diff
@@ -141,6 +141,10 @@
 
     get "/tags" => "tags#index"
     get "/tags.json" => "tags#index", :format => "json"
+    get "/tags/new" => "tags#new", :as => "new_tag"
```
pushcx: Convert this to `resource` instead of listing actions individually.

app/controllers/application_controller.rb
```diff
@@ -88,6 +88,19 @@ def require_logged_in_moderator
     end
   end
 
+  def require_logged_in_admin
+    require_logged_in_user
+
+    if @user
```
pushcx: `unless @user.try(:is_admin?)`

app/views/tags/edit.html.erb
masongup: I agree, and perhaps a suggested range for `hotness_mod` as well. Speaking of which, I don't really know what they do right now either, besides `is_media`. I can try and look things up and figure it out, but can anyone chime in on what the intent of these flags is and how `hotness_mod` works?

pushcx: * `hotness_mod` is a multiplier on score in rankings. `1` is normal. Lower values reduce score, higher raises it. Normal range is probably `0.5` to `1.5`.
 * `is_media` tags the medium, nice for noting pdf or videos that some people don't like to see. A story must have at least one non-media tag.
 * `privileged` only moderators can apply this tag, useful for `announce` and other staff threads
 * `inactive` can't be applied to new articles. Useful so you don't delete history if you decide to change your tagging scheme.

masongup:
Thanks for the review. I'll be looking over these over the next few days or so. These suggestions seem to also call for adding some specs at the model and possibly controller level as well.

app/controllers/application_controller.rb
masongup: Tried that, it gives a double render error if you try to access one of the protected pages when there is no user logged in. As long as we're using `require_logged_in_user` as-is, it needs to only render when `@user` is not `nil` and also not admin. I agree that this is pretty awkward, but I think it needs a broader refactoring of how authorization is done that seems outside the scope of this PR.

app/views/tags/edit.html.erb
masongup: On the `hotness_mod`, I got concerned when I noticed that all of the tag `hotness_mod`s on my site were `0.0`, so I checked the JSON output on production Lobsters and the code. It appears that `0.0` is normal, and a couple of tags have `-0.25`, so I'll put those ranges in the text.

config/routes.rb
masongup: As I feared when I did it that way, the resourceful route generator really wants to name the route I named `update_tag` as `tag`. However, the existing routes already use the `tag` route name for the `home#tagged` route, which is used all over the site. So to use that, I'd have to change the name of that route, and every reference to that route throughout the app. This feels like something that ought to be done as part of a general refactor to use more resourceful routes throughout the app.

pushcx: I'd expect the resource route to be `POST` or `PUT` instead of `GET`. Rails puts [multiple actions at the same URL](http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions) as long as they have different verbs.

masongup: Sorry if I didn't describe it well enough. What's happening is the following routes:

```
tag GET      /t/:tag(.:format)                             home#tagged {:format=>/html|rss|json/}
    GET      /t/:tag/page/:page(.:format)                  home#tagged {:format=>/html/}
```

Get created by the following declarations in `routes.rb`:

```
get "/t/:tag" => "home#tagged", :as => "tag", :format => /html|rss|json/
```

While adding `resources :tags, except: [:show, :destroy]` generates the routes:

```
       tags GET      /tags(.:format)                               tags#index {:format=>/html/}
            POST     /tags(.:format)                               tags#create {:format=>/html/}
    new_tag GET      /tags/new(.:format)                           tags#new {:format=>/html/}
   edit_tag GET      /tags/:id/edit(.:format)                      tags#edit {:format=>/html/}
            PATCH    /tags/:id(.:format)                           tags#update {:format=>/html/}
            PUT      /tags/:id(.:format)                           tags#update {:format=>/html/}
            GET      /tags.json(.:format)                          tags#index {:format=>/json/}
```

If the above declarations are commented out, then you get the resources routes like follows:

```
       tags GET      /tags(.:format)                               tags#index {:format=>/html/}
            POST     /tags(.:format)                               tags#create {:format=>/html/}
    new_tag GET      /tags/new(.:format)                           tags#new {:format=>/html/}
   edit_tag GET      /tags/:id/edit(.:format)                      tags#edit {:format=>/html/}
        tag PATCH    /tags/:id(.:format)                           tags#update {:format=>/html/}
            PUT      /tags/:id(.:format)                           tags#update {:format=>/html/}
            GET      /tags.json(.:format)                          tags#index {:format=>/json/}
```

The `tag` route identifier is getting assigned to the heavily-used `/t/:tag` route which connects to `home#tagged`, and so is not available for the `/tags/:id` route, which connects to `tags#update`. I don't see any way to change the route name generated by `resources`, so the only way to resolve the conflict and allow the use of `resources` routes is to change the name of the other tag routes, which are used in several other places in the application.

app/views/tags/index.html.erb
masongup: Trouble is, the all-tags display looks like this:

image

So where do I put the edit link? An extra row of text in each bubble? I don't think we have a tag list view where it would look reasonable to add an extra link in.

app/views/layouts/application.html.erb
masongup: I was thinking about this, and the page I used for it is a nice screen of all tags sized by popularity that's been in the app for a while. Oddly enough, it doesn't seem to be linked anywhere in the app that I've seen. What if we left the link to that page there for all users instead of restricting to admin? Just a thought, deleting it is fine if we don't want to do that.

masongup:
Added in leaving entries in the moderation log as well.

app/views/tags/index.html.erb
pushcx: Ah, I misread and thought this template was for `/filters`, not `/tags`, my mistake. I don't have any better suggestion.

app/views/layouts/application.html.erb
pushcx: Yeah, let's just delete.

config/routes.rb
pushcx: Oof, thanks for breaking that down. Too much magic and convention, let's definitely just stick with the way you've got it now.

spec/controllers/tag_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "spec_helper"
+
+describe TagsController do
+  before { allow(controller).to receive(:require_logged_in_admin) }
+
+  context 'create' do
+    it 'creates new tags' do
+      post :create, params: { tag: { tag: 'my tag' } }
```
pushcx: Great to have a couple specs. Could you change this to `'mytag'`? I'm pretty sure the site assumes tags to be single words. Some of the models only edited by the admin in console are missing validations (mostly presence of references).

db/migrate/20171018034551_drop_default_tag_name.rb
```diff
@@ -0,0 +1,5 @@
+class DropDefaultTagName < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default(:tags, :tag, from: '', to: nil)
```
pushcx: Good catch.

spec/controllers/tag_controller_spec.rb
```diff
@@ -15,21 +15,50 @@
...
     end
+
+    it 'rejects updates with unpermiited params' do
+      expect { post :update, params: { id: tag.id,  tag: { is_media: true } } }
+        .to raise_error ActionController::UnpermittedParameters
+    end
+
+    it 'updates with all permitted params' do
+      post :update, params: { id: tag.id,  tag: {
+        tag: 'my tag', description: 'desc', hotness_mod: 1.5, privileged: true, inactive: true
```
pushcx: Same `my tag` -> `mytag` here.

pushcx:
@masongup Sorry to be so slow reviewing this, but I'm really excited to see the tests and mod logging. Looks like we're down a couple very small issues before merging.

masongup:
@pushcx Hey, thanks for getting to it. I believe I've got all of the comments fixed related to these changes, however, I merged master in to fix a conflict on the readme, and Travis doesn't seem to like the changes to the database made on master branch.

The Travis build breaks on trying to create that `replying_comments` view, I think because the query is hard-coded to use the database schema `lobsters` while the travis DB is named `lobsters_dev`. I think fixing this to not depend on the schema name would fix Travis, though I haven't dug into it too far yet.

I was also getting an error trying to migrate my local MySQL DB to latest. It seems to not like migration `20180124143340_add_doffed_at_to_hat.rb`, giving error `Mysql2::Error: Invalid default value for 'doffed_at': ALTER TABLE `hats` ADD `doffed_at` datetime DEFAULT 0`. It seems this invalid default was fixed in a later migration `20180131203555_fix_doffed_at.rb`, however IMO the earlier migration should be fixed so that older DB versions can be cleanly migrated to latest.

How do you want to handle these database issues? Redoing all of that stuff on this PR seems out of scope. I could open another PR off of master to clean it up, though it may take me a few days or so to get to it. Or would you like to fix it yourself or have someone else do it? We can wait until it gets fixed, then merge the fix into this PR so that we can merge it with a clean CI build.

pushcx:
The MySQL issue is caused by it changing a default about how much silly invalid stuff it should do. I've pushed an edit. We're tracking the Travis issue over in #450. It doesn't sound like that's blocking you, so you should be good if you merge in from master now. If not, let me know and I'll take a look

masongup:
I merged master in again to pick up the new migration, but Travis still fails though. If you're okay to merge with a failing CI, we can go ahead, or we can wait until that bug is fixed.

pushcx:
@masongup Sorry I forgot about this for a while, but I'm glad to merge it this morning. Thank you!

masongup:
Excellent, thanks!

PR #427, opened 2017-10-23
user: add method is_heavy_self_promoter?

talklittle:
Per discussion https://lobste.rs/s/2somjr/5_ways_developers_waste_more_than_20_their#c_odxdsb and on IRC.

Future work (future PRs) could include checking domain names.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -101,6 +101,12 @@ class User < ActiveRecord::Base
   # minimum karma required to process invitation requests
   MIN_KARMA_FOR_INVITATION_REQUESTS = MIN_KARMA_TO_DOWNVOTE
 
+  # proportion of posts authored by user to consider as heavy self promoter
+  HEAVY_SELF_PROMOTER_PROPORTION = 0.51
+
+  # minimum number of submitted stories before checking self promotion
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -382,6 +388,20 @@ def is_new?
     Time.now - self.created_at <= NEW_USER_DAYS.days
   end
 
+  def is_heavy_self_promoter?
+    total_count = self.stories_submitted_count
+
+    if total_count < MIN_STORIES_CHECK_SELF_PROMOTION
+      false
+    else
+      authored = Story.where(
```
pushcx: Rails allows you to add to an association scope, so this could be `self.stories.where(:user_is_author => true).count`.

talklittle:
@pushcx Okay I added something to the user profile visible to moderators. Ready for another review.

For now I actually don't think anything should be user-facing, and I'm against doing a hard error on story submission. We would not want to discourage users from posting original content that is high quality and fits in with Lobsters.

IMO having this visible to moderators just adds one more tool to figuring out if a user is submitting content that doesn't belong at Lobsters.

pushcx:
Good points. We can definitely take this one very slowly, as the topic generally only comes up infrequently when someone starts posting stories that attract spam flags and critical comments.

Issue #414, opened 2017-10-05
more realistic fake_data


pushcx:
Improve `lib/tasks/fake_data` to add:
 - [x] text-only stories
 - [x] stories with text and links
 - [x] comments
 - [x] replies to comments
 - [x] user-deleted stories and comments
 - [x] saved stories
 - [x] hats
 - [x] moderators
 - [x] moderations of users: ban, unban, disable + enable invites, grant moderatship
 - [x] moderations of stories: merge, edit, delete
 - [x] moderations of comments: delete, undelete
 - [x] moderations of hats: add, delete

It's not important that `stories_count` and `users_count` be accurate or even preserved.

This is a great first issue for new contributors. These don't need to all be done in one PR, you can do just one or two.

pushcx:
This is done thanks to all our contributors!

PR #431, opened 2017-11-12
Adding text only stories to fake_data task

ponyta:
This is for issue #414 (text only stories).

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -15,11 +15,18 @@ class FakeDataGenerator
         username: user_name
     end
 
-    @stories_count.times do
+    @stories_count.times do |i|
       user = users[Random.rand(users.length-1)]
       title = Faker::Lorem.sentence(3)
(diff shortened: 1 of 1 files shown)


lib/tasks/fake_data.rake
```diff
@@ -15,19 +15,26 @@ class FakeDataGenerator
...
+      end
     end
   end
 end
 
 desc 'Generates fake data for testing purposes'
 task fake_data: :environment do
   fail "It's not intended to be run outside development environment" unless Rails.env.development?
-  unless (User.count + Tag.count + Story.count) == 0
+  unless User.count == 1 and Tag.count == 1 and Story.count == 0
```
pushcx: I don't see a reason to change this behavior, and definitely not as part of this work.

ponyta:
@pushcx fixed; PTAL

pushcx:
Thanks, nice to start checking things off in 414.

PR #432, opened 2017-11-13
Fix broken downvote menu on bad comments

sometoby:
Here's a screenshot of the broken downvote menu:

bug

The reduced opacity from comments with the 'bad'-class affects the
stacking order of its child elements and breaks the downvote menu. This
is avoided by appending the downvote reason menu below the comment
instead of inside it.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -115,22 +115,25 @@ var _Lobsters = Class.extend({
       d.append(a);
     });
 
-    if (thingType == "story")
-      li.after(d);
-    else
+    if (thingType == "story") {
       $(voterEl).after(d);
-
-    d.position({
-      my: "left top",
-      at: "left bottom",
-      offset: "-2 -2",
-      of: $(voterEl),
-      collision: "none",
-    });
-
-    /* XXX: why is this needed? */
-    if (thingType == "story")
+      d.position({
+        my: "left top",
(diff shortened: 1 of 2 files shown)


talklittle:
Does this break the y-offset of the popup menu for long comments?

In your screenshot example, changing "OpenBSD sucks!" to a multiline comment instead, I think you'll see the popup menu appearing too far down instead of directly below the downvote arrow.

sometoby:
Unfortunately it does break the y-offset. I hacked up and pushed a new fix that should work properly.

talklittle:
Cool. Does the new commit still preserve the placement of the "flag" popup on stories? (With and without submission text)

Looks like you can also remove the old `XXX` code below, since your new code does the same thing.

sometoby:
It does now. Did I catch everything this time?

talklittle:
Can't think of anything else. Just make sure it's tested on mobile too (e.g. using Firefox's Responsive Design Mode)

sometoby:
I found two more problems unrelated to my changes:

screen shot 2017-11-14 at 15 18 40

I tested the desktop and mobile view for both stories and comments of varying text lengths with the new changes, everything seems to work fine now.

pushcx:
Thanks for digging into this fiddly set of little bugs!

sometoby:
My pleasure! Glad I had an opportunity to contribute.

PR #435, opened 2017-11-20
New story URL validation update

pgl:
- should now support custom ports (eg http://example.com:443/)

----

I tried submitting the following URL:

 * https://er.educause.edu:443/blogs/2017/10/time-for-password-expiration-to-die

Unfortunately I received an error saying that it wasn't a valid URL:

image

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -67,7 +67,7 @@ class Story < ActiveRecord::Base
     if self.url.present?
       check_already_posted
       # URI.parse is not very lenient, so we can't use it
-      unless self.url.match(/\Ahttps?:\/\/([^\.]+\.)+[a-z]+(\/|\z)/i)
+      unless self.url.match(/\Ahttps?:\/\/([^\.]+\.)+[a-z]+(:\d+)?(\/|\z)/i)
         errors.add(:url, "is not valid")
       end
     elsif self.description.to_s.strip == ""


jcs:
I don't think arbitrary ports should be allowed, since the Lobsters server has to fetch content from them and it shouldn't be poking at random ports.  Also, specifying https:// and :443 is redundant.

pgl:
I've got plenty of bookmarks for webservers running on arbitrary ports. And you're right, it's redundant - but it fetches fine, because the URL fetcher uses a proper URL validation regex. :)

pushcx:
@pgl Thanks for running this down and opening a PR. I agree with jcs that I don't want the lobsters server hitting random ports on arbitrary servers. So there's two approaches for this PR:

1. Check the regexp to filter for common http ports (80, 443, 8000, and 8080 come to mind; definitely not 3000 as I see that used for dev instances).
2. Or leave the regexp as-is and expand the PR to have the title fetcher only invoked on well-known ports.

I'd prefer the latter (I have also seen webservers on odd ports) but would be happy to merge the easier thing now.

pgl:
I don't have much say here, but I thought I'd add my opinion anyway: I don't really see the problem with accessing random ports. They're just ports: the potential for abuse is just as great with port 8000 etc.

I think rate-limiting the form would be a better way of preventing potential abuse, if that's the issue.

nickhammond:
@pgl Fetching random ports is going to be a headache to support for future operations and since Lobsters focuses on sharing links that are accessible via the web whitelisting to those common ports would be preferred.

There's a comment about URI.parse not being very lenient in the Story model, does anyone recall a few examples? 

Something along the lines of this would support http, https, 80 & 443.

```
url = URI.parse(url) rescue false
url.kind_of?(URI::HTTP) || url.kind_of?(URI::HTTPS)
```

Source: https://coderwall.com/p/ztig5g/validate-urls-in-rails

pushcx:
This PR petered out, so I addressed it in 49e678ea3919a8b56f5d1c3bc840c94881eb9059, adding a bunch of tests of the various clever things `Story` does to `url`.

PR #436, opened 2017-11-23
Add light styles for code/pre html elements

xkpb:
I always found it hard to discern code samples from normal text; this
commit adds a faint grey background to both inline code elements and
code sections.


pushcx:
Hey, thanks for taking the initiative on a pet peeve. I see how this makes it clear where the code block is, but it also makes it the most visually heavy thing on the page, which distracts from regular reading and skimming. Could you talk more about the problem you're having and we'll see if there's an approach without that tradeoff?

xkpb:
@pushcx sure! The issue I was trying to solve is that simply using a monospace font was not immediately recognizable as code due to the fact that lobsters uses a small font to begin with. Blockquotes are much more recognizable as a different type of text due to a vertical bar and italics, so I was trying to bring equal distinction to code, but may have been over-zealous in my approach. :)

nickhammond:
#eee seems to be a bit strong, what about making it a bit lighter? Github uses #f6f8fa for instance. Another option being drop the background and just have a light outline.

#eee:
screen shot 2018-01-17 at 8 07 21 am

#f6f8fa:
screen shot 2018-01-17 at 8 07 07 am

xkpb:
The light version looks great to me! I can make that change on the PR
sometime today or tomorrow.

pushcx:
Alternately, a light bar to the left like blockquote's heavy bar might be a very nice display. What looks good to you?

talklittle:
@pushcx Doesn't work for inline code

pushcx:
I tried copying the `blockquote` styling to `pre` and tinkering a bit and it never stopped looking off. Without a good alternative and strong demand, I'm going to close this PR.

PR #438, opened 2017-12-19
comments: hide top-level sibling tree lines

talklittle:
since they make it easy to confuse siblings for replies

-----

Example **before:**

User `test2` is replying to the original submission, but the tree lines connecting the top-level comments make it look like a reply to the first comment instead, causing unnecessary confusion.

screenshot-2017-12-18 something to discuss

Example **after:**

screenshot-2017-12-18 something to discuss 1


pushcx:
Nice little improvement, thank you.

Issue #317, opened 2016-07-24
Feature request: tag-scoped search


pushups:
I think being able to restrict searches to specific tags would be a very helpful feature. I did find a few old threads asking for it:
https://lobste.rs/s/gg4d7o/ask_jcs_can_we_get_tag_search
https://lobste.rs/s/uprcao/feature_request_narrow_limit_searching

I'm particularly interested in being able to search ask. There have been a few posts recently asking about tools the community uses, and books that people have found helpful that would be great to go back to if I happen not to bookmark them immediately. Of course I think the utility extends way beyond that.

I'm willing to look into building this if it would be a useful feature, and I'm not duplicating work.

pushcx:
This would be very straightforward to add to `Search#search_for_user!` `app/models/search.rb` now that the site searches via SQL. I'd like to see syntax like `tag:python django' to search for stories about "django" tagged with "python".

pushups:
I'll take a stab at implementing this.

pushups:
Closing because this is merged!

pushcx:
Thanks, @pushups. I have already used this on prod to find a story. :)

PR #439, opened 2017-12-21
search: add tag scoped search

pushups:
When searching for a story a user can include a single tag:<tag_name>
element which will limit the scope of the search to that tag name.

This works similarly to the domain:<domain_name> feature which
limits the scope of the search to stories from that domain.

Examples:
"tag:python django" as a search query will search for stories
tagged with the python tag which reference the word "django".

"tag:python" will return all of the stories tagged with python
in the search results - which is essentially the /t/python page
but you can now order by Newest or Points as well.

Technical details:
* Added query parsing at the beginning of the search
to set a tag_scope string.

* When a tag_scope is set the relation gets another where clause
for only the tags that match exactly that scope.

* Modified the relevance queries to use literal SQL for the select
statements and the ranked ordering. See: title_match_sql, etc.
When the query was limited to tags ActiveRecord was re-writing
it by substituting the ORDER clauses into the columns for selecting
and the 'AS rel_title' was getting dropped. This caused the query
to fail.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -53,18 +53,26 @@ class Search
     end
   end
 
+  def with_tags(base, tag_scopes)
+    base
+      .joins({ :taggings => :tag }, :user)
+      .where(:tags => { :tag => tag_scopes })
+      .having("COUNT(stories.id) = ?", tag_scopes.length)
+      .group("stories.id")
+  end
+
   def search_for_user!(user)
     self.results = []
     self.total_results = 0
 
     # extract domain query since it must be done separately
     domain = nil
-    tag_scope = nil
+    tag_scopes = []
     words = self.q.to_s.split(" ").reject{|w|
       if m = w.match(/^domain:(.+)$/)
         domain = m[1]
       elsif m = w.match(/^tag:(.+)$/)
-        tag_scope = m[1]
+        tag_scopes << m[1]
       end
     }.join(" ")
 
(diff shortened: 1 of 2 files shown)


pushups:
Forgot to mention that this is meant to address: https://github.com/lobsters/lobsters/issues/317

pushcx:
This is solid and I hope to merge it soon. Thank you!

I do see two bugs related to multiple tags. Imagine there's a story "test story" tagged "foo" and "bar".

1. If I search for "test tag:foo", the story is correctly listed, but only shows the "foo" tag, not also "bar".
2. If I search for "test tag:foo tag:bar", it only searches for stories with the "bar" tag instead of stories tagged "foo" and "both".

It'd also be great to have a couple tests - one of the basic functionality and one of searching for two tags not turning up a story with only one, and any other behavior you think would be valuable.

pushups:
@pushcx Thanks for taking a look at this! I'm working on the tests and bug fixes when I have some free time and hope to get it over to you probably in the next couple weeks.

pushups:
@pushcx I think this is ready for another review when you get a chance. 
* Tests passed.
* I added search_spec.rb and a few tests for regular search behavior, and also the tag specific behavior.
* I fixed the issues you pointed out: 
1. All tags are now displayed correctly
1. Searching for `baz tag:foo tag:bar` gives you stories with `foo` and `bar` tags that include the term `baz`.

pushcx:
I'm sorry I lost track of this PR. I'm glad to merge and deploy it. Thanks, @pushups!

pushups:
@pushcx I'm really excited to use this feature. Thanks for merging it!

PR #441, opened 2018-01-03
submit: fix resubmitting old URLs

### LINKED ISSUE #440: I can't resubmit link
I am trying to resubmit this link since they added the new 2017 videos of the conference and it is not possible:
http://bangbangcon.com/recordings.html

talklittle:
also now only requires hitting Submit a single time,
instead of twice, if you already see the AJAX warning

Fixes #440


pushcx:
Thanks for solving this, @talklittle.

PR #445, opened 2018-01-20
Set `Cache-Control` header for the assets

avinassh:
Since we use asset digests in the production environment, I think its beneficial if we also set max allowed Cache Control headers.

Reference: http://edgeguides.rubyonrails.org/asset_pipeline.html#cdns

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -32,6 +32,11 @@ Lobsters::Application.configure do
   # Generate digests for assets URLs.
   config.assets.digest = true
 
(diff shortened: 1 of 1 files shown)


pushcx:
We've got nginx serving assets directly with `Cache-Control: public` at the moment, so this wouldn't have an effect. I'm planning to get that nginx conf into the lobsters-ansible repo in the next day or two. A PR there that sets `max-age` on the assets with content hashes would be very welcome.

PR #446, opened 2018-01-20
Make SMTP settings configurable from the env

avinassh:
The changes made here does not affect how the older SMTP settings were configured. I have kept the earlier values as default and code won't break even if none of the environment values is not set.

I am curious how main lobsters site is running in production. Does this file is manually overwritten during deployment?


pushcx:
This looks totally reasonable, thanks. In prod, postfix is listening to the local port.

PR #449, opened 2018-01-26
Avoid require memory leak with pathnames in load path

jrafanie:
Any ruby requires that occur in your application will leak with
Pathnames in the $LOAD_PATH in ruby 2.3/2.4.  Beyond that, it's faster
to use strings in the load path.  Please join this bug and add your
support for a fix:

https://bugs.ruby-lang.org/issues/14372

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,7 +16,7 @@ module Lobsters
     #config.autoload_paths += %W(#{config.root}/extras)
 
     # Eager load everything in extras/
-    config.eager_load_paths << Rails.root.join("extras")
+    config.eager_load_paths << Rails.root.join("extras").to_s
 
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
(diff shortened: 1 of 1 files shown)


jrafanie:
screencast 2018-01-26 11-35-31

pushcx:
Thanks for this small workaround. I'd just bump to Ruby 2.5, but Ubuntu doesn't yet have a package for the LTS we're on (xenial).

jrafanie:
@pushcx Good point.  Honestly though, there are optimizations/caching done for string $LOAD_PATH objects that aren't done for pathname so even on 2.5.0, [it's faster to use strings in the load path](https://github.com/ruby/ruby/blob/v2_3_6/load.c#L43-L47).

pushcx:
Oh, cool. Thank you for pointing that out, I'm happy to have a little perf improvement.

PR #453, opened 2018-02-02
Fix migrations from new instance

jstoja:
I just tried to deploy a lobsters instance from the ground up and the migrations are failling.
I found this issue: https://github.com/RolifyCommunity/rolify/issues/444 and modifying the migration files, fixed the migrations.

I think this needs the confirmation from someone that knows a bit more Rails than I do.


pushcx:
The normal way (noted by Rails at the top of `db/schema.rb`) to start a new instance is to load the schema, not run all the migrations. They're just supposed to be used by people with existing databases to update, and they're working for that.

PR #455, opened 2018-02-03
Feature: Configurable open signups

avinassh:
This PR lets you have complete open sign ups and is configurable with an env var. Set `OPEN_SIGNUPS=true`, any other string or absence of this env will be considered as false. So, existing Lobsters instance will run as expected. 

Changes:
1. If `OPEN_SIGNUPS` is not set or set to any value other than `true`, app works like earlier.
2. If `OPEN_SIGNUPS` is set to `true`, in the header there will be an item `Sign Up Now` and in footer `Join <site name>`, they both point to `/invitations/open`. 
3. I added a new CSS class `signup-text` which can be used to customize `Sign Up Now` text in header.

### DIFF
--- a/app/controllers/signup_controller.rb
+++ b/app/controllers/signup_controller.rb
@@ -7,7 +7,9 @@ class SignupController < ApplicationController
       flash[:error] = "You are already signed up."
       return redirect_to "/"
     end
-
+    if Rails.application.open_signups?
+      redirect_to action: :invited, invitation_code: 'open' and return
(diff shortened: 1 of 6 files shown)


avinassh:
Screenshot:

screen shot 2018-02-03 at 2 39 24 pm

avinassh:
@pushcx any update on this PR?

pushcx:
Thanks!

PR #463, opened 2018-03-05
fix migration datetime value

### LINKED ISSUE #462: Broken migration "20180131203555_fix_doffed_at.rb" on MySQL 5.7.21
`ActiveRecord::StatementInvalid: Mysql2::Error: Incorrect datetime value: '0000-00-00 00:00:00' for column 'doffed_at' at row 1: UPDATE `hats` SET `hats`.`doffed_at` = NULL WHERE (doffed_at = '0000-00-00 00:00:00')`

talklittle:
fixes #462


pushcx:
After 826c88f I'm really ready to be done with this migration (or mysql itself).

PR #476, opened 2018-03-28
settings: account deletion inputs initially disabled

### LINKED ISSUE #466: Prefilled password fields for "delete my account"
In Firefox, the `autocomplete="off"` for "Delete Account" at https://lobste.rs/settings is ignored, which lets it prefill the password, making account deletion just a single click away.

I'd propose to rename the field so this doesn't happen, or to show a popup where one needs to enter the password (ala Twitter).

talklittle:
prevents Firefox autofilling password for deletion

fixes #466

### DIFF
--- a/app/views/settings/index.html.erb
+++ b/app/views/settings/index.html.erb
@@ -1,3 +1,15 @@
+<script>
+  $(document).ready(function() {
+    $("#delete_account_checkbox").change(function(e) {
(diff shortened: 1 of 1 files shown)


pushcx:
I don't want to make this form depend on JavaScript. If the user agent fills, this is their feature and not our bug.

PR #477, opened 2018-03-28
settings: "I am sure" checkbox to delete account

talklittle:
Alternative to #476 not requiring JavaScript.

Meant to address issue #466.


pushcx:
Thanks, this is a solid fix.

PR #479, opened 2018-04-08
Fixed padding issue for some dialogs on mobile devices

phikal:
Pages like [search](https://lobste.rs/search) don't pad certain parts of the page on mobile devices (at least my browser), as can be seen in the following page:

alter

My patch just adds the same margin as `.box`'s receive on mobile devices, to instances of `.help`. This means that the desktop UI stays the same. The only thing I had to do, was to add the `.help` tag to `div`s outside of a `div.box`, so that they also get the necessary padding. This might slightly change the appearance of some boxes, if this is a problem, I can try and find another way to fix it.\

**Note:** I couldn't test this "properly" since I don't have all the rails stuff installed, and basically have no idea what is necessary _to be installed_. My suggestions are purely based on playing around with the inspection tools on Firefox and Chromium.

### DIFF
--- a/app/assets/stylesheets/mobile.css
+++ b/app/assets/stylesheets/mobile.css
@@ -49,7 +49,7 @@
 		margin: 0;
 	}
 
-	.box {
+	.help, .box {
 		margin: 0 8px;
 	}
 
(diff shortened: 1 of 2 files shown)


pushcx:
@phikal Thanks for a great little fix!

PR #483, opened 2018-04-26
add link to '/tags' on search page

david-wolgemuth:


pushcx:
Thanks for taking a look at this, I'm glad to point people to a list of all their options. Why the `target="_blank"`, though? I don't see a compelling value - searches are generally short and there's no clever js on this page, so anything entered wouldn't be lost by going forward and back to the search form.

Issue #480, opened 2018-04-17
Don't delete invitations


pushcx:
Invitations are deleted when they are used or the sender is banned. I've been playing whack-a-mole with a spammer who likes to invite himself extra accounts to upvote his spam. Seeing the record of invites would've saved me some time picking out his accounts and ignoring the similar accounts that were legitimately registered at the same time.

I'd like to have a `used_at` datetime to mark the use (and prevent it from being used twice), and `new_user_id` linking to the new `User` created.

cosmicbuffalo:
Hi Peter, I just submitted a quick fix for this issue in PR #485, let me know if it needs any changes. I wasn't sure if I should put a test in there or not since it's super basic and I didn't see any other existing invitation model tests.

pushcx:
A test would've been nice, but I'm happy to merge it without one as the spammer who prompted this has tried twice to get back onto the site. I added the `new_user` tracking.

PR #485, opened 2018-05-01
Used Invitations

cosmicbuffalo:
- adds used_at datetime to Invitation model
- adds `used`/`unused` scopes based on presence of the `used_at` field
- uses the `unused` scope when looking up valid invitations
- `@invitation.destroy` changes to `@invitation.update(used_at: Time.current)`


cosmicbuffalo:
Addresses issue #480

pushcx:
Thanks for implementing this! I didn't expect anyone to pick this up, so it was a delight to see this PR. I landed this in c371ca2.

Issue #277, opened 2016-03-31
support hats on private messages


jcs:
mostly for moderators to send messages with or without that

PR #489, opened 2018-05-03
Add hats to private messages

david-wolgemuth:
Users with hats can add a hat to a private message.  Hat is visible in inbox next to username.


pushcx:
Thanks, this is a great PR. I slightly tweaked the tests to make it clearer that they're checking validity.

Close #277

Issue #482, opened 2018-04-24
Fix config/routes.rb specifying `scope :format => "html"`


pushcx:
A user asked for a list of stories they've upvoted, preferably in json. I pointed out the `/upvoted` endpoint, but `/upvoted.json` was throwing 404s.

I worked around this in 9eae3e3c9daed5b52aaa60f5835c91542a951037 but it looks like this has been the case since 981759c34ff6eeb629b0cb54b029b1fc0001a611 in Jan 2014. I'm puzzled by this commit, because Rails doesn't try to do anything to support other formats unless there's a `(.format)` in the Route and code in the controller. I think this `scope` can be simply removed, but I don't have the attention to read more history for clues and test a fix now so I'm writing up a quick issue.

pushcx:
Removed in d24e123 after discussion in #490.

pushcx:
I realized a minute later it was silly not to clean up the redundant format requirements while it was fresh in my mind, so I did that in a59cf79.

NinjaTrappeur:
Hey, I was checking out this PR, it actually introduce a "weird" behaviour: in case the controller cannot render the requested format, instead of returning a 404 (previous behavior), it will now return an empty template.

I don't think this is a big deal, just wanted to warn you. See https://lobste.rs/upvoted/page/1.wut .

We could prevent that by keeping the scope and changing its value with the array [html, json,rss]. Again, I'm not a rails dev, this could be wrong.

pushcx:
I get an HTTP 406, so at least the header's right. I'm not sure under what other conditions Rails might throw 406 so I hesitate to share that template between the two. I'll leave it for now.

PR #490, opened 2018-05-08
Exposing most pages as JSON and RSS

NinjaTrappeur:
Hi,

This PR expose most of lobsters routes as HTML, JSON and RSS. I basically just added both the JSON and RSS formatters to every relevant routes. I haven't modified those for which the HTML formatter is not the default one.

Note: I did not refactored the routes, a lot of legacy/duplicate one seems to still be present.

I needed to test this change locally and had some trouble setting up the project on my system. I added a shell.nix file which will help future nix users to setup their dev environment and prevent most of the struggles I gone though.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,38 +9,38 @@ Lobsters::Application.routes.draw do
     get "/rss" => "home#index", :format => "rss"
     get "/hottest" => "home#index", :format => "json"
 
-    get "/page/:page" => "home#index"
+    get "/page/:page" => "home#index", :format => /html|json|rss/
 
     get "/newest" => "home#newest", :format => /html|json|rss/
-    get "/newest/page/:page" => "home#newest"
+    get "/newest/page/:page" => "home#newest", :format => /html|rss|json/
     get "/newest/:user" => "home#newest_by_user", :format => /html|json|rss/
-    get "/newest/:user/page/:page" => "home#newest_by_user"
-    get "/recent" => "home#recent"
-    get "/recent/page/:page" => "home#recent"
-    get "/hidden" => "home#hidden"
-    get "/hidden/page/:page" => "home#hidden"
+    get "/newest/:user/page/:page" => "home#newest_by_user", :format => /html|json|rss/
+    get "/recent" => "home#recent", :format => /html|rss|json/
+    get "/recent/page/:page" => "home#recent", :format => /html|rss|json/
+    get "/hidden" => "home#hidden", :format => /html|rss|json/
+    get "/hidden/page/:page" => "home#hidden", :format => /html|rss|json/
     get "/saved" => "home#saved", :format => /html|rss|json/
-    get "/saved/page/:page" => "home#saved"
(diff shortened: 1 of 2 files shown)


pushcx:
Hey, thanks for taking a look at the source of some 404s. The last time this came up I wrote up issue #482 to remove the html `scope` to remove all the workaround `:format`s, rather than add more of them. Did you try that and get some kind of failure out of it?

As for nix - I don't mind adding it, but (like the Vagrantfile) I don't use it to know how to maintain it and would expect it to break when we update Ubuntu for the new version of Ruby. Maybe add a warning?

NinjaTrappeur:
No, this is actually the first time I ever write ruby, I was a bit
afraid of breaking anything :).

I'll have a look at this scope by the end of the week. I'll also add a
warning as well as my handle on the shell.nix file.

On Tue, May 8, 2018, at 21:09, Peter Bhat Harkins wrote:
> Hey, thanks for taking a look at the source of some 404s. The last
> time this came up I wrote up issue #482[1] to remove the html scope to
> remove all the workaround :formats, rather than add more of them. Did
> you try that and get some kind of failure out of it?> As for nix - I don't mind adding it, but (like the Vagrantfile) I
> don't use it to know how to maintain it and would expect it to break
> when we update Ubuntu for the new version of Ruby. Maybe add a
> warning?> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/lobsters/lobsters/issues/482
  2. https://github.com/lobsters/lobsters/pull/490#issuecomment-387510180

NinjaTrappeur:
So, I just made a bit of archeology in the git logs :)

The html scope has been here for a long time, actually since 981759c34ff6eeb629b0cb54b029b1fc0001a611 (log: "be explicit about which route formats are accepted). There hasn't been any major rewrite of the routes since then. The commit you mentionned in #482 was actually just a revert which confused git somehow. The html scope was already there.

I guess we could add json and rss to all the routes, but, according my very limited understanding of rails, we still need to implement these renderers in the controllers.

Some controllers seems to be already implementing those formats even though they were not exposed on the routes. For instance, the /upvoted json and rss controller renderers have been introduced in ab8f8f3a , but it seems like the author forgot to add the associated routes.

So here's a potential plan:

1. Check I did not screw things up: check the controllers of the routes I exposed JSON and RSS for are actually handling those formats. Then we could merge this PR: I need the paginated /newest data.
2. On another PR, check we can safely write json formatted for the POST actions. I did not modified these routes in this PR.
3. On arother PR, check the duplicate routes and see if we can decommission them: the routing scheme seems a bit messy.

pushcx:
Thanks for digging in, this gave me the confidence to tackle #482. I ran `rails routes > pre`, removed the scope, and ran `rails routes > post`. Diffing the two files showed the only thing that changed was the removal of the redundant `format` on the routes.

NinjaTrappeur:
Nice! Thanks.

Issue #491, opened 2018-05-13
Comment search doesn't use tag/domain scope. 


pushcx:
I did [a comment search](https://lobste.rs/search?utf8=%E2%9C%93&q=federation+tag%3Ameta&what=comments&order=newest) for "federation tag:meta" and found a bunch of comments on stories that are not tagged `meta`. A similar search with a `domain:` also won't limit the search to comments on stories from the given domain.

pushcx:
The functionality was merged in #502, but a test will come in #501, so this issue is still open.

pushcx:
With #501 merged this is done. Thanks, @jtopgi and @david-wolgemuth.

PR #501, opened 2018-06-11
filter by tag and domain when searching for comments

david-wolgemuth:
Fix to issue [#491 Comment search doesn't use tag/domain scope.](https://github.com/lobsters/lobsters/issues/491).

PR #502, opened 2018-06-12
Comment search doesn't use tag scope.

jtopgi:
This fixes the tag search for the issue: #491. Also enable byebug for development.


pushcx:
Hey @jtopgi, great to see another contribution from you! @david-wolgemuth also attacked this in #501 at the same time, which is a first for Lobsters! So there's a bit to untangle here.

After reviewing both and chatting with David about an alternate approach for domain search, I'm going to merge this PR when it's ready. To get ready, this needs to remove the code changes for byebug and to `schema.rb` (@hmadison here's that scenic bug/misconfig that keeps cropping up). A PR needs to be focused on solving an issue or implementing a feature so it's a coherent unit of work.

Also, I wouldn't be merging this if it didn't have tests, because it looks like a feature that will be easy to break in the future. Luckily David wrote some over in #501 and he's agreed to include them in his upcoming revamp of `Story` domains + searching them.

So, thanks again for your contribution - as soon as this is focused on the changes to `Comment` and `Search` to enable tag searching, I'll merge this PR.

jtopgi:
Thanks, I made your suggested changes. I'm trying to write the test but for some reason, when I use rspec, test data persists even after the test is done. This is causing me difficulty in writing my tests. I tried methods in clearing the cache like adding `config.cache_store = :null_store` to the environment files but that didn't work. Any help would be appreciated.

david-wolgemuth:
I was running into the same issue when I wrote tests for the comment search.  It was a dependent-destroy issue, (can't destroy user, because comments require a user).  To fix it I reversed the order of the destroys.

```ruby
after :all do
  @comments.each(&:destroy!)
  @messages.each(&:destroy!)
  @user.destroy!
end
```

jtopgi:
@david-wolgemuth do you know a way to clear out RSpec? After every failed test, I get the error the `User is already defined`. This is persistent throughout the tests. The only way I know how is to reset the database. Is there a quicker way?

david-wolgemuth:
I think you could run `RAILS_ENV=test rails db:console` then delete the user?

pushcx:
It's strange to me that these issues are happening, given that `spec/rails_helper.rb` sets `config.use_transactional_fixtures = true`.

@jtopgi, it looks like this commit missed out on the replacement of Machinist with FactoryBot in #506. If you check out your branch and run `git merge master` it'll pull in those changes, probably complain about merge conflicts, and then you can have a passing test suite. But if it's easier, you can just remove your test code and I'll let @david-wolgemuth handle it in #501 as mentioned before.

pushcx:
It's strange to me that these issues are happening, given that `spec/rails_helper.rb` sets `config.use_transactional_fixtures = true`.

@jtopgi, it looks like this commit missed out on the replacement of Machinist with FactoryBot in #506. If you check out your branch and run `git merge master` it'll pull in those changes, probably complain about merge conflicts, and then you can have a passing test suite. But if it's easier, you can just remove your test code and I'll let @david-wolgemuth handle it in #501 as mentioned before.

pushcx:
Thanks, @jtopgi. This is a solid solution and I'm glad to merge it.

PR #504, opened 2018-06-13
add byebug to development

david-wolgemuth:
@jtopgi and I (and others I assume) use byebug for development / testing.


pushcx:
Happy to add useful tools that make things easier for contributors, thanks for this!

PR #510, opened 2018-06-13
Use `scenic-mysql_adapter` over `scenic-mysql`

hmadison:
[scenic-mysql](https://github.com/rennanoliveira/scenic-mysql) hasn't been updated in a while and is missing code [to remove the table name from the view in the schema dump](https://github.com/EmpaticoOrg/scenic-mysql_adapter/blob/master/lib/scenic/adapters/my_sql.rb#L86). This should resolve:

* The db/schema.rb churn for new contributes
* Issues with setting up a new database from the schema

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem "mysql2", "~> 0.3.20"
 # gem "pg"
 
 gem 'scenic'
-gem 'scenic-mysql'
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks for solving this persistent annoyance.

@david-wolgemuth, @jtopgi, I know you've both run into this recently.

Issue #509, opened 2018-06-13
Turn on rails 5.2 defaults after upgrade


thomasdziedzic-pd:
Restart after each item:

- [x] Rails.application.config.active_record.cache_versioning = true
- [x] Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
- [x] Rails.application.config.active_support.use_authenticated_message_encryption = true
- [x] Rails.application.config.action_controller.default_protect_from_forgery = true
- [x] Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
- [x] Rails.application.config.active_support.use_sha1_digests = true

pushcx:
Also:
 - [ ] [set `config.load_defaults`](https://github.com/lobsters/lobsters/pull/498#discussion_r195209699)

And as an implementation note, if someone wants to tackles this, you certainly don't have to do all of them. I'd be very happy to merge PRs for these one-by-one.

pushcx:
I thanked Thomas in #528, but I want to note for any future readers of this issue that Thomas drove this migration and checked everything off the list in his many PRs.

PR #516, opened 2018-06-20
enable default_protect_from_forgery and use_authenticated_message_encryption

thomasdziedzic-pd:
https://github.com/lobsters/lobsters/issues/509

from http://guides.rubyonrails.org/configuring.html

> config.action_controller.default_protect_from_forgery determines
> whether forgery protection is added on ActionController:Base. This is
> false by default, but enabled when loading defaults for Rails 5.2.

rails pull request: https://github.com/rails/rails/issues/29193

### DIFF
--- a/config/initializers/new_framework_defaults_5_2.rb
+++ b/config/initializers/new_framework_defaults_5_2.rb
@@ -25,7 +25,7 @@ Rails.application.config.active_record.cache_versioning = true
 
 # Add default protection from forgery to ActionController::Base instead of in
 # ApplicationController.
-# Rails.application.config.action_controller.default_protect_from_forgery = true
+Rails.application.config.action_controller.default_protect_from_forgery = true
 
 # Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
 # 'f' after migrating old data.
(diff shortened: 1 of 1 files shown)


thomasdziedzic-pd:
I didn't know if I should remove the `protect_from_forgery` line at https://github.com/lobsters/lobsters/blob/master/app/controllers/application_controller.rb#L2 since it uses a different option.

The default is `:null_session` as explained here: http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html#method-i-protect_from_forgery

But removing the line in the application controller would switch the method to `:exception` which is the default method generated in rails now.

Relevant docs:

> Valid unverified request handling methods are:
> 
> :exception - Raises ActionController::InvalidAuthenticityToken exception.
> 
> :reset_session - Resets the session.
> 
> :null_session - Provides an empty session during request but doesn't reset it completely. Used as default if :with option is not specified.

http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html#method-i-protect_from_forgery for full details.

thomasdziedzic-pd:
Added another commit which should enable use_authenticated_message_encryption. I realized that this can be safely enabled since lobsters doesn't use the MessageEncryptor class.

thomasdziedzic-pd:
I opened a rails issue for the outdated configuration docs: https://github.com/rails/rails/issues/33170

pushcx:
Hey @thomasdziedzic-pd, thanks for this one. I'm moving slowly on this one because it's security-sensitive and I haven't been able to schedule a couple distraction-free hours to give this the deliberative attention it deserves.

thomasdziedzic:
Do you have any questions that I could help you out with this PR? Would like to continue working on some other things that are blocked by this PR

pushcx:
Spent time this morning reading the docs and kicking the tires, looks good.

PR #526, opened 2018-08-14
enable use_authenticated_cookie_encryption

thomasdziedzic:
Fixes all of the individual configuration settings on https://github.com/lobsters/lobsters/issues/509
The next step will be to enable 5.2 defaults and remove these initializer files.

from the comments:
```
Use AES-256-GCM authenticated encryption for encrypted cookies.
Also, embed cookie expiry in signed or encrypted cookies for increased security.

This option is not backwards compatible with earlier Rails versions.
It's best enabled when your entire app is migrated and stable on 5.2.

Existing cookies will be converted on read then written with the new scheme.
```


thomasdziedzic:
I also went ahead, and cleaned up the rubocop warnings introduced in the previous few commits. I also enabled automatic schema migration for the test environment since it's a feature built into rails now.

pushcx:
Looks good in local testing, thank you.

PR #535, opened 2018-08-22
Send Webmentions

### LINKED ISSUE #383: Feature: Send Webmentions
### Idea

In the lights of the very good discussions that take place on Lobsters and the problem with discovering that discussions if one isn't aware of it haven taken place on Lobsters I think it would be interesting if support could be added for sending the submission of the link as a [Webmention](https://webmention.net/) for any site that advertises that it accepts Webmentions.

That way the original site will know about the discussion going on here and people from elsewhere will be aware of the discussion. (Whether that is wanted or not is another thing, this feature suggestion assumes that more people finding the discussions would generally be a good ting)

### Implementation

As Lobsters already fetches and parses a page for the existence of eg. a rel-canonical, adding a check for a rel-webmention as well would be easy.

To then send a POST request with the Lobster page as "source" and the submitted page as a "target" should be easy (although maybe one wants to limit it to when it has received some kind of activity rather than just a plain submission).

### Icing on cake

An icing on the cake would be if any comment to that submission could then also trigger Webmention pings (as a simple variant of [Salmentions](https://indieweb.org/Salmentions), an advanced way to make webmentions travel upstream when used in a microblog/social scenario of indie-sites). That way the entire comment tree could be presented on the Webmention accepting page for those who accept and show full comment trees.

To make the icing on the cake work and to make the initial Webmention a bit clearer about what mention it actually is, it would be good to mark up the bookmark with [Microformats](https://indieweb.org/h-entry): http://indieweb.org/bookmark#How_to_markup And the comments as: http://indieweb.org/comments#How_to_markup

That's not necessary for Webmentions in itself, but the [IndieWeb](http://indieweb.org/) community that has spearheaded Webmentions suggests that one does it, to convey as much useful information as possible. Sending a Webmention without Microformat data is better than sending none though, one can then use other metadata to present it in a nice way.

NinjaTrappeur:
Hi!

This PR implements the feature discussed [here](https://github.com/lobsters/lobsters/issues/383#issuecomment-411827459).

The general idea was to be as little invasive a possible. 

# Changelog

I made some changes to the initial plan.

I ended up not storing the webmentions endpoint in the database: we won't have the webmention endpoint synchronously as we need to download and parse the link page. We also can't put this webmention discovery in the fetchAttributes function as I first intended to do: there's no way to retrieve the content downloaded by this function when saving the story.

So instead, everything is done in a cron-triggered task. I think this is a better design, everything stays compartmentalized.

The task:

* For any new story.
  * Download the web-page using Sponge.
  * Parse the web-page and look for a web-mention endpoint. The regexes used here are extracted from the [official webmention-client gem](https://github.com/indieweb/webmention-client-ruby). I did not directly used this gem: it was impossible to use the parser without the whole client.
  * Send a webmention to this endpoint (if any).

I also added some markup necessary to make the repost parseable by the webmention endpoint.

The markup has been tested against the official validator. The parser has been manually tested against the webmentions.rocks examples. The post method has been manually tested.

# Potential Improvement

* We could markup the comments, it would allow the mentioned websites to retrieve them.
* We could send a new webmention on every comment, it would let know the mentioned websites it needs to parse again the comment page in order to retrieve the new comments.

I tried to markup the comments but did not find an immediate way to do so. The current template structure makes this tricky.

Fixes #383

I plan to maintain the changes introduced by this PR.

We'll also need to add this cron task to the ansible playbook. Should I open another PR there?

[Edit] Looks like the linter is angry. I'm fixing it.

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -7,7 +7,7 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
 <%= story.is_hidden_by_cur_user ? "hidden" : "" %>
 <%= story.is_saved_by_cur_user ? "saved" : "" %>
 <%= story.is_expired? ? "expired" : "" %>">
-<div class="story_liner">
+<div class="story_liner h-entry">
   <div class="voters">
     <% if @user %>
       <a class="upvoter"></a>
@@ -17,9 +17,9 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
     <div class="score"><%= story.score %></div>
   </div>
   <div class="details">
-    <span class="link">
+    <span class="link h-cite u-repost-of">
       <% if story.can_be_seen_by_user?(@user) %>
-        <a href="<%= story.url_or_comments_path %>"><%=
+        <a class="u-url" href="<%= story.url_or_comments_path %>"><%=
           break_long_words(story.title) %></a>
       <% end %>
       <% if story.is_gone? %>
@@ -73,7 +73,7 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
             <% else %>
               via
             <% end %>
-            <a href="/u/<%= ms.user.username %>" class="<%=
+            <a href="/u/<%= ms.user.username %>" class="h-card u-url <%=
               ms.html_class_for_user %>"><%= ms.user.username %></a>
 
             <%= time_ago_in_words_label(ms.created_at) %>
@@ -106,7 +106,7 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
         <% else %>
           via
         <% end %>
-        <a class="<%= story.html_class_for_user %>"><%=
+        <a class="h-card u-url <%= story.html_class_for_user %>"><%=
           story.user.username %></a>
         just now
       <% else %>
@@ -115,7 +115,7 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
         <% else %>
           via
         <% end %>
-        <a href="/u/<%= story.user.username %>" class="<%=
+        <a href="/u/<%= story.user.username %>" class="h-card u-url <%=
           story.html_class_for_user %>"><%= story.user.username %></a>
 
         <%= time_ago_in_words_label(story.created_at) %>
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -1,3 +1,8 @@
+<% if false %>
+  The h-entry, u-report-of, u-url and h-card css classes are necessary
+  for the webmentions support. See <http://microformats.org/wiki/h-entry>
+  for more details.
+<% end %>
 <li id="story_<%= story.short_id %>" data-shortid="<%= story.short_id %>"
 class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
 <%= story.vote && story.vote[:vote] == -1 ? "downvoted" : "" %>
--- a/extras/sponge.rb
+++ b/extras/sponge.rb
@@ -96,7 +96,8 @@ class Sponge
     end
   end
 
-  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10)
+  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10,
+            return_full_res = false)
     raise ArgumentError.new("http redirection too deep") if limit <= 0
 
     uri = URI.parse(url)
@@ -218,7 +219,11 @@ class Sponge
 
     case res
     when Net::HTTPSuccess
-      return res.body
+      if return_full_res
+        return res
+      else
+        return res.body
+      end
     when Net::HTTPRedirection
       # follow
       newuri = URI.parse(res["location"])
--- a/extras/sponge.rb
+++ b/extras/sponge.rb
@@ -96,8 +96,7 @@ class Sponge
     end
   end
 
-  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10,
-            return_full_res = false)
+  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10)
     raise ArgumentError.new("http redirection too deep") if limit <= 0
 
     uri = URI.parse(url)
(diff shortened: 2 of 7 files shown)


app/views/stories/_listdetail.html.erb
```diff
@@ -7,7 +7,7 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
 <%= story.is_hidden_by_cur_user ? "hidden" : "" %>
 <%= story.is_saved_by_cur_user ? "saved" : "" %>
 <%= story.is_expired? ? "expired" : "" %>">
-<div class="story_liner">
+<div class="story_liner h-entry">
```
pushcx: I'm sure in a year or two I'm going to forget what these extra classes are for, notice they have no rules in the css, and remove them. Could you add a comment with all four classes (so it shows up in any grep) and linking to the webmention docs?

extras/sponge.rb
```diff
@@ -96,7 +96,8 @@ def cookies(host)
     end
   end
 
-  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10)
+  def fetch(url, method = :get, fields = nil, raw_post_data = nil, headers = {}, limit = 10,
```
pushcx: This method already has too many arguments, let's not add one more. Let's return `res` and change existing call sites to `fetch(...).body`.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+require File.expand_path('../../config/boot', __FILE__)
+require APP_PATH
+Rails.application.require_environment!
+
+LAST_STORY_KEY = "webmentions:last_story_id".freeze
+
+def endpoint_from_body(html)
+  doc = Nokogiri::HTML(html)
+
+  if !doc.css('[rel~="webmention"]').css('[href]').empty?
```
pushcx: Can this turn into a single `return doc.at_xpath()`? I'm not sure if it supports regexp to combine these variations or if you'd have to [define a function](https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri%2FXML%2FSearchable:xpath). But I'd like to see something like that, it took me a bit of staring back and forth to recognize that only the `rel` value was changing between the different branches.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+    doc.css('[rel="http://webmention.org/"]').css('[href]').attribute('href').value
+  elsif !doc.css('[rel="http://webmention.org"]').css('[href]').empty?
+    doc.css('[rel="http://webmention.org"]').css('[href]').attribute('href').value
+  end
+end
+
+def endpoint_from_headers(header)
+  return unless header
+
+  if (matches = header.match(/<([^>]+)>; rel="[^"]*\s?webmention\s?[^"]*"/))
```
pushcx: Similarly, it's gotta be possible to make this two regexps. This could use some tests so there's samples of valid data, I'm puzzled by the shape of this being "url; webmation" *and* "webmention; url", that's a surprisingly flexible grammar.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+def send_webmention(source, target, endpoint)
+  sp = Sponge.new
+  sp.timeout = 5
+  sp.fetch(endpoint.to_s, :post, {
+    "source" => URI.encode_www_form_component(source),
+    "target" => URI.encode_www_form_component(target),
+  }, nil, {}, 3)
+end
+
+last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last.id).to_i
```
pushcx: This can wrap in `if __FILE__ == $PROGRAM_NAME` so you can test `endpoint_from_headers`.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+Story.where("id > ? AND is_expired = ?", last_story_id, false).order(:id).each do |s|
+  if s.url.present?
```
pushcx: `next unless s.url.present?` to save an indent and save me looking for an `else` :) Same with the other two `if`s.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+    begin
```
pushcx: Why `begin` with no `rescue`?

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+      sp = Sponge.new
+      sp.timeout = 3
+      response = sp.fetch(s.url, :get, nil, nil, {
+        "User-agent" => "#{Rails.application.domain} webmention endpoint lookup",
```
pushcx: Great way to make this obvious to other admins.

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+      }, 3, true)
+      if !response.nil?
+        wm_endpoint_raw = endpoint_from_headers(response['link']) ||\
+                          endpoint_from_body(response.body.to_s) || nil
```
pushcx: The `|| nil` can be removed. Ruby is very enthusiastic about implicitly suppling `nil`s. :grimacing:

script/send_new_webmentions
```diff
@@ -0,0 +1,86 @@
...
+        if !wm_endpoint_raw.nil?
+          wm_endpoint = uri_to_absolute(wm_endpoint_raw, URI.parse(s.url))
+          send_webmention(s.short_id_url, s.url, wm_endpoint)
+        end
+      end
+    end
+  end
+  last_story_id = s.id
```
pushcx: This doesn't persist to the `Keystore`

script/send_new_webmentions
NinjaTrappeur: As I said in the PR body, this code is not mine, it has been extracted from [here](https://github.com/indieweb/webmention-client-ruby/blob/master/lib/webmention/endpoint/discover.rb#L30)

This code has been extensively used and tested, I don't really feel like refactoring it without an extensive test-suite.

Is it a real deal breaker for you?

extras/sponge.rb
NinjaTrappeur: This is the first time I ever write some Ruby, I'm really not familiar with the tooling. 

Is there any goto refactoring tool for this kind of job or do you just grep the code looking for some fetch occurrences?

script/send_new_webmentions
pushcx: No, I can live with it. It's ugly, but at least it's not a core feature and it's at the end of the call graph.

pushcx:
GitHub is all freaked out because I made some style tweaks and squash-merged this manually.

Thanks for doing this, @NinjaTrappeur. I'm deploying it now. Hopefully it'll help connect us to the wider web.

PR #538, opened 2018-08-30
Remove duplicated before action

longkt90:
these lines of codes appeared twice


pushcx:
Huh, what a weird bit of duplication. Thanks for spotting it and offering a PR!

Issue #522, opened 2018-07-14
tags missing during preview


nickpsecurity:
Although I might be misremembering, I think the site used to show the tags in the preview of a new submission. The tags aren't there anymore. Was that intentional or a bug?
![no tags during preview](https://user-images.githubusercontent.com/17504074/42725347-baaddbda-8747-11e8-9b71-1180069929b3.png)

pushcx:
It's a bug.

btreim:
I'd love to take a crack at this.

pushcx:
@btreim Help yourself! I typically review PRs and issues on Wednesday and Thursday mornings, so if you puzzle it out in the next day or two your fix will be seen quickly. Thanks for offering.

btreim:
Just an update, my development laptop crashed shortly after taking this on. If someone else wants to jump on this issue in the meantime, by all means!

Issue #525, opened 2018-08-13
Multi Tag based Feeds


Church-:
As per user suggestion, a feature I thought i'd recommend would be multi tag feeding, useful both to members and non member users of the site. Allowing you to search tags like so: lobste.rs/t/assembly+javascript+culture.

This feature would work both via web view, and via RSS as well.

pushcx:
We had a chat about this on IRC and I accepted the feature request.

sohaibbhatti:
Just thinking, would it make sense to show the intersection of tags? or the union?
I.e. for assembly+javascript+culture, would we displaying all feeds tagged as either assembly OR javascript OR assembly, or display feeds containing assembly AND javascript AND assembly?

I'd love to have a stab at this

pushcx:
In chat I understood it to be union, so assembly OR javascript OR culture.

sohaibbhatti:
Awesome thanks!

noahhex:
@sohaibbhatti Yeah it's union, although adding intersection for more fine grained control would be interesting I think.

pushcx:
In chat, rdivyanshu pointed out that we have the tag `c++` that will not work happily with a `.split('+')`. Because tags are only added by the administrator, the name field (`tag`) never got any validation limiting it. We'll have to find another joining character; I really don't want to rename that tag to `cplusplus` or something.

crertel:
`|` perhaps?

pushcx:
`|` is not a valid character in URLs, but `;` is, the markdown engine's `autolink` option recognizes it, and it has the meaning of separating things. Seems a reasonable color to paint this bikeshed.

briankung:
Tempted to work on this next after a short break. Would it be okay if we allowed escaped plus signs (`%2B`) and kept the `+` delimiter? A la: https://stackoverflow.com/a/7146153/1042144

pushcx:
Yeah, that sounds fine. I don't have anywhere near a strong enough opinion to force it on whomever is volunteering to implement this. :)

PR #541, opened 2018-09-06
Fix story tags not being rendered in preview

sohaibbhatti:
https://github.com/lobsters/lobsters/issues/522

Accessing tags from a story involves a database query. For records that haven't been persisted yet, resorting to fetch tags from the instantiated taggings.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -624,8 +624,7 @@ class Story < ApplicationRecord
           # we can't lookup whether the user is allowed to use this tag yet
           # because we aren't assured to have a user_id by now; we'll do it in
           # the validation with check_tags
-          tg = self.taggings.build
-          tg.tag_id = t.id
+          self.tags << t
         end
       end
     end


pushcx:
It's strange to me that the unsaved associated tags are not available through the usual `.tags`, do you know why that is? And were you able to track down why this broke?

sohaibbhatti:
I'd looked at a commit that was for resolving a n+1 query problem which caused the issue to pop up. 7dfbbea3b542e431161d662ee889131c3a64d5fc 

To reproduce
```
git checkout 7dfbbea3b542e431161d662ee889131c3a64d5fc^1  #wouldnt exist
git checkout 7dfbbea3b542e431161d662ee889131c3a64d5fc #will exist
```

I believe its because of a has_many :through association, so I'm assuming a query is always needed, i.e. we can access taggings from a built story, but not tags as it needs to go through taggings.(plus there is a default order scope on the association as well). I'll look into it further

pushcx:
Ah, so the issue is that `StoryController#story_params` is setting the tags as `Story#tags_a=` which uses `taggings` instead of `tags`, so Rails doesn't create objects in `tags` for the `has_many :through` to use. Can you edit `tags_a` to maintain that usual behavior? That's a lot more maintainable than knowing not to a custom method instead of one association of one model before it's persisted.

sohaibbhatti:
Definitely! I'll do it in a bit!

pushcx:
Thanks for running this one down and finding a clean fix that maintains Rails idioms. I've just deployed your fix.

sohaibbhatti:
Thank you for being on it with the reviews :)
Off topic, but whenever you have a moment could you have a look at the comment in  https://github.com/lobsters/lobsters/issues/525

PR #543, opened 2018-09-10
senic.rb -> scenic.rb

xtian:


pushcx:
Thanks for catching this typo.

PR #544, opened 2018-09-10
README.md: Remove secret_key_base instructions

xtian:
secret_key_base is set via secrets.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,14 +75,7 @@ running tests:
     ```sh
     lobsters$ rake db:schema:load
     ```
-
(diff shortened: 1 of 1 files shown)


pushcx:
Do new users of the codebase automatically get that created for them, or do we need to update these instructions?

xtian:
They do when deploying but otherwise the values here get used:

https://github.com/lobsters/lobsters/blob/337d4747fb8ac48fabdbbc5c3415cc1263f0b623/config/secrets.yml#L13-L22

PR #460, opened 2018-03-02
Rubocop

pushcx:
I'm adding rubocop to enforce style issues to save human attention (and patience) in code reviews. The default config found 4,849 offenses and this branch will be merged when there are zero.

Like everybody I have opinions on style issues, but my preferred style for any codebase is whatever the codebase is already doing. With that in mind, I'm configuring [cops](https://rubocop.readthedocs.io/en/latest/cops/#available-cops) to enforce the common usage and then tweaking outliers to match, eg. 988aa22 and fda61bb.

If you have opinions on how best to paint this bikeshed, this PR is the place to speak up. Once this is merged I'd like to change style config as rarely as possible because it's distracting churn when reading through history during bughunts. Speak now or forever hold your peace or, in the future, make a compelling, metrics-driven argument that a style change will materially improve code quality.

The current 3,382 offenses may seem like a lot, but if you sort by offense count you'll of course see a [pareto-like distribution](https://en.wikipedia.org/wiki/Pareto_distribution) of 135 cops, so this PR hopefully won't linger for months. `rubocop | grep -o '[A-Z][^/]*/[A-Z][^ ]*' | sort | uniq -c | sort -rn | head -n 20`

TODO: run rubocop as part of the test suite so humans see it easily and Travis fails the build.

PR #545, opened 2018-09-10
.rubocop.yml: Enable Rails/Blank

xtian:


pushcx:
As discussed in #460, there isn't a compelling reason to repaint this bikeshed.

PR #546, opened 2018-09-10
.rubocop.yml: Enable Style/FrozenStringLiteralComment

xtian:
`frozen_string_literal` improves compatibility with future Ruby versions and may improve performance: https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/


xtian:
Ok that makes sense. In the case of #545 I would say it makes sense from a clarity perspective, and as I noted for this one there are compatibility and performance ramifications.

What problem are you specifically trying to avoid by avoiding style changes in the version control history?

I would argue that any resulting churn is worth the cost in cases where 1) the check can be made to pass using `--auto-correct` and thus the effort to enable it is low, and 2) the current state of the check is `Enabled: false` because any extant consistent style has to be enforced through manual review which has a high cost and failure rate over time.

pushcx:
As I mentioned in 8138d9e7a43e1da29bd565527f75da61de9c94f0 I'd prefer to let future me suffer if this becomes a default in Ruby 3 (I anticipate pretty minimal suffering, tho). And our config is `EnforcedStyle: never`, not `Enabled: false`, so we are getting that checked now.

Issue #478, opened 2018-04-04
rake db:schema:load doesn't generate foreign key in database


WanCC1995:
I followed the readme to build up the applications. (build on MySQL)
I found that the foreign key only appears at app level, while db level doesn't have any foreign keys. It's kind of weird that the database don't do such validation.
I'm wondering how to make up it. One straight forward idea is to manually add such foreign keys to the database. But I guess that it may not be a wise idea.

Looking forward to your reply.

pushcx:
The Lobsters codebase predates Rails's support for foreign keys. Confusingly, [the documentation](http://guides.rubyonrails.org/association_basics.html) uses the tern "foreign key" for the column assocating one record with the `id` column of another, though they are not database-level foreign keys.

Now that Rails supports FKs, it would be good to add a migration using [`add_foreign_key`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_foreign_key) for these. In the `rubocop` branch that recently landed I added a lot of the missing `:dependent => :destroy` and to associations, so this should be straightforward.

WanCC1995:
@pushcx Thanks for your timely reply! I'll add a new migration.

pushcx:
This was done in #548

PR #548, opened 2018-09-13
Add foreign keys

tansengming:
This is an initial fix for https://github.com/lobsters/lobsters/issues/478

There are a few things to note,

- the migration was generated by running [Immigrant](https://github.com/jenseng/immigrant), which creates the migration by inferring it from the models,
- I've had to comment out a few lines on the migration because of column mismatches. E.g. [User#id](https://github.com/lobsters/lobsters/blob/5b39f579ef4ecefd4818a75957b5d7983fcf43fc/db/schema.rb#L221) is an unsigned integer, while [HatRequest#user_id](https://github.com/lobsters/lobsters/blob/5b39f579ef4ecefd4818a75957b5d7983fcf43fc/db/schema.rb#L44) uses an integer column. MySQL has issues with having foreign key constraints on different column types.
- I've left them there for reference, in case you'd like to deal with them in the future.

I look forward to hearing to hearing comments from this!


pushcx:
Hey, thanks for the PR! It sounds like you've found a lurking bug with the different column definitions, and user ids should never be negative. I'm going to write a migration to correct those, then hopefully you should be able to uncomment all of these.

pushcx:
I just pushed a commit normalizing all table ids + foreign keys pointing at them as unsigned bigints, adding a few missing "not null" constraints. Now, if my double- and triple-checking that I got everything correctly paid off, we should have a totally consistent set of columns.

@tansengming, would you mind rebasing this PR on current master? A `git push -f` is no big deal here. Or closing + reopening a new PR, if that's easier for you.

Either way, thanks for making it clear how much cruft we had to clear out. This prevents a bunch of difficult and irritating future issues, so I'm very glad for your contribution.

tansengming:
@pushcx that sounds great! I'll get this fixed up in a day or two. 
https://github.com/lobsters/lobsters/commit/12644b27136eaca9feec3a5fe4ea99b7a04809f4 looks like a big change I hope there weren't any migration issues.

pushcx:
No, migration was surprisingly painless. MariaDB was a total champ, too - requests were slow, but we didn't have downtime. Only 5-6 requests failed to acquire a write lock and 500'd. I was impressed.

Also, hi, I see you maintain stripe-rails, which I'm integrating into another project this afternoon Thanks for that!

tansengming:
> Also, hi, I see you maintain stripe-rails, which I'm integrating into another project this afternoon. Thanks for that!

you're welcome! I hope you have a good experience with it. Do let me know if you have any issues with it.

tansengming:
FYI I've started a new PR for this change over here https://github.com/lobsters/lobsters/pull/561 and will be closing this one.

Also, there seems to be a mismatch of the schema vs migration that I've tried to fix over here https://github.com/lobsters/lobsters/pull/560 . I recommend merging this before the foreign key PR

Issue #508, opened 2018-06-13
Turn on belongs_to_required_by_default (rails 5.0 default).


thomasdziedzic-pd:
As mentioned in https://github.com/lobsters/lobsters/pull/498/files#diff-b1fe55db50c712fef0673345e5b9c0d9R30

pushcx:
Reviewing issues... how did this not break/start throwing warnings when I merged the `config.load_defaults 5.2`?

thomasdziedzic:
I turn it off right after loading the 5.2 defaults: https://github.com/lobsters/lobsters/blob/master/config/application.rb#L29

PR #550, opened 2018-09-20
Remove redundant validations

xtian:
These are all covered by the `belongs_to` directive.


pushcx:
Ah, yeah, these became redundant when #508 was done. Thanks!

PR #551, opened 2018-09-20
EmailParser: Make content-type regex case-insensitive

xtian:


pushcx:
Yep, I see in RFC 1341 that this is correct. Thanks!

PR #552, opened 2018-09-20
InvitationRequest: Validate presence of email

xtian:
`InvitationRequest#email` is used by `InvitationRequestMailer`


pushcx:
Good catch, thanks.

PR #553, opened 2018-09-24
Update CONTRIBUTING.md

sean-public:
rdivyanshu pointed out on IRC that this link needs to be fixed


pushcx:
Thanks for reporting and chipping in a fix, I'm pushing this live now.

Issue #422, opened 2017-10-16
Backend-independent full text search


pushcx:
The search dropped its dependency on sphinx in favor of SQL's full text search. This is [currently using](https://github.com/lobsters/lobsters/blob/0ad6dffbe705f1d446461061f855a95b1df233d6/app/models/search.rb#L88) a MySQL-specific syntax. Maybe there's an Proper ActiveRecord Way to do this but maybe it'll have to be hand-coded (only need support for MySQL and PostgreSQL).

evexoio:
@pushcx the code you referred to no longer exists... does this mean this issue should be closed?

or do you want to get rid of ~any `MATCH AGAINST` hard coded queries in the code?

norpol:
Is this ticket still relevant?

pushcx:
This was implemented with the new search backend in September 2023.

Issue #361, opened 2017-05-24
Feature Request: Continuing work on JSON API


lukefiji:
Wondering if you would like to make any progress on the JSON API from #42? I'm working on a Lobste.rs React web app as a practice project and would love to move onto creating a React Native version later down the road. 

I'm wondering how difficult it would be to be able to feed more than just the first 25 hottest/newest posts; and if it is possible in a reasonable manner, I would really like to work with you and give you my input to help make this API flourish

pushcx:
@lukefiji If you'd like this please open a PR with code. Me building out an API by guessing at someone else's needs would likely only be of marginal use (and adds a maintenance burden).

If you do embark on this, please don't make one giant PR for all possible endpoints. It would be very helpful to be able to see them one at a time and give suggestions for future improvements so you wouldn't waste your own time writing code that wouldn't be a good fit for the project.

Issue #150, opened 2014-05-06
Opening up more API endpoints.


jcspencer:
Opening up the following resources via JSON would greatly improve the ability to create 3rd party applications around Lobste.rs
- Search
- Tags
- Users
- Login/Logout
  - Comment
    - Upvote/Downvote
  - Stories
    - Upvote/Downvote
  - Settings(?)
- Signup

To keep things simple, I presume Login would just take a username/email and password and return a session token?

Just a thought :)

Issue #42, opened 2012-12-02
JSON API for third-party apps / Lobsters iPhone App


rhysforyou:
I'd absolutely love a JSON API so people can build apps on top of Lobste.rs. I've been wanting to build an iPhone client for the site for a while but the lack of an easily accessible API has prevented that so far. I'd be happy to help implement this myself, I've got a bit of Rails experience, but I thought a feature this large warranted a bit of discussion.

To avoid having to deal with authentication at this point it'd probably be best to start it out as a read-only API. The main things I'd want access to as an app developer right now are the top and newest story sections and individual posts.

Have you thought much about the API so far?

PR #555, opened 2018-09-27
Feature: Add /stories JSON endpoint

theoretick:
Hey all, thanks for the awesome site!

I was looking for some JSON endpoints to enable searches across lobste.rs and found these previous discussions: https://github.com/lobsters/lobsters/issues/42, https://github.com/lobsters/lobsters/issues/361, https://github.com/lobsters/lobsters/issues/150. It looks like no progress has moved forward recently so I took a look and added a `/stories.{html|json}` endpoint.

I originally considering modifying the `SearchController` to support both MIME Types but treating stories as a RESTful resource felt more correct. In addition, I added some HTML handling to provide more intuitive access to endpoint as well: redirecting unscoped `/stories` requests to root and scoped `/stories?=q` requests to `/search`.

#### Additional Notes

 1. The behavior of `Search#page_count` seemed a bit odd when result sets were empty or nil, returning either 0 or 5. IMHO a page_count should always be a positive integer so I corrected that behavior and added some `Search` tests.
 2. The unscoped `/stories.json` endpoint requires a query param. This is mostly to keep things simple and avoid the additional questions of sort order, caching, etc... If this seems overly restrictive, happy to change it.
 3. I initially started writing a more fully-featured test that actually used `Search` but due to the way MySQL full text search works, `MATCH(..) AGAINST(..)` does not function inside of transactions :( (https://dev.mysql.com/doc/refman/5.6/en/innodb-fulltext-index.html#innodb-fulltext-index-transaction). This side-tracked me a bit into working on https://github.com/lobsters/lobsters/issues/422 but that’s quite a bit more involved to keep the queries reasonably optimized. TBD!
 4. As a result of (3), I’m not a huge fan of the mocked Search results in the new controller spec but it seems sufficient for now.
 5. Should these JSON endpoints (and the existing `stories/:id.json` endpoint) be documented anywhere? Would it be worth adding a page, API spec, or some section to the readme?

Let me know your thoughts!

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -11,6 +11,49 @@ class StoriesController < ApplicationController
   before_action :find_story!, :only => [:suggest, :submit_suggestions]
   around_action :track_story_reads, only: [:show], if: -> { @user.present? }
 
+  def index
+    @title = "Stories"
+    @cur_url = "/stories"
+
+    query = params[:q].to_s
+
+    respond_to do |format|
+      if query.blank?
+        format.html { redirect_to "/" }
+        format.json {
+          render json: { message: "Missing parameter: q" }, status: 422
+        }
+      else
+        format.html { redirect_to search_path, params: request.query_parameters }
+        format.json {
+          @search = Search.new
+          @search.what = "stories"
+          @search.q = query
+
+          if params[:order].present?
+            @search.order = params[:order]
+          end
+          if params[:page].present?
+            @search.page = params[:page].to_i
+          end
+
+          if @search.valid?
+            @search.search_for_user!(@user)
+          end
+
+          render json: {
+            results: @search.results.map(&:as_json),
+            meta: {
+              page: @search.page,
+              page_count: @search.page_count,
+              total_results: [@search.total_results, 0].max
+            }
+          }
+        }
+      end
+    end
+  end
+
   def create
     @title = "Submit Story"
     @cur_url = "/stories/new"
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -36,12 +36,15 @@ class Search
 
   def page_count
     total = self.total_results.to_i
+    per_page_count = self.per_page.to_i
 
-    if total == -1 || total > self.max_matches
+    if total > self.max_matches
       total = self.max_matches
+    elsif total < per_page_count
+      total = 1
     end
 
-    ((total - 1) / self.per_page.to_i) + 1
+    ((total - 1) / per_page_count) + 1
   end
 
   def what
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -65,6 +65,7 @@ Rails.application.routes.draw do
   get "/search/:q" => "search#index"
 
   resources :stories do
+    get "index"
     post "upvote"
     post "downvote"
     post "unvote"
--- a/spec/controllers/stories_controller_spec.rb
+++ b/spec/controllers/stories_controller_spec.rb
@@ -1,0 +1,90 @@
+require "rails_helper"
+
+describe StoriesController do
+  let(:story) { create(:story, title: "mytitle") }
+  let(:other_story) { create(:story, title: "myothertitle") }
+
+  let(:json_response) { JSON.parse(response.body) }
+
+  context 'index' do
+    context 'html' do
+      it 'redirects to search when scoped' do
+        get :index, params: { q: "myquery" }
+
+        expect(response).to redirect_to search_path
+      end
+
+      it 'redirects to root when unscoped' do
+        get :index
+
+        expect(response).to redirect_to root_path
+      end
+    end
+
+    context 'json' do
+      it 'returns Search#results when scoped' do
+        search = Search.new
+
+        allow(search).to receive(:results) { Story.all }
+        allow(Search).to receive(:new) { search }
+
+        get :index, format: :json, params: { q: story.title }
+
+        expect(response).to be_successful
+
+        results = json_response.fetch("results")
+        expect(results.count).to eq(1)
+        expect(results.first["title"]).to eq(story.title)
+
+        expect(json_response.fetch("meta")).to eq({
+          "page" => 1,
+          "page_count" => 1,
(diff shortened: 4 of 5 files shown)


pushcx:
Thanks for jumping in. I'm curious, what prompted this? I'm asking because I'm not really sure why the site has json endpoints; I just skimmed down `application.js.erb` and none of our code uses it. I'm happy to help curious users read the site programmatically without scraping, my concern is the cost of maintaining something we don't need. I don't want to commit to maintaining code we don't use, especially the stability of quirks like whether an API with zero results has zero pages or one empty page.

I'd rather add json output to the SearchController: doing this in StoriesController duplicates that controller action. We'll never remember to update that if the logic changes. Our controllers are not particularly RESTful (one of those things I'd like to slowly improve as I have excuse to touch them), but I'd rather see new/create semantics around a many-fielded Search model than a very complicated interface to indexing Story models. Kinda heading into the weeds where I'm not sure there's a lot of semantic bang-for-the-buck.

How did `Search#page_count` return 5 when there's no results? Zero sounds like the correct answer when there are no pages, though I can see why you prefer 1.

For documentation, my first instinct is to say that I'd like the pagination cleanup of #394 first. Or json-api to normalize output. Maybe that's letting "perfect" get in the way of "good". And maybe if they're irregular and undocumented people won't expect too much from them. :)

theoretick:
Thanks @pushcx!  Apologies for the late reply, been a bit busy :)

> Thanks for jumping in. I'm curious, what prompted this? I'm asking because I'm not really sure why the site has json endpoints; I just skimmed down application.js.erb and none of our code uses it. I'm happy to help curious users read the site programmatically without scraping, my concern is the cost of maintaining something we don't need. I don't want to commit to maintaining code we don't use, especially the stability of quirks like whether an API with zero results has zero pages or one empty page.

Totally fair points. My initial motivation came from an issue someone opened for one of my personal projects (https://github.com/theoretick/discuss-it/issues/86).  tl;dr - http://discussitapp.com is a search tool for finding online discussions across reddit, hackernews, slashdot, etc, and lobste.rs would be an awesome addition.

When I initially started looking into this I figured I would have to scrape the site and I wasn’t really a fan of that approach so I took a look at a more programmatic solution. creating an entire API seemed to be going down a pretty big path as you noted, but I realized that there was already some initial JSON support [via the `StoriesController#show` action](https://github.com/lobsters/lobsters/blob/b43a5999b4b2140971ddc48dc5221f512e112e4d/app/controllers/stories_controller.rb#L167). As such, it seemed like an additional endpoint would be gradual enough to not be dumping too much complexity.

That said, I completely understand the added maintenance cost. Especially if it results in a significant change in semantics (see below). If after this discussion it seems like a bad direction, we can close this PR out. No worries at all :)

> I'd rather add json output to the SearchController: doing this in StoriesController duplicates that controller action. We'll never remember to update that if the logic changes. Our controllers are not particularly RESTful (one of those things I'd like to slowly improve as I have excuse to touch them), but I'd rather see new/create semantics around a many-fielded Search model than a very complicated interface to indexing Story models. Kinda heading into the weeds where I'm not sure there's a lot of semantic bang-for-the-buck.

👍 I took that initial approach but changed my mind. Happy to add a simpler `format.json { ... }` and keep it simple. One less controller!

The only issue I had is that it gets a bit weird to return different resource types through the same endpoint (i.e. stories vs comments). Perhaps nesting the resources in the response could be the most flexible so if we wanted to ever return both matching stories and comments it would be simple:

```
200 OK

{
  “stories”: [ ... ],
  “comments”: [ ... ],
  “meta”: {
  }
}
```

> How did Search#page_count return 5 when there's no results? Zero sounds like the correct answer when there are no pages, though I can see why you prefer 1.

The 5 count was an interesting one. So the way the logic appears to work, if `total` results is -1, then `total = max_matches` (100). Then the page count is calculated as `((total - 1) / per_page) + 1` or `((100 - 1) / 20) + 1` resulting in 5. See relevant logic: https://github.com/lobsters/lobsters/blob/b43a5999b4b2140971ddc48dc5221f512e112e4d/app/models/search.rb#L40-L44

RE: zero counts, I’d argue that returning a result set at all is enforcing a minimum page_count of 1. From an API perspective, the alternative would be a 404 condition where there is no idea of a page_count. From a service perspective (i.e. `Search` here), a page_count is a virtual attribute of the `results`, which is always an empty array: https://github.com/lobsters/lobsters/blob/b43a5999b4b2140971ddc48dc5221f512e112e4d/app/models/search.rb#L21. I’d probably feel differently if the result sets was nullified.  Regardless, that’s my 2 cents, if you’d prefer 0, I’m happy to change it.


> For documentation, my first instinct is to say that I'd like the pagination cleanup of #394 first. Or json-api to normalize output. Maybe that's letting "perfect" get in the way of "good". And maybe if they're irregular and undocumented people won't expect too much from them. :)

Hah, that one is always a tough call but think it really falls to the discretion of the maintainers. I'm all for iterative solutions but public documentation of an API brings up a non-trivial maintenance cost as you noted before.

Cheers,
Lucas

pushcx:
No worries on the slow response; I tend to check in on the codebase on Wed + Thu mornings rather than throughout the week.

From poking around your app, it sounds like what would really fit your needs is an endpoint that takes a URL and returns a list of links to submissions of that URL, rather than an interface to the full search engine. The full-text search engine is the slowest, most resource-intensive database query on the site and I hesitate to encourage people to hit it in a loop - but url, hey, we have an index on that column, that's really cheap. I wouldn't even bother with pagination: the most-submitted link talk has only been submitted four times (Bret Victor's "The Future of Programming") and even if listing five hundred pairs of title + url this would be a cheap endpoint to serve. The natural place to put this would be `StoriesController#check_url_dupe`. There's some related work happening now in #557, even.

pushcx:
#557 was implemented yesterday, so now is a good time to looking at making the `check_url_dupe` endpoint serve json. It looks like a really good fit for your needs.

theoretick:
@pushcx thanks! Just opened https://github.com/lobsters/lobsters/pull/594 as a potential replacement and closing this one out

PR #560, opened 2018-10-05
Schema fix for unsigned bigints migration

tansengming:

PR #561, opened 2018-10-05
Add foreign keys (2nd Try)

tansengming:
This is a retry of https://github.com/lobsters/lobsters/pull/548 after @pushcx pushed a [migration](https://github.com/lobsters/lobsters/commit/12644b27136eaca9feec3a5fe4ea99b7a04809f4) to fix the column mismatch issues.

I've had to do a schema fix over https://github.com/lobsters/lobsters/pull/560 to get this to work, but other than that things are looking good!

Lastly, I'd like to note again that this migration was generated by running [Immigrant](https://github.com/jenseng/immigrant), which creates the migration by inferring it from the models


tansengming:
Note that the travis ci failure is an unrelated rubocop line length warning that was propagated from 5e6f1e6 . I'll rebase from master once it's been fixed there.

pushcx:
#564 fixed the style violation on master.

pushcx:
Found a couple orphan records violating these new constraints; I'm investigating and manually cleaning data before I deploy this.

pushcx:
`messages.author_user_id` is set to `0` for messages from the system like "your story was edited by user suggestions". I think I'd rather drop that FK constraint than assign them to `inactive-user` or create another system user that might get replied to (or need a bunch of special cases for), etc. I'm rewriting the commit, sorry to anyone who's pulled in the last 15 minutes.

tansengming:
That sounds like a good call @pushcx . I'm curious how were you checking for FK constraint violations before deploying this?

I'll be happy to add a test case for: 

> messages.author_user_id is set to 0 for messages from the system

in case someone else starts wondering about this FK exception later.

PR #564, opened 2018-10-08
Fix Rubocop violation on master

jmks:
There is a RuboCop violation that is broken on [master](https://travis-ci.org/lobsters/lobsters/builds/436689242?utm_source=github_status&utm_medium=notification).

This fixes it 😄


pushcx:
Thanks. I had to patch the Webmention code on its deployment and missed this violation.

Issue #559, opened 2018-09-30
Flick scrolling performance stutters


wting:
When attempting to scroll through the comments page it stutters quite a bit ([mp4](https://gfycat.com/FloweryHappygoluckyAiredale)):

![floweryhappygoluckyairedale-size_restricted](https://user-images.githubusercontent.com/622170/46262504-27c3c200-c4b7-11e8-9973-cd6630feef75.gif)

I'm using an iPhone 8 and Chrome

pushcx:
Huh, that's weird. Thanks for the clear demo.

Could you log out and see if it persists? The site only serves js to logged-out users. There's not much, but that might narrow down the possible causes.

wting:
I can't reproduce in Incognito ([mp4](https://gfycat.com/gifs/detail/HarshTatteredGrayfox)):

![loggedout](https://user-images.githubusercontent.com/622170/46263323-88a4c780-c4c2-11e8-8fa1-3f6660ce54a9.gif)

pushcx:
Well, we've only got [~600 lines of js](https://github.com/lobsters/lobsters/blob/master/app/assets/javascripts/application.js.erb), so that narrows things down considerably. Any chance there's someone know knows performance bugs in mobile Safari well reading this?

What's puzzling is that I don't think we have any events connected to scrolling or visibility; it's a dozen click handlers and a couple form editing niceties.

```
$ ag '\.on' app/assets/javascripts/application.js.erb 
423:  $(document).on("click", "a.comment_replier", function() {
464:  $(document).on("click", "button.comment-cancel", function() {
485:  $(document).on("click", "a.comment_editor", function() {
493:  $(document).on("click", "a.comment_deletor", function() {
502:  $(document).on("click", "a.comment_undeletor", function() {
512:  $(document).on("click", "a.comment_disownor", function() {
522:  $(document).on("click", "a.comment_moderator", function() {
536:  $(document).on("click", "div.markdown_help_toggler .markdown_help_label",
542:  $(document).on("click", ".comment-post", function(event) {
547:  $(document).on("click", "button.comment-preview", function() {
551:  $(document).on("click", "button.story-preview", function() {
555:  $(document).on("blur", "#story_url", function() {
578:  $(document).on("blur change", "#story_title", Lobsters.checkStoryTitle);
581:  $(document).on("click", "#story_guidelines_toggler", function() {
```

knixeur:
[...]

PR #565, opened 2018-10-10
Fix small markup error en listdetail.html.erb

knixeur:
Just a small typo when closing quotes.


knixeur:
Lol, I was actually trying to reproduce #559 (I wanted to earn an invite to lobste.rs :), did I? ) but couldn't since I don't have an iOS device. As I was checking the generated output I caught that :)

pushcx:
Sure, happy to send one for a fix. I'd still love help with that scrolling thing - it's a weird one, and I don't have any iOS devices to try to replicate it.

PR #566, opened 2018-10-10
Make migrations runnable 

Paxa:
In new rails we need to specify migration version, also some tables have custom mysql collation in schema.rb, but not in migrations

Also I added `ENV["APP_DOMAIN"]` to make it easier run in docker (so no need to change any files)


pushcx:
Welcome! Did the codebase get some press? We've had an influx of people setting up new sites.

Looks like there's a lot of overlap between this and #567.

The README doesn't suggest creating a database by running all migrations to start a site, it suggestions loading the `schema.rb`. Did that not work for you?

Paxa:
Hi, I was thinking if I use schema.rb then later I can not run new migrations to update.
Would it be nice to keep both migrations and `schema.rb`

Btw I have idea to add welcome story in seeds.rb with test@example.com user credentials and some introduction on how to add posts, manage users, etc. What you think?

Paxa:
Also some indexes are missing in migration but exists in `schema.rb`, should we add migration for that?

pushcx:
If load the schema you'll be at the same point as any new migrations and able to run them as they arrive. `schema.rb` matches production and that's what I care about maintaining. A migration to add an index that already exists would fail and prevent production from running any later migrations.

I made some small improvements to `README.md` that I've been meaning to for a few weeks, making it clear what to run in development vs. production. I think some more introduction to the site in a seeded story sounds fine.

Paxa:
Is there a way to automatically skip `db:schema:load` when database not empty?
Now if I restart docker container then it trying to delete all tables:
```

-- create_table("comments", {:id=>:bigint, :unsigned=>true, :options=>"ENGINE=InnoDB DEFAULT CHARSET=utf8mb4", :force=>:cascade})
rake aborted!
ActiveRecord::StatementInvalid: Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails: DROP TABLE IF EXISTS `comments` CASCADE
```

Paxa:
If you consider to use migrations then we can make initial migration (that I made) skip when older migrations already ran. (make it check if some table is exists, e.g. `users`)

pushcx:
I don't know about docker besides the existence of the docker repo someone created, that's linked in the readme. If it's trying to delete your database when you restart the container, that sounds like a bug in that config.

Paxa:
That docker container is kinda outdated and slow to start, I use my fork... Normally for rails I use command:
```
rake db:create db:migrate && rails s
```
But because of missing migration, docker image have to check migration version
https://github.com/jamesbrink/docker-lobsters/blob/master/docker-assets/docker-entrypoint.sh#L15

PR #567, opened 2018-10-10
Fix migrations rails5

knixeur:
After cloning the project and trying to run it (my first time with Ruby/Rails) I had troubles with migrations. It turns out, there were some migrations missing and also there's a change in how ActiveRecord migrations are declared.
This commits try to fix this so migrate can be used.

Feel free to reject this PR if this is a wrong approach on Rails.

### DIFF
--- a/db/migrate/00000000000000_create_users.rb
+++ b/db/migrate/00000000000000_create_users.rb
@@ -1,0 +1,18 @@
+class CreateUsers < ActiveRecord::Migration[4.2]
+  def change
+    create_table "users", id: :bigint, unsigned: true, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+      t.string "username", limit: 50, collation: "utf8mb4_general_ci"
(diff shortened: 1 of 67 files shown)


pushcx:
Welcome! Did the codebase get some press? We've had an influx of people setting up new sites.

Looks like there's a lot of overlap between this and #566.

The README doesn't suggest creating a database by running all migrations to start a site, it suggestions loading the `schema.rb`. Did that not work for you?

knixeur:
Lol, I stumbled the site on some random day a few weeks ago and kept reading it daily. I don't plan to setup a new site, just had some time and I enjoy fixing code.
I'll close this PR as it seems #566 does the same and better.
Loading schema.rb did work, as I said, I don't have experience with Ruby/Rails so I was toying with db:drop/create db:schema:load and when ran into db:migrate it failed so I went ahead to to try to fix it.
I guess I'm too used to not have a "base" definition like schema.rb and work only with migrations.

PR #568, opened 2018-10-10
Add flag reason for stories that don't work

notriddle:
[Triggered by this post](https://lobste.rs/s/6y2zsf/elm_compiler_for_erlang_virtual_machine)

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -16,6 +16,7 @@ class Vote < ApplicationRecord
     "O" => "Off-topic",
     "A" => "Already Posted",
     "S" => "Spam",
+    "U" => "Broken Link",
     "" => "Cancel",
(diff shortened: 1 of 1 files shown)


app/models/vote.rb
```diff
@@ -16,6 +16,7 @@ class Vote < ApplicationRecord
     "O" => "Off-topic",
     "A" => "Already Posted",
     "S" => "Spam",
+    "U" => "Broken Link",
```
pushcx:
```suggestion
    "B" => "Broken Link",
```

This'll help me instantly recognize it in ad-hoc queries.

pushcx:
Ah, yeah, this is definitely useful for getting mod attention instead of having people @ mention me.

Issue #557, opened 2018-09-29
partial to display stories with duplicate URLs


pushcx:
A feature I have wanted for [a while](https://lobste.rs/s/H6fIor/some_new_features#c_ws2nz8) is a partial listing stories with duplicate URLs, eg:

 * 2012-08-16 Bit-twiddling to improve bloom filter performance – 8 points, 0 comments
 * 2012-07-08 Bit-twiddling to improve bloom filter performance – 16 points, 22 comments
 * 2012-05-28 8 secrets your bloom filter needs that the liberal media won’t tell you – -3 points, 0 comments
 * 2011-12-17 Bit-twiddling to improve bloom filter performance – 2 points, 1 comment

It’d be nice both on /stories/new when it warns of a duplicate URL, and below the comments in story pages to connect discussions. I haven't seriously thought about the implementation, but check out  `Story.find_similar_by_url` for the definition of 'duplicate' URLs and maybe you just want to build on that?

pushcx:
Implemented by @brianpgerson in #557

PR #569, opened 2018-10-11
Add partial for duplicate stories [ISSUE-557]

brianpgerson:
This change for https://github.com/lobsters/lobsters/issues/557 now uses the built-in `Story.find_similar_by_url` method to display duplicate stories where the URLs are considered identical.

We display in two places: 1 is a modified version of the error flash when you try to submit a duped story:

image

Also if it is NOT recent:

image

And when you view a story that has matches for the same URL:

image

The title is the link to the domain and the comment links to the discussion.

Thanks for reviewing! And if anyone is uh, a little more stylistically inclined than I am and wants to contribute any markup/css tweaks, feel very free to do so.


app/views/stories/_form.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= render :partial => "stories/form_errors", :locals => {
-  :f => f, :story => f.object } %>
+<%= render :partial => "stories/form_errors", :locals => {:f => f,
```
pushcx:
```suggestion
<%= render :partial => "stories/form_errors", :locals => { :f => f, 
```

There's a trailing space at ` }`, so it's nice to balance.

app/views/stories/_form_errors.html.erb
```diff
@@ -1,24 +1,25 @@
 <div class="form_errors_header">
-  <% if story.errors.count == 1 && story.already_posted_story %>
+  <% if story.errors.count == 1 && !story.already_posted_stories.empty? %>
     <div class="flash-error">
-    <h2>Error: This story was already submitted <%=
-      time_ago_in_words_label(story.already_posted_story.created_at) %></h2>
+    <h2>Error: This story was most recently submitted <%=
```
pushcx:
```suggestion
    <h2>Error: This story was submitted <%=
```

This is more accuracy than the submitter needs to get what's going on, let's just drop it.

app/views/stories/_form_errors.html.erb
```diff
@@ -1,24 +1,25 @@
...
+      time_ago_in_words_label(story.already_posted_stories.first.created_at) %></h2>
     <p>
-    Please view the <a href="<%= story.already_posted_story.comments_path %>"
-    target="_blank">previous discussion</a> for this story.
+    Please view the previous discussions for this story: <br>
```
pushcx: Visual styling: please move the help text + partial render down out of the flash. The screenshots (thanks for including those, super helpful!) show it's too darn much red and blue for our very low-intensity site.

app/views/stories/_similar.html.erb
```diff
@@ -0,0 +1,25 @@
+<% similar.each do | story | %>
```
pushcx:
```suggestion
<% similar.each do |story| %>
```

Style tweak; dunno why Rubocop didn't catch this.

app/views/stories/_similar.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <a href="<%= story.url %>" target="_blank">
+        <%= story.title %><em></em></a>
```
pushcx:
```suggestion
        <%= story.title %></a>
```

Dead code.

app/views/stories/_similar.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <span class="comments_label">
+        |
```
pushcx: We do two-space indents in views, too.

app/views/stories/show.html.erb
```diff
@@ -76,3 +76,12 @@
     <% end %>
   </ol>
 <% end %>
+
```
pushcx: Please wrap the whole thing in `<% if @similar.any? %>` rather than always printing the `h4`.

app/views/stories/_similar.html.erb
```diff
@@ -0,0 +1,25 @@
+<% similar.each do | story | %>
+    <a href="<%= story.url %>" target="_blank">
```
pushcx: Please make this an `<ol>` with each story an `<li>`.

brianpgerson:
thanks for the review! i'll take care of this stuff soon, maybe tonight.

app/views/stories/_form_errors.html.erb
brianpgerson: so do you just want it to be like a the ordered list on the comments page but between the flash and the submit form?

pushcx: Yep!

brianpgerson:
okay I ended up changing up my approach a bit. i think this is much cleaner/clearer code-wise and helps make things simple re: moving the dupes OL down. lemme know what you think! 

updated screens:

image

image

(no duplicates -> no header)
image

brianpgerson:
this should be ready for a final review!

pushcx:
Thanks for your patience - I was traveling two weeks ago and had food poisoning last week. Digging in now!

pushcx:
This is great! Thanks for helping out, @brianpgerson.

pushcx:
Made some tweaks in 3ed2b8e5af0ce3d7d13c0e88104cacc045f3ed10 like using the word 'similar' instead of 'duplicate', just fiddly stuff not worth making you do.

PR #572, opened 2018-10-15
Fix #571 handle new comment

### LINKED ISSUE #571: can't remove upvote from just-posted comment
User reports:

> Upon posting a new comment, I was not able to remove the default upvote from the comment until I had reloaded the page. The upvote (and also, the downvote) arrows on the new comment had no effect. This was seen on an iPhone, Safari, iOS 12.0.1.

This is probably because the jQuery isn't binding to the new elements created after the ajax post, but I haven't looked into this at all.

#554 is related but definitely not required. #500 is similar but very unlikely to be causally related.

knixeur:
Bind click event to comments OL container instead of existing elements.
Otherwise when creating the new comment it won't have the event attached
to it.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -385,11 +385,12 @@ var _Lobsters = Class.extend({
 var Lobsters = new _Lobsters();
 
 $(document).ready(function() {
-  $(".comment a.downvoter").click(function() {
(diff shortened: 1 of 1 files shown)


pushcx:
Ah, great! I'm glad this was such a straightforward fix. Thanks, knixeur.

PR #574, opened 2018-10-18
Fix failing test data cleanup due to foreign key constraint

ur5us:
Before this change, running `Tag.destroy_all` would fail due to existing
`Moderation` records with a foreign key on some of those tags with the
following message:

```
Mysql2::Error:
  Cannot delete or update a parent row: a foreign key constraint fails
  (`lobsters_test`.`moderations`, CONSTRAINT `moderations_tag_id_fk`
  FOREIGN KEY (`tag_id`) REFERENCES `tags` (`id`))
```

Thus, one had to manually cleanup some test data before re-running the test suite.


pushcx:
Thanks, I appreciate the quick fix!

PR #576, opened 2018-10-18
operational wording of error messages on HTTP error pages.

alanpost:
The error message wording on the 500 page isn't suitable for a site with the traffic lobste.rs now receives.  This patch makes the wording more descriptive and idiomatic, as well as providing guidance for reporting server side errors.

### DIFF
--- a/public/422.html
+++ b/public/422.html
@@ -1,11 +1,10 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>422</title>
+  <title>Lobsters HTTP 422 status: client error</title>
 </head>
 <body>
-  <h1>rejected</h1>
-  <p>
-  it's not me, it's you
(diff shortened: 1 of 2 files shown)


alanpost:
If you'd rather have a PR moving these pages from lobsters.git to lobsters-ansible.git let me know; I could make the case they belong in that repository by comparison to the about page.  Unlike that page, however, these must exist on any deployment of the codebase.

pushcx:
Thanks for these.

I'm actually tempted to go the other direction, bring in the about and other pages from the ansible repo to here. There's already a hundred places someone needs to remove the name "Lobsters" and otherwise customize to set up a sister site; a couple more is no big deal and probably actually helpful for understanding what files to put where to have a running site.

pushcx:
Based on this + our other conversations I moved the assets + history over in 2fe67dc.

PR #578, opened 2018-10-19
Fix story author microformat tag

NinjaTrappeur:
The author entity was specified as a u-url instead of a u-author.

See https://chat.indieweb.org/dev/2018-10-18#t1539875247122800


pushcx:
Thanks for this fix!

(Readers: this got noticed and discussed in the chat room, so there's not much to say here.)

PR #579, opened 2018-10-19
Search via elasticsearch

hmadison:
Right now, search on the site is slow. If the hosting can accommodate it, I feel that moving to ElasticSearch gives that portion of the site a speed boost and moves the load off of the database and onto a service designed to handle the kinds of queries which are thrown at it.

The proposed changes are backwards compatible with the SQL based search, we are only swapping out a service object and adding code to index the documents.


alanpost:
@hmadison and I went through a test plan for this feature today.

We've provisioned a dedicated host for the Elasticsearch daemon, owing to it's unusually large memory demand.  Since it will be production infrastructure, it needs a backup plan.  As middleware it needs a reconciliation or consistency plan.  These can both largely be done with the elastic.rake functions: the search database can be completely wiped, a new index created, and all production data idempotently synced.  We'll still want to run tarsnap to capture log and configuration file, but the additional backup overhead should be modest.

I'll want to add this host to my monitoring system and will do so before we deploy.  I'm not yet sure how much memory or vCPUs we'll need--I'll need to fully populate the search table and run some test searches.  Like monitoring I'll do so before we deploy.  Finally, the reconcile script should run as a cron job to catch any stories, comments, deletions, or edits that are not added to search in the originating request.

To test, we submitted a few stories and comments with unique and overlapping words in the story titles and comment bodies.  We then deleted a subset of these stories and comments to create a reasonable matrix of positive and negative search results.

With this test data, we ran searches for stories and comments that included strings appearing in single, multiple, and deleted comments and stories.  We ran these tests as the admin user, a normal user, and logged out.  The admin user was able to retrieve search results that included deleted stories and comments.  The regular user and non-user requests did not return these deleted results.

We then edited both a story and comment to both include and remove words not previously in either.  Searching for these added and removed words produced the expected results (removing results when a word was deleted, adding a result when the word was added) for all three classes used for testing deleted comments and stories.

In every case I also fetched the entire query database and manually verified it corresponded to the comments and stories as submitted or edited.

If a user is filtering a tag or tags, results with those tags do show up in search.  I would prefer they did not, but would further like them to show up for the set of tags specified in the search query.

If a user has hidden a story, that story will show up in search.  Probably it should not.

Finally, if there is a comment that has not been deleted for a story that has been deleted, it is returned in search.  i'm not sure what the appropriate behavior is here, though I would lean toward not displaying comments associated with deleted stories.

Small improvements that could be made:

If the search server is offline, we get a 500 error.  We could report an error on the search page instead.

I will proceed with the monitoring and backup changes, as well as trial-loading production database for the purpose of sizing this host.

alanpost:
> I will proceed with the monitoring and backup changes, as well as trial-loading production database for the purpose of sizing this host.
 
@hmadison and I were able to do load testing of production data recently.

When it loads, the search middleware takes about 1.5GiB of memory.  Loading the production database doesn't change the memory used.  This is not surprising, as our database size is reasonably modest and my understanding is that the initial memory used is in service to 'warming up' the daemon by preinitializing internal data structures and caches.

There is plenty of room on the production system to install this program.

It does take a few hours or so to load the database using the backfill procedure.  This is true regardless of whether any new data is added to the search daemon.  This is because all of the work transits the work queue, so each story and comment must be added to the queue, pushed to search, and only then do we know if the work was necessary.  Certainly it's possible to optimize this, but I had imagined this procedure only on a periodic basis.

The backfill procedure does absolutely require beanstalk and the attendant backburner beanstalk client--otherwise the process terminates before the queue is flushed.  I don't mind full out requiring beanstalk but it does add another daemon to production.  I'd frankly like to then use it to mail stories and comments rather than doing so from cron.

alanpost:
@hmadison,

This PR consists now of 14 separate commits, some of which were clearly temporary or debugging in nature.   (to include "cleanup" and "rebase fixes".)

Would you mind cleaning up these commits in to something we can review?

hmadison:
@alanpost Done.

hmadison:
We are going to see about trying to use postgres full text search before trying elasticsearch.

PR #581, opened 2018-10-20
Fix deprecation warnings for upcoming Rails 6

ur5us:
DEPRECATION WARNING: Delegating exists to arel is deprecated and will be
removed in Rails 6.0. (called from block in <class:Story> at
./lobsters/app/models/story.rb:75)


pushcx:
Ah, thanks for digging in to these warnings and squaring them away.

PR #583, opened 2018-10-21
Switch to database cleaner approach for cleaning up test data

ur5us:
While the previous approach of hand-rolling test data cleanup + setup works
it’s prone to become insufficient with regards to future test scenarios. Recent
commits 6a6da094 and later 4620a922 highlight the issue.

Using the database cleaner gem provides a cleaner (pun intended) approach to
setup a clean room environment before and after each specific test.

Annotating specs with with `:js` or `:truncate` will switch from a transaction
based cleanup strategy to a truncation based one to enable feature/request specs
for which a web server is spun up in separate process by RSpec, in other words
the process executing the spec is not the same as the process handling the
request so RSpec/DatabaseCleaner wouldn’t know when to rollback the transaction.

The overall downside of this approach might be that RSpec takes a few more seconds 
to run all specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ gem "sitemap_generator" # for better search engine indexing
 group :test, :development do
   gem 'bullet'
   gem 'capybara'
+  gem 'database_cleaner'
   gem 'good_migrations'
   gem "listen"
   gem "rspec-rails"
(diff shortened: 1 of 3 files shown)


pushcx:
I think your reasoning is dead on. With the extra complexity of enforced foreign keys, our current approach has become brittle. Good tradeoff. Thanks for this.

PR #589, opened 2018-10-30
Use code tags instead of backticks

bsima:
This was bothering me so i fixed it real quick


pushcx:
Total brainfart for me to use Markdown instead of HTML there. Thanks, merging.

PR #592, opened 2018-11-08
Move closing `a` element

mozfreddyb:
This patch avoids a double-closing `a` element.
It patch aligns _listdetail.html.erb with _similar.html.erb, where the closing `a` element is in each if-branch. Compare with https://github.com/lobsters/lobsters/blob/3ed2b8e5af0ce3d7d13c0e88104cacc045f3ed10/app/views/stories/_similar.html.erb#L16,L18


pushcx:
Ah, yep, this is my editing mistake. Thanks for fixing it!

Issue #539, opened 2018-08-31
Migrate to SQLite


pushcx:
We've kicked around the idea of migrating the site to postgresql. I'd like to discuss it more thoroughly and permanently. Let's make a list some pros and cons with small/medium/large effect:

* Medium pro: common table expressions so the Replies feature can be turned off when any-level parent goes to a negative score instead of just the immediate parent. A bad reply rarely goes negative fast enough to prevent prompting for replies. This could also be extended to email notifications.
* Large con: The time and effort to migrate. I merged a bunch of changes to make the codebase portable between the two when I started [Barnacles](https://github.com/pushcx/barnacl.es), but I haven't pulled changes into there since a few months after. The views with the scenic gem will certainly need attention.
* Unknown: Better full-text searching?
* Unknown: Lower RAM usage?

I'm certain I'm forgetting things we've discussed. Pinging @hmadison and @alanpost who've talked about this.

pushcx:
Wow, that's wonderful. Thank you, that lowers both our cost of migrating and our number of unknowns.

If you didn't see it on Lobsters, someone just launched [a site with a similar focus](https://lobste.rs/s/ggrw0h/std_bz_lobsters_install_for_blockchain).j

renchap:
If you want to migrate the content (not schema, it is best to let Rails recreate-it from migrations if possible) to PG, you can use https://pgloader.io
It has a lot of features (type conversions, ...) and makes this quite easy.

xtian:
I was able to get `rake db:schema:load` to work with Postgres on my fork without too much effort: https://github.com/xtian/obelisk.link/commit/62a8138a319381522232e7dc562e71af4ae60635

brigriffin:
I would also like to see the Lobsters project function with PostgreSQL instead of MySQL. Hence as a proof of concept I forked this project and created the following branch which makes it possible to use Lobsters out-of-the-box with PostgreSQL:

https://github.com/brigriffin/lobsters/commit/ce8494297b47e35db76599242f475288752edd79

I would like to thank here @pyk and @xtian for their forks mentioned in this issue which I based my fork on. The advantage of my fork is that it is complete, up-to-date, and there should be (hopefully) nothing missing.

Now I know it is not possible to use my fork/branch as PR because it basically makes it not compatible  with MySQL anymore. Is anybody willing to help to integrate my fork/branch so that it can be both MySQL and PostgreSQL compatible? I am sure that would be a great gain for this project.

[...]

PR #593, opened 2018-11-08
Lobsters on PostgreSQL

brigriffin:
This PR adapts the whole lobsters installation to make it compatible with and use PostgreSQL instead of MySQL. It should work out-of-the-box but has the downside that it then only works with PostgreSQL. Some more work is required here to make it both compatible with PostgreSQL and MySQL.

Credits also go to @pyk and @xtian as this work is based on their work too (see #539).

For further details see my comment here: https://github.com/lobsters/lobsters/issues/539#issuecomment-433749476


pushcx:
Hey, I'm catching up on old PRs and (embarrassingly), this is the oldest.

I think for this to work and stay working we'll need to regularly test that we haven't accidentally broken postgresql, because it's very easy for us to do. Indeed, so easy that we've managed to do it several times since you opened this PR.

I've created #861 to track that work, and hopefully someone will be able to take it on soon so that we can revisit your PR and land it. It seems very much worth the small investment of our time to make things convenient for the sister sites.

brigriffin:
@pushcx just wanted to thank you for your feedback. I am glad to hear that my PR can be useful and that hopefully someone will tackle soon the tests part in #861 as you mention.

pushcx:
Closing out an inactive PR. Please reopen if this is something you'd like to pick back up, though it'd be a lot of work.

PR #594, opened 2018-11-12
Add JSON MIME to StoriesController#check_url_dupe

theoretick:
Replacement for https://github.com/lobsters/lobsters/pull/555

Adds JSON support to the `check_url_dupe` endpoint allowing searches for discussions about previously submitted URLs.

PR #595, opened 2018-11-21
Allow check_url_dupe to be queried for non-logged-in-users

theoretick:
Follow-up to https://github.com/lobsters/lobsters/pull/594 where I missed the authentication prevents queries against this endpoint for non-authenticated users.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -2,8 +2,7 @@ class StoriesController < ApplicationController
   caches_page :show, if: CACHE_PAGE
 
   before_action :require_logged_in_user_or_400,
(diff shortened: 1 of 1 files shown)


app/controllers/stories_controller.rb
```diff
@@ -368,7 +367,7 @@ def story_params
       :merge_story_short_id, :is_unavailable, :user_is_author, :tags_a => [],
     )
 
-    if @user.is_moderator?
+    if @user&.is_moderator?
```
pushcx:
```suggestion
    if @user && @user.is_moderator?
```

We don't use `&.` in the codebase. Unfortunately rubocop does not offer a cop to prohibit it.

theoretick: no worries, done!

PR #597, opened 2018-11-22
Refocus lobsters more on links than comments

notriddle:
I'd rather do something like this than [lock invites](https://lobste.rs/s/sfzmwr/proposal_set_everyone_s_invite_count_zero). Specifically, notice a few problems with how [the thread that prompted that discussion went](https://lobste.rs/s/egbgwb/introducing_conjure_palantir_s):

* The original post received 21 upvotes, and the first comment on it, which the author posted in an attempt to tell people to avoid the software, received 95. This doubles the article's score (since you're already capping it).
* Lobsters counts upvoted comments positively with the linked page, but actually, that poster probably would have wanted that link to be hidden.
* Because downvoting is so much harder than upvoting, people upvote comments they strongly agree with and ignore comments they disagree with, which, again, counts in favor of the link.

Lobsters then, essentially, optimizes for controversy (mostly because you copied Reddit in design, and Reddit is also optimized to generate controversy). Don't do that.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -311,7 +311,7 @@ class Story < ApplicationRecord
     i_cpoint = self.tags_a.include?('ask') ? 0 : -0.0625
     cpoints = self.merged_comments
       .select(:downvotes)
-      .map {|c| c.downvotes == 0 ? i_cpoint : -0.25}
+      .map {|c| c.downvotes == 0 ? i_cpoint : -0.25 }
       .inject(&:+).to_f
 
     # mix in any stories this one cannibalized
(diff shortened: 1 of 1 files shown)


pushcx:
I just read your comment with [more thoughts on why you suggest this](https://lobste.rs/s/sfzmwr/proposal_set_everyone_s_invite_count_zero#c_ukxw7m), thanks for elaborating.

I'm not accepting this PR because it's a huge change that doesn't at all have consensus on the site. My opinion is strongly the opposite, that the discussion is usually [the best thing on the site](https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate#c_2ztjbd), so I'm not at all tempted to BDFLYOLO it in.

Issue #596, opened 2018-11-21
write a custom cop to prohibit &.


pushcx:
The "safe navigation" operator `&.` makes it easier to work with and propagate `nil` values. Lobsters predates it so it was never previously used, but also I don't *want* it to be easy to add and pass around more `nil`s than we already have to. Unfortunately, its use is becoming common, so it's been a distraction in several recent PRs.

Rubocop has [clear instructions for creating a cop](https://github.com/rubocop-hq/rubocop/blob/master/manual/development.md) and its existing [SafeNaviation cop](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/style/safe_navigation.rb) is probably a great starting point. I believe the [require](https://docs.rubocop.org/en/latest/extensions/) config can load a cop (I really prefer not to maintain a fork of the rubocop repo) but haven't tested.

Bonus points and high praise if you can get `Enforce: not-used` or something added to the official SafeNavigation cop so we're not maintaining anything, but I have no idea what that process is like.

tansengming:
this sounds fun I'll pick it up. I'll try to keep this thread updated as I get it through Rubocop.

alanpost:
Thank you @tansengming.  @hmadison did a mock-up you might find helpful to start with:

```
module RuboCop
  module Cop
    module Style
      # Disallow the safe navigation operator
      #
      # @example
      #   # bad
      #   foo&.bar
      #
      class DisallowSafeNavigation < Cop
        MSG = 'Do not use &.'.freeze

        def on_csend(node)
          add_offense(node)
        end
      end
    end
  end
end
```

If you cared to clarify, improve, or better guide the user in the error message as well as test and put this outline in to a PR that would be phenomenal.

tansengming:
I've got an initial version of this ready at https://github.com/tansengming/rubocop-disallow-safe-navigation/pull/1

I'll be starting a PR on lobsters to get it integrated and we shall see how that goes.

tansengming:
The new custom cop should work with travis with the new configuration at https://github.com/lobsters/lobsters/pull/598 . Note that I had to fix a couple of Rubocop issues to get that PR to pass.

pushcx:
On the bonus part, I meant getting the rubocop project itself to expand the SafeNavigation cop they ship now with a config setting to prohibit use of the SafeNavigation operator.

PR #598, opened 2018-11-24
integrate rubocop-disallow-safe-navigation

tansengming:
This finishes https://github.com/lobsters/lobsters/issues/596 by integrating a new custom Rubocop Cop to check for the safe navigation operator. Let me know what you think!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 # Project setup:
-require: rubocop-disallow-safe-navigation
+require: ./extras/rubocop/cop/style/disallow_safe_navigation
 Rails:
   Enabled: true
 AllCops:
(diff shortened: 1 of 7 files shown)


tansengming:
Note that I had to fix a `&.` problem for travis to pass 
https://github.com/lobsters/lobsters/pull/598/commits/cf9fdaf04df09812a67af15cc87d059f9984f894 so I guess that means it's working 😸

pushcx:
Did this strictly require a new, custom gem? I'd really prefer to avoid that. I'd hoped with `require` we could pull in the cop from this repo's `extras/` directory.

tansengming:
>  I'd really prefer to avoid that. I'd hoped with require we could pull in the cop from this repo's extras/ directory.

@pushcx I've updated the PR for this.

pushcx:
This is great, thank you. I hadn't realized `&.` had snuck in, so I'm glad to have rubocop guarding the door from now on.

PR #602, opened 2018-11-26
fix #599: skip robocop on server start in production env

### LINKED ISSUE #599: TypeError TypeError: superclass must be a Class (Module given)
Hey ! First, thank you for your work.

I've got this error since yesterday:

```bash
TypeError: superclass must be a Class (Module given)
  /lobsters/extras/prohibit_safe_navigation.rb:14:in `<module:Style>'
  /lobsters/extras/prohibit_safe_navigation.rb:3:in `<module:Cop>'
  /lobsters/extras/prohibit_safe_navigation.rb:2:in `<module:RuboCop>'
  /lobsters/extras/prohibit_safe_navigation.rb:1:in `<top (required)>'
```

Do you know how to fix it ?

Thank you

evexoio:
fixes #599 

Rails tries to use robocop in production env where robocop gem isn't available.

add a conditional to prevent using robocop in production.

### DIFF
--- a/extras/prohibit_safe_navigation.rb
+++ b/extras/prohibit_safe_navigation.rb
@@ -1,3 +1,6 @@
+require 'rails'
+return if Rails.env.production?
+
 module RuboCop
   module Cop
     module Style


guillaumebriday:
LGTM 👍

pushcx:
Why was the `require "Rails"` needed? I thought the problem was that Rails was loading it? In which case, why would `Rails.env` not exist?

Could you unindent this code and just do a `return if Rails.env.production?` guard clause at the top, please? It's already nested enough.

evexoio:
@pushcx `require "rails"` is need when rubocop is launched with `bundle exec rubocop`, there for rails won't be loaded.

check https://travis-ci.org/lobsters/lobsters/builds/459895212

> Could you unindent this code and just do a return if Rails.env.production? guard clause at the top, please? It's already nested enough.

Sure, I've pushed the changes.

guillaumebriday:
With just a return it throw this error : 

```
/extras/prohibit_safe_navigation.rb: /lobsters/extras/prohibit_safe_navigation.rb:3: Invalid return (SyntaxError)
```

The syntax : 

```ruby
unless Rails.env.production?
 # code
end
```

works fine

evexoio:
@guillaumebriday Oops, my bad. I ran it without making sure I'm in production env.

I've tried to return using one liners but it just won't work.

PR #603, opened 2018-11-26
make `rails test` run `rails spec` instead

### LINKED ISSUE #585: Make 'rails test' run our test suite
For a while Rails has defaulted to the command `rails test` for running the test suite. On our codebase, that runs zero tests and exits; you have to run `rspec` to run the suite. There must be some little piece of missing plumbing that would let us match the default.

evexoio:
fix #585 

make `rails test` run `rails spec` instead


pushcx:
Ah, what a lovely little fix. Thanks!

PR #604, opened 2018-11-26
Adding Puma as the default server

guillaumebriday:


evexoio:
@guillaumebriday Thanks for your pull request :)

Can you please provide which issue does this PR solve?

guillaumebriday:
Hey, hum none. Should I open one for that ?

After a month of using lobsters every day I wanted to open multiple PR.

So I will make few changes and open PRs if you want to check 👍

evexoio:
Not necessarily an issue, if you can just leave the reasoning behind the PR that would be great ❤️

btw was WEBrick doing any problems for you?

guillaumebriday:
Ok I will for the next PRs.

Nope but I like to have the closest default configuration from rails. Moreover, WEBrick is single threaded unlike puma.

More information here : https://devcenter.heroku.com/articles/ruby-default-web-server#why-not-webrick

evexoio:
@guillaumebriday makes sense, I just tested your PR locally, puma works great.

alanpost:
I cannot accept this patch as-is.

In production we use unicorn, deployed via a role in lobsters-ansible.git:

https://github.com/lobsters/lobsters-ansible/tree/master/roles/unicorn

Changing the worker would minimally involve replacing the unicorn role with a puma role, including all necessary due diligence demonstrating that change worked.  That said, I'm not aware of any open tickets, present problems, or pressing need to change our worker pool.  Doing the technical work outlined here is necessary, but not sufficient.

What problem are you solving for us here?

PR #610, opened 2018-11-28
Fix quoted-printable subject question-mark not being encoded

### LINKED ISSUE #396: Question marks must be encoded when formatted as quoted-printable in mail headers
https://github.com/lobsters/lobsters/blob/c8d2582fcd9b6074cd6925633c40a7108b8747d1/script/mail_new_activity#L16

Probably needs to be:

```ruby
l.gsub(/=*$/, "").tr(" ", "_").gsub('?', '=3F') << "?=" }.join("\n\t")
```

evexoio:
Fixes #396 

screen shot 2018-11-28 at 11 10 48 am

 TESTED ON THUNDERBIRD


evexoio:
@pushcx I did the separation to require the String class into the tests, now if I put the code back in script/mail_new_activity, It will be mixed with other code and `require` won't work.

Do you have any method in mind to test the String class while having it in the script/mail_new_activity file?

alanpost:
The script itself could recognize a --test flag that short-circuits the regular codepath and runs unit tests.  The body of the script could also (I believe?) be moved to a function that is only executed when the script is invoked but not executed when the script is require'd.

I'll let @pushcx comment on the best approach here.

evexoio:
@pushcx Thanks, I could require the file with ease 👍

pushcx:
I hope this nails this frustrating bug, thanks!

PR #612, opened 2018-11-29
return empty array when there's no similar stories

### LINKED ISSUE #611: similar_stories returning empty array
I was trying to edit a story as an admin, I got this error.

screen shot 2018-11-29 at 11 21 53 am

Possible fix is to change
`@_public_similar_stories ||= similar_stories.base`
to
`@_public_similar_stories ||= similar_stories.base unless similar_stories.empty? else []`

evexoio:
fixes #611

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -236,7 +236,7 @@ class Story < ApplicationRecord
   end
 
   def public_similar_stories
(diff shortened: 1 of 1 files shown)


pushcx:
Nice way to fix this by making the method consistent, thanks.

PR #613, opened 2018-11-29
Migrate deprecated form_tag and form_for to form_with

### LINKED ISSUE #527: Convert form_for & form_tag into form_with.
This is a migration due to changes in rails 5.1.

The original form specifiers will be [replaced in the future](https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a). Finally, you will probably want to convert them into remote forms, or specify `local: true`. Then it will be safe to enable `config.action_view.form_with_generates_remote_forms`.

evexoio:
Fixes #527

I've tested every view while migrating it, additional testing would still be appreciated 👍

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,7 @@
 
 # Project setup:
 require: ./extras/prohibit_safe_navigation
+require: ./extras/prohibit_form_for_and_form_tag
 Rails:
   Enabled: true
 AllCops:
--- a/.ruumba.yml
+++ b/.ruumba.yml
@@ -1,0 +1,9 @@
+require: ./extras/prohibit_form_for_and_form_tag
+
+AllCops:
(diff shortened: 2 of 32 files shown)


evexoio:
@pushcx I'll look into adding a robocop cop that requires the use of `form_with`.

Thanks for looking into this, if you need further help on this one please let me know.

evexoio:
@pushcx I pushed a rubocop cop to disallow the use of `form_tag` and `form_for`,
I installed a new gem called ruumba, its an extension of ruumba for .erb files.
can you please add `ruumba -D -e app/views -c .ruumba.yml` to the CI?

pushcx:
Wow! I expected it existed already and we could deal if not, thanks for figuring out how to check the views and write one.

PR #617, opened 2018-12-10
Fixing production environment 

guillaumebriday:


pushcx:
I caught this independently when I pushed the change backlog yesterday. So thank you for this fix (I did the exact same thing!) but we don't need to merge it.

I'm back from work travel + having the flu, so I'm catching up on the PR and issue queue.

PR #619, opened 2018-12-11
Make /recent leader text more accurate

enaeher:
The old text implied that this page contains only those stories which
have not reached the front page. In fact, it contains all recent
stories in order of submission, regardless of whether they are or are
not on the front page.


pushcx:
Huh. If `/recent`/ includes stories that have made the front page, it's a bug. I refactored the `StoryRepository` and some queries not too long ago, it's possible this broke. Investigating.

pushcx:
Yeah, this is a bug introduced in 0ebe0663d2ac479c6808800ffbb989b8f9448c92.

enaeher:
OK, I didn't realize it was actually meant to work that way. I'll close this.

PR #626, opened 2018-12-24
Notes mysql2 gem installation workaround for macOS

colindean:
There is apparently a bug in the `mysql_config` script included in `mysql-connector-c` that the mysql2 gem uses to build the native extensions. This theoretically would trip up anyone using macOS 10.12+ without having MySQL or MariaDB running locally.


pushcx:
Nice little doc improvement, thank you!

PR #625, opened 2018-12-24
Drop erroneous hash member to fix editing comments

colindean:
This caused the content of `comment.comment` to go into a `body` HTML attribute instead of the body of the textarea. This restores the functionality prior to 089f3475.

Fixes #624

PR #628, opened 2018-12-26
Fix commentbox text_area

### LINKED ISSUE #624: 'Edit' button seems broken
When I click `edit` on a comment, I'm presented with the comment's textarea but the comment is not in the textarea.

I'm seeing this on Chrome Version 71.0.3578.98 on macOS 10.14.1 as well as Chrome on Android 9 (latest of both, presumably).

evexoio:
This caused the content of `comment.comment` to go into a `body` HTML attribute instead of the body of the textarea. This restores the functionality prior to 089f3475.

Closes #624 #625

### DIFF
--- a/app/views/comments/_commentbox.html.erb
+++ b/app/views/comments/_commentbox.html.erb
@@ -6,13 +6,14 @@ data-shortid="<%= comment.short_id if comment.persisted? %>">
   <% end %>
 
   <%= f.hidden_field "story_id", value: comment.story.short_id %>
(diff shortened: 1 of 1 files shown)


pushcx:
OK, I finally had the time to sit down and tinker with this for a while. Thanks for running this down. I can see we're going to have some hassles with the patch/post thing in the future...

evexoio:
@pushcx 👍 thanks for taking the time to merge this :)

PR #629, opened 2018-12-30
Fake Data: Hats, Comments

nikhiljha:
Small PR to add some more fake data. I'm learning Rails and making progress on #414 at the same time!

(Also I reduced the number of stories, 1000 takes forever and doesn't make the site look any more realistic.)


pushcx:
Great, thanks for making progress to help folks develop on the codebase. Good call on reducing the number of stories.

Issue #554, opened 2018-09-26
Remove jQuery


pushcx:
In the time the site's been around, cross-browser js functionality and compatibility has improved to the point that we can drop jQuery as a dependency.

We only have ~600 lines of JavaScript and none of it's very complex. We don't serve JS to logged-out users, but for logged-in users it adds up to 140 KB (45KB compressed) of data (standard Rails asset caching, so caches very well). Removing jQuery is not a huge win, so this is pretty far down on my to-do list. Still, it's come up in chat a couple times in the last few months and I thought I'd post it for anyone who might like the exercise of replacing jQuery with no-framework js.

We have two libraries in use. Autosize resizes textareas rather than scroll them and doesn't depend on jQuery - so there's no changes here. Select2 is used for filtering and tab-completing the tag dropdown when creating and editing story changes. It depends on jQuery and will need to be [rewritten](https://github.com/select2/select2/issues/3652) or replaced. (Ideally with identical behavior, because we haven't touched it in six years and submitters have muscle memory for it.)

daau:
[Choices](https://github.com/jshjohnson/Choices) seems like a good Select2 alternative which doesn't depend on jQuery.

pushcx:
From their demo, Choices looks like a good fit, thanks for the suggestion. We'd have to add a hook so that tab completes the currently-selected choice, but that's probably not too bad.

daau:
Great! I'll submit a PR shortly with this feature.

Carpetsmoker:
I did some work on s/select2/choices/: https://github.com/Carpetsmoker/lobsters/tree/choices

Can't say I'm too impressed with it. The API is annoying. The documentation isn't very well written, and several things are non-obvious how to do (like not have it muck about with the width of the input), it's a lot larger than select2 (56k vs. 213k), and I had to heavily modify the shipped CSS to make it vaguely behave.

I see if I can find something better later, or maybe I'll write my own simple function. It's not a very hard problem and probably would have finished it by now instead of making that Choices thing behave.

drakmail:
How about https://github.com/yairEO/tagify ?

Carpetsmoker:
Thanks. I spent some time with that as well, and eventually just gave up.

I did some other work using "pure" JavaScript a while ago, and the experience wasn't a happy one. The DOM API offends a lot of my sensibilities on how a good API should work, which is basically just a nice way of saying that I think it's pretty terrible. `elem.parentNode.removeChild(elem)`, seriously?

[...]

PR #630, opened 2018-12-31
Remove jQuery from application.js.erb

chrislloyd:
Why is addressed in #554. My approach here is to make this a slightly larger PR broken into individual atomic commits. Adding test plans for each commit considering that the JS codebase isn't particularly well tested. If this becomes too big (or I abandon it) I'll just cut the scope of the PR to whatever I get done.

TODO:

- [ ] Remove explicit jQuery references from `application.js.erb`
- [ ] Remove `jquery_class.js`
- [ ] Replace `select2.js`

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -4,6 +4,34 @@
 
 "use strict";
 
+// Polyfill Element.matches()
+// https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#polyfill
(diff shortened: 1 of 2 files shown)


pushcx:
Hi Chris,

It looks like this hasn't been touched in a long while - do you plan to take it back up in the future, or should we close out this PR?

chrislloyd:
Closed. Would love to get back around to it but don’t have a lot of time to spare.

PR #644, opened 2019-02-23
Use conversations instead of individual messages

### LINKED ISSUE #226: Merge Received and Sent into a unified "Conversations"
It'd be nice if we could see Received/Sent as part of a unified single conversation. A conversation would be initialized by one user sending a message to another user, and any replies to that conversation would appear one after another.

edwardloveall:
Closes #226 

- [x] Add deleted_by_author and deleted_by_recipient
- [x] Only fetch non-deleted conversations
- [x] Convert all messages into conversations
- [x] Figure out what hats have to do in conversations
- [ ] Delete old message code?
- [x] Mark conversations as unread if message inside of them are unread by the user
- [x] Update (touch) the conversation when a message is created
- [x] Add short_ids to conversations
- [x] Create mod note on a message `app/views/messages/_form.html.erb`
- [x] The current author and recipient are not correct half the time. It uses the conversation author and recipient, but it needs to respond to who's actually talking.
- [x] Ability to mark conversations as deleted by author/recipient
- [x] Auto-delete conversations once both sides have marked them as deleted
- [x] Remove testing file `create_messages.rb`


app/models/conversation.rb
```diff
@@ -0,0 +1,28 @@
+class Conversation < ApplicationRecord
+  belongs_to(
+    :author,
+    class_name: User.name,
```
pushcx: You have to use a string like `"User"`, because if `Conversation` loads `User` and `User` loads `Conversation` the Rails autoloader breaks.

app/models/conversation.rb
```diff
@@ -0,0 +1,28 @@
...
+    :recipient,
+    class_name: User.name,
+    foreign_key: "recipient_user_id",
+  )
+  has_many :messages
+
+  validates :short_id, presence: true, uniqueness: true
+  validates :subject, presence: true
+
+  scope :involving, ->(user) do
```
pushcx: Nice scope. I'm glad to see fewer and fewer direct `where` calls outside of models.

app/views/conversations/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="box wide">
+  <ul>
+    <% @conversations.each do |conversation| %>
+      <li class="conversation <%= conversation_status_classes(conversation) %>">
+        <%= link_to conversation.partner(of: @user).username, conversation_path(conversation) %>
```
pushcx: This `partner` method reads very nicely.

config/routes.rb
```diff
@@ -95,6 +95,8 @@
   get "/comments/page/:page" => "comments#index"
   get "/comments" => "comments#index"
 
+  get "/conversations" => "conversations#index", as: :conversations
```
pushcx: If this can be `resource :conversations, only: [:index, :show]`, you don't need to follow the convention of the rest of this file.

pushcx:
This is off to a nice start! I tossed in some little comments. Nice to get a little closer to the Rails resource/rest convention, and thanks for starting with a test.

edwardloveall:
Thanks for the comments! I'm working through some failing tests at the moment trying to keep messages intact enough to convert to conversations. I'll keep pushing progress.

app/models/conversation.rb
edwardloveall: Done in ab986b0

config/routes.rb
edwardloveall: Done in ab986b0

edwardloveall:
@pushcx I had some questions that would help guide this feature along, whenever you get the chance:

* Do you have a list of browsers you support for styling purposes? Asking because I'm using [CSS grid](https://caniuse.com/#feat=css-grid) which has about 85% global browse share, but maybe not for lobste.rs
* How many messages are currently in the system? Asking to start thinking about how I'll convert messages to conversations
* When we do convert messages to conversations, I wonder how we'll make sure that message are all linked to a conversation. I have some proofs of concept that group by authotrecipient/normalized subject because that's the only way I can think of. Am I missing anything?
* We may have to make this transition in two steps, creating conversations with one migration, then later removing old fields from messages. Seem right to you?

alanpost:
> Do you have a list of browsers you support for styling purposes? Asking because I'm using CSS grid which has about 85% global browse share, but maybe not for lobste.rs

Excluding crawlers, RSS feeds, and unrecognized or miscellaneous browsers, here is a summary from our logs over the past ~year (13 Jan 2018 through 09 Jan 2019).  I've removed the tail from each browser once absolute traffic fell below 100k hits for the year or total percentage of hits fell below 0.5%

     Hits         h%    Vis.     v% Data                                                                                                                 
     -------- ------ ------- ------ ----                                                                                                                 
     50239586 37.33% 1264374 23.89% Firefox                         |||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
     8406300   6.25%  190226  3.59%  ├─ Firefox/60.0                ||||||||||||||                                                                       
     8073598   6.00%  183295  3.46%  ├─ Firefox/61.0                |||||||||||||                                                                        
     7280739   5.41%  161800  3.06%  ├─ Firefox/59.0                ||||||||||||                                                                         
     7229524   5.37%  145701  2.75%  ├─ Firefox/63.0                ||||||||||||                                                                         
     4073410   3.03%  107249  2.03%  ├─ Firefox/58.0                ||||||                                                                               
     3696845   2.75%   84411  1.60%  ├─ Firefox/62.0                ||||||                                                                               
     3677211   2.73%   75840  1.43%  ├─ Firefox/64.0                ||||||                                                                               
     2569622   1.91%   68139  1.29%  ├─ Firefox/52.0                ||||                                                                                 
     1592801   1.18%   48446  0.92%  ├─ Firefox/57.0                ||                                                                                   
     755554    0.56%   15631  0.30%  ├─ Firefox/65.0                |                                                                                    
     537831    0.40%   14389  0.27%  ├─ Firefox/56.0                |                                                                                    
     447495    0.33%    8662  0.16%  ├─ Firefox/45.0                |                                                                                    
     240465    0.18%   49381  0.93%  ├─ Firefox/40.1                |                                                                                    
     179498    0.13%    4037  0.08%  ├─ Firefox/66.0                |                                                                                    
     166754    0.12%    6951  0.13%  ├─ Firefox/50.0                |                                                                                    
     135315    0.10%     115  0.00%  ├─ Firefox/1.0                 |                                                                                    
     114211    0.08%    4169  0.08%  ├─ Firefox/52.9                |                                                                                    
     102232    0.08%    6266  0.12%  ├─ Firefox/38.0                |                                                                                    
     84141     0.06%    4702  0.09%  ├─ Firefox/54.0                |                                                                                    
     71465     0.05%    2727  0.05%  ├─ Firefox/12.0                |                                                                                    
     67606     0.05%    6993  0.13%  ├─ Firefox/53.0                |                                                                                    
     60155     0.04%    2843  0.05%  ├─ Firefox/55.0                |                                                                                    
     56839     0.04%    1269  0.02%  ├─ Firefox/36.0                |                                                                                    
     55622     0.04%    2652  0.05%  ├─ Firefox/31.0                |                                                                                    
    
     Hits         h%    Vis.     v% Data                                                                                                                 
     -------- ------ ------- ------ ----                                                                                                                 
     33986456 25.26% 1302688 24.62% Chrome                          |||||||||||||||||||||||||||||||||||||||||||||||||||||||||                            
     2500174   1.86%   79687  1.51%  ├─ Chrome/65.0.3325.181        ||||                                                                                 
     2279193   1.69%   73100  1.38%  ├─ Chrome/67.0.3396.99         |||                                                                                  
     1849359   1.37%   81606  1.54%  ├─ Chrome/67.0.3396.87         |||                                                                                  
     1646308   1.22%   49341  0.93%  ├─ Chrome/68.0.3440.106        ||                                                                                   
     1575861   1.17%   56004  1.06%  ├─ Chrome/63.0.3239.132        ||                                                                                   
     1464712   1.09%   49700  0.94%  ├─ Chrome/70.0.3538.110        ||                                                                                   
     1423495   1.06%   46023  0.87%  ├─ Chrome/69.0.3497.100        ||                                                                                   
     1352676   1.01%   47549  0.90%  ├─ Chrome/66.0.3359.181        ||                                                                                   
     1275168   0.95%   37986  0.72%  ├─ Chrome/66.0.3359.139        ||                                                                                   
     1208253   0.90%   36812  0.70%  ├─ Chrome/70.0.3538.77         ||                                                                                   
     1156773   0.86%   38715  0.73%  ├─ Chrome/71.0.3578.98         |                                                                                    
     1016363   0.76%   32571  0.62%  ├─ Chrome/64.0.3282.186        |                                                                                    
     876612    0.65%   42094  0.80%  ├─ Chrome/65.0.3325.109        |                                                                                    
     818587    0.61%   23880  0.45%  ├─ Chrome/70.0.3538.102        |                                                                                    
     680098    0.51%   22473  0.42%  ├─ Chrome/64.0.3282.167        |                                                                                    
     675203    0.50%   29950  0.57%  ├─ Chrome/70.0.3538.80         |                                                                                    
    
     Hits         h%    Vis.     v% Data                                                                                                                 
     -------- ------ ------- ------ ----                                                                                                                 
     16724598 12.43%  757279 14.31% Safari                          ||||||||||||||||||||||||||||                                                         
     9770417   7.26%  476404  9.00%  ├─ Safari/604.1                ||||||||||||||||                                                                     
     3774643   2.80%  145878  2.76%  ├─ Safari/605.1.15             ||||||                                                                               
     879634    0.65%   34026  0.64%  ├─ Safari/604.5.6              |                                                                                    
     704684    0.52%   24376  0.46%  ├─ Safari/602.1                |                                                                                    
     298180    0.22%   12472  0.24%  ├─ Safari/604.4.7              |                                                                                    
     242146    0.18%    7474  0.14%  ├─ Safari/601.1                |                                                                                    
     147816    0.11%     889  0.02%  ├─ Safari/533.18.5             |                                                                                    
     132408    0.10%    6126  0.12%  ├─ Safari/538.1                |                                                                                    
     102861    0.08%    3608  0.07%  ├─ Safari/603.3.8              |                                                                                    
     99669     0.07%    6207  0.12%  ├─ Safari/9537.53              |                                                                                    
     70238     0.05%    3012  0.06%  ├─ Safari/600.1.4              |                                                                                    
     55768     0.04%    2827  0.05%  ├─ Safari/537.21               |                                                                                    
    
     Hits         h%    Vis.     v% Data                                                                                                                 
     -------- ------ ------- ------ ----                                                                                                                 
     1266635   0.94%   58325  1.10% MSIE                            ||                                                                                   
     389131    0.29%    4831  0.09%  ├─ Edge/17.17134               |                                                                                    
     177626    0.13%    4110  0.08%  ├─ Edge/16.16299               |                                                                                    
     176087    0.13%    9414  0.18%  ├─ MSIE/11.0                   |                                                                                    
     162518    0.12%    8354  0.16%  ├─ MSIE/7.0                    |                                                                                    
     64881     0.05%    1124  0.02%  ├─ Edge/14.14393               |                                                                                    
    
     Hits         h%    Vis.     v% Data                                                                                                                 
     -------- ------ ------- ------ ----                                                                                                                 
     883915    0.66%   42879  0.81% Opera                           |                                                                                    
     364965    0.27%   27338  0.52%  ├─ Opera/537.36                |                                                                                    
     181788    0.14%    3255  0.06%  ├─ Opera/12.16                 |

alanpost:
> How many messages are currently in the system? Asking to start thinking about how I'll convert messages to conversations

Fewer than 5k.

edwardloveall:
Thanks @alanpost!

Of the hits you gave me (representing 76.62% of all your traffic) about 82.99% of that traffic supports the grid properties (if I copied the data over correctly). That's 63.59% of all traffic to lobste.rs, which surprised me! I thought it would be higher as the [global reported stats on caniuse.com](https://caniuse.com/#feat=css-grid) are 85.75%.

I may go for flex-box instead as that will have [a bit better browser coverage](https://caniuse.com/#feat=flexbox) and likely serve my purposes just as well.

Please let me know if I'm missing anything here.

alanpost:
> That's 63.59% of all traffic to lobste.rs

Of the traffic I elided the majority of it are bots and RSS readers, bringing us closer to your 85.75% global stat.  ~80% is the absolute minimum I would want to see for coverage.  Given that we're on that margin I like the idea you present of using flex box instead of CSS grid.  If that turns out to present a maintenance or development burden we can resample just the prior quarter, look at user counts rather than hits, or limit ourselves to requests involving direct/private messages.  Precision will only change that number so much, however.  Let me know if you encounter any unexpected problems.

pushcx:
Showing my gray hair here, but why not continue using `<table>` to display the tabular list of conversations?

I think we can do it in one migration. Your strategy to `group by` author/recipient pairs and normalized subject sounds exactly right to me. For a couple thousand records it'd be overkill to try to update the data in a single SQL query, might as well select out the messages, normalize, save them back. (Remember that Rails migrations shouldn't reference models because that'll break in the future if the model is renamed/removed.) So I guess the migration looks like:

1. add `conversations` able
2. add `conversation_id` to `messages`, nullable
3. select all the messages, loop them to find or create the `conversation`, save that `conversation_id`
4. make `messages.conversation_id` `not null` and drop the extra fields from `messages`

edwardloveall:
It's a fair question @pushcx! 😄 I think the conversation list is tabular to use a `<table>`. Barring any unforeseen styling weirdness, I'll do that. There's still the conversation view itself which I don't think it's appropriate to display with a table.

Thanks for the guidance on the migration. I'll brush up on my SQL and try to get something up soon.

Sorry this is going slowly. I don't have a ton of time to dedicate to it. Thanks for your patience.

pushcx:
There's no rush, nothing to apologize for. I appreciate you volunteering.

edwardloveall:
I updated the migration that takes a stab at this. It seems to be pretty good, but I'm only using test data. I'd love it if someone could try it with real data.

alanpost:
I've deployed your branch to our test system.

In the Conversations area, the column From has been renamed to Partner.  That should go back to being the From column.

I notice receiving a message doesn't change the sort order.  There is also not a message count, leaving one to read the Received field to find recent, read messages.

I propose then to rename the received field to last modified, and sort the page descending so the most recent conversations that have received messages are near the Create Conversation port of the page.

We may also want a created field as a complement to the last modified field, but I don't have a use case for that so I'm hesitant bringing it up.

edwardloveall:
Thanks for the suggestions @alanpost! UI is definitely still a work in progress and I even have some of your ideas already on my list 🙂 Sounds like the data migration worked alright though?

alanpost:
> Sounds like the data migration worked alright though?

We need to do that still; in an abundance of excitement I both wanted to take your branch for a test drive and make it possible (via installing it on the test system) to provide a platform for multiple people to QA together.  Particularly given the nature of this change.

edwardloveall:
Gotcha. I'll be working on more of those features soon so hopefully you'll have something to test. I'm glad your excited!

app/models/conversation.rb
```diff
@@ -0,0 +1,42 @@
+class Conversation < ApplicationRecord
+  before_validation :assign_short_id, :on => :create
+
+  belongs_to(
```
pushcx: Style nit: check out the top of `app/models/user.rb` to see how these are formatted.

db/migrate/20190218153814_create_conversations.rb
```diff
@@ -0,0 +1,99 @@
...
+    drop_table :conversations
+  end
+end
+
+class MessageConverter
+  include ActiveRecord::ConnectionAdapters::Quoting
+
+  def run
+    config = ActiveRecord::Base.configurations[Rails.env].symbolize_keys
+    mysql = Mysql2::Client.new(config)
```
pushcx:
```suggestion
    mysql = ActiveRecord::Base.connection
```

pushcx:
I ran the migration locally on a backup of prod data, it ran cleanly and my messages look totally correct. Really nice to read a conversation on a single page, I'm looking forward to this landing.

app/views/conversations/index.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      link_to(
+        conversation_path(conversation),
+        class: "conversation #{conversation_status_classes(conversation)}"
+      ) do
+      %>
+        <span class="partner">
+          <%= conversation.partner(of: @user).username %>
+        </span>
+        <span class="received"><%= time_ago_in_words_label(conversation.last_message.created_at) %></span>
+        <span class="subject"><%= conversation.subject %></span>
```
pushcx: Please include `conversation.messages.count` somewhere in this table (which is a a 1 + n query, so needs to be preloaded).

db/migrate/20190218153814_create_conversations.rb
edwardloveall: Thanks this is much better.

app/views/conversations/index.html.erb
edwardloveall: Done. I added a `preload(:messages)` but after doing some quick perf tests (just looking at loading times in the log) I didn't see a difference between preloading and not. I also didn't get any kind of warning from bullet. If you know something I don't, I'll happily put a preload in there 🙂

edwardloveall:
@alanpost Partner is back to `From` in 1347417, conversations are ordered by recently received messages at the top in aa47240. Although now that I read your suggestion again, did you mean to the conversations with recent messages to be at the bottom of the page?

> I propose then to rename the received field to last modified

I went with `Last Messages` instead of `Last Modified` because I felt that most users probably aren't thinking about the underlying conversation model and when it was modified, they're thinking about new messages. Again, easy to change if you feel strongly 🙂

edwardloveall:
I wanted to check on deletion. Do we want conversation deletion? That's fine but then I have a couple more questions:

* If a single message in a "thread" is delete but others aren't, should I resurface that message in the conversation, or just never assign it to a thread, effectively deleting it?
* If a user deletes a conversation, but their conversation partner does not, and then the partner sends another message, should it resurface the conversation for the user that deleted it?

---

Regardless of that, I still need to add hats and mod notes to conversations, fix up the CSS so it's using flexbox instead of css grid and remove old message code, then I think this is ready for a full review.

app/views/conversations/index.html.erb
pushcx: If you don't see a `select count...` query in the output of your `rails server` for every conversation, and bullet isn't complaining there, you're fine.

(Though maybe the query output I'm used to seeing comes from my `~/.pryrc`?)

```
require 'logger'
if Object.const_defined? 'Rails'
  Rails.logger = Logger.new STDOUT
end
# print SQL to STDOUT
if Object.const_defined? 'ActiveRecord'
  ActiveRecord::Base.logger = Logger.new STDOUT
  ActiveRecord::Base.clear_reloadable_connections!
end
```

pushcx:
Hm, thanks for laying out the corner cases of deleting, that's tricky. How does this sound: move the `deleted_by_author`/`recipient` flags up to `conversations`, and add `_at` to their name, making them timestamps. So rather than flipping a boolean to true, set it to the timestamp that someone clicks "delete" on the conversation. When showing the list, only `find` conversations where `updated_at > deleted_by_author_at`/`recipient_at`; and in a single conversation only select messages newer than the timestamp. (`Message#check_for_both_deleted` will also need to move up to `Conversation`.)

app/views/conversations/index.html.erb
edwardloveall: Grabbed your pryrc No bullet, but I do see the count for every query.

```sql
SELECT COUNT(*) FROM `messages` WHERE `messages`.`conversation_id` = 16
```

I just can't get a performance difference between preloading and not with ~100 conversations for a user. If there's another thing I should test, let's do it!

edwardloveall:
Clever! I'll try that out.

edwardloveall:
Alright, I think this is ready to be looked at so I removed its draft mode status. A couple of notes:

* I couldn't make a foreign key on the conversations for `author_user_id` or `recipient_user_id` because the primary key was `bigint(20) unsigned` on the `users` table but only `bigint(20)` on the foreign key column.
* I did add one on messages for `conversation_id` though because they matched
* Hats remain on messages instead of moving up to conversations because I'm not super confident I got the hat part correct and it took way less time
* I'm not sure what a mod note should look like on a conversation so it also remains on a messsage

And some things I wasn't sure that I didn't want to spend time on unless y'all thought they were important: 
* I didn't remove any message code, columns, or routes. Since this is such a big change, I was wary of spending my time on being destructive. I figure you'll want to deploy this, make sure no one has problems, then remove it after. However, if you all want me to remove stuff, I'll give it a shot!
* You can't make mod notes from existing messages, only new ones
* You can't bulk edit conversations
* You can't mark conversations as unread

app/models/conversation.rb
```diff
@@ -7,7 +7,7 @@ class Conversation < ApplicationRecord
   belongs_to :recipient,
              class_name: "User",
              foreign_key: "recipient_user_id"
-  has_many :messages
+  has_many :messages, dependent: :destroy
```
pushcx: Yep, good catch.

pushcx:
I ran out of review time this morning, will try to make time an evening this week.

edwardloveall:
Thanks @pushcx! I'll be out of town over the weekend so no rush.

app/views/conversations/index.html.erb
pushcx: I wouldn't expect any performance difference in your local dev, but it'll cause problems in prod because of the sheer number of queries all kicking off at once while there's other traffic on the database. I'll take a look on Wednesday on how to preload the count properly; I always have to tinker to get the right syntax.

pushcx:
Regarding the foreign key over to `users.id` - they column definitions should match. We did a big cleanup in #538 but this one must've slipped through the cracks. Let's update the columns to match `users.id` in this migration so we can use foreign keys.

Regarding everything else, I'll reread + experiment on Wednesday. :)

pushcx:
Of course, now that I reread #561 I remember why there isn't a foreign key. But we can still match the column definition properly. Diving in to the rest of the code...

edwardloveall:
Hey all! I wanted make sure this doesn't get lost to time. I rebased off the current master so it's still ready to be reviewed 🙂

briankung:
I can take a look if you want a reviewer? The only thing that looks odd to me are the tabs in `application.css`, but I guess that's in keeping with the rest of the file. Otherwise, the tests and the code looks good. I'm pulling it down for manual testing now.

pushcx:
Please do, I'm happy to have the help. I plan to look at this tomorrow morning at my usual Lobsters coding time otherwise.

briankung:
Some notes:
- BUG: A user can't send messages by visiting someone's profile (the old messages code barfs when it butts up against [requiring a conversation to exist](https://user-images.githubusercontent.com/2836167/58487294-f1fdd880-812c-11e9-90d3-cb337b5704f9.png)). I think that would be my primary entrypoint for using this feature.
- BUG?: Deleted messages reappear if the other user continues to send messages
- BUG: Conversations from other users appear in the list of conversations (and just as a sanity check, the user I'm using to view the conversations isn't a moderator) EDIT - for whatever reason, I can't recreate this anymore, even after blowing the database away and recreating everything
- BUG: I can also add messages to a conversation that I wasn't originally a part of. EDIT - see above note about not being able to recreate anymore
- SUGGESTION: it might be nice to autocomplete user names in the "to" field. MAYBE IT'S TIME FOR A DATALIST https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist#Examples. I just found out about these, so I'm excited 😂
- BUG: Messages that are too long (the limit seems to be around `> 2^16` bytes, regardless of the limit in the schema) fail silently
- BUG: Messages with titles that are too long throw a 500 with `ActiveRecord::ValueTooLong in ConversationsController#create`
- BUG?: Users can send a messages to themselves
- BUG?: Users can force the creation of Moderator Notes with [Web-Parameter Tampering](https://www.owasp.org/index.php/Web_Parameter_Tampering) i.e. adding `<input name="message[mod_note]" type="hidden" value="1">` to the message form
- BUG (but not really important): Trying to delete someone else's conversation with parameter tampering results in hitting the equivalent of this code in `conversations_controller#destroy` which throws a 500:
    ```
    if @conversation.present?
      # ...
    else
      flash[:error] = "Could not delete message"
      redirect_to conversation_path(@conversation) # but @conversation is nil!
    end
    ```
    Something similar happens with injecting a different hat id.

Well, that's about it for now.

edwardloveall:
Thank you @briankung! I'm taking a look at these now 👍

briankung:
I'd be happy to assist if you'd like to split up the work. We'd just need to figure out how to do the splitting.

pushcx:
To help with the contextual ones - yeah, all the "BUG?" items should be considered bugs. Thanks for searching out all these corner cases, Brian.

I'd rather not use a datalist, as we'd have to send down the list of all 10k+ users. This would be ~250kb of `<option>` tags.

app/models/conversation_creator.rb
```diff
@@ -0,0 +1,45 @@
...
+          create_modnote: create_modnote?(message_params),
+        )
+      end
+    end
+  end
+
+private
+
+  def recipient(username)
+    @_recipient ||= User.find_by(username: username)
```
briankung: Is there an underscore on this instance variable because it's not used directly?

app/views/conversations/_hat_inputs.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if user.wearable_hats.any? %>
```
briankung: Note: unlikely to ever occur, but a moderator without wearable hats won't get the option to add a modnote

app/models/conversation_creator.rb
edwardloveall: Correct, it's a convention I picked up.

edwardloveall:
> BUG?: Deleted messages reappear if the other user continues to send messages

I think this one is not a bug. It's a weird corner case of how do you delete conversations when two parties are involved. We discussed it here: https://github.com/lobsters/lobsters/pull/644#issuecomment-481713399

app/views/conversations/_hat_inputs.html.erb
edwardloveall: Yup, I should use the `is_moderator` flag instead.

edwardloveall:
I addressed a couple of these changes today, and also did a rebase to bring the commits up to date. I still need to address @briankung's bugs around message body and title length, but those are the only problems I know about.

edwardloveall:
Oh, and of note: I think my mysql version is newer than everyone else's because it keeps trying to add a `COLLATE=utf8mb4_0900_ai_ci` to all my tables whenever I migrate or rollback. I'm happy to add this or just keep rolling those lines back.

I also see that since there's a new migration, technically my migration will come after the new one. I'll rename it so it come sequentially afterward in the DB.

edwardloveall:
I did a significant refactor here to fix the issue @briankung found. Invalid messages:
* Should not save the parent conversation
* Should show the validation errors so users can fix them

This involved moving the creator objects to become form objects. I think overall this is nicer, although a bit less straight forward.

pushcx:
I haven't forgotten this, it's just a little larger than I generally have time free for. Sorry for the delay; I still want to get this merged soon.

edwardloveall:
No worries @pushcx and thanks for the checkin. If it would be helpful to walk through it in real time (for whatever reason) lemme know. I get that it's just big, also.

app/models/user.rb
```diff
@@ -206,11 +206,19 @@ def disable_invite_by_user_for_reason!(disabler, reason)
...
 
-      msg = Message.new
+      conversation = Conversation.new
+      conversation.author_user_id = disabler.id
+      conversation.recipient_user_id = self.id
+      conversation.subject = "Your invite privileges have been revoked"
+      conversation.save!
+      msg = conversation.messages.new
+
+      # TODO: Remove these 3 lines after message no longer has these fields
```
pushcx: Time for these fields to go, the migration works. :)

app/views/conversations/index.html.erb
pushcx: I looked it up. You'll need to add a `messages_cache` column to `Conversations` and [`counter_cache` to the `belongs_to`](https://guides.rubyonrails.org/v5.2/association_basics.html#options-for-belongs-to-counter-cache), then Rails will keep the count up-to-date without doing a `count(*)` query for every conversation.

app/controllers/conversations/messages_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        render "conversations/show"
+      end
+    end
+
+  private
+
+    def message_params
+      params
+        .require(:message)
+        .permit(:body, :hat_id, :mod_note)
```
pushcx: Only moderators should be permitted to set the `mod_note` param. There's an example of this in the old `MessagesController`.

pushcx:
Made time to experiment. It's looking *great* and I think we're down to the last few tweaks.

 * Needs a way to modnote incoming messages. A link or button below the username of the sender would be perfect.
 * Remove old /messages controller and views, and have their routes redirect to the appropriate place in /conversations.
 * `ConversationsController#show` needs to mark all messages in the conversation as read before calling `@user.update_unread_message_count!`.

app/controllers/conversations/messages_controller.rb
edwardloveall: I believe we have that logic here: https://github.com/lobsters/lobsters/pull/644/files#diff-3ac3bb9392913059755d80f258ad66ffR37

Care to confirm?

edwardloveall:
>Needs a way to modnote incoming messages. A link or button below the username of the sender would be perfect.

Apologies, but I'm not sure I understand how this would work. Right now, there's a Mod Note checkbox moderators can enable when responding to a conversation. What would the button do? Would it take you to a different page with a mod note form? Sorry, it's probably a silly question. I'm not a mod and don't really understand mod notes 🙂 

> `ConversationsController#show` needs to mark all messages in the conversation as read before calling `@user.update_unread_message_count!`.

I believe we're doing this already [here](https://github.com/lobsters/lobsters/blob/7c51edcc6e7c42b1e926e25affc2a3cfab8034ee/app/controllers/conversations_controller.rb#L13-L17). Am I misunderstanding?

Removing old message stuff is going to be a blast!

pushcx:
I got to spend a half-hour with this this morning and it looks great. Unless work runs long, I'm going to spend time this evening to merge this by doing the fiddly little things that are quicker to do than explain. :D You're right about that 'marking read' issue, and I'll take care of mod notes.

edwardloveall:
Awesome @pushcx. Sorry this dragged out for so many months. Please do let me know if there's anything you'd like me to fix up

edwardloveall:
Bumping this again. Let me know if there's anything you'd like me to do.

pushcx:
Can you give me write access to your repo? I've hacked up some stuff on top of this branch and have notes for more small stuff, I can at least push that.

edwardloveall:
Done!

pushcx:
OK, there's a pile of commits. The last two things on my todo list are:

  fix 'delete conversation' button and tests
  fix conversations index 1 + n

edwardloveall:
Hi! It's been a bit. Is there anything I can do to help move this along? I'd hate to see it get too far out of date and then abandoned. Let me know!

edwardloveall:
I'm going to guess that this won't get merged since it's been 3+ years. I'll close this and someone else can pick this up in the future.

Issue #645, opened 2019-03-05
link to inline images


pushcx:
If someone uses the markdown for an inline image, it's rendered to an image tag and then removed from the html (see `extras/markdowner.rb`). That works, but is confusing output, especially for old stories when inline images were permitted. ([more context](https://lobste.rs/s/lepnje/i_made_programming_language_based_on#c_11i2ac))

Instead of removing the image tag, let's convert `<img src="[url]">` to `<a href="[url]">[url]</a>`. I prefer this conversion happen to the HTML rather than to the Markdown to avoid a corner case where we don't parse the image markup/node the same as CommonMarker, or that it could sneak in as a raw html tag.

PR #647, opened 2019-03-06
Replace images in story description with text links [#645]

briankung:
Feedback appreciated! 👍

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -14,8 +14,8 @@ class StoriesController < ApplicationController
     @title = "Submit Story"
     @cur_url = "/stories/new"
 
-    @story = Story.new(story_params)
-    @story.user_id = @user.id
+    @story = Story.new(user: @user)
+    @story.attributes = story_params
 
     if @story.valid? && !(@story.already_posted_recently? && !@story.seen_previous)
       if @story.save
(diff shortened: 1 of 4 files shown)


briankung:
There is currently a way to sneak past the limitation on inline images if you go to `stories#edit` and then `#update` with no actual changes, so don't merge just yet!

briankung:
^ I was actually wrong, the test user I was testing with manually was actually a moderator.

 Funny story, I couldn't get the 2nd scenario working in `submit_story_spec.rb`. I thought I was going crazy, but no matter how often I tried it, I just couldn't get it to pass - there was always a link instead of an img tag. I was also trying something cute in `stories_controller.rb`:

`@story = Story.new(story_params.merge(user: @user))`

I banged my head for a good hour (?) at least, until I realized that the attribute assignment was probably done in order, and that `#description=` depends on the story's user. With that, I decided to try this:

`@story = Story.new({user: user}.merge(story_params))`

And sure enough, the test started passing. As frustrating as that was, I didn't want to leave a time bomb for the next person, so I made it more explicit with:

```
@story = Story.new(user: @user)
@story.attributes = story_params
```

pushcx:
Clean approach, and I really like that you tested what should and shouldn't appear in the output, both in an end-to-end test and at the model level. Thanks!

PR #651, opened 2019-03-14
CI: Get Bundler 2 first

olleolleolle:
This PR is about trying to **make the build pass:**

Before this change: it says "the Gemfile.lock _demands_ Bundler 2+" and quits. Example: https://travis-ci.org/lobsters/lobsters/builds/505896930

After this change: builds green.


pushcx:
Thanks, glad to have the fix!

Issue #328, opened 2016-10-31
mails: suboptimal quoted-printable generation


leahneukirchen:
lobste.rs sends mail where the entire value of a header is encoded as a single quoted-printable token and then split across lines.  Possibly, this results in a UTF-8 multibyte sequence split across tokens,
which according to my reading of the RFCs is allowed, but resulted in at least two mail clients as rendering errors.

Perhaps a more clever version of qp can be implemented (or leveraged from some stdlib, but I couldn't find it.)

pushcx:
Could you give an example of the bug and what a valid version of that header would look like, please?

chneukirchen:
```
puts ("foö "*30).quoted_printable(true)
=?UTF-8?Q?fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_f?=
	=?UTF-8?Q?o=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo?=
	=?UTF-8?Q?=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3?=
	=?UTF-8?Q?=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_fo=C3=B6_?=
```

Line 3 is split between a multibyte encoding.

Instead, something like

```
Subject: =?UTF-8?Q?fo=C3=B6?= =?UTF-8?Q?_fo=C3=B6?=
 =?UTF-8?Q?_fo=C3=B6?= =?UTF-8?Q?_fo=C3=B6?= =?UTF-8?Q?_fo=C3=B6?=
 =?UTF-8?Q?_fo=C3=B6?= =?UTF-8?Q?_fo=C3=B6?= =?UTF-8?Q?_fo=C3=B6?=
```
could be used.

tinyplasticgreyknight:
I may take a stab at this, it's vaguely related to something I was working on recently.

evexoio:
I was just trying to reproduce this, but I didn't get any false rendering in both gmail and thunderbird
[...]

PR #653, opened 2019-03-22
Forces encoding to utf-8 when fetching url titles [Resolves #648]

### LINKED ISSUE #648: Fetched story titles are missing apostrophes and other special characters
## The problem

Recently I noticed two stories, each of whose titles on Lobsters was missing an apostrophe that was in the title of the linked page. The most recent of those stories was [Sometimes **its** better for tests to hit the real API](https://lobste.rs/s/wortjv/sometimes_its_better_for_tests_hit_real), which links to an article titled [Sometimes **it’s** better for tests to hit the real API](https://www.codewithjason.com/sometimes-better-tests-hit-real-api/). If you paste that URL in the [Submit Story](https://lobste.rs/stories/new) page and click Fetch Title, the fetched title is missing the apostrophe, so that’s probably why those stories were submitted with inaccurate titles.

The title fetched by the Fetch Title button on the Submit Story page should not omit any visible characters. There may be a security argument for omitting some characters such as the Unicode LTR character, but the apostrophe `’` (and the straight single quote `'`) should definitely be preserved.

I haven’t tested whether the button also strips other special characters that should be preserved, such as `"`.

## Investigating the cause

The Fetch Title button seems to be currently powered by this JavaScript:

https://github.com/lobsters/lobsters/blob/fdd85dbe3d05984398d019d048d812c960e79a89/app/assets/javascripts/application.js.erb#L351-L377

I don’t see any explicit escaping code in that JavaScript. Maybe jQuery does the excessive escaping, or maybe it’s an encoding issue.

I don’t see anything wrong in the Fetch Title button’s view code:

https://github.com/lobsters/lobsters/blob/fdd85dbe3d05984398d019d048d812c960e79a89/app/views/stories/_form.html.erb#L9-L10

https://github.com/lobsters/lobsters/blob/fdd85dbe3d05984398d019d048d812c960e79a89/app/views/stories/_form.html.erb#L155-L160

briankung:
~~I found it moderately difficult to write a test case for this, since it seems like all the tests I wrote seemed to assume UTF-8 encoding and passed the \xe2\x80\x99 apostrophe through just fine.~~ Regression test added.

Manual testing confirms that it works for the URL in #648 (https://www.codewithjason.com/sometimes-better-tests-hit-real-api/)

I can't locate Abdullah Samman, the last person to touch this line in 0e198cb9bf39211c22e5293d846e3e6be348e4f1, on GitHub, or I'd tag them as a reviewer.

~~Set to draft PR awaiting feedback on testing! I could spend a bit longer working on getting a test case written or we could just merge it.~~

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -920,9 +920,8 @@ class Story < ApplicationRecord
       begin
         s = Sponge.new
         s.timeout = 3
-        @fetched_content = s.fetch(self.url, :get, nil, nil, {
-          "User-agent" => "#{Rails.application.domain} for #{self.fetching_ip}",
-        }, 3).body.force_encoding('utf-8')
+        user_agent = { "User-agent" => "#{Rails.application.domain} for #{fetching_ip}" }
+        @fetched_content = s.fetch(url, :get, nil, nil, user_agent, 3).body.force_encoding('utf-8')
       rescue
         return @fetched_attributes
       end
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -922,7 +922,7 @@ class Story < ApplicationRecord
         s.timeout = 3
         @fetched_content = s.fetch(self.url, :get, nil, nil, {
           "User-agent" => "#{Rails.application.domain} for #{self.fetching_ip}",
-        }, 3).body.encode!(invalid: :replace, undef: :replace, replace: '')
+        }, 3).body.force_encoding('utf-8')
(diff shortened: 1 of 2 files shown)


pushcx:
Phew, what a corner case to run down, thanks for the deep dive and clear test.

(You know, if you like obscure encoding errors, #328 is a finely-aged sample of the genre...)

PR #655, opened 2019-03-24
Fixes moderation logs for newly created Tags [Resolves #650]

### LINKED ISSUE #650: Modlog of tag creation is bugged
The [modlog](https://lobste.rs/moderations?utf8=%E2%9C%93&moderator=%28All%29&what%5Btags%5D=tags) lists the last three tag creations as updating all fields from `''` (almost certainly `nil.to_s`) to their values instead of saying "Created..."

briankung:
Turns out `#new_record?` is only `true` before the save, which is not true in an `after_save` callback. I also made the unit tests a bit more specific.

Resolves #650

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -59,8 +59,8 @@ class Tag < ApplicationRecord
 
   def log_modifications
     Moderation.create do |m|
-      if self.new_record?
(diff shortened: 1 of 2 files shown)


pushcx:
Thanks for running this down!

PR #658, opened 2019-04-01
Add a title attribute to the upvote button

notriddle:
https://lobste.rs/s/bnze9m/suggestion_title_attributes_on


pushcx:
Seems like a reasonable little improvement, thanks.

PR #659, opened 2019-04-04
Q-encode characters as multiple encoded words

### LINKED ISSUE #328: mails: suboptimal quoted-printable generation

briankung:
Unless we parse the text and figure out where the glyph boundaries are, we might end up splitting a glyph between bytes. Considering that was the original complaint in #328, I'm not sure what to do. Granted, the failing case I'm working with - a _really_ long word with multibyte characters - is probably unlikely, but still technically a possibility.

There's a failing test: `it 'handles really long words'` that covers this.

~~Submitting to publish the research / work I've done so far, but not particularly happy with it.~~

Resolves #328

### DIFF
--- a/script/mail_new_activity.rb
+++ b/script/mail_new_activity.rb
@@ -9,15 +9,25 @@ Rails.application.require_environment!
 
 class String
   def quoted_printable(encoded_word = false)
-    s = [self].pack("M")
+    string = [self].pack("M")
 
-    if encoded_word
-      s.split(/\r?\n/).map {|l|
-        "=?UTF-8?Q?" + l.gsub(/=*$/, "").gsub('?', '=3F').tr(" ", "_") << "?="
-      }.join("\n\t")
-    else
-      s
-    end
+    return string unless encoded_word
+
+    q_encode_word = ->(w) { "=?UTF-8?Q?#{w}?=" }
+
+    string \
+      # Undo linebreaks from #pack("M") because we'll be adding characters
+      .gsub("=\n", "") \
+      # Question marks are delimiters in q-encoding so must be escaped
+      .gsub("?", "=3F") \
+      # Spaces are insignificant in q-encoding so must be escaped
+      .gsub(' ', ' _') \
+      # Take each space-separated word, then q-encode
+      .split(' ').map(&q_encode_word) \
+      # Recombine words then word wrap at 75 characters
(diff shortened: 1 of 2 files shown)


pushcx:
I spent a few minutes tinkering with this PR and I'd like to merge it. I really appreciate the way you've added clear comments, links, and tests to make future understanding that much easier.

If I understand correctly, there's now only an issue if we have a single word with multibyte characters that's so long it encodes to more than 75 characters *and* one of its multibyte characters is encoded to fall right on that boundary? That seems so implausible for a story title that I'm OK with the bug and noting the corner case in a comment. At the least, maybe changed behavior will prompt affected users to engage more on the issue tracker. (And maybe one of the non-english sister sites where this might be more common will pick up where you left off to fix even the corner cases.)

briankung:
> If I understand correctly, there's now only an issue if we have a single word with multibyte characters that's so long it encodes to more than 75 characters and one of its multibyte characters is encoded to fall right on that boundary?

Yes, that's how I understand it as well. To be clear, the current behavior is not to truncate the title whatsoever.

I guess I'm just nervous about it! It probably doesn't follow the RFC. The other thing that isn't particularly great about this is that spaces are sometimes encoded as their own encoded word, e.g. this line: https://github.com/lobsters/lobsters/pull/659/files#diff-6036a6c06b7487e9e3986dcaf3d36c15R7

`"Why  Use Pointers?"` (two spaces after "Why") turns into `"=?UTF-8?Q?Why?= `**=?UTF-8?Q?_?=**` =?UTF-8?Q?_Use?= =?UTF-8?Q?_Pointers=3F?="`. Functionally, it doesn't matter, but it makes it look a bit ugly. It would be nice to handle those.

I'll see what I can do to resolve that and then mark the draft PR as ready for review.

pushcx:
You solved our oldest open bug! Thanks, Brian. I really appreciate you puzzling out these complex and subtle bugs.

Issue #376, opened 2017-07-10
Add downvote category for inappropriate/hostile comments


jonhoo:
Currently, the available downvote categories for comments [are](https://github.com/jcs/lobsters/blob/108baedb39206198e442a5bb0828c644862a79e0/app/models/vote.rb#L6):

 - Off-topic
 - Incorrect
 - Me-too
 - Troll
 - Spam

It is unclear from this list how comments that can be construed as inappropriate, hostile, or abusive should be tagged. As with all kinds of downvoting there's a fine line to walk here, but it doesn't feel right to tag such posts as "Troll" (which is probably the most related of the categories above). I'm not entirely sure what this new category should be called ("Abuse" may be too strong), but I do feel like one should be added.

jcs:
I think troll is fine for that, but either way, this is something that would need to be discussed in a `meta`-tagged post on the site.

jonhoo:
Posted at https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate

jonhoo:
Following on from https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate#c_pryqvi, I think the current list is fine.

jonhoo:
Re-opening following discussion in https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate#c_2ztjbd.

jonhoo:
I'm going to try and give a fair and (relatively) concise summary of the discussion thus far so people won't have to read through the 154 comments currently on the linked thread. I'll do this in the form of linked quotes, as that may also give additional context. I have tried to include relevant comments from both sides of this debate, but have probably failed to do so. My own, more compact, but also more subjective, summary is at the bottom of this post.

> Scorn, showing off, and put-downs shut people down fast, especially the newbies and minorities. Kindness is the fuel of a good community.
— [pushcx](https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate#c_2ztjbd)

> I think uncharitable should be a new downvote flag. I’m okay with dissent, I’m okay with people being assholes, I’m even somewhat okay with outright negativity, as long as the comment is also charitable to the other viewpoint.
— [peter](https://lobste.rs/s/wei3a9/downvote_category_for_inappropriate#c_86eabt)

> Better to drop [such comments] to bottom of thread hidden by default with a plus or profile setting that can display them. That keeps them out of sight with votes but avoids total censorship where people can still read what was said.
[...]

PR #663, opened 2019-04-22
Add "Unkind" option

monokrome:
Add a new option "Unkind" when down-voting comments.

As brought up here: https://lobste.rs/s/xnjo8g/add_downvote_reason_unkind

As mentioned in https://lobste.rs/s/xnjo8g/add_downvote_reason_unkind#c_0lzrxb, this is maybe related to https://github.com/lobsters/lobsters/issues/376?

Does the phrasing need to be 100% figured out before any merging? The intent - which seems generally clear - seems important, and if it's being stored as a single character then we can easily adjust language as necessary later?

Note: I used "me-too" as a search term to see if there are any specific tests for this, and the me-too case didn't seem to have any. So, no tests here but happy to add some if people have ideas for what is testable here!

Also, I'm wondering if this requires consideration here?

https://github.com/lobsters/lobsters/blob/d2963d4b80b47ac9f34bda65154a37cdde0a2174/app/controllers/mod_controller.rb#L33

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -9,6 +9,7 @@ class Vote < ApplicationRecord
     "M" => "Me-too",
     "T" => "Troll",
     "S" => "Spam",
+    "U" => "Unkind",
     "" => "Cancel",
   }.freeze
 


Carpetsmoker:
My 2¢ on this is that this may not actually be that useful. You can *already* downvote comments that are excessively hostile ("troll").

I think the key issue is tracking long-term consistently abrasive people. Everyone is a bit too abrasive sometimes, and that's kind of okay. The problem is usually a small minority who are abrasive a lot of the time.

Also see my comments here:

- https://lobste.rs/s/xnjo8g/add_downvote_reason_unkind#c_kqtuqr
- https://lobste.rs/s/xnjo8g/add_downvote_reason_unkind#c_6lfmsz

So I would prefer a different patch which adds a new "unkind" flag mechanism. This wouldn't -1 a post, just track it so a user's profile page can show something like `476 unkind flags (23% of total posts`, and/or a special admin page with users with the most unkind flags.

I know this is a bit more work, but IMHO it's also much more useful.

spacejam:
This seems to have quite a lot of support. One reason I think it should be a -1 is that we should rank objectively true yet unkind posts below objectively true posts that contain less toxicity. Avoiding a special case for this flavor of community-degrading content keeps the system easier to maintain and improve over time.

pushcx:
I ran out of review time this morning, will try to make time an evening this week.

danielcompton:
> One reason I think it should be a -1 is that we should rank objectively true yet unkind posts below objectively true posts that contain less toxicity. Avoiding a special case for this flavor of community-degrading content keeps the system easier to maintain and improve over time.

I agree with this. I'm not in favour of adding a special flag for unkind comments that avoids downranking the comment. 

I do like the idea of showing a user their aggregate comment history and downvote percentages, possibly alongside the quartiles of active Lobsters commenters so you have some context to put the numbers in. However I don't think this needs to block this much smaller PR.

monokrome:
For what it's worth, I think that the value of a post is denoted just as much by how it affects people in as well as from whether something is true or not. As an example, if there were two posts with similar true content, I'd hope that the person who isn't calling people "idiots" or being otherwise abusive would have better karma. Just a thought.

I'd be totes happy to change it in whichever direction makes the most sense to y'all, though 💟

monokrome:
@Carpetsmoker I think that "troll" is different than "unkind". Trolls are generally intending to aggro someone, while people may be "unkind" and not "trolling". I think that if we want "troll" to account for this, then maybe "troll" is a poor choice of words, but I can also see enough semantic difference that it may be valuable to separate the two? 🤷‍♀

monokrome:
Meow? :smile_cat:

monokrome:
@carpetsmoker @arp242 @danielcompton Hey! Where is this at?

danielcompton:
I think this would be a great addition to Lobsters, I’d love to see it added.

angersock:
It's come up a couple of times in the past couple of years, and still hasn't been implemented. Maybe come back next year and try again. :)

FWIW I think it'll just lead to abuse and damage the community, but I've covered that [elsewhere](https://lobste.rs/s/xnjo8g/add_downvote_reason_unkind#c_xcun5m).

angersock:
> Does the phrasing need to be 100% figured out before any merging? The intent - which seems generally clear - seems important, and if it's being stored as a single character then we can easily adjust language as necessary later?

In the intervening time since filing, have we made any progress on the definition of the "unkind" flag? With that in hand it'd probably be a lot easier to carry it into a `meta` thread again. Otherwise, it's may be (rightly) criticized as pulling a fast one.

pushcx:
[Added](https://lobste.rs/s/s1gzie/adding_politics_related_tag#c_gfr5my)

pushcx:
Some kind of bug occurred - this isn't appearing. It's in `Vote::COMMENT_REASONS` in prod console but not prod js. Not sure what's wrong here. I manually ran `RAILS_ENV=production bundle exec rails assets:precompile` in prod and it didn't appear.

pushcx:
...huh, this has gone live. First flag of a comment for being unkind happened at 2019-12-27 14:42:18. I don't know what would've unstuck this on that day, the code changes I merged + pushed were minor and unrelated. Maybe the 'bundle update' that would've gone with 57e0010 bounced unicorn and the constant wasn't refreshed? I don't like not understanding this.

monokrome:
What happened here?

PR #666, opened 2019-04-28
remove links from malicious actor

rain-1:


pushcx:
It links to an explanation of what's up with Stylish with a warning. If you don't want to see links to the existing dark styles, please ask their authors to move them over to another site and re-open this PR with links to their new homes.

PR #668, opened 2019-04-29
Better colours for error boxes

Carpetsmoker:
The boxes are hard to read due to the black text on dark red background
colour (contrast ratio: 4.04).

This replaces the background colour with a light red one (ratio: 14.97)
for better readability.

Alternatively, could also muck about with the text colour, but this
looked better IMHO and it was easier to get a good contract ratio
(changing the text to #fff is more readable, but contrast ratio is only
5.19).

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1157,8 +1157,7 @@ div.errorExplanation a {
 }
 div.flash-error p a,
 div.flash-notice p a,
-div.flash-success p a,
-div.errorExplanation a {
+div.flash-success p a {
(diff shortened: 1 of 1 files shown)


Carpetsmoker:
Before and after screenshots (I added in a link in the second one to ensure that shows up readable as well):

2019-04-30-031906_1450x293_scrot

2019-04-30-032742_1429x299_scrot

pushcx:
Looks great, thank you. I'm glad to merge this.

(I see your other PR but will actually have to think about that one when I'm fresh, instead of just look at nice screenshots and :+1: .)

PR #671, opened 2019-05-01
Add homepage field

### LINKED ISSUE #669: Homepage field?
What are the thoughts on adding a homepage field for users?

Right now you can already put a link in your profile description, but the advantage of a separate field would be that we can add [rel="me"](http://microformats.org/wiki/rel-me), which can be an alternative to e.g. Keybase. It was mentioned by some people in the Keybase discussion; I had never heard of it before, but turns out many services actually add rel="me" (e.g. Twitter, GitHub, etc.) to home page links.

Should be easy to do, and I'll gladly make a PR. Just wanted to check in first to see if there are any objections.

Thanks!

Carpetsmoker:
Add new homepage field so a rel=me link can be added. Also add rel=me to
the Twitter/GitHub links.

See: http://microformats.org/wiki/rel-me

Closes #669


Carpetsmoker:
Travis fails with:

```
app/validators/http_url_validator.rb:6:23: W: Lint/UriEscapeUnescape: URI.encode method is
 obsolete and should not be used. Instead, use CGI.escape, URI.encode_www_form or 
URI.encode_www_form_component depending on your specific use case.

      uri = URI.parse(URI.encode(value))
                      ^^^^^^^^^^^^^^^^^
```

`CGI.escape()` behaves different though; need to check what the easiest way to do this is when I got some time (only needed to make IDN domains work, perhaps there's a better way to start with).

briankung:
This answer was pretty helpful for my understanding of the `CGI.escape` vs `URI.encode` conundrum: https://stackoverflow.com/a/13059657/1042144

> The only problem left is that we didn't escape our query parameters properly, which brings us to the conclusion: we should not use a single method for the entire URI, because there is no perfect solution (so far). As you see `&` was not escaped from "My Blog & Your Blog". We need to use a different form of escaping for query params, where users can put different characters that have a special meaning in URLs.

Carpetsmoker:
Yes, I was aware of what the differences/issues are, just not what the solution is.

As far as I can find there is no stdlib way of doing it so I just opted for a regexp validation like the email one. This should be good enough and not reject any valid URLs.

pushcx:
Looks great. Thanks for being so thorough with styling, js help, and tests, this is wonderful.

Issue #670, opened 2019-05-01
case sensitive username under postgresql


jwjwyoung:
Hi, 

This might be a dumb problem. 

I am deploying the lobsters on my local machine using PostgreSQL as the backend database. And there is some inconsistency of the case sensitivity of username when sending a message. 

https://github.com/lobsters/lobsters/blob/d88202c677543dd4f8d6952ca378b7b37fab8d01/app/models/user.rb#L81 shows username is case insensitive, but 

https://github.com/lobsters/lobsters/blob/bdd821504358b114af3d925d1cc5e7d95f97d23e/app/views/messages/_form.html.erb#L8 
https://github.com/lobsters/lobsters/blob/01f10b9fa4701a0c08d87b98958deb88629e9b20/app/models/message.rb#L101 
shows that it differentiate between different cases for recipient_username, which results in that if I want to send a message to `IVory`, then I can only input the name as   `IVory` instead of `Ivory`,

A quick fix would be `User.where(‘lower(username) = ?’, username.downcase).first`

While this works well under MySQL since there is no case sensitive problem for the column stored into the database, just want to post the issue to say the problem under PostgreSQL.

Best,
JW.

jwjwyoung:
A quick follow up is that the same thing happens for the email https://github.com/lobsters/lobsters/blob/01f10b9fa4701a0c08d87b98958deb88629e9b20/app/models/user.rb#L74 which is case insensitive. 

But login shows that email can be of a different case, and IVory@gmail.com cannot be loginned through ivory@gmail.com, which is annoying under PostgreSQL.

pushcx:
MySQL let us be really sloppy about these, thanks for reporting this. If you've fixed these, please do send back a PR so we can share it. In any case - please tell us when you launch your site so we can add it to the [list of sister sites](https://github.com/lobsters/lobsters/wiki).

Deimos:
PostgreSQL has an (included) module you can enable named `citext` that you can use as a column type and is case-insensitive for comparisons: https://www.postgresql.org/docs/current/citext.html

pushcx:
I'm grooming the backlog and I'm going to close out this old bug. We don't have the developer capacity to support alternate databases.

PR #672, opened 2019-05-01
Update message.rb

jwjwyoung:
make find_by username is case insensitive

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -270,7 +270,7 @@ class CommentsController < ApplicationController
 
   def threads
     if params[:user]
-      @showing_user = User.find_by!(username: params[:user])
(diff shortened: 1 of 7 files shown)


briankung:
See comment here: https://github.com/lobsters/lobsters/pull/673#discussion_r280587856

Basically, from my reading of the extant code, the username and email fields are case insensitive MySQL columns, so `#where` clauses will return case insensitive, anyway. I've also manually tested this and found it to be case insensitive with usernames.

jwjwyoung:
Thanks for your reply, again I need to say it's okay under MySQL, but not works well under PostgreSQL,  as discussed in https://github.com/lobsters/lobsters/issues/670.

briankung:
Ah, my bad! ✌️

pushcx:
After [discussion in the other PR](https://github.com/lobsters/lobsters/pull/673#issuecomment-496917141), we're going to take a different strategy.

PR #673, opened 2019-05-01
Update login_controller.rb

jwjwyoung:
make username case insensitive

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -24,9 +24,9 @@ class LoginController < ApplicationController
 
   def login
     if params[:email].to_s.match(/@/)
-      user = User.where(:email => params[:email]).first
+      user = User.where('lower(email) = ?', params[:email].downcase).first
     else
-      user = User.where(:username => params[:email]).first
+      user = User.where('lower(email) = ?', params[:email].downcase).first
     end
 
     fail_reason = nil


app/controllers/login_controller.rb
```diff
@@ -24,9 +24,9 @@ def index
 
   def login
     if params[:email].to_s.match(/@/)
-      user = User.where(:email => params[:email]).first
+      user = User.where('lower(email) = ?', params[:email].downcase).first
     else
-      user = User.where(:username => params[:email]).first
+      user = User.where('lower(email) = ?', params[:email].downcase).first
```
briankung: You probably meant to write `lower(username) = ?` in this SQL statement, which is why Rubocop is going after you to move the identical statements out of the conditional.

Also, oddly enough while I was testing out the case insensitivity, I found that I can already log in with a case insensitive username, likely due to MySQL column type `utf8mb4_general_ci` (https://github.com/lobsters/lobsters/blob/master/db/schema.rb#L245)

Carpetsmoker:
Is it not possible to use a case-insensitive collation or some such for this on PostgreSQL? It seems to me that this is very prone to breakage in the future.

[Reading this](https://stackoverflow.com/a/7005656/660921) it seems there are at least a few other solutions than sprinkling lower() all over the place.

jwjwyoung:
For the speed concern, maybe adding an extra index for the lower cased username:
`CREATE INDEX ON groups (LOWER(username))` could help.

pushcx:
Thanks for the pointer, @arp242. I agree that a case-insensitive collation is a better approach. It means there will one place we can clearly label as existing for postgresql compatibility and not have to continue spreading calls to `lower` or add unused indexes.

PR #674, opened 2019-05-02
Lobsters + Scout & Nate Berkopec Performance Collab

caser:
This is a follow-up on the collaboration between Lobsters and Scout + Nate Berkopec. 

Lobste.rs discussion thread here:
https://lobste.rs/s/5hshvd/proposal_lobste_rs_performance_analysis

In this PR, we've added Scout to the gemfile (referenced by git hash) and config/scout_apm.yml to .gitignore. 

The config file includes the organization's Scout key, so is not included here (e-mailing directly to @pushcx). It also includes the privacy features we've enabled as discussed in the Lobste.rs thread linked above, specifically the disabling of IP collection and query parameters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ config/database.yml
 config/initializers/secret_token.rb
 config/*.sphinx.conf
 config/unicorn.conf.rb
+config/scout_apm.yml
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem "sitemap_generator" # for better search engine indexing
 
 gem "ruumba"
 
-gem "scout_apm"
+gem "scout_apm", :git => 'git://github.com/scoutapp/scout_apm_ruby.git', :ref => '574b0a6'
 
 group :test, :development do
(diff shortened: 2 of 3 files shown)


pushcx:
So, I was just poking around the dashboard as data started to trickle in and the first thing I saw was that /search ate 427 MB of RAM. We knew it was slow, but not that it was a memory hog. I immediately suspect it was retrieving all hits instead of just the first page. When I clicked in there was this tempting "Backtrace" button, so I clicked that and it showed the search code, highlighting the exact line of code with the bug. Then I couldn't help but [fix it](https://github.com/lobsters/lobsters/commit/e23227f49e8ac632978a670b8289aaff92303557). So, thanks for making so easy to find the [mysterious year-old bug that crashes the site](https://github.com/lobsters/lobsters-ansible/issues/34). I diagnosed and fixed it when I wasn't even looking for it.

But, uh, sorry, Nate, for crossing off what was sure to be item #1 in your report.

caser:
Amazing! Love to see quick wins :)

PR #677, opened 2019-05-03
Add mention of the karma threshold for downvoting

yaymukund:
I added a bit in the about page about needing 50 karma to downvote. 😊

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -97,12 +97,12 @@
 
   <h2 id="downvotes">Downvote Explanations</h2>
   <p>
-  Users may downvote stories and comments after their account has
-  <%= User::MIN_KARMA_TO_DOWNVOTE %> karma. Often on other sites, a user would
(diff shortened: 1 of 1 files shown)


app/views/home/about.html.erb
```diff
@@ -97,6 +97,7 @@
 
   <h2 id="downvotes">Downvote Explanations</h2>
   <p>
+  Users may downvote stories and comments after their account has 50 karma.
```
Carpetsmoker: This should probably use the `User.MIN_KARMA_TO_DOWNVOTE` constant.

yaymukund: I've done the suggestion you suggested

yaymukund:
done!

pushcx:
Thanks for adding this. Could you move the sentence to the end of the paragraph? It reads better to introduce the overall system and then talk about the corner cases at the end. Thanks!

pushcx:
Thanks for this addition.

PR #684, opened 2019-05-30
comment_deletor/comment_undeletor: add missing event parameter to callback

jordigh:
comment_deletor/comment_undeletor: add missing event parameter to callback

This appears to be a C&P error introduced in
003d47ce221a4bc988b122804a8b7888e007c5f3 due to copying similar code
around here and forgetting to also modify the function parameters:

https://github.com/lobsters/lobsters/blob/003d47ce221a4bc988b122804a8b7888e007c5f3/app/assets/javascripts/application.js.erb#L545

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -491,7 +491,7 @@ $(document).ready(function() {
     });
   });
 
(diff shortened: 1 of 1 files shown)


pushcx:
Ha! Funny how adding a bug also prevented the default event handler. Thanks for the fix.

Issue #683, opened 2019-05-14
Comment count on merged story doesn't include comments from source story


alanpost:
This morning I merged story 5o7wvj in to qlgw0n.  After doing so and as of right now, the comment count as displayed on the home page and the story page shows a count of two (2) whereas manually counting shows three (3) comments.

I expected the comment count to be three (3) rather than two (2) to correspond to the sum of comments on both stories.

pmueller:
It appears this happens due to `Story#update_comments_count!` not running callbacks. (https://github.com/lobsters/lobsters/blob/627b6c5df7e3e70883c54fff1595e3bf9b00c92a/app/models/story.rb#L809)
If a comment originally on story 5o7wvj was e.g deleted, it would update the comments count for 5o7wvj but not run the callback for `update_merged_into_story_comments`, so qlgw0n would be left with an incorrect comments count

PR #685, opened 2019-06-02
Bug fix: merged story can end up with an incorrect comments count

pmueller:
This should fix the issue described in #683

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -806,8 +806,8 @@ class Story < ApplicationRecord
     comments = self.merged_comments.arrange_for_user(nil)
 
     # calculate count after removing deleted comments and threads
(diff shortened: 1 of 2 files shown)


pushcx:
Hey, thanks for taking a run at this. I really appreciate that you added a test to prevent regression.

What do you think about just explicitly calling `update_merged_into_story_comments` from `update_comments_count!`? That saves a bunch of extra queries, including touching `updated_at` that's not really appropriate.

Also we can remove the redundant `self.comments_count = ` in either case.

pmueller:
Hi @pushcx, that makes total sense to me. I've made those changes 👍

pushcx:
Looks great, thanks for solving this and sending a fix. :D

PR #687, opened 2019-06-13
set action_view.form_with_generates_remote_forms to default (true)

thomasdziedzic:
Now that https://github.com/lobsters/lobsters/issues/527 has been closed, we can use the default value for `action_view.form_with_generates_remote_forms` which is true, and conforms to the defaults of rails. This should remove surprises for people browsing the code.

I also took the time to make the style more consistent.

This also opens up the opportunity for someone to experiment with turning 1 form into a remote form without having to convert them all.

### DIFF
--- a/app/views/comments/_commentbox.html.erb
+++ b/app/views/comments/_commentbox.html.erb
@@ -1,6 +1,6 @@
 <div class="comment comment_form_container"
 data-shortid="<%= comment.short_id if comment.persisted? %>">
-<%= form_with url: comment, :id => "edit_comment_#{comment.short_id}" do |f| %>
+<%= form_with url: comment, id: "edit_comment_#{comment.short_id}", local: true do |f| %>
   <% if comment.errors.any? %>
     <%= errors_for comment %>
   <% end %>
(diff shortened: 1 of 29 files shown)


pushcx:
Why change the default and disable it on every form instead of adding `remote: true` if and when we want it remote forms?

thomasdziedzic:
I did it so that the code base conforms to the rails default to remove surprises for newcomers.

This will also align the code base with the rails documentation [0]:

> By default form_with attaches the data-remote attribute submitting the form via an XMLHTTPRequest in the background if an Unobtrusive JavaScript driver, like rails-ujs, is used. See the :local option for more.

The `remote: true` way was the old way of doing things and isn't available for `form_with` [0].

But there is an alternative of using `local: false` which I think would allow the use of remote forms, although I haven't tested if it works as expected.

[0] https://api.rubyonrails.org/v5.2.3/classes/ActionView/Helpers/FormHelper.html#method-i-form_with

thomasdziedzic:
keeping form_with_generates_remote_forms turned off until lobste.rs converts to remote forms

pushcx:
Why close this?

thomasdziedzic:
reopened

thomasdziedzic:
What are the next steps for this?

pushcx:
I'll take the next step.

PR #689, opened 2019-06-13
update scout apm gem source to https

thomasdziedzic:
when running bundle I kept getting the following error:

```
The git source `git://github.com/scoutapp/scout_apm_ruby.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
```


pushcx:
Yeah, no reason not to. Thanks for fixing this.

Issue #434, opened 2017-11-20
Sponge vulnerable to MITM


arkadiyt:
`Sponge` is used to fetch url content in several places:
- handling github/twitter oauth flows (particularly sensitive)
- sending push notifications via pusher
- fetching story attributes
- fetching user avatars
- handling story caching with diffbot

However it has the following code: https://github.com/lobsters/lobsters/blob/master/extras/sponge.rb#L154
```
host.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
which makes it vulnerable to MITM attacks.

jcs:
> edit: I believe you are doing this because you are directly hitting an ip address to avoid server side request forgery issues

Partially, but also because it's fetching content from arbitrary websites where SSL validation doesn't matter and trying to enforce validation would break title fetching more often than a MITM would accomplish anything.

It should probably enable validation by default but allow disabling it from a setting, which would only get set during title fetching.

evexoio:
@pushcx @alanpost can we close this issue? there doesn't look like any action needs to be taken here.

evexoio:
> We rarely if ever see this on Lobster.rs, probably since no one wants to follow such links.

So is it really an issue? what is HN's take on this one btw?

arkadiyt:
> So is it really an issue?

Yes, because the same http library is being used for github and twitter oauth flows - anyone on the network path between Lobste.rs and Twitter/Github can intercept those credentials and gain the same level of account access.

> what is HN's take on this one btw?

HN doesn't fetch anything from submitted urls.

evexoio:
I see, then we should add https support for oauth flows and leave the title fetching without https.

let's see what @pushcx or @alanpost think about this.

jbbarth:
This issue is a bit old but I think it deserves fixing. Everybody could agree I think that hitting trusted APIs should be done in HTTPS, even more if you transmit credentials/secrets like done in the GitHub OAuth flow. I gave a shot at a solution with good defaults in #692.

jcs:
#692 was merged so this can be closed.

pushcx:
Thanks for catching the stale issue, closing.

PR #692, opened 2019-06-17
Turn SSL certificate verification on for critical services

jbbarth:
The `Sponge` class used to skip SSL certificate verification entirely
for all HTTPS calls for all HTTPS calls, which makes Lobsters
vulnerable to some MITM attacks that are easily exploited on private or
corporate LANs. We still want to skip certificate verification when
there's no specific security risk, but we do want to validate
certificates when communicating with well-known APIs, such as
Archive.is, GitHub, Gravatar and PushHover. Note that some APIs are
accessed outside of Sponge, such as Twitter's one.

Turning this verification on will flat out break requests to those
websites in case their certificate expired or is incorrect. This may
trigger a 500 for the users of Lobsters, which is what we want. Note
that for Gravatar the 500 would be limited to avatar serving, and
wouldn't impact the main page of Lobsters.

This PR is supposed to fix concerns detailed in #434. The review should be trivial, but feel free to ask and I'll do my best to answer or improve this proposal.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -915,6 +915,9 @@ class Story < ApplicationRecord
       begin
         s = Sponge.new
         s.timeout = 3
+        # User submitted URLs may have an incorrect https certificate, but we
+        # don't want to fail the retrieval for this. Security risk is minimal.
+        s.ssl_verify = false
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks for finding a really nice balance on this. I'm going to deploy it and watch my inbox for 500s for a few days before closing #434.

Issue #675, opened 2019-05-02
Replies in posts that have been hidden are not shown even after the post has been unhidden


gustafe:
I commented in a post, but I decided to hide it after the comment.

When checking my thread view, I saw an answer to my comment, and decided to unhide the post. 

Replies to my comment after unhiding the post did not appear under my "Replies" view, however.

pushcx:
Looks like that's because unhiding a post doesn't create a `ReadRibbon`, only visiting the story page does.

jbbarth:
I tried to look into this one: 
- the ReadRibbon should already exist since it's created when visiting a story _or_ hiding it
- but it is switched to `is_following = false` when hiding (via `ReadRibbon.hide_replies_for()`), and never gets back to `is_following = true` when unhiding
- the fix is pretty trivial, we can just have a `ReadRibbon.unhide_replies_for()` counterpart called when unhiding a story

I'll try to submit a PR with the fix in the next few days.

pushcx:
Fixed by @jbbarth in #693.

PR #693, opened 2019-06-17
Fix replies missing after unhiding story

jbbarth:
This PR aims at fixing #675.

As said in the issue, ReadRibbon for a given story is marked as "is_following = false" when hiding the story, making replies not displayed in the "Replies" section. But when unhiding a story we don't mark it back as "is_following = true", hence the problem.

The PR also has 2 small refactor commits that are neutral from a behavior point of vue but make the code a bit more coherent.

HiddenStory and ReadRibbon related behaviors were not tested previously. We may want to add a test, it might be a bit tautological but let me know, I can give it a try.

Of course feel free to tell me what to improve in this proposal!

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -310,6 +310,7 @@ class StoriesController < ApplicationController
     end
 
     HiddenStory.where(:user_id => @user.id, :story_id => story.id).delete_all
+    ReadRibbon.unhide_replies_for(story.id, @user.id)
 
     render :plain => "ok"
   end
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -299,7 +299,6 @@ class StoriesController < ApplicationController
     end
 
     HiddenStory.hide_story_for_user(story.id, @user.id)
-    ReadRibbon.hide_replies_for(story.id, @user.id)
 
     render :plain => "ok"
   end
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks for digging in here, I hope my misdiagnosis didn't cost you too much time. I like the way you encapsulated the work by creating `unhide_story_for_user` and `unhide_replies_for`.

In `unhide_replies_for`: why `find_or_create_by`? I think that might be a bug: if someone clicks "hide" on a story without entering it, then `/hidden` and unhides it, this creates a ribbon for them, which seems inappropriate. Is there another corner case I'm not seeing that's fixed by this, or should this just no-op without error if the user doesn't have a ribbon already?

jbbarth:
You're right, it's definitely a corner case I didn't consider correctly, pure overlook on my side, nothing was specifically fixed by this.

I think the same goes for `ReadRibbon.hide_replies_for()`: you can hide without entering a story and creating a read ribbon at this point doesn't make sense. It used to have no practical consequence because of the bug we're fixing in this PR, but now if you unhide, you'll end up activating the read ribbon with a timestamp that doesn't correspond to any real visit of the story. So it's purely made up and we shouldn't create a read ribbon in that case too.

I will iterate and ping you again when things are better.

jbbarth:
Iteration was trivial, I don't know what I really was thinking in my last sentence.

Small nitpick: in fact our discussion is also true if you click hide/unhide from within a story page. The user may do that 3 hours after opening the page, the action is async, and creating a read ribbon at this point may make him miss some comments in the future. So in the end it's definitely better in all cases to not confuse hiding/unhiding with a visit of a story page.

pushcx:
Great, thanks for working through all these different circumstances so we have a robust fix.

PR #694, opened 2019-06-18
mass bump

thomasdziedzic:
After this `bundle outdated` shows only 2 packages:

1. capybara which depends on ruby >= 2.4, should be bumped for the rails 6 update
2. tzinfo which is locked to 1.x by activesupport


amerf1:
Hi Thomas I hope you are great! Firstly thanks for keeping this code alive really appreciate it. What is the best way to get in touch with you by email just wanted some help lobsters

thomasdziedzic:
Is your question related to this PR? If so, just ask here.
If not, maybe a more appropriate place would be to open a new issue with your question.

amerf1:
No its not sorry sure will do that thanks Thomas

pushcx:
Thanks for updating this. Because we still have pretty thin test coverage I try to deploy this sort of thing when I know I can be around for a few hours in case of 500s, so I'll probably merge this weekend.

Issue #697, opened 2019-06-26
 Remove inline scripts from lobste.rs


thomasdziedzic:
Related to: https://github.com/lobsters/lobsters/issues/486#issuecomment-437617883

Removing inline scripts from the code will allow us to enable a stricter content security policy.

pushcx:
Thanks for creating these two issues, I appreciate the care you've put into setting a reasonable baseline and making sure we're set up to keep making progress on it. Thank you.

magikid:
I did a search through the code looking for script tags and here's what I found:

```
λ ~/projects/personal/lobsters/ master rg '<script>' app/views
app/views/layouts/application.html.erb
28:    <script>

app/views/messages/index.html.erb
1:<script>

app/views/stories/show.html.erb
29:    <script>
43:      <% if @user %><script>autosize($("#comment")[0]);</script><% end %>

app/views/stories/_form.html.erb
145:  <script>

app/views/stories/new.html.erb
39:<script>
```

pushcx:
We also have some `onclick` handlers that probably count as inline scripts.

Issue #696, opened 2019-06-26
Remove inline styles from lobste.rs


thomasdziedzic:
Related to: https://github.com/lobsters/lobsters/issues/486#issuecomment-437449313

Removing inline styles from the code will allow us to enable a stricter content security policy.

magikid:
I ran a quick search through the code looking for inline styles and here's what I found:

```
λ ~/projects/personal/lobsters/ master rg 'style=' app/views
app/views/comments/_commentbox.html.erb
15:  <div style="width: 100%;">
29:        style="display: none;">
46:        <div style="display: inline-block; margin-left: 1em;">
54:      <div style="clear: both;"></div>

app/views/filters/index.html.erb
25:      <th width="5%" style="text-align: right; padding-right: 1em;">Stories</th>
38:        <td style="text-align: right; padding-right: 1em;"><%=

app/views/global/_markdownhelp.html.erb
1:<div class="markdown_help" style="display: none; padding-top: 0.5em;">
21:      style="color: inherit;">linked text</a></td>
30:    <td><pre style="margin: 0px;">pre

app/views/home/about.html.erb
266:  <li id="emoji" style="list-style: none; position: relative; left: -1em;">
267:  <span style="float: left; line-height: 1.5em;">🦞</span>
268:  <p style="padding-left: 1em;">

app/views/hats/requests_index.html.erb
43:        <p style="clear: both;">

app/views/hats/index.html.erb
20:    <th style="width: 130px;">User</th>

app/views/messages/index.html.erb
11:  <div class="legend" style="float: right;">
39:            <div style="white-space:nowrap;">

app/views/messages/show.html.erb
2:  <div class="legend" style="float: right;">
35:    <div style="float: left;">
41:    <div style="float: left; padding-left: 1em;">
48:      <div style="float: left; padding-left: 1em;">
56:  <div style="clear: both;"></div>

app/views/layouts/application.html.erb
47:        <a id="l_holder" style="<%= @traffic_style %>"

app/views/mod/commenters.html.erb
4:  <table class='data zebra' style='margin: 0 auto;'>

app/views/settings/index.html.erb
58:      <div class="boxline markdown_help_toggler" style="margin-left: 9em;
64:        <div style="clear: both;"></div>
165:        <span style="color: green; font-weight: bold;">

app/views/signup/invited.html.erb
66:        style="margin-left: 9em; width: 600px;">
71:        <div style="clear: both;"></div>

app/views/stories/_form.html.erb
[...]

PR #698, opened 2019-06-26
implement content security policy

### LINKED ISSUE #486: Add a Content Security Policy
When we update to Rails 5.2 we'll have access to its [support for CSP](http://guides.rubyonrails.org/security.html#content-security-policy).

thomasdziedzic:
Fixes  #486

Created #696 & #697 to remove inline styles and scripts.

For now, let's just get the ball rolling for content security policies so we can have a baseline to start from.

I didn't know where to save the reports so I decided to take the simplest route first which is to just log the violations.

Another option is to save the reports into the database like the approach that can be found in:
https://bauland42.com/ruby-on-rails-content-security-policy-csp/

Let me know if there is a preference.


pushcx:
I agree with all your choices to get started here. This is great, thank you.

Issue #700, opened 2019-07-05
adding 2fa throws 500


pushcx:
```
A NoMethodError occurred in settings#twofa_enroll:

  undefined method `random_base32' for ROTP::Base32:Class
  app/controllers/settings_controller.rb:87:in `twofa_enroll'
```

Started happening a couple days ago after #694 but I've been under the weather and unable to look at it.

PR #701, opened 2019-07-05
Fix 2fa

thomasdziedzic:
Closes https://github.com/lobsters/lobsters/issues/700

I wrote a test to reproduce the issue and then bumped rotp to 5.1.0 and the tests still pass.


pushcx:
Great, thank you for fixing this - and being so thorough by adding a regression test and tidying our test helpers.

PR #702, opened 2019-07-05
use released scout_apm gem

thomasdziedzic:
Not sure why the git version was used in the first place, but there has been a new release that includes the commit that it was pinned to.


pushcx:
We deliberately pinned a version so I could code-review it ensure it met our privacy expectations and not have to worry about updates changing behavior in a way we're not OK with.

PR #703, opened 2019-07-06
Enable rubocop rspec

thomasdziedzic:
I think this will have a positive effect on the code base in the long run.

I enabled some smaller, less controversial changes to begin with.
I also discovered an issue because the replying_comment test didn't actually get picked up because it didn't match *_spec.rb
I fixed this and discovered that the test was broken, so I decided to fix it as well.

This example is a good reason to have this enabled!


pushcx:
Thanks for catching the filename bug! I agree that these are some nice, safe cops to enable.

If you're thinking of enabling more cops, can you grab me on Freenode or something? I tried to set a pretty high bar for them (PR #460) to keep code churn down and I don't want you to waste your time. Especially if that might discourage you from continuing to make so many great contributions. :)

PR #707, opened 2019-07-19
PDF parsing support for fetch title

NattyNarwhal:
I submit a lot of PDFs, and currently, I have to manually scrape the
title of the PDF. This involves (often troublesome due to poor OCR or
layout by the producing software) copying it from the document, or its
metadata. I developed this feature to automate it, at least for the PDFs
nice enough to have proper metadata filed.

This uses the pdf-reader gem, instead of doing the heavy lifting. It
also refactors the story model's attribute handling to account for items
that aren't HTML. (We can theoretically also parse image/audio/video
metadata too.)

With more refactoring, we can add other information too, like tags or
checking the self-post checkbox, but that's for another day.

This is the first Ruby code I've written, so this might not fly to style
guidelines or whatever.


NattyNarwhal:
It appears the test suite pulls some weird shenanigans. If you look in the first test, it returns null, while the second test returns... a double? With a body property?

NattyNarwhal:
Ok, I think what's happening is that the tests just inject into the content and they aren't given real HTTP responses, which fouls up the function in non-real-world scenarios. I could work around this by doing something like `if res <= Net::HTTPResponse`, but that doesn't seem correct either (mostly for cheating the test suite). What would be your recommended approach?

This cheat device tricks the second (unicode) test, but not the first one, likely because it doesn't override sponge, confusing it more.

```diff
diff --git a/app/models/story.rb b/app/models/story.rb
index 79dfc03..bb09631 100644
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -997,10 +997,17 @@ class Story < ApplicationRecord
         s.ssl_verify = false
         user_agent = { "User-agent" => "#{Rails.application.domain} for #{fetching_ip}" }
         res = s.fetch(url, :get, nil, nil, user_agent, 3)
-        case res["content-type"]
-        when /pdf/
-          @fetched_content = res.body
-          return fetched_attributes_pdf
+        # in the test suite, we aren't given real HTTP responses, but the
+        # content is injected in instead.
+        if res and res.class <= Net::HTTPResponse
+          case res["content-type"]
+          when /pdf/
+            @fetched_content = res.body
+            return fetched_attributes_pdf
+          else
+            @fetched_content = res.body.force_encoding('utf-8')
+            return fetched_attributes_html
+          end
         else
           @fetched_content = res.body.force_encoding('utf-8')
           return fetched_attributes_html
```

pushcx:
To capture the conversation from #lobsters - the tests are failing because the `if !@fetched_content` section only exists to retrieve the content and then the code that used to be below it was filling in and returning `@fetched_attributes`. In the current version of this, it's not doing that.

I think the right fix here would be to rename `@fetched_content` to `@fetched_response` and move the case statement below, so that object can be injected by the test.

NattyNarwhal:
Ok, so I have in my working tree the model changed over to work with responses instead of just the content; now I'm wondering how I would change the tests. The unicode one has the idea of replacing sponge, but I don't know enough Ruby to know how I would emulate the array indexer semantics with the fake object it's returning.

This here:

```ruby
allow_any_instance_of(Sponge).to receive(:fetch).and_return double(body: content) # where would the ["content-type"] go?
```

pushcx:
I'm tagging 'good first issue' in the hopes that some kind hacktoberfester has extra attention to come help work through the test issues. :)

NattyNarwhal:
Merge conflicts resolved, hopefully. Might be worth merging this if future refactors could make this hairier.

NattyNarwhal:
Apologies, I clearly forgot how to write Ruby in the span of of a few months :grimacing:

pushcx:
I fixed the merge conflict and merged this in 5f50c79. Deploying now.

Thanks for adding this!

Issue #704, opened 2019-07-13
Update Specs to current style


thomasdziedzic:
I've been reading through rspec-rails documentation.

I found 2 interesting things that relate to the lobsters code base:

According to https://github.com/rspec/rspec-rails#request-specs

> Rails and RSpec teams discourage directly testing controllers in favor of functional tests like request specs

Should lobsters follow this advice and accept only request specs in the future and have a migration ticket to move controller specs to request specs?

Another similar question that is a little bit less obvious from https://github.com/rspec/rspec-rails#feature-specs

> Before Rails introduced system testing facilities, feature specs were the only spec type for end-to-end testing. While the RSpec team now officially recommends system specs instead, feature specs are still fully supported, look basically identical, and work on older versions of Rails.
>
> On the other hand, feature specs require non-trivial configuration to get some important features working, like JavaScript testing or making sure each test runs with a fresh DB state. With system specs, this configuration is provided out-of-the-box.

Should we open an issue to migrate existing feature specs to system specs and only accept system specs moving forward?

pushcx:
These both sound like worthwhile ways to keep ahead of the Rails deprecation curve. I'm tweaking the issue title to track the work.

If someone reading this would like to contribute, you can open a PR to upate a single test, you don't need to do all of them at once.

petrenkorf:
@pushcx Is there any on going process for this issue? I'd love to help migrating existing tests if any

PR #711, opened 2019-07-30
enforce content security policy

thomasdziedzic:
Now that enough time has passed since https://github.com/lobsters/lobsters/pull/698 we should enable the enforcement of the content security policy.

I also added a requests (as discussed in https://github.com/lobsters/lobsters/issues/704) test for csp so that the headers get their own tests.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -22,4 +22,4 @@ end
 # Report CSP violations to a specified URI
 # For further information see the following documentation:
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
-Rails.application.config.content_security_policy_report_only = true
+Rails.application.config.content_security_policy_report_only = false
--- a/spec/requests/content_security_policy_spec.rb
+++ b/spec/requests/content_security_policy_spec.rb
@@ -1,0 +1,18 @@
+require 'rails_helper'
+
+describe 'content security policy' do
+  it 'sets the correct headers' do
+    get '/'
+
+    directives = [
+      "default-src 'none'",
+      "img-src * data:",
+      "script-src 'self' 'unsafe-inline'",
+      "style-src 'self' 'unsafe-inline'",
+      "form-action 'self'",
+      "report-uri /csp-violation-report"
+    ].join('; ')
+
+    expect(response.headers['Content-Security-Policy']).to eq(directives)
+  end
+end
--- a/spec/requests/content_security_policy_spec.rb
+++ b/spec/requests/content_security_policy_spec.rb
@@ -10,7 +10,7 @@ describe 'content security policy' do
       "script-src 'self' 'unsafe-inline'",
       "style-src 'self' 'unsafe-inline'",
       "form-action 'self'",
-      "report-uri /csp-violation-report"
+      "report-uri /csp-violation-report",
     ].join('; ')
 
     expect(response.headers['Content-Security-Policy']).to eq(directives)


pushcx:
I think we'll need to land #697 before then - looking at the log, we get a lot of reported violations:

```
grep "csp-report" production* | cut -d'{' -f 2- | sort | uniq -c | sort -rn | head
   4661 "csp-report":{"blocked-uri":"data","document-uri":"https://lobste.rs/","original-policy":"default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report","referrer":"","violated-directive":"default-src"}}
   2013 "csp-report":{"blocked-uri":"eval","document-uri":"https://lobste.rs/","original-policy":"default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report","referrer":"","violated-directive":"script-src"}}
   1475 "csp-report":{"blocked-uri":"chrome-extension","document-uri":"https://lobste.rs/","original-policy":"default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report","referrer":"","violated-directive":"default-src"}}
    491 "csp-report":{"document-uri":"https://lobste.rs/","referrer":"","violated-directive":"script-src 'self' 'unsafe-inline'","effective-directive":"script-src","original-policy":"default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri /csp-violation-report","blocked-uri":"","status-code":0,"source-file":"safari-extension://com.bumblebee.wastenotime-e6q24juvvz","line-number":1,"column-number":889}}
```

Not sure if there's anything to be done for that browser extension. At a guess, it's injecting js into the page to show a timer or collect statistics.

thomasdziedzic:
> I think we'll need to land #697 before then - looking at the log, we get a lot of reported violations:
> ...
> Not sure if there's anything to be done for that browser extension. At a guess, it's injecting js into the page to show a timer or collect statistics.

I don't think the mentioned errors are related to #697 

Here is a formatted output of those errors:

```json
[
    {
        "csp-report": {
            "blocked-uri": "data",
            "document-uri": "https://lobste.rs/",
            "original-policy": "default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report",
            "referrer": "",
            "violated-directive": "default-src"
        }
    },
    {
        "csp-report": {
            "blocked-uri": "eval",
            "document-uri": "https://lobste.rs/",
            "original-policy": "default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report",
            "referrer": "",
            "violated-directive": "script-src"
        }
    },
    {
        "csp-report": {
            "blocked-uri": "chrome-extension",
            "document-uri": "https://lobste.rs/",
            "original-policy": "default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri https://lobste.rs/csp-violation-report",
            "referrer": "",
            "violated-directive": "default-src"
        }
    },
    {
        "csp-report": {
            "blocked-uri": "",
            "column-number": 889,
            "document-uri": "https://lobste.rs/",
            "effective-directive": "script-src",
            "line-number": 1,
            "original-policy": "default-src 'none'; img-src * data:; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; form-action 'self'; report-uri /csp-violation-report",
            "referrer": "",
            "source-file": "safari-extension://com.bumblebee.wastenotime-e6q24juvvz",
            "status-code": 0,
            "violated-directive": "script-src 'self' 'unsafe-inline'"
        }
    }
]
```

I used https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only as a reference in understanding these issues.

I can not reproduce the issue by visiting lobste.rs since firefox and chrome report warnings in the console for violated policies.

I'm guessing this is related to chrome/safari extensions (3rd & 4th items) injecting content like the 1st & 2nd items because I grepped for eval in the code base and couldn't find anything.

From reading around the internet I found a couple of likely examples. These browser extensions might be missing a proper manifest file for the extension and this is probably not a lobste.rs content security policy issue.

According to https://stackoverflow.com/a/25868498 chrome extensions have their own CSP that you need to supply for the extension to work with a site's CSP.

I found something similar with firefox at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy

So I think we could merge this now, and we'll just have to live with badly configured browser extensions for CSP.

pushcx:
Thanks for digging in. I'm OK with breaking old/sloppy extensions.

And thinking more about this, our failure mode is that we break voting or some other key function that relies on js and we do an easy revert. Merging and deploying.

timmc:
Looks like this broke commenting.

```
Content Security Policy: The page’s settings blocked the loading of a resource at https://lobste.rs/comments/hmhvwy/reply (“default-src”).
```

PR #714, opened 2019-08-02
Update story.rb

deniz0:
Added another Tracking URL to story.rb


pushcx:
Thanks, this definitely looks like the kind of thing we should block. Could you put it in alphabetically rather than at the end?

gtgl:
I created pull request which that link put alphabetically.
Duplicate of #715

pushcx:
Not sure what happened here but the user has deleted their GitHub account, so I added this site in 2b1bfa3.

PR #720, opened 2019-08-10
Typo: replaced a dot with a slash

claudey:


pushcx:
Thanks for catching this and sending a fix. It's especially important to get all the details right here because this instruction is for newbies, so I really appreciate this.

PR #725, opened 2019-08-20
Fix bug preventing the inactive-user user to reset their password

arandomandy:
Fix bug preventing the inactive-user user account from being created by *rake db:seed* as well as preventing the inactive-user user to reset their password.


pushcx:
Thanks for taking this on. It was fun talking through possible approaches in #lobsters with you, and this is a tidy solution.

PR #727, opened 2019-08-21
Logged in users can no longer visit login page.

### LINKED ISSUE #726: /login shows login form to logged-in users
If a logged-in user visits `/login`, they see the login form. Should probably just redirect to `/` without any flash message.

claudey:
This resolves #726


thomasdziedzic:
I think that there is another PR which fixes this bug at https://github.com/lobsters/lobsters/pull/728

PR #728, opened 2019-08-22
Fix bug letting already logged in users see login and forgot password forms

arandomandy:

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -17,13 +17,11 @@ class LoginController < ApplicationController
   end
 
   def index
-    if @user
-      redirect_to "/"
-    else
-      @title = "Login"
-      @referer ||= request.referer
-      render :action => "index"
-    end
+    return redirect_to "/" if @user
+
+    @title = "Login"
+    @referer ||= request.referer
+    render :action => "index"
   end
 
   def login
@@ -102,12 +100,10 @@ class LoginController < ApplicationController
   end
 
   def forgot_password
-    if @user
-      redirect_to "/"
-    else
-      @title = "Reset Password"
-      render :action => "forgot_password"
-    end
+    return redirect_to "/" if @user
+
+    @title = "Reset Password"
+    render :action => "forgot_password"
   end
 
   def reset_password
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -17,9 +17,13 @@ class LoginController < ApplicationController
   end
 
   def index
-    @title = "Login"
-    @referer ||= request.referer
-    render :action => "index"
+    if @user
+      redirect_to "/"
+    else
+      @title = "Login"
+      @referer ||= request.referer
+      render :action => "index"
+    end
   end
 
   def login
@@ -98,8 +102,12 @@ class LoginController < ApplicationController
   end
 
   def forgot_password
-    @title = "Reset Password"
-    render :action => "forgot_password"
+    if @user
+      redirect_to "/"
+    else
+      @title = "Reset Password"
+      render :action => "forgot_password"
+    end
   end
 
   def reset_password
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -116,7 +116,7 @@ class LoginController < ApplicationController
     end
 
     if @found_user.is_wiped?
-      flash.now[:error] = "It's not possible to reest your password " <<
+      flash.now[:error] = "It's not possible to reset your password " <<
                           "because your account was deleted before the site changed admins " <<
                           "and your email address was wiped for privacy."
       return forgot_password
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -119,7 +119,7 @@ class LoginController < ApplicationController
     end
 
     if @found_user.is_banned?
-      flash.now[:error] = "Your acocunt has been banned."
+      flash.now[:error] = "Your account has been banned."
       return forgot_password
     end
 
--- a/spec/controllers/login_controller_spec.rb
+++ b/spec/controllers/login_controller_spec.rb
@@ -68,6 +68,12 @@ describe LoginController do
         expect(flash[:error]).to match(/wiped/)
       end
     end
+
+    it "doesn't show login form if user is already logged in" do
+      post :login, params: { email: user.email, password: 'asdf' }
+      get :login
+      expect(response).to redirect_to('/')
+    end
   end
 
   describe "/login/reset_password" do
@@ -97,6 +103,12 @@ describe LoginController do
         post :reset_password, params: { email: deleted_wiped.email }
         expect(flash[:success]).to be_nil
       }.not_to(change { User.find(deleted_wiped.id).password_reset_token })
+    end
+
+    it "doesn't show reset form if user is already logged in" do
+      post :login, params: { email: user.email, password: 'asdf' }
+      get :login
+      expect(response).to redirect_to('/')
     end
   end
 
--- a/spec/controllers/login_controller_spec.rb
+++ b/spec/controllers/login_controller_spec.rb
@@ -71,7 +71,7 @@ describe LoginController do
 
     it "doesn't show login form if user is already logged in" do
       post :login, params: { email: user.email, password: 'asdf' }
-      get :login
+      get :index
       expect(response).to redirect_to('/')
     end
   end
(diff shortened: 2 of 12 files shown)


spec/controllers/login_controller_spec.rb
```diff
@@ -68,6 +68,12 @@
         expect(flash[:error]).to match(/wiped/)
       end
     end
+
+    it "doesn't show login form if user is already logged in" do
+      post :login, params: { email: user.email, password: 'asdf' }
+      get :login
```
thomasdziedzic: This should be testing the :index rather than :login.

spec/controllers/login_controller_spec.rb
```diff
@@ -98,6 +104,12 @@
         expect(flash[:success]).to be_nil
       }.not_to(change { User.find(deleted_wiped.id).password_reset_token })
     end
+
+    it "doesn't show reset form if user is already logged in" do
+      post :login, params: { email: user.email, password: 'asdf' }
+      get :login
```
thomasdziedzic: This should be testing the :forgot_password rather than :login

spec/controllers/login_controller_spec.rb
```diff
@@ -98,6 +104,12 @@
         expect(flash[:success]).to be_nil
       }.not_to(change { User.find(deleted_wiped.id).password_reset_token })
     end
+
+    it "doesn't show reset form if user is already logged in" do
```
thomasdziedzic: Since this is testing forgot password, it would be more appropriate to say something like `doesn't show forgot password form ...`

spec/features/comment_spec.rb
```diff
@@ -83,6 +83,7 @@
           created_at: 90.days.ago,
           comment: "Cool story.",
         )
+        reset_session!
```
thomasdziedzic: Rather than resetting the session which uses an internal method, I would prefer that you visit the login path to perform the actual login in a feature spec.

pushcx:
Thomas, thanks for jumping in to guide this PR, I appreciate you working through it with arandomandy.

spec/features/comment_spec.rb
arandomandy: How can I go about this? Should I create a feature spec to test: User A logs in --> User A visits the logout path --> User B logs in?

If so, what about this test? I'd still need to somehow log out from user, so stub_login_as user1 wouldn't fail.

If it's too much trouble to explain, feel free to push the changes you'd make. Hopefully, I'll learn from that.

thomasdziedzic: To logout from the user, you can do something like:

```ruby
visit '/settings'
click 'Logout'
```

arandomandy: I actually tried doing that before and it made another feature spec fail: `rspec ./spec/features/homepage_spec.rb:7`

From what I could see, **page** does not contain the expected **story.title** value when all tests are executed `bundle exec rspec`.

1. Running all tests
 a) using `reset_session!`
   **story.title** equals **story.title.18** which is contained in page 
 b) using the logout path visit `"/settings" and click_on "Logout"`
   **story.title** equals **story.title.16** which is not contained in page 

I'm not actually sure what might be causing this, but it looks like it is due to some dependency to stories previously created other tests. Perhaps we're not tearing them down?

I'll leave the error message down below just in case anyone wants to take a look at it. 

```
Failures:

  1) Reading Homepage when logged out reading a story
     Failure/Error: expect(page).to have_content(story.title)
       expected to find text "story title 18" in "Lobsters Recent Comments Search\nLogin\n1\nstory title 16 tag1 tag2 example.com\nvia username53 just now | cached | 1 comment\nModeration Log Hats Wiki Privacy About"
     # ./spec/features/homepage_spec.rb:11:in `block (3 levels) in <top (required)>'

Finished in 40.67 seconds (files took 2.67 seconds to load)
176 examples, 1 failure

Failed examples:

rspec ./spec/features/homepage_spec.rb:7 # Reading Homepage when logged out reading a story
```

thomasdziedzic: This looks like a potential source for a lot of future headaches if we don't figure it out in this PR.
@pushcx any idea what might be causing this?

thomasdziedzic: I believe I got this down to a minimal test case:

having the following click_on will fail the homepage_spec:

```diff
diff --git a/spec/features/comment_spec.rb b/spec/features/comment_spec.rb
index fe8909f..0e37460 100644
--- a/spec/features/comment_spec.rb
+++ b/spec/features/comment_spec.rb
@@ -83,7 +83,8 @@ RSpec.feature "Commenting" do
           created_at: 90.days.ago,
           comment: "Cool story.",
         )
-        reset_session!
+        visit '/settings'
+        click_on 'Logout'
 
         stub_login_as user1
         visit "/s/#{story.short_id}"
```

with the following error, which needs to be reproduced with the entire suite:

```
$ bundle exec rspec
...................................................F............................................................................................................................

Failures:

  1) Reading Homepage when logged out reading a story
     Failure/Error: expect(page).to have_content(story.title)
       expected to find text "story title 18" in "Lobsters Recent Comments Search\nLogin\n1\nstory title 16 tag1 tag2 example.com\nvia username53 just now | cached | 1 comment\nModeration Log Hats Wiki Privacy About"
     # ./spec/features/homepage_spec.rb:9:in `block (3 levels) in <top (required)>'

Finished in 13.91 seconds (files took 1.72 seconds to load)
176 examples, 1 failure

Failed examples:

rspec ./spec/features/homepage_spec.rb:7 # Reading Homepage when logged out reading a story
```

But then changing the following in the home_controller fixes the issue:

```diff
diff --git a/app/controllers/home_controller.rb b/app/controllers/home_controller.rb
index da2d18f..c3ca659 100644
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -66,9 +66,8 @@ class HomeController < ApplicationController
   end
 
   def index
-    @stories, @show_more = get_from_cache(hottest: true) {
-      paginate stories.hottest
-    }
+    @stories = Story.all
+    @show_more = false
 
     @rss_link ||= {
       :title => "RSS 2.0",
```

This change passes the entire suite. So I think it is related to caching not properly being cleared between runs, I will have to investigate how lobse.rs performs caching in order to see the best way to clear it between runs.

thomasdziedzic: Here is the final diff which works for me, I had to disable the rails cache in the test environment. I think that's what we want anyways.

```diff
diff --git a/config/environments/test.rb b/config/environments/test.rb
index 0a4a643..b059bb7 100644
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -7,6 +7,9 @@ Rails.application.configure do
   # and recreated between test runs. Don't rely on the data there!
   config.cache_classes = true
 
+  # disable caching when running tests
+  config.cache_store = :null_store
+
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
diff --git a/spec/features/comment_spec.rb b/spec/features/comment_spec.rb
index fe8909f..0e37460 100644
--- a/spec/features/comment_spec.rb
+++ b/spec/features/comment_spec.rb
@@ -83,7 +83,8 @@ RSpec.feature "Commenting" do
           created_at: 90.days.ago,
           comment: "Cool story.",
         )
-        reset_session!
+        visit '/settings'
+        click_on 'Logout'
 
         stub_login_as user1
         visit "/s/#{story.short_id}"
```

app/controllers/login_controller.rb
```diff
@@ -17,9 +17,13 @@ def logout
   end
 
   def index
-    @title = "Login"
-    @referer ||= request.referer
-    render :action => "index"
+    if @user
```
pushcx: Please do the guard-clause style of `return redirect_to "/" unless @user` followed by an empty line.

spec/features/comment_spec.rb
pushcx: Good research! Removing the story cache use from `HomeController` is a non-starter, the site would fall over. I suspect disabling caching like this would break tests of the `Keystore` model... but we don't have any of those, so this is a solid fix for now.

pushcx:
Thanks for taking this on. It's satisfying to squash these small bugs. If you bring in @thomasdziedzic's suggestion for the logout test, this is ready to merge.

pushcx:
Deployed this fix. Thanks for contributing so many, and thanks to @thomasdziedzic for puzzling out our test issue.

Issue #731, opened 2019-08-30
mail_new_activity undefined method on empty database


alanpost:
I've got a new Lobsters instance, with a single user but no stories, tags, comments, etc.  When I run the mail_new_activity.rb script, I get the following error:

    script/mail_new_activity.rb:69:in `<main>': undefined method `id' for nil:NilClass (NoMethodError)
    error at line 14, in call to /usr/local/bin/bundle exec ruby script/mail_new_activity.rb

This is line 69:

    last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last.id).to_i

Is it possible to make this script no-op silently when there are no stories?

alanpost:
Upon submitting stories, a similar error related to missing comments:

    script/mail_new_activity.rb:146:in `<main>': undefined method `id' for nil:NilClass (NoMethodError)
    error at line 14, in call to /usr/local/bin/bundle exec ruby script/mail_new_activity.rb

This is lines 145-146:

    last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) ||
      Comment.last.id).to_i

PR #732, opened 2019-09-04
Make mail_new_activity no-op silently

arandomandy:

### DIFF
--- a/script/mail_new_activity.rb
+++ b/script/mail_new_activity.rb
@@ -66,7 +66,7 @@ if __FILE__ == $PROGRAM_NAME
 
   mailing_list_users = User.where("mailing_list_mode > 0").select(&:is_active?)
 
-  last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last.id).to_i
+  last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last && Story.last.id).to_i
 
   Story.where("id > ? AND is_expired = ?", last_story_id, false).order(:id).each do |s|
     s.fetch_story_cache!
@@ -142,8 +142,7 @@ if __FILE__ == $PROGRAM_NAME
 
   # repeat for comments
 
-  last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) ||
-    Comment.last.id).to_i
+  last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) || Comment.last && Comment.last.id).to_i
 
   Comment.where(
     "id > ? AND (is_deleted = ? AND is_moderated = ?)",


arandomandy:
> I think this is related to #731
> 
> I would suggest adding a test for this fix at https://github.com/lobsters/lobsters/blob/master/spec/scripts/mail_new_activity_spec.rb so that in the future this issue doesn't reappear and so that we can show that the fix does what it says on the tin.

arandomandy:
> I think this is related to #731
> 
> I would suggest adding a test for this fix at https://github.com/lobsters/lobsters/blob/master/spec/scripts/mail_new_activity_spec.rb so that in the future this issue doesn't reappear and so that we can show that the fix does what it says on the tin.

I thought about this, but I couldn't figure out a way to test it since I am so new to RoR. If anyone knows how to go about testing it, feel free to jump in.

cc @thomasdziedzic @pushcx @hmadison

script/mail_new_activity.rb
```diff
@@ -66,7 +66,7 @@ def story_subject(story, prefix = "")
 
   mailing_list_users = User.where("mailing_list_mode > 0").select(&:is_active?)
 
-  last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last.id).to_i
+  last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last && Story.last.id).to_i
```
thomasdziedzic: I think you can use the safe navigation operator here http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

```suggestion
  last_story_id = (Keystore.value_for(LAST_STORY_KEY) || Story.last&.id).to_i
```

script/mail_new_activity.rb
```diff
@@ -142,8 +142,7 @@ def story_subject(story, prefix = "")
 
   # repeat for comments
 
-  last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) ||
-    Comment.last.id).to_i
+  last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) || Comment.last && Comment.last.id).to_i
```
thomasdziedzic: Same comment here with the safe navigation operator.

```suggestion
  last_comment_id = (Keystore.value_for(LAST_COMMENT_KEY) || Comment.last&.id).to_i
```

script/mail_new_activity.rb
pushcx: We have a custom rubocop linter preventing it.

pushcx:
I'm OK without a test - someone can only run into this bug once when first starting a site, and it's harmless.

Thanks for fixing this, @arandomandy and thanks for the review, @thomasdziedzic.

PR #738, opened 2019-09-11
csp (attempt 2)

thomasdziedzic:
Hopefully fixing the issues described at https://github.com/lobsters/lobsters/pull/737

Added `connect_src :self` to fix the commenting issues.

Also found an additional case that we have to handle when submitting stories.
Unfortunately it performs an eval when previewing a story submission, so we have to enable that as well.

Finally, I learned that we have a comment spec which should have caught the posting issue, but it succeeded so my guess is that capybara tests do not enforce content security policies during a test which is a bummer.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -26,4 +26,4 @@ end
 # Report CSP violations to a specified URI
 # For further information see the following documentation:
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
-Rails.application.config.content_security_policy_report_only = false
+Rails.application.config.content_security_policy_report_only = true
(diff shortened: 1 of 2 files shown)


thomasdziedzic:
@pushcx I gave myself some time to think of the solution and I would be more comfortable if we kept it report only and modified the csp to match what we would have deployed to production.

This will allow more time to gather metrics on blocked content and allow us to catch any further major issues.

pushcx:
Sorry, I'm not sure I understand your suggestion here. I see you set this PR to report-only - so you want me to merge + deploy this to generate some metrics for us to analyze to inform later enforcing the CSP?

thomasdziedzic:
Yup, I want to generate some more metrics with this updated policy. I know there are some browser plugins which might cause false positives, but I want to make sure we can explain all the metrics before merging this in.

Maybe at some point in a couple of weeks, we can revisit all the generated reports and see if there is any need for further policy adjustment.

pushcx:
Deployed.

PR #741, opened 2019-09-15
Prevent submission of login and reset password forms from logged in users

arandomandy:

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -17,13 +17,11 @@ class LoginController < ApplicationController
   end
 
   def index
-    if @user
-      redirect_to "/"
-    else
-      @title = "Login"
-      @referer ||= request.referer
-      render :action => "index"
-    end
+    return redirect_to "/" if @user
+
+    @title = "Login"
+    @referer ||= request.referer
+    render :action => "index"
   end
 
   def login
@@ -102,12 +100,10 @@ class LoginController < ApplicationController
   end
 
   def forgot_password
-    if @user
-      redirect_to "/"
-    else
-      @title = "Reset Password"
-      render :action => "forgot_password"
-    end
+    return redirect_to "/" if @user
+
+    @title = "Reset Password"
+    render :action => "forgot_password"
   end
 
   def reset_password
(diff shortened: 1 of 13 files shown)


app/controllers/login_controller.rb
```diff
@@ -7,6 +7,8 @@ class LoginFailedError < StandardError; end
 class LoginController < ApplicationController
   before_action :authenticate_user
   before_action :check_for_read_only_mode, :except => [:index]
+  before_action :require_no_user_or_redirect,
+                only: [:index, :login, :forgot_password, :reset_password]
```
arandomandy: Should `:only => [:index, :login, :forgot_password, :reset_password]` be used here rather than `only: [:index, :login, :forgot_password, :reset_password]`?

I checked the rails documentation on archive.org and it seems that `:only =>` is the older version of `only:`

cc @pushcx

spec/controllers/login_controller_spec.rb
```diff
@@ -110,6 +116,12 @@
       get :forgot_password
       expect(response).to redirect_to('/')
     end
+
+    it "doesn't allow submission from logged in users" do
+      post :login, params: { email: user.email, password: 'asdf' }
+      post :reset_password, params: { email: user.email, password: 'asdf' }
```
thomasdziedzic: reset_password doesn't take a password

```suggestion
      post :reset_password, params: { email: user.email }
```

app/controllers/login_controller.rb
pushcx: Ruby hashes with symbol keys used to look like this `{ :only => [] }`. A few years after the Lobsters codebase started, Ruby added a new style for this, `{ only: [] }` in 2007 and it took a long time to become popular. Some people don't like the inconsistency that Ruby requires `=>` for any key that's not a symbol.

The Lobsters codebase [does not consistently](https://github.com/lobsters/lobsters/pull/460/commits/69df28173823908708480da0dc7b6d8741268765) enforce [one or the other](https://www.rubydoc.info/gems/rubocop/0.46.0/RuboCop/Cop/Style/HashSyntax). It's not worth the churn to change them all. Try to use whichever is in use in that place or your personal preference if you're changing the line anyways.

pushcx:
Looks great. You've developed this into a clean, idiomatic solution to the bug, thank you.

PR #743, opened 2019-09-17
Fix broken anchor link in about page.

aterweele:


pushcx:
Ah, nice to have this little fix, thank you!

PR #744, opened 2019-09-20
Fixes #733 by refactoring the tagged query

### LINKED ISSUE #733: Stories with multiple tags displayed multiple times
The site allows viewing stories by tag, and accepts a list of tags. Eg, `https://lobste.rs/t/elixir,erlang` will show all stories that have either of those tags.

However, any story that has more than one of those tags will be listed multiple times. In this screenshot from `https://lobste.rs/t/elixir,erlang,browsers,compilers` you can see that a story with two of the requested tags appears twice and a story with all four appears four times.
Screen Shot 2019-09-04 at 8 02 51 AM

lachie:
Also adds a simple spec for `.tagging`.

### DIFF
--- a/app/models/story_repository.rb
+++ b/app/models/story_repository.rb
@@ -41,7 +41,7 @@ class StoryRepository
   def tagged(tags)
     tagged = Story.base.positive_ranked.joins(:taggings).where(taggings: { tag_id: tags.map(&:id) })
 
-    tagged.order(created_at: :desc)
+    tagged.distinct.order(created_at: :desc)
   end
 
   def top(length)
--- a/app/models/story_repository.rb
+++ b/app/models/story_repository.rb
@@ -39,9 +39,9 @@ class StoryRepository
   end
 
   def tagged(tags)
-    taggings = Tagging.select(:story_id).where(tag_id: tags.map(&:id))
+    tagged_story_ids = Tagging.select(:story_id).where(tag_id: tags.map(&:id))
 
-    Story.base.positive_ranked.where(id: taggings).order(created_at: :desc)
+    Story.base.positive_ranked.where(id: tagged_story_ids).order(created_at: :desc)
   end
 
   def top(length)
--- a/app/models/story_repository.rb
+++ b/app/models/story_repository.rb
@@ -39,9 +39,9 @@ class StoryRepository
   end
 
   def tagged(tags)
-    tagged = Story.base.positive_ranked.joins(:taggings).where(taggings: { tag_id: tags.map(&:id) })
+    taggings = Tagging.select(:story_id).where(tag_id: tags.map(&:id))
 
-    tagged.distinct.order(created_at: :desc)
+    Story.base.positive_ranked.where(id: taggings).order(created_at: :desc)
   end
 
   def top(length)
--- a/spec/models/story_repository_spec.rb
+++ b/spec/models/story_repository_spec.rb
@@ -29,6 +29,19 @@ describe StoryRepository do
     end
   end
 
+  describe ".tagged" do
+    it "selects unique tagged stories" do
+      tag1 = create(:tag)
+      tag2 = create(:tag)
+      story = create(:story, user: submitter, title: "A story", tags: [tag1, tag2])
+
+      tagged = repo.tagged([tag1, tag2])
+
+      expect(tagged.count).to be 1
+      expect(tagged.first).to eq story
+    end
+  end
+
   def story_titles_from(user)
     repo.newest_by_user(user).map(&:title)
   end


pushcx:
I'm thrilled to see the spec, thanks especially for adding a test. And to add on to @thomasdziedzic's comment, I've slowly learned to be suspicious when I see `distinct` - when I've used it, I often later learn the hard way that I was papering over a misunderstanding in what the query was doing. I think the important thing about the second query is that it expresses a slightly different meaning. The first one is "give me the stories and their taggings", the second is "give me the stories in the set of stories with these tags". We have dupes because the first one is not exactly what we mean to do.

lachie:
Hey,
Thanks for the feedback, I've switched it to the subquery version.

I didn't do any irl benchmarking, but here's a quick explain of both queries (the current query generated via `#to_sql`)
```
MariaDB [lobsters_dev]> explain SELECT DISTINCT `stories`.* FROM `stories` INNER JOIN `taggings` ON `taggings`.`story_id` = `stories`.`id` WHERE `stories`.`merged_story_id` IS NULL AND `stories`.`is_expired` = FALSE AND ((CAST(upvotes AS signed) - CAST(downvotes AS signed)) >= 0) AND `taggings`.`tag_id` IN (1, 2) ORDER BY `stories`.`created_at` DESC;
+------+-------------+----------+--------+-------------------------------------------------------------------------------+--------------------+---------+--------------------------------+------+--------------------------------------------------------+
| id   | select_type | table    | type   | possible_keys                                                                 | key                | key_len | ref                            | rows | Extra                                                  |
+------+-------------+----------+--------+-------------------------------------------------------------------------------+--------------------+---------+--------------------------------+------+--------------------------------------------------------+
|    1 | SIMPLE      | taggings | range  | story_id_tag_id,taggings_tag_id_fk                                            | taggings_tag_id_fk | 8       | NULL                           | 4    | Using index condition; Using temporary; Using filesort |
|    1 | SIMPLE      | stories  | eq_ref | PRIMARY,is_idxes,index_stories_on_is_expired,index_stories_on_merged_story_id | PRIMARY            | 8       | lobsters_dev.taggings.story_id | 1    | Using where                                            |
+------+-------------+----------+--------+-------------------------------------------------------------------------------+--------------------+---------+--------------------------------+------+--------------------------------------------------------+

MariaDB [lobsters_dev]> explain select s.title from stories s where (s.upvotes-s.downvotes) >= 0 and s.id in (select story_id from taggings where tag_id in (1, 2));
+------+--------------+-------------+--------+------------------------------------+--------------------+---------+--------------------------------+------+-----------------------+
| id   | select_type  | table       | type   | possible_keys                      | key                | key_len | ref                            | rows | Extra                 |
+------+--------------+-------------+--------+------------------------------------+--------------------+---------+--------------------------------+------+-----------------------+
|    1 | PRIMARY      | <subquery2> | ALL    | distinct_key                       | NULL               | NULL    | NULL                           | 4    |                       |
|    1 | PRIMARY      | s           | eq_ref | PRIMARY                            | PRIMARY            | 8       | lobsters_dev.taggings.story_id | 1    | Using where           |
|    2 | MATERIALIZED | taggings    | range  | story_id_tag_id,taggings_tag_id_fk | taggings_tag_id_fk | 8       | NULL                           | 4    | Using index condition |
+------+--------------+-------------+--------+------------------------------------+--------------------+---------+--------------------------------+------+-----------------------+
3 rows in set (0.001 sec)
```
I haven't used mysql in a minute so don't really have an intuition on which is better from eyeballing.

The subquery one is more elegant in any case, so here it is!

pushcx:
Yep, they're basically equivalent... except the big difference is that `distinct` prompts MySQL to do an extra sort pass, that's what the `Using filesort` [note is about](https://dev.mysql.com/doc/refman/8.0/en/explain-output.html).

This looks like it'll execute two queries, so there's an extra round-trip to the server. Can you express this in one, perhaps with [ActiveRecord#merge](https://api.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-merge)?

lachie:
Hi, 
I did a `#to_sql` to verify that its one query which resembles the subquery suggestion:
```ruby
# assuming submitter, tag1, tag2
repo = StoryRepository.new(submitter)
tagged = repo.tagged([tag1, tag2])
puts tagged.to_sql
#=>
```
```sql
SELECT `stories`.* FROM `stories` 
WHERE `stories`.`merged_story_id` IS NULL 
AND `stories`.`is_expired` = FALSE 
AND ((CAST(upvotes AS signed) - CAST(downvotes AS signed)) >= 0) 
AND `stories`.`id` IN (SELECT `taggings`.`story_id` FROM `taggings` WHERE `taggings`.`tag_id` IN (3, 4)) 
ORDER BY `stories`.`created_at` DESC
```

pushcx:
Huh. Yeah, that's exactly what I meant, and I pulled your branch locally to test it out. I'm surprised that the `Tagging.select` doesn't trigger an independent query. Rails must be doing something very clever under the covers to recognize the query hasn't been evaluated yet and that it can fold it in.

Well, wonderful! Merging. Thanks so much for running this all the way down with a test.

Issue #745, opened 2019-09-24
database_consistency


pushcx:
I ran the [database_consistency](https://github.com/djezzzl/database_consistency) gem to look for issues. I'm not committing it to the `development` group in the `Gemfile` because took a little fiddling to get running, but the output looks correct.

If anyone would like to write migrations/validations to deal with one or more of these, they'd be some nice little improvements in reliability.

 - [x] fail Comment short_id column has limit in the database but do not have length validator
 - [x] fail Comment comment column is required in the database but do not have presence validator
 - [ ] fail Comment comment column has limit in the database but do not have length validator
 - [x] fail Comment markeddown_comment column has limit in the database but do not have length validator
 - [x] fail Hat hat column is required in the database but do not have presence validator
 - [x] fail Hat hat column has limit in the database but do not have length validator
 - [x] fail Hat link column has limit in the database but do not have length validator
 - [x] fail HatRequest hat column has limit in the database but do not have length validator
 - [x] fail HatRequest link column has limit in the database but do not have length validator
 - [x] fail HatRequest comment column has limit in the database but do not have length validator
 - [x] fail Invitation email column has limit in the database but do not have length validator
 - [x] fail Invitation code column has limit in the database but do not have length validator
 - [x] fail Invitation memo column has limit in the database but do not have length validator
 - [x] fail InvitationRequest code column has limit in the database but do not have length validator
 - [x] fail InvitationRequest email column has limit in the database but do not have length validator
 - [x] fail InvitationRequest name column has limit in the database but do not have length validator
 - [x] fail InvitationRequest memo column has limit in the database but do not have length validator
 - [x] fail InvitationRequest ip_address column has limit in the database but do not have length validator
 - [ ] fail Keystore key column has limit in the database but do not have length validator
 - [ ] fail Message short_id column has limit in the database but do not have length validator
 - [ ] fail ModNote note column has limit in the database but do not have length validator
 - [ ] fail ModNote markeddown_note column is required in the database but do not have presence validator
 - [ ] fail ModNote markeddown_note column has limit in the database but do not have length validator
 - [ ] fail Moderation action column has limit in the database but do not have length validator
 - [ ] fail Moderation reason column has limit in the database but do not have length validator
 - [ ] fail ReplyingComment user_id column is required in the database but do not have presence validator
 - [ ] fail ReplyingComment story_id column is required in the database but do not have presence validator
 - [ ] fail ReplyingComment comment_created_at column is required in the database but do not have presence validator
[...]

PR #749, opened 2019-10-01
Add validation for comments, hats, hat requests

magikid:
This adds validation and tests for the Comment, Hat, and HatRequests model to fix some parts of #745

### DIFF
--- a/spec/models/hat_request_spec.rb
+++ b/spec/models/hat_request_spec.rb
@@ -1,0 +1,21 @@
+require 'spec_helper'
+
+describe HatRequest do
+  it "has a limit on the hat field" do
+    hat_request = HatRequest.make!
+    hat_request.hat = "a" * 256
+    expect(hat_request).to_not be_valid
+  end
+
+  it "has a limit on the link field" do
+    hat_request = HatRequest.make!
+    hat_request.link = "a" * 256
+    expect(hat_request).to_not be_valid
+  end
+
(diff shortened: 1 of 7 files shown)


spec/models/hat_request_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+describe HatRequest do
+  it "has a limit on the hat field" do
+    hat_request = build(:hat_request, hat: "a" * 256)
+    expect(hat_request).to_not be_valid
```
thomasdziedzic: I just realized that I opened a PR that fixes a subset of your work.

I used the shoulda-matchers gem to get some cleaner specs written.

https://github.com/lobsters/lobsters/pull/750/files#diff-65783f9ae3d2c150b4912f3ebf364df4R5

Thoughts on switching to that gem for these tests?

magikid: I would rather not use that for testing these validations.

I definitely think they read better but I feel like they don't show what the validation is actually doing as well since with the shoulda matchers you don't construct an invalid object.

thomasdziedzic: Ok, since you plan on doing more than me, I'll close out my PR.

pushcx:
Thanks for starting to burn down this list. I know it's unexciting, but I'm glad to prevent the inevitable future bugs.

Issue #746, opened 2019-09-26
/search.rss and /search?format=rss returns HTTP 500


da2x:
The following links return a HTTP 500. I realize that RSS from search results isn't an officially documented and supported feature, but the two first URLs should probably return the HTML page instead if RSS isn't supported.

https://lobste.rs/search?order=newest&q=domain:arstechnica.com&format=rss
https://lobste.rs/search?order=newest&q=interesting&format=rss
https://lobste.rs/search.rss?order=newest&q=domain:arstechnica.com

Update: also applies to &format=atom and search.atom.

finnoleary:
Could you give an example of a search that it does work on?

Also, wouldn't it be correct functionality, though? I do not know the conventions around Atom or RSS, but trying to parse HTML as RSS will fail, which is likely to do unexpected things depending on how each feed reader chooses to handle it. Whereas leaving it as a 500 makes any possible feed reader behaviour entirely predictable. 

I guess what I'm really asking is, what's your use case, and is this normal expected behaviour for RSS feeds?

pushcx:
Yeah. We should probably fail with 400 (bad request).

pushcx:
I merged #765 to address this and noticed there's a different exception if the action uses render without respond_to and added a couple tests, one of which is not behaving: 6c804ee

emrox:
yes, I'll have to adapt it. Hope I can find some time this week

pushcx:
I spent a while tinkering with this. If there's a `render` call that named an `action` or `template`, but no `format` block, Rails tries to look up a template for any format requested and fails, throwing `MissingTemplate`. The implict render doesn't throw. This difference is weird and possibly buggy.

In 891aabbf I removed the `render` calls that exhibit this behavior and are redundant because they give the name of the action, which the implict render finds anyways. This still leaves the problem for `/recent` because it renders the `index` template. So, solutions?

We could add `:constraints => { :format => [:html]}` to nearly every route, so much for convention over configuration.

We could add a `format` block to every `render`, so much for convention over configuration.

We could wrap `render` in `ApplicationController` to add a `format` block if not present. It sounds straightforward, but `render` has a complicated interface so it feels likely this would get bitten by unknowns.

We could symlink templates, e.g. from `recent.html.erb` to `index.html.erb`, but that seems like a setup to confuse someone who doesn't notice the templates are all the same file, or renames a template and breaks the symlinks.

[...]

PR #751, opened 2019-10-01
return 404 for search when requested format is not html

emrox:
this solves https://github.com/lobsters/lobsters/issues/746 and returns a 404 when the search is requested with a different format than html

room for improvement: don't execute the search when the requested format is not html. This would may require another method to keep the code cleaner and IDK if it's worth the effort

### DIFF
--- a/spec/controllers/search_controller_spec.rb
+++ b/spec/controllers/search_controller_spec.rb
@@ -2,16 +2,20 @@ require 'rails_helper'
 
 describe SearchController do
   describe '#index' do
-    it 'response okish for html requests' do
-      get :index
+    context 'when requesting html view' do
+      it 'responses with oki' do
+        get :index
 
-      expect(response).to have_http_status(:ok)
+        expect(response).to have_http_status(:ok)
+      end
     end
 
-    it 'responses with a 404 for rss requests' do
-      get :index, format: :rss
+    context 'when requested with another format' do
+      it 'responses with a 404' do
+        get :index, format: :rss
 
-      expect(response).to have_http_status(:not_found)
+        expect(response).to have_http_status(:not_found)
+      end
     end
   end
 end
--- a/spec/controllers/search_controller_spec.rb
+++ b/spec/controllers/search_controller_spec.rb
@@ -1,0 +1,17 @@
+require 'rails_helper'
+
+describe SearchController do
+  describe '#index' do
+    it 'response okish for html requests' do
+      get :index
+
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'responses with a 404 for rss requests' do
+      get :index, format: :rss
+
+      expect(response).to have_http_status(:not_found)
+    end
+  end
+end
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -2,7 +2,6 @@ class SearchController < ApplicationController
   def index
     respond_to do |format|
       format.html { render_search }
-      format.all { render status: :not_found, plain: '404 not found' }
     end
   end
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,5 +1,14 @@
 class SearchController < ApplicationController
   def index
+    respond_to do |format|
+      format.html { render_search }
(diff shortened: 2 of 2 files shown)


pushcx:
Running the search when we can't provide the output is painfully wasteful - search is our most expensive page on the site.

Why 404 instead of 400?

Is there a better global solution here, to catch all unavailable formats from the outside? I suspect we may have goofed something in Rails config that this bug can occur at all.

emrox:
> Why 404 instead of 400?

400 signals just "something was wrong here" and could be used as general error. 404 is more specific and tells the remote client the absence of the requested resource. It's like when requesting `file.rss` from a static webserver - when it's not found the webserver returns a 404, even when a `file.html` exists.

emrox:
> Is there a better global solution here

yes, but rss is a valid format for Lobsters, so it may should not be disallowed globally. AFAIK rss is allowed & valid for the most routes so it may make sense to just block it for individual routes. But it also could be done in the other way to not accept it globally and just allow it on certain routes

pushcx:
I was trying to be a bit coy when I asked about a global solution, but to be clearer and put my inbox at risk (I get an email for every site 500) - this is not a problem for this controller action, it's a bug for every route. (I promise it's real, please quit testing this bug in prod, folks.) That's why I suspect we have some kind of Rails misconfiguration.

emrox:
> this is not a problem for this controller action, it's a bug for every route. (I promise it's real, please quit testing this bug in prod, folks.) That's why I suspect we have some kind of Rails misconfiguration.

I'll take a look into it. Should not happen in general

emrox:
updated after merging https://github.com/lobsters/lobsters/pull/765

pushcx:
With #765 merged, what's left to do here?

emrox:
This PR just ensures the search is not executed when requested with an unknown format.

F.e. when someone requests the search with RSS the controller performs all actions just to figure out at the end that the rss view is missing.

spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+describe SearchController do
+  describe '#index' do
+    context 'when requesting html view' do
+      it 'responses with oki' do
```
pushcx:
```suggestion
      it 'responds with 200' do
```

spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        get :index
+
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    context 'when requested with another format' do
+      it 'responses with a 404' do
```
pushcx:
```suggestion
      it 'responds with 404' do
```

spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        get :index, format: :rss
+
```
pushcx: Please add a mock to test that a search didn't run, maybe that `Search.new` or `.search_for_user` wasn't called.

pushcx:
Wait, I left some small notes, but I just realized:

> This PR just ensures the search is not executed when requested with an unknown format.

If this is the case, then this is an issue for *every* action when a missing format is requested. This can't be fixing it at the right layer of abstraction, we must need to fix it in routes or `ApplicationController`. We shouldn't edit every individual action like this.

pushcx:
Hey @emrox, checking in on the PR queue. Any thoughts on how we could solve this at a higher layer for all actions?

emrox:
@pushcx in theory this already merged PR should do the most for all other things? https://github.com/lobsters/lobsters/pull/765

When you just want to have RSS for some routes it may should be solved in a different way. But I'm not sure. Maybe just have Html rendered, and RSS only when there is a view file for it?

pushcx:
It still runs the action. If I put `raise "dang"` at the top of `SearchController#index` I get that exception instead of a 404.

pushcx:
It's been a few months - were you still investigating, or should I close this PR and file a new issue to track the broader issue of running the action for unavailable formats?

emrox:
IMHO it can be closed. And it may be an edge case?

PR #752, opened 2019-10-03
add initial requests spec for the users controller

thomasdziedzic:
I wanted to learn more about request specs (due to https://github.com/lobsters/lobsters/issues/704) and how they differ from controller specs.


pushcx:
Nice. I quite like this style of "smoke test" that checks an important page loads and includes some information that no redesign will ever remove - it catches embarrassing bugs with negligible maintenance burden. eg ff1dc75523e537e1d3571befa96b2cdf598fe961

PR #753, opened 2019-10-03
bump rubocop

thomasdziedzic:
Keeping up with rubocop.


pushcx:
Nice to catch up, thanks. I don't know how rubocop manages to generate so many new opinions every few months; seems exhausting.

PR #754, opened 2019-10-03
Add validation for invitation and invitation request

AminArria:
This adds validation and tests for the `Invitation` and `InvitationRequest` models to fix some parts of #745

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -11,6 +11,8 @@ class Invitation < ApplicationRecord
     end
   end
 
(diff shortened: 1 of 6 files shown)


pushcx:
Nice to have factories and tests, thanks for such a thorough take on these.

pushcx:
I should've read more closely before I merged - would you mind calling `.valid?` and then asserting that the specific field is invalid rather than the whole model? This prevents false positive tests where this validation fails but something unrelated about the model is invalid and the test continues passing.

Issue #755, opened 2019-10-04
Accessibility and readability


dreadnip:
Hi,

I wanted to take a swing at improving the accessibility (good contrast, safer colours, correct attribute usage, etc..) and readability (fonts, font-size, spacing, whitespace use, colours also) of the site. But before I do this I'd like to know if you're open to this idea. If you guys are hard set on the current design then there would be no point.

pushcx:
I'm generally positive about things that make the site easier to read and use, especially to folks using screen readers or other assistive tech (as in the linked issue). But I'm pretty averse to [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality). So the answer is sure, wherever there's a clear justification that a change is an improvement. Sounds like we should start there for specific changes to make sure you don't waste your time.

dreadnip:
I'm not a Ruby dev so sadly I won't be able to contribute to the core functionality of the website anyway. What I can do is help clean up the front-end.

While doing a quick run through of the HTML/CSS I found that while all of it is still valid, improvements could definitely be made. Let's look at the header for example. Right now the structure is this: div > div (floated) > span > a + clearfix div. This could be rewritten to: header (flexbox) > a.

There are small inconstancies throughout the page, like the main list being an ordered list while its order is by definition variable and can change at any time.

Some devs scoff at these things, but having up-to-date, modern and semantically correct HTML is really important imo. I'll put something together for a PR.

svintbel:
As for me it's a good idea what you offer, but without good knowledge of Rails architecture you will face a lot of troubles cause of ruby integrations into HTML and covering actual views by tests.

dreadnip:
> As for me it's a good idea what you offer, but without good knowledge of Rails architecture you will face a lot of troubles cause of ruby integrations into HTML and covering actual views by tests.

That's true. I don't mind setting up a basic Ruby dev environment just so I can play with the site source myself, but I'd have to look into it a bit.

I've done some work on a static HTML snapshot of the homepage and uploaded it to my Github Pages: https://sanderdlm.be/lob/. I've kept the structure and style as close to the original as possible, while still improving on the things I thought could use some love. Does this look like something you guys could get behind?

pushcx:
I see how you've cleaned up the markup, but I don't know how to tell what's unfinished from what's an intentional change here. The spacing of stories is really funky, and the mobile view is very limited. Could you talk through what changes you were making, maybe separate changes to markup, accessibility, and style choices from each other?

[...]

PR #756, opened 2019-10-05
Add aria label to make comment text unique

HParker:
I saw https://github.com/lobsters/lobsters/issues/755 and was interested if there where any improvements Lobsters could make to accessibility without design changes. I found one that seems like it might help:

> In some situations, designers may choose to lessen the visual appearance of links on a page by using shorter, repeated link text such as "read more". These situations provide a good use case for aria-label in that the simpler, non-descriptive "read more" text on the page can be replaced with a more descriptive label of the link. The words 'read more' are repeated in the aria-label (which replaces the original anchor text of "[Read more...]") to allow consistent communication between users.
https://www.w3.org/TR/WCAG20-TECHS/ARIA8

This will allow screen readers to understand the context on these identical links.

Caviate:
I am not an accessibility expert and I found a conversation here: https://github.com/bryanbraun/anchorjs/issues/63
suggesting that it might be likely that this would make screen readers on lobsters more chatty and duplicate information in a not-strictly-helpful way. This contradicts the ARIA8 idea that all links with identical text should have an `aria-labelledby` or `aria-label` that uniquely identify it. So if others have more knowledge about how screen readers would work on Lobsters, I would be excited to learn from you!

I do expect that this is better for screen readers than today, but I am totally open to being wrong!

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -124,51 +124,60 @@ class="story <%= story.vote && story.vote[:vote] == 1 ? "upvoted" : "" %>
 
         <% if story.is_editable_by_user?(@user) %>
           |
-          <a href="<%= edit_story_path(story.short_id) %>" class="<%=
-            story.has_suggestions? ? "story_has_suggestions" : "" %>">edit</a>
+          <a href="<%= edit_story_path(story.short_id) %>"
+             class="<%= story.has_suggestions? ? "story_has_suggestions" : "" %>">edit</a>
         <% end %>
         <% if story.can_have_suggestions_from_user?(@user) %>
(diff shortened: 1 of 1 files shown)


dreadnip:
Another option for more screen reader friendly links is to include more text inside the link tag, but hide it with css. Something like

<a href="somewhere">read more <span class="hidden">about whatever this link is about</span></a>

pushcx:
Sounds like it'd be a good to hear from someone who knows ARIA very well, or uses a screen reader before merging. Maybe a meta thread if you don't know someone?

pushcx:
I'm closing this to keep our PR list tidy, but please feel free to reopen if you find a person/resource we can rely on for ARIA best practices.

PR #759, opened 2019-10-06
change several controller specs to request specs

yaaase:
👋 Hi! Long time lobste.rs users, first time contributor.

This is re: https://github.com/lobsters/lobsters/issues/704

I'm just incrementally doing these as time permits; if I'm not matching the intent of the issue, please let me know 😸 

Thanks!


spec/requests/merge_stories_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+describe 'merged stories', type: :request do
+  let(:user) { create(:user) }
+  let(:mod) { create(:user, :moderator) }
+  let(:merged_story) { create(:story) }
+  let(:story) { create(:story) }
+
+  context 'user votes on merged story' do
+    before {
```
thomasdziedzic: Multi line blocks tend to use the `do ... end` notation rather than the `{ ... }` notation.

yaaase: Thanks! Changed that before block

spec/requests/csp_violation_report_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+          "form-action 'self'",
+          "report-uri http://localhost:3000/csp-violation-report",
+        ].join('; '),
+        "referrer" => "http://localhost:3000/",
+        "violated-directive" => "img-src",
+      },
+    }
+
+    post '/csp-violation-report', params: { body: body.to_json, format: :json }
+    expect(response).to have_http_status(:ok)
```
jaysonvirissimo: Although this is strictly better than no test at all, this isn't really testing that it records the violation, so the it-message is somewhat misleading. How about changing the it-message to "responds appropriately" or adding an expectation that ensures the violation is being logged, maybe like this:
```ruby
it 'records the violation' do
  allow(Rails.logger).to receive(:info).with(expected_body)
  post '/csp-violation-report', params: { body: actual_body.to_json, format: :json }
  expect(Rails.logger).to have_received(:info).with(expected_body)
end
```
`expected_body` and `actual_body` could be set in let-blocks above.

spec/requests/keybase_proofs_spec.rb
```diff
@@ -14,12 +12,12 @@
   let(:create_params) { { keybase_proof: valid_kb_params } }
 
   before do
-    stub_login_as user
+    sign_in user
```
jaysonvirissimo: Nice 👍

spec/requests/csp_violation_report_spec.rb
jaysonvirissimo: LGTM 👍

spec/requests/csp_violation_report_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          "form-action 'self'",
+          "report-uri http://localhost:3000/csp-violation-report",
+        ].join('; '),
+        "referrer" => "http://localhost:3000/",
+        "violated-directive" => "img-src",
+      },
+    }
+  end
+
+  it 'response appopriately' do
```
pushcx:
```suggestion
  it 'responds appropriately' do
```

pushcx:
Typo fix to click 'accept' on and this is ready to merge! Thanks for working through these.

yaaase:
Thanks! Should be all set!

PR #762, opened 2019-10-08
switch 'cached' link back to the Internet Archive

girst:
Feedback welcome, commit message below. I'll post this to the lobsters thread as well. I haven't been able to test this out yet, but the CI seems happy with it (famous last words?). 

-----

As discussed in
https://lobste.rs/s/yadlxj/pointing_cached_at_internet_archive, point
the 'cached' link for a story to the Internet Archive's Wayback Machine.
The closest date to the submitted date will be selected, the newest one
if that is unavailable.
Also fixes the error message for failed archival (the URL pointed to
diffbot.com)

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -278,7 +278,8 @@ class Story < ApplicationRecord
   end
 
   def archive_url
-    "https://archive.is/#{CGI.escape(self.url)}"
+    datecode = (self.created_at || Time.current).utc.strftime("%Y%m%d%H%M%S")
+    "https://web.archive.org/web/#{datecode}/#{self.url}"
   end
 
   def as_json(options = {})
--- a/extras/story_cacher.rb
+++ b/extras/story_cacher.rb
@@ -40,12 +40,14 @@ class StoryCacher
       Rails.logger.error "error fetching #{db_url}: #{e.message}"
     end
 
+    ia_url = "https://web.archive.org/save/#{story.url}"
+
     begin
       s = Sponge.new
       s.timeout = 45
-      s.fetch(story.archive_url)
+      s.fetch(ia_url)
     rescue => e
-      Rails.logger.error "error caching #{db_url}: #{e.message}"
+      Rails.logger.error "error caching #{ia_url}: #{e.message}"
     end
 
     nil


cnst:
There's absolutely no reason for this commit.

I don't understand why you're even bothering to make a pull request when the discussion on lobste.rs has quite a number of good arguments against this change.

In addition to what's already been said on Lobsters — that we use the `cached` button as a short-term archival search engine (where archive.today shines), not as a long-term preservation tool (where archive.org shines), just compare something like the following two links:

* http://web.archive.org/web/*/www.apple.com
* http://archive.is/www.apple.com

Results:

* archive.org takes 11 seconds to load (and it's unusable until fully loaded), doesn't work without JavaScript at all (doesn't work in lynx at all), doesn't show any thumbnails for the individual snapshots (just an annoying popups from dates to times that take 11 seconds to load initially, does it even work on mobile without `:hover`? ).

* archive.is loads in about a second, doesn't seem to require JavaScript at all (works just fine in lynx), has thumbnail screenshots for each snapshot, better and faster UI

There are some other reasons, too; and both pros and cons for each service, depending on usecase; including the fact that archive.is can search archive.org, but not the other way around.

As such, given our requirements, I don't support changing from archive.is to archive.org for lobste.rs.

Please note that our usecase is different than Wikipedia.  I wouldn't support using the database-driven non-deterministic URL shortener links from archive.is on Wikipedia, because they could indeed disappear at any point without any alternatives, breaking thousands of articles without a deterministic way of fixing the issues.  But our usecase at Lobsters is different, and we could make this commit on a very short notice, easily fixing all broken links all at once should anything happen to archive.is, so, the decision on the present selection should be different as well.

girst:
On Tue, Oct 08, 2019 at 02:28:38PM -0700, Constantine A. Murenin wrote:
>There's absolutely no reason for this commit.
>
>I don't understand why you're even bothering to make a pull request 
>when the discussion on lobste.rs has quite a number of good arguments 
>against this change.

I really don't get why you come across this hostile in any discussion 
about this. You, and pretty much only you, have brought up "good 
arguments against this change," while nickp's post supporting this 
change has 12 upvotes.

girst:
reason for opening the PR now: the thread has dropped from the front 
page and I had a few spare minutes to implement it. and please don't 
forget that I don't have the final say in whether this gets merged, the 
maintainers do.

cnst:
Yes, and he seemingly changed his mind once I pointed out the issues.  Am I supposed to ask him to delete or edit his comment if my observation about him changing his mind is correct?  (BTW, why did you post a second comment above instead of correcting your original comment if you expect other to edit theirs?)

pushcx:
Tidying PRs - there's no consensus in favor of a change and not really even a consensus around what criteria we should use to decide so I'm going to default to leaving it as-is.

PR #763, opened 2019-10-09
Improved Invitation and InvitationRequest tests

AminArria:
Modified `Invitation` and `InvitationRequest` tests according to @pushcx comments on #754 

Also included tests to check if factory generates good models (fixed factory as well)

Should I add tests for all the validations on the models or is this enough?


pushcx:
Thank you so much for revisiting this to make the tests more reliable. This is plenty. Looks like there's just one rubocop rule to satisfy before this can merge.

PR #765, opened 2019-10-09
handle request for unknown formats properly

emrox:
from the comment in https://github.com/lobsters/lobsters/pull/751#issuecomment-539983909


pushcx:
With tests, no less! Thank you, and my inbox thanks you.

pushcx:
I ran out of time this morning, but I noticed there's a different exception if the action uses `render` without `respond_to` and added a couple tests, one of which is not behaving: 6c804eeea641d11fbbd9992435cbaee2197d9c32

PR #766, opened 2019-10-10
Add validations into models(#745)

svintbel:
- [x]  fail Keystore key column has limit in the database but do not have length validator
- [x]  fail Message short_id column has limit in the database but do not have length validator
- [x]  fail ModNote note column has limit in the database but do not have length validator
- [x]  fail ModNote markeddown_note column is required in the database but do not have presence validator
- [x]  fail ModNote markeddown_note column has limit in the database but do not have length validator
- [x]  fail Moderation action column has limit in the database but do not have length validator
- [x]  fail Moderation reason column has limit in the database but do not have length validator

===================================================
- [x]  fail Message author model should have proper foreign key in the database
- [x]  fail User sent_messages associated model should have proper index in the database
- [x]  fail User moderations associated model should have proper index in the database

===================================================
Also can be marked as done, as i told [before](https://github.com/lobsters/lobsters/issues/745#issuecomment-539024662):

- [x] fail User lower(username) model should have proper unique index in the database
- [x]  fail User lower(email) model should have proper unique index in the database

### DIFF
--- a/app/models/keystore.rb
+++ b/app/models/keystore.rb
@@ -1,7 +1,7 @@
 class Keystore < ApplicationRecord
   self.primary_key = "key"
 
-  validates :key, presence: true
+  validates :key, presence: true, length: { maximum: 50 }
 
   def self.get(key)
     self.find_by(:key => key)
@@ -12,6 +12,7 @@ class Keystore < ApplicationRecord
   end
 
   def self.put(key, value)
+    validate_input_key(key)
     if Keystore.connection.adapter_name == "SQLite"
       Keystore.connection.execute("INSERT OR REPLACE INTO " <<
         "#{Keystore.table_name} (`key`, `value`) VALUES " <<
@@ -25,7 +26,6 @@ class Keystore < ApplicationRecord
       kv.value = value
       kv.save!
     end
-
     true
   end
 
@@ -34,6 +34,7 @@ class Keystore < ApplicationRecord
   end
 
   def self.incremented_value_for(key, amount = 1)
+    validate_input_key(key)
     Keystore.transaction do
       if Keystore.connection.adapter_name == "SQLite"
         Keystore.connection.execute("INSERT OR IGNORE INTO " <<
@@ -90,5 +91,10 @@ class Keystore < ApplicationRecord
 
   def self.decremented_value_for(key, amount = -1)
     self.incremented_value_for(key, amount)
+  end
+
+  def self.validate_input_key(key)
+    exception = ActiveRecord::ValueTooLong.new('50 characters is the maximum allowed for key')
+    raise exception if key.length > 50
   end
 end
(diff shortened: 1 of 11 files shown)


db/migrate/20191008123620_add_foreign_key_for_messages_and_comments.rb
```diff
@@ -0,0 +1,13 @@
+class AddForeignKeyForMessagesAndComments < ActiveRecord::Migration[5.2]
```
pushcx: Messages may not have an `author_user_id` pointing to the `users` table; it can be `0` to indicate the message is from the system. (See `Message.author_username`.)

app/models/keystore.rb
```diff
@@ -91,4 +92,9 @@ def self.decrement_value_for(key, amount = -1)
   def self.decremented_value_for(key, amount = -1)
     self.incremented_value_for(key, amount)
   end
+
+  def self.validate_input_key(key)
+    exception = ActiveRecord::ValueTooLong.new('50 characters is the maximum allowed for key')
```
pushcx: I don't have a better solution, but it concerns me that now this length is validated here and on line 4. They could get out of sync. (Also, it's weird to assign an exception to a variable before raising it, I've never seen that before. Was it just for line length?)

svintbel: Yeap, I've looked through the app and how this class is using and found that it is using as service module, without existing any class instances so adding validation on line 4 for class instances is useless, but that stub database_consistency gem issue. By the way, that issue is right, as we have a limit in db for this field, so I decided to add this exception as we inserting data here through SQL without AR. If this gets out of sync we may remove it, no matter)

db/migrate/20191008123620_add_foreign_key_for_messages_and_comments.rb
svintbel: It is confusing for me, in this case, maybe it needs to add a partial index, but MySQL doesn't support this kind of index. May someone gives advice?

app/models/keystore.rb
pushcx: Maybe just move the `50` into a constant the class. Then it's not defined multiple places that can get out of sync, and anyone touching one place can't help but notice it's a shared value and that there's other uses to look at.

svintbel: Resolved

db/migrate/20191008123620_add_foreign_key_for_messages_and_comments.rb
svintbel: Maybe someone can give me a piece of advice on how to resolve this issue ```fail Message author model should have proper foreign key in the database``` in such case?

pushcx:
Sorry for the delay getting back to this, but I appreciate you adding these and your clear explanations of which we don't need to do. :)

PR #767, opened 2019-10-11
add replying comments and deleted stories

yaaase:
This contributes to https://github.com/lobsters/lobsters/issues/414 by adding replies to comments and user-deleted stories


pushcx:
Great, thanks for improving this, it'll help every developer of the codebase.

Issue #368, opened 2017-06-12
Don't use bad canonical links


pushcx:
On a story that was posted twice with a bad link, [emmab](https://lobste.rs/u/emmab) suggested:

> Modify Lobsters to not follow `rel=canonical` unless it can resolve a page with status 2**?

I think it's a good idea. I'm posting this issue for discussion and hopefully someone with free time can implement it.

pushcx:
For more context: a lot of sites have bad `<link rel=canonical>` tags in their headers. The two most common behaviors are pointing it at the homepage (they didn't understand what the tag is for) and not updating one that comes in a template (like in this story's case, pointing at `username.github.com`). I contact site owners to tell them about the bug when I can easily find contact info, but it would be nice if Lobsters could catch the second case.

Now that I write this up, I think we should probably also ignore any canonical link that points right to a domain name. It's the most popular failure mode, and if a Lobsters user is submitting a homepage with one the system would correctly use the URL they entered anyways.

simmel:
So this can be closed @pushcx ?

pushcx:
Yep, thanks for noticing it.

PR #769, opened 2019-10-11
do not use bad canonical links

yaaase:
This addresses https://github.com/lobsters/lobsters/issues/368 

Do not use the rel=canonical attribute if it is
1) unparseable by URI
2) parseable but resolves to the domain root (eg "google.com/")
3) parseable, but returns a non-200 level response code when retrieved via GET

This PR adds 2 fixture files which are the same aside from their rel=canonical attribute; if confusing, it is trivial to use different source sites for 3.html and 4.html (or perhaps they can be stripped down to merely their canonical attributes).

(This page was chosen out of autocomplete laziness / clicking what at the moment was the top-leftmost thing on mcsweeneys)

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -977,17 +977,35 @@ class Story < ApplicationRecord
 
     @fetched_attributes[:title] = title
 
-    # now get canonical version of url (though some cms software puts incorrect
-    # urls here, hopefully the user will notice)
+    # attempt to get the canonical url if it can be parsed,
+    # if it is not the domain root path, and if it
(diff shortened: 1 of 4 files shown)


pushcx:
Looks like some extra code snuck in the `fake_data` task? Otherwise this is perfect, thank you for digging in, this will save a regular hassle with broken links.

yaaase:
Ah sorry, did not check out a clean branch. Will fix this in the morning!

yaaase:
Ok, should be set now.

Issue #764, opened 2019-10-09
Mod editing: repost story description as comment


pushcx:
Pretty regularly, story submitters write a personal comment in the story text. A mod has to be the no-fun police, it's pretty frustrating for the submitter to repost a comment, and the modlog entry is pretty long and unreadable. Let's formalize this and make the experience better for everyone:

On the story editing form, add a checkbox (only for mods) below the story text labeled "repost description as comment". When this is checked, the text should be removed from the story and posted as a new comment by the story submitter with the same timestamp as the story. Rather than include the entire text of the comment, the modlog should say something like `"reposted description as comment https://lobste.rs/c/#{comment.short_id}"`.

Optional bonus points: make it possible for user suggestions to trigger this, similar to #519.

PR #772, opened 2019-10-17
Feature: Mod editing: repost story description as comment

shveikus:
Resolution for #764 :
_On the story editing form, add a checkbox (only for mods) below the story text labeled "repost description as comment". When this is checked, the text should be removed from the story and posted as a new comment by the story submitter with the same timestamp as the story. Rather than include the entire text of the comment, the modlog should say something like "reposted description as comment https://lobste.rs/c/#{comment.short_id}"._

Changes:
1. added checkbox into edit story view for moderators;
2. added service method into StoriesHelper;
3. added unit tests for feature points.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -465,7 +465,7 @@ private
 
   def repost_story_description
     if ActiveRecord::Type::Boolean.new.deserialize(story_params[:repost_description])
-      @story.description= ""
+      @story.description = ""
       comment = @story.comments.build
       comment.comment = story_params[:description].to_s
       comment.user = User.find(@story.user_id)
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -244,11 +244,12 @@ class StoriesController < ApplicationController
     @story.editor = @user
 
     if @story.url_is_editable_by_user?(@user)
-      @story.attributes = story_params
+      @story.attributes = story_params.except(:repost_description)
     else
-      @story.attributes = story_params.except(:url)
+      @story.attributes = story_params.except(:url, :repost_description)
     end
 
+    StoriesHelper.repost_story_description(@story, story_params) if @user && @user.is_moderator?
     if @story.save
       return redirect_to @story.comments_path
     else
@@ -385,7 +386,7 @@ private
     p = params.require(:story).permit(
       :title, :url, :description, :moderation_reason, :seen_previous,
       :merge_story_short_id, :is_unavailable, :user_is_author, :user_is_following,
-      :tags_a => [],
+      :repost_description, :tags_a => []
     )
 
     if @user && @user.is_moderator?
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -484,7 +484,7 @@ private
     m.story_id = @story.id
     comment = @story.comments.find(&:new_record?)
     comment.try(:save)
-    m.action = "reposted description as comment https://lobste.rs/c/#{comment.short_id}"
+    m.action = "Reposted description as comment https://lobste.rs/c/#{comment.short_id}"
     m.reason = @story.moderation_reason
     m.save
     @story.is_moderated = true
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -471,14 +471,14 @@ private
       comment.user = User.find(@story.user_id)
       comment.created_at = @story.created_at
       comment.updated_at = @story.created_at
-      log_moderation_repost_repost_description
+      log_mod_repost_description
       true
     else
       false
     end
   end
 
-  def log_moderation_repost_repost_description
+  def log_mod_repost_description
     m = Moderation.new
     m.moderator_user_id = @story.editor.try(:id)
     m.story_id = @story.id
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -250,6 +250,7 @@ class StoriesController < ApplicationController
     end
 
     StoriesHelper.repost_story_description(@story, story_params) if @user && @user.is_moderator?
+
     if @story.save
       return redirect_to @story.comments_path
     else
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -244,12 +244,12 @@ class StoriesController < ApplicationController
     @story.editor = @user
 
     if @story.url_is_editable_by_user?(@user)
-      @story.attributes = story_params.except(:repost_description)
+      @story.attributes = story_params
     else
-      @story.attributes = story_params.except(:url, :repost_description)
+      @story.attributes = story_params.except(:url)
     end
 
-    StoriesHelper.repost_story_description(@story, story_params) if @user && @user.is_moderator?
+    repost_story_description if @user && @user.is_moderator?
 
     if @story.save
       return redirect_to @story.comments_path
@@ -393,7 +393,7 @@ private
     if @user && @user.is_moderator?
       p
     else
-      p.except(:moderation_reason, :merge_story_short_id, :is_unavailable)
+      p.except(:moderation_reason, :merge_story_short_id, :is_unavailable, :repost_description)
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/controllers/stories_controller.rb
```diff
@@ -384,7 +386,7 @@ def story_params
     p = params.require(:story).permit(
       :title, :url, :description, :moderation_reason, :seen_previous,
       :merge_story_short_id, :is_unavailable, :user_is_author, :user_is_following,
-      :tags_a => [],
+      :repost_description, :tags_a => []
```
pushcx: This should go down on line 395 with the other mod-only attributes. Then you don't need the `story_params.except` above.

app/views/stories/edit.html.erb
```diff
@@ -10,6 +10,18 @@
     <% if @user.is_moderator? %>
       <br />
       <div class="box">
+        <!-- Because description can't be empty if URL is absent -->
```
pushcx: Good note, and I see how the existing validation covers this server-side.

spec/controllers/stories_controller_spec.rb
```diff
@@ -146,4 +146,69 @@
       end
     end
   end
+
+  describe '#UPDATE' do
```
pushcx: It's convention to match case of the method name. Solid tests, you got everything here.

spec/controllers/stories_controller_spec.rb
```diff
@@ -146,4 +146,69 @@
...
+    let(:attr) do
+      {
+        url: 'http://newexample.com/',
```
pushcx: Please use `example.com` - it's set aside for tests and examples. `newexample.com` is an SEO/marketing firm with an expired SSL cert.

spec/factories/story.rb
```diff
@@ -3,6 +3,7 @@
     association(:user)
     sequence(:title) {|n| "story title #{n}" }
     sequence(:url) {|n| "http://example.com/#{n}" }
+    sequence(:description) {|n| "#{n} Lorem ipsum dolor sit amet" }
```
pushcx: Remove this. If the test doesn't pass, it needs to explicitly create the `story` record it needs, not rely on an attribute of the global test setup.

app/models/story.rb
```diff
@@ -990,4 +993,10 @@ def fetched_attributes
 
     @fetched_attributes
   end
+
+  def check_for_description_repost(all_changes)
```
pushcx: Add an `attr_accessor` to `Story` (see line 141) to explicitly record + know here that this edit happened rather than infer it by checking for effects.

app/models/story.rb
```diff
@@ -569,11 +569,14 @@ def log_moderation
       m.moderator_user_id = self.editor.try(:id)
     end
     m.story_id = self.id
-
     if all_changes["is_expired"] && self.is_expired?
       m.action = "deleted story"
     elsif all_changes["is_expired"] && !self.is_expired?
       m.action = "undeleted story"
+    elsif check_for_description_repost(all_changes)
+      comment = self.comments.find(&:new_record?)
```
pushcx: Please move comment saving up to the controller. It's bad enough that `Story` creates `Moderation` records, let's not make it more of a god object.

app/helpers/stories_helper.rb
```diff
@@ -27,4 +27,18 @@ def is_unread?(comment)
 
     (comment.created_at > @ribbon.updated_at) && (comment.user_id != @user.id)
   end
+
+  def self.repost_story_description(story, story_params)
```
pushcx: The helpers in `StoriesHelper` are for the view, please move this up to a private method in the controller.

app/controllers/stories_controller.rb
svintbel: Oh, I've not found it before. Thanks.

spec/controllers/stories_controller_spec.rb
svintbel: done

app/helpers/stories_helper.rb
svintbel: Resolved.

app/models/story.rb
svintbel: done

app/models/story.rb
svintbel: Done.

spec/controllers/stories_controller_spec.rb
svintbel: Done.

spec/controllers/stories_controller_spec.rb
svintbel: Done.

spec/factories/story.rb
svintbel: Removed.

app/controllers/stories_controller.rb
```diff
@@ -459,4 +461,31 @@ def track_story_reads
...
+    end
+  end
+
+  def log_mod_repost_description
+    m = Moderation.new
+    m.moderator_user_id = @story.editor.try(:id)
+    m.story_id = @story.id
+    comment = @story.comments.find(&:new_record?)
+    comment.try(:save)
+    m.action = "reposted description as comment https://lobste.rs/c/#{comment.short_id}"
```
pushcx:
```suggestion
    m.action = "Reposted description as comment https://lobste.rs/c/#{comment.short_id}"
```

app/controllers/stories_controller.rb
pushcx: You have to remove it from here or regular users can set the param.

svintbel: Sorry, but I think they can't do that, as this param excepts on line 395 when user is regular with other moderation params.

shveikus:
@pushcx , sorry for the delay, could we resolve this issue?

app/controllers/stories_controller.rb
```diff
@@ -460,4 +462,31 @@ def track_story_reads
     yield
     @ribbon.touch
   end
+
+  def repost_story_description
+    if ActiveRecord::Type::Boolean.new.deserialize(story_params[:repost_description])
+      @story.description= ""
```
pushcx, line 468:
```suggestion
      @story.description = ""
```

pushcx: Please also set `@story.editor = @user.id`, then you can remove the `try` below on line 484. If a moderator editing a story doesn't create an entry in the moderation log, we should error out.

app/models/story.rb
```diff
@@ -571,6 +572,12 @@ def log_moderation
     all_changes = self.changes.merge(self.tagging_changes)
     all_changes.delete("unavailable_at")
 
+    # condition because of duplicating this logic in controller
```
pushcx, line 575:
```suggestion
    # the controller creates a more explicit modlog message for this
```

app/models/story.rb
```diff
@@ -571,6 +572,12 @@ def log_moderation
...
+    if self.repost_description == "1"
```
pushcx, line 576: Can this be `if self.repost_description`?

app/models/story.rb
pushcx: You can delete this method.

spec/models/user_spec.rb
```diff
@@ -111,7 +111,7 @@
   it "shows a user is recent or not" do
     user = create(:user, :created_at => Time.current)
     expect(user.is_new?).to be true
-    user = create(:user, :created_at => Time.current - (User::NEW_USER_DAYS.days + 1))
+    user = create(:user, :created_at => Time.current - (User::NEW_USER_DAYS.days + 1.day))
```
pushcx, line 114: Sorry about this bug - it was fixed in another commit, so if you want to merge master into your PR you won't need to make this change.

pushcx:
I found a couple small changes to make and then, yes, I'd be very happy to merge this PR down and make a chore of mine easier. :)

app/models/story.rb
shveikus: Yes, sorry)

app/models/story.rb
shveikus: I think, only with some extra logic, because view helper method ```check_box``` which used for repost description feature has default value.
```ruby
 check_box(object_name, method, options = {}, checked_value = "1", unchecked_value = "0")
``` 
[Docs](https://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-check_box)

spec/models/user_spec.rb
shveikus: Yep, sure.

shveikus:
Hi @pushcx, maybe we could resolve this PR, what do you think about it?)

PR #773, opened 2019-10-18
Update Scout gem to the latest version

matthewchigira:
Update the Scout APM gem to the latest version so that the Timeline view can be utilised.

https://docs.scoutapm.com/?_ga=2.117993099.324453617.1571089242-2135995770.1562204706#updating-to-the-newest-version

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem "ruumba" # tests views
 gem "sitemap_generator" # for better search engine indexing
 gem 'transaction_retry' # mitigate https://github.com/lobsters/lobsters-ansible/issues/39
 
-gem "scout_apm", :git => 'https://github.com/scoutapp/scout_apm_ruby.git', :ref => '574b0a6'
+gem "scout_apm"
 
 group :test, :development do
   gem 'bullet'
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem "ruumba" # tests views
 gem "sitemap_generator" # for better search engine indexing
 gem 'transaction_retry' # mitigate https://github.com/lobsters/lobsters-ansible/issues/39
 
-gem "scout_apm"
+gem "scout_apm", "2.6.2"
 
 group :test, :development do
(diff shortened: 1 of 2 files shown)


pushcx:
Context: this came out of discussion with Nate Berkopec, who thinks the timeline view would be useful to [his performance analysis](https://lobste.rs/s/5hshvd/proposal_lobste_rs_performance_analysis). 

@matthewchigira Please pin a specific version of the gem as discussed in the thread.

matthewchigira:
> @matthewchigira Please pin a specific version of the gem as discussed in the thread.

@pushcx Ok no problem, I've done that now.

pushcx:
Thanks!

BTW, some quick notes from reading:
 * you've slipped some tabs into your `lib/scout_apm/auto_instrument/parser.rb` and `lib/scout_apm/layer_converters/slow_request_converter.rb`; `lib/scout_apm/utils/backtrace_parser.rb` and `scout_apm.gemspec`, and `test/unit/auto_instrument/controller-instrumented.rb` gained trailing whitespace - eh, bunch more, I'm gonna stop listing, as you can add git/rubocop rule to find + prevent these
 * `lib/scout_apm/instant/middleware.rb` has a typo, `s/noise/noisy/`
 * the sig on `PayloadSerializer.serialize`, `SlowTransaction`, and related methods are kinda ridiculous, you should consider defining a shared interface on these args and passing an array (though this means eg. making a collection classes, it means one dumb organizer that calls map and per-collection small smart things, instead of vice-versa)
 * it's weird that `TrackedRequest.layer_insignificant?` isn't a method on `Layer` given that all the data to determine it could live in that class, and the `0`/`1` magic numbers really want to be constants
 * in `lib/scout_apm/utils/unique_id.rb`, `TransactionId` is a random string, but the comment in `TrackedRequest` says it's a UUID. One of these is wrong.
 * `AutoInstrument.local_assignments?`'s memoization could be more explicitly expressed as `@local_assignments[node] ||= if ...`, no explicit return

Issue #771, opened 2019-10-17
An inactive user has -1 stories submitted, rather than 0 stories.


alanpost:
A user invited to the site submitted a single story, which was flagged more than 
MIN_KARMA_TO_SUBMIT_STORIES times.  This user then deleted the story and their account.

The Stories Submitted field on their users page now reads "-1 (+1 deleted)".  It's not possible to have submitted negative stories, this field should read "0 (+1 deleted)" instead.

I expect the act of deleting the story is and was sufficient to produce this bug, regardless of whether the story was flagged 4+ times or not.  The bug can be seen here:

https://lobste.rs/u/soufianesakhi

PR #774, opened 2019-10-22
Fix error refreshing stories count

alexrs:
Resolves https://github.com/lobsters/lobsters/issues/771

When a user is inactive, [`User#refresh_counst!`](https://github.com/lobsters/lobsters/blob/791b21f60a2e952cabe5187559c8fb7724b7b8aa/app/models/user.rb#L344) is called. This method updates `user:#{self.id}:stories_submitted` with the value from `self.stories.not_deleted.count`.

When a user deletes a story and then deactivates the account, the value for `user:#{self.id}:stories_submitted` will be the number of stories submitted minus the number of stories deleted.

When the info of the user is displayed, [`stories_displayed`](https://github.com/lobsters/lobsters/blob/master/app/helpers/users_helper.rb#L7) already takes into account the stories that have been deleted.

So if a user submits 1 story and then removes it, `User#refresh_counst!` will update the value of `user:#{self.id}:stories_submitted` to `1 - 1`, this is, 0.

But then, when displaying the info, `User#stories_deleted_count` will be substracted to `User#stories_submitted_count`, so the deleted story will be substracted again, displaying the value of -1.

I am not familiar enough with the code to claim that this is the best fix for this issue nor it does not have any side effects in other parts of the codebase. Let me know if my approach is correct, and point me in the right direction if it is not.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -342,7 +342,7 @@ class User < ApplicationRecord
   end
 
   def refresh_counts!
-    Keystore.put("user:#{self.id}:stories_submitted", self.stories.not_deleted.count)
+    Keystore.put("user:#{self.id}:stories_submitted", self.stories.count)
     Keystore.put("user:#{self.id}:comments_posted", self.comments.active.count)
     Keystore.put("user:#{self.id}:comments_deleted", self.comments.deleted.count)
   end


alanpost:
Thank you for this PR @alexrs.  I really appreciate your attending to Issue #771 and you have my gratitude.  I need another reviewer to check this work, but your description of the fix is outstanding and the patch appropriately modest.

app/models/user.rb
```diff
@@ -342,7 +342,7 @@ def fetched_avatar(size = 100)
   end
 
   def refresh_counts!
-    Keystore.put("user:#{self.id}:stories_submitted", self.stories.not_deleted.count)
```
alanpost: [I'm transcribing information that was given to me, forgive any lossiness in the process.]

As I understand it, changing this settings will effect how karma is calculated:

https://github.com/lobsters/lobsters/blob/791b21f60a2e952cabe5187559c8fb7724b7b8aa/app/models/user.rb#L202-L208

In that we will no longer account for karma on deleted stories.  That's not appropriate, deleted stories should still have their karma accounted for.  Is it possible to update the user page to use stories_displayed, rather than changing the way we count stories_submitted?

alexrs: Aren't we using `stories_displayed` already?

 https://github.com/lobsters/lobsters/blob/2d1887cb35f761b7c7fdf95adce74003e19511d0/app/views/users/show.html.erb#L90

As far as I understand, the issue here is that when `UsersHelper#stories_submitted_content` is called, we compute the stories displayed as `stories_displayed = stories_submitted - stories_deleted`, but `User#refresh_counts!` already _substracted_ the deleted stories by counting only the not deleted ones.

https://github.com/lobsters/lobsters/blob/2d1887cb35f761b7c7fdf95adce74003e19511d0/app/models/user.rb#L342

Another option is to subtract `User#stories_deleted_count` when the karma is calculated. 

```ruby
k.to_f / (self.stories_submitted_count - self.stories_deleted_count + self.comments_posted_count)
```

This approach would make the way the karma is calculated clearer.

alexrs: By the way, thanks for the quick feedback!

alexrs:
Hey @alanpost. Have we reached an agreement on this?

pushcx:
Hi @alexrs,

Sorry for the very slow response. I dug in on your change and I'm happy to merge it as-is. The `stories_submitted` name is now clearer (as opposed to meaning "stories submitted minus stories deleted"), and the change to interpretation of `average_karma` is equally fair.

PR #775, opened 2019-10-29
Updating not null constraint on columns with presence validator

AminArria:
Issue #745 

Setting not NULL constraint for:
- Table 'hat_requests', columns 'hat', 'link', 'comment'
- Table 'invitation_requests', columns 'name', 'email'

Don't know why `collation: "utf8mb4_general_ci"` disappeared, should I add it by hand?

Also don't know why `replying_comments` changed, do you prefer that I remove that change from the commit or should I leave it?

### DIFF
--- a/db/migrate/20191029021735_fix_database_consistency_with_validators.rb
+++ b/db/migrate/20191029021735_fix_database_consistency_with_validators.rb
@@ -1,0 +1,9 @@
+class FixDatabaseConsistencyWithValidators < ActiveRecord::Migration[5.2]
+    def change
+        change_column_null :hat_requests, :hat, false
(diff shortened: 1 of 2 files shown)


pushcx:
I ran down the `replying_comments` change, it's a [bug fix from scenic](https://github.com/scenic-views/scenic/commit/07b1580b44d56f2d2af5bfc5d61e1083f7ea1b4a#diff-28afa2cc5a1525843d06dec026729ac0).

pushcx:
I don't see a collation change in the `db/schema.rb` and it didn't show up for me when I locally ran the migration, so I'm going to shrug and hope it's nothing.

PR #776, opened 2019-10-29
Convert several controller specs to request specs

yaaase:
I am not sure how to meaningfully reproduce the application controller spec however. See: https://github.com/rspec/rspec-rails/issues/2024


pushcx:
Worth updating these specs even if it's not 100%, thank you.

PR #777, opened 2019-10-30
add moderators and moderations of comments

yaaase:
moderators as a mod of users, moderations section began below


pushcx:
Thanks for adding these, I've checked them off in #414.

Issue #734, opened 2019-09-04
Extract 'domain' model from Story


pushcx:
The `Story` model has a `domain` method extracts the domain from a url. We're increasingly wanting a `Domain` model that `Story` can have a foreign key to for quicker searching, analytics, and feedback at story submit time.

So we'll need a migration to add the table (with a single not-null string column `name`) and add the FK for all existing stories. `Story#url=` can `find_or_create_by!` the association for new stories. As in `Story.find_similar_by_url`, it should normalize case and remove a leading `www.` Then tests for submitting a story with a never-seen before domain, a seen-before domain, and editing a story to change the URL.

emrox:
Should the domain be split into subdomain and domain or just the whole hostname?

f.e. for medium there could be
```
{ subdomain: 'mypage', name: 'medium.com' }
```

or it could be just

```
{ hostname: 'mypage.medium.com' }
```

pushcx:
I'd rather have a single field, it'll make the basic features a lot easier to build on top of. I know it won't capture aggregators like Medium.com and Dev.to well, but they're already pretty iffy on quality.

emrox:
When it's just one field maybe just add a `domain` row to the `Story` model?

pushcx:
I don't remember a compelling reason to have a separate `Domain` model if we use Rails 5's attributes API to extract an object for it to centralize that code a bit. Joins to look at flags by domain are going to wind up touching `stories` anyways. Not sure what I was thinking here.

pushcx:
This was completed a little while ago.

PR #778, opened 2019-10-31
Store normalized domain for story

emrox:
approach to save a normalized version (without `www` or `m` in the beginning) of the domain into the database - https://github.com/lobsters/lobsters/issues/734

existing stories can be updated after release:
```ruby
Story.where.not(url: ['', nil]).each(&:save)
```


app/models/story.rb
```diff
@@ -1008,4 +1013,15 @@ def valid_canonical_uri?(url)
   def canonical_target(parsed)
     parsed.at_css("link[rel='canonical']").attributes["href"].text
   end
+
+  def build_normalized_domain
+    begin
+      hostname = URI.parse(url).hostname.squish
```
pushcx: See `URL_RE` in this story - we don't use `URI.parse` much because it's not lenient enough. We'll need to use that like `set_domain` did.

db/migrate/20191031090746_add_normalized_domain_to_stories.rb
```diff
@@ -0,0 +1,5 @@
+class AddNormalizedDomainToStories < ActiveRecord::Migration[5.2]
+  def change
+    add_column :stories, :normalized_domain, :string
```
pushcx: Please name this `domain` and have the migration store the domain for all stories and then set the field to be `not null`. That'll let us run queries, remove the complexity of the `domain` method and instance variable, etc.

spec/models/story_spec.rb
```diff
@@ -76,29 +76,31 @@
...
 
-  it "has domain straight out of the db, when Rails doesn't use setters" do
-    s = create(:story, url: 'https://example.com/foo.html')
-    s = Story.find(s.id)
-    expect(s.domain).to eq('example.com')
-    s.url = 'http://example.org'
-    expect(s.domain).to eq('example.org')
-    s.url = 'invalid'
-    expect(s.domain).to be_nil
+    it "has domain straight out of the db, when Rails doesn't use setters" do
```
pushcx: This test can be deleted.

spec/models/story_spec.rb
```diff
@@ -76,29 +76,31 @@
...
-      "http://example.com" => "example.com",
-      "https://example.com" => "example.com",
-      "http://example.com:8000" => "example.com",
-      "http://example.com:8000/" => "example.com",
-      "http://www3.example.com/goose" => "example.com",
-      "http://flub.example.com" => "flub.example.com",
-    }.each_pair do |url, domain|
-      story.url = url
-      expect(story.domain).to eq(domain)
+  describe '#domain' do
```
pushcx: No need to add a `describe` block around this test.

pushcx:
We closed this out in #797; I'll post some issues to build on that soonish.

PR #780, opened 2019-10-31
fix: don't try to match empty title

emrox:
When visiting https://lobste.rs/stories/new there is a JS console error

```
TypeError: e.match(...) is null
```

this happens because it tries to match stories title even when the title is empty. This is a simple fix to not try to match against empty titles


pushcx:
Thanks for noticing and fixing this.

Issue #642, opened 2019-02-17
Upvote buttons on mobile awfully small


faebser:
Sometimes it takes me a three or four tries to successfully upvote a post on mobile. I looked at the html/css and the button is a small 18x18 pixel arrow made with border-radius. 

IMHO it would make sense to have the upvote button to be whole width between the border on the left and the headline on the right. To achieve this I would like to enlarge the existing *a* element and add a child element that is going to hold the arrow. This provides a bigger clickable surface while maintaining the same look.

pushcx:
Sounds good to me. Are the comment arrows similarly hard to click?

faebser:
They are slightly easier to hit, maybe because there is more white space around them. If you want I can also extend them the same way.

Are there any tests for the frontend?

pushcx:
If the style carries over, yeah, let's increase their target size as well.

There are no tests for the frontend.

faebser:
Actually let me backtrack on the following quote

> They are slightly easier to hit

They are really easier to hit. Please let me look into that.

magikid:
@faebser Are you still looking into this?

faebser:
hi @magikid, yes I will fix this in the following days

kconner:
I'm also having this problem. To upvote a story or comment on my phone usually takes me three attempts at hitting the tap target.

pushcx:
#1508 is cleaning up a lot of story + comment layout issues that will make this a lot easier.

Unicode fonts have gotten pervasive, so we could replace the current upvote arrow that's built out of CSS border with a unicode upward-pointing triangle. And then that would make it easy to broaden the <a> tag so the whitespace around the upvote arrow and the score are all part of the clickable target.

That will be a straightforward fix for this bug, but there's also an opportunity for a layout change. It'd be disruptive but probably worth it. We could move the comment folder (the `[-]` and `[+]`) underneath the score to narrow the controls to the left side of comments. This would save horizontal space that's especially valuable on deeply-nested replies on desktop, or all the time on mobile. I call it disruptive because it would be a very visible change to core UI, but I think it's worth it in this instance.

PR #782, opened 2019-11-08
Upvote buttons mobile

faebser:
This is the pull request for #642. I changed both the voting buttons on the link as the ones for the comments. They are both slightly bigger and easier to hit on mobile.

Please have a look and let me know if I any more changes are needed.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -389,11 +389,11 @@ var Lobsters = new _Lobsters();
 
 $(document).ready(function() {
   var $olcomments = $("ol.comments");
-  $olcomments.on("click", ".comment .voters > a + div + a", function() {
+  $olcomments.on("click", ".comment .voters > a.downvoter-tap", function() {
     Lobsters.downvoteComment(this);
     return false;
   });
-  $olcomments.on("click", ".comment .voters > a", function() {
+  $olcomments.on("click", ".comment .voters > a.upvoter-tap", function() {
     Lobsters.upvoteComment(this);
     return false;
   });
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -402,7 +402,7 @@ $(document).ready(function() {
     Lobsters.flagStory(this);
     return false;
   });
-  $("li.story .voters > a").click(function() {
+  $("li.story .voters a.upvote-tap").click(function() {
     Lobsters.upvoteStory(this);
     return false;
   });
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -389,11 +389,13 @@ var Lobsters = new _Lobsters();
 
 $(document).ready(function() {
   var $olcomments = $("ol.comments");
-  $olcomments.on("click", ".comment a.downvoter", function() {
+  $olcomments.on("click", ".comment .voters > a + div + a", function() {
```
arp242: These kind of selectors seem very fragile to me. Actually, looking at the HTML I'm not even sure if I fully understand what this is matching exactly. Is it not possible to keep the simple `a.downvoter` match?

app/assets/javascripts/application.js.erb
```diff
@@ -389,11 +389,13 @@ var Lobsters = new _Lobsters();
...
+    console.log("downvote");
```
arp242: Looks like you left this in by accident, as well as another `console.log("upvote");` a bit further down.

faebser: silly me, will get removed asap

app/assets/javascripts/application.js.erb
faebser: I had to add an additional container to the page to get the desired result of having a bigger tap area and out of a personal preference I did not assign a class to this container. The `.comment .voters > a + div + a` just matches on an `a` that is located after an `a` and a `div` on the first level of the voters `div`. I do not expect that the voters container will see much changes in the future. But you not immediately groking the selector shows that it is too complicated so I'm gonna add a class to the outer container and clean up the selector.

faebser:
Is there something stopping this request from being merged?

pushcx:
This adds a block-level `div` to the inline-level `a`, so it's invalid markup... but then you change `a` to be laid out as a block, so why add the div at all?

faebser:
You are right, that is invalid markup. I'm going to fix that.

The ```div``` contains the graphics (the arrow) while the surrounding ```a``` actually catches the click. I wanted to keep the ```a``` for accessibility reasons.

pushcx:
Ping @faebser - anything I can help with?

faebser:
Sorry, my life is taking up most of my time right now. I will try to do this next week.

pushcx:
Our CSS and JS have been substantially rewritten since this PR was last active, so I'm going to close it out. But anyone is welcome to pick it up and reopen this or a new PR, the inconvenience still exists.

PR #785, opened 2019-11-18
Improve database consistency by setting a length limit

rubynho:
### Enhancement

Set a length limit on the string and text fields of the invitation request column of the database to improve the consistency.

Solves #745


pushcx:
`InvitationRequest` was already updated. I don't see a need for this.

PR #792, opened 2019-12-07
Make title width validation consistent with Model

ilimeza:


pushcx:
Thanks for catching this inconsistency. Could you make it impossible to reintroduce by using a shared constant like `Keystore.MAX_KEY_LENGTH` does?

pushcx:
ping @ilimeza - any interest in finishing this off?

Issue #793, opened 2019-12-08
app/views/_comment.html.erb: Consider adding role='button' for <label ... class='comment_folder' ...


mini-eggs:
While using Qutebrowser or Vimium for FF/Chrome you cannot collapse comments with the keyboard alone. This is due to the comment collapsing label not being treated as a button by Qutebrowser/Vimium. Adding a role of button to the `label.comment_folder` makes this much more enjoyable to use. I'm not sure of the a11y implications. If anyone has knowledge on this portion, if it might hurt accessiblity, I would appreciate to hear your thoughts.

Would a PR implementing this be accepted? 

Here's some screenshots of with and without the role added. Notice the `sd` on the collapsing icon in the second photo. This enables me to `click` the button without a mouse. 

[Without role of button.](https://i.imgur.com/JNkaaBy.png)
[With role of button](https://i.imgur.com/RsP7ny7.png)

pushcx:
Sounds worth a shot.

pushcx:
Sorry for the slow response, my usual dev schedule has been seriously disrupted by coronavirus.

Thanks for contributing this. I suggested a small name improvement and then this is good.

However, there's a limitation here that means the PR won't fix your problem. We only serve javascript to logged-in users. I can send you an invite to solve it for you individually, but otherwise - do you see another way to address this for everyone who's not logged in?

mini-eggs:
> Sorry for the slow response, my usual dev schedule has been seriously disrupted by coronavirus.
> Thanks for contributing this. I suggested a small name improvement and then this is good.
No sweat. 

I'll do the name change some time soon. 

> do you see another way to address this for everyone who's not logged in?
Huh, I'll have to look into this more for a non-js solution. First thing that comes to mind is wrapping each comment (and that comments children) in a [details/summary](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) and having all opened by default. I'll post my findings here after digging into that.

PR #794, opened 2019-12-08
Feat(aria='button' for label.comment_fold): Adding an aria button for

mini-eggs:
Adding an aria button for folder button on comments + aria expanded to play well w/ a11y.

This PR is in reference to https://github.com/lobsters/lobsters/issues/793

> While using Qutebrowser or Vimium for FF/Chrome you cannot collapse comments with the keyboard alone. This is due to the comment collapsing label not being treated as a button by Qutebrowser/Vimium. Adding a role of button to the label.comment_folder makes this much more enjoyable to use. I'm not sure of the a11y implications. If anyone has knowledge on this portion, if it might hurt accessiblity, I would appreciate to hear your thoughts.

Referenced [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) for the accessibility bit. Looks like `aria-exanded` fits this case perfectly.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -67,6 +67,10 @@ var _Lobsters = Class.extend({
     Lobsters._showDownvoteWhyAt("comment", voterEl, function(k) {
       Lobsters.vote("comment", voterEl, -1, k); });
   },
+  foldComment: function(folderEl) {
+    var input = document.getElementById(folderEl.getAttribute("for"));
+    folderEl.setAttribute("aria-expanded", input.checked ? "true" : "false");
+  },
   _showDownvoteWhyAt: function(thingType, voterEl, onChooseWhy) {
     if (!Lobsters.curUser)
       return Lobsters.bounceToLogin();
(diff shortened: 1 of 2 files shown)


mini-eggs:
Hey @pushcx. Pinging you because you replied to the issue I raised for this PR a couple months ago. Hope I got the right person. 

Any chance this PR can be reviewed sooner rather than later? Would love for this aria button to work on Lobsters.

app/assets/javascripts/application.js.erb
```diff
@@ -67,6 +67,10 @@ var _Lobsters = Class.extend({
     Lobsters._showDownvoteWhyAt("comment", voterEl, function(k) {
       Lobsters.vote("comment", voterEl, -1, k); });
   },
+  foldComment: function(folderEl) {
```
pushcx, line 70: Could you rename this to `updateAriaLabel`? It doesn't fold and unfold the comment.

pushcx:
If you'd like to make this rename, I can go ahead and merge this as a partial solution and we can come back to the js issue in #793 later.

pushcx:
Closing inactive PR to keep the list tidy, feel free to reopen if you pick this up again.

PR #795, opened 2019-12-16
Use archive.md instead of archive.is

tie:
The archive.is website is not accessible in Russia. Using archive.md is fine.

See https://blocklist.rkn.gov.ru/ (enter archive.is in URL field).
Have no idea why that happened though.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -278,7 +278,7 @@ class Story < ApplicationRecord
   end
 
   def archive_url
-    "https://archive.is/#{CGI.escape(self.url)}"
+    "https://archive.md/#{CGI.escape(self.url)}"
   end
 
   def as_json(options = {})


NattyNarwhal:
Since there's been some small debate over what archiving service to use, wouldn't offering multiple in a dropdown be an option?

tie:
Personally, I’m fine with `archive.md`/`.today`/`.is`/etc. (I think they use about 5-6 TLDs) as long as it works over HTTPS, and the problem is that `archive.is` domain does not work with Russian ISPs.

Also, `archive.md`/`.is` gives an option to search on other web archives (Google Cache, Wayback Machine, Webcite and Megalodon) so I don’t see any need in the dropdown menu.

pushcx:
Without getting into #762, this seems like a worthwhile minor change.

PR #797, opened 2019-12-28
Search improvements

hmadison:
Addresses the two issues we saw last night by:

- Add a full text index to the `story_cache` column. This should improve the search performance and clean up one of the larger issues the site has been having for a while.
- Migrates the domain method over to a full table. This will let us start to flag heavy promoting stories some point down the road. 
- Add a page for domains that mirrors what we have for tags. This lets us use an index over a regexp search on the search controller when loading a list of all stories for a given domain.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
 ---
 language: ruby
-dist: xenial
 services:
   - mysql
 env:
   global:
-    - DATABASE_URL=mysql2:///lobsters_test
+    - DATABASE_URL=mysql2://root:@localhost/lobsters_dev
     - RAILS_ENV=test
 before_install:
   - gem install bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
   - gem install bundler
   - gem install bundler-leak
 before_script:
-  - sudo mysql -e 'create database lobsters_test;'
+  - ./bin/rails db:create
   - ./bin/rails db:schema:load
 script:
   - bundle exec rspec
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,17 @@
 ---
 language: ruby
+dist: xenial
 services:
   - mysql
 env:
   global:
-    - DATABASE_URL=mysql2://root:@localhost/lobsters_dev
+    - DATABASE_URL=mysql2:///lobsters_test
     - RAILS_ENV=test
 before_install:
   - gem install bundler
   - gem install bundler-leak
 before_script:
-  - ./bin/rails db:create
+  - sudo mysql -e 'create database lobsters_test;'
   - ./bin/rails db:schema:load
 script:
   - bundle exec rspec
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -219,19 +219,19 @@ class HomeController < ApplicationController
   end
 
   def for_domain
-    @domain = Domain.first_by(domain: params[:domain])
+    @domain = Domain.where(domain: params[:domain]).first
 
     @stories, @show_more = get_from_cache(domain: @domain.domain) do
       paginate @domain.stories
     end
 
     @heading = params[:domain]
-    @title = "Stories submitted for #{domain.domain}."
+    @title = "Stories submitted for #{@domain.domain}."
     @cur_url = for_domain_url(params[:domain])
 
     @rss_link = {
-      title: "RSS 2.0 - For Domain #{tags_with_description_for_rss(@tags)}",
-      href: "/t/#{params[:tag]}.rss",
+      title: "RSS 2.0 - For #{@domain}",
+      href: "/domain/#{params[:domain]}.rss",
     }
 
     respond_to do |format|
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -218,6 +218,29 @@ class HomeController < ApplicationController
     end
   end
 
+  def for_domain
+    @domain = Domain.first_by(fdqn: params[:fqdn])
+
+    @stories, @show_more = get_from_cache(tags: tag_params.sort.join(',')) do
+      paginate @domain.stories
+    end
+
+    @heading = params[:fqdn]
+    @title = "Stories submitted for #{domain.fqdn}."
+    @cur_url = for_domain_url(params[:domain])
+
+    @rss_link = {
+      title: "RSS 2.0 - For Domain #{tags_with_description_for_rss(@tags)}",
+      href: "/t/#{params[:tag]}.rss",
+    }
+
+    respond_to do |format|
+      format.html { render :action => "index" }
+      format.rss { render :action => "rss", :layout => false }
+      format.json { render :json => @stories }
+    end
+  end
+
   def top
     @cur_url = "/top"
     length = time_interval(params[:length])
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -221,7 +221,7 @@ class HomeController < ApplicationController
   def for_domain
     @domain = Domain.first_by(fdqn: params[:fqdn])
 
-    @stories, @show_more = get_from_cache(tags: tag_params.sort.join(',')) do
+    @stories, @show_more = get_from_cache(domain: domain.fqdn) do
       paginate @domain.stories
     end
 
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -219,14 +219,14 @@ class HomeController < ApplicationController
   end
 
   def for_domain
-    @domain = Domain.first_by(fdqn: params[:fqdn])
+    @domain = Domain.first_by(domain: params[:domain])
 
-    @stories, @show_more = get_from_cache(domain: domain.fqdn) do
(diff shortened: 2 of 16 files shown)


.travis.yml
```diff
@@ -1,16 +1,17 @@
...
   global:
-    - DATABASE_URL=mysql2://root:@localhost/lobsters_dev
+    - DATABASE_URL=mysql2:///lobsters_test
     - RAILS_ENV=test
 before_install:
   - gem install bundler
   - gem install bundler-leak
 before_script:
-  - ./bin/rails db:create
+  - sudo mysql -e 'create database lobsters_test;'
```
pushcx: What happened here?

app/models/domain.rb
```diff
@@ -0,0 +1,3 @@
+class Domain < ApplicationRecord
+  has_many :stories, dependent: :destroy
```
pushcx: Was this prompted by a rubocop rule or something?

I'm unsettled by having an accessory models like Domain that's created-in-time and not very important with a `dependent: :destroy` pointing at Story. If we deleted one of these records to recreate it, oops, we'd delete a bunch of stories + comments, our core features. It feels like a gun waiting to go off.

Any ideas for making this less of a landmine? Can we remove the `dependent: :destroy`, override the `destroy` method on the class, anything else? I'd just rather `stories` have some broken FKs than have to restore from database backups.

db/migrate/20191227232955_create_domains.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDomains < ActiveRecord::Migration[5.2]
+  def change
+    create_table :domains do |t|
+      t.string :fqdn, unique: true
+      t.boolean :is_tracker
```
pushcx: not null, default false

app/models/story.rb
```diff
@@ -830,13 +824,9 @@ def update_merged_into_story_comments
...
-  def domain
-    return @domain if @domain
-    set_domain self.url.match(URL_RE) if self.url
-  end
-
-  def set_domain match
-    @domain = match ? match[:domain].sub(/^www\d*\./, '') : nil
+  def set_domain(match)
+    name = match ? match[:domain].sub(/^www\d*\./, '') : nil
+    self.domain = name ? Domain.where(fqdn: name).first_or_initialize : nil
```
pushcx: Nicely done.

db/migrate/20191227234358_add_domain_to_stories.rb
```diff
@@ -0,0 +1,5 @@
+class AddDomainToStories < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :stories, :domain, foreign_key: true
```
pushcx: `Story.all.each {|s| s.url = s.url; s.save! }`

And can just use
`%w{ 1url.com 7.ly... }.each { |d| Domain.create! domain: d, tracking: true }`

app/controllers/home_controller.rb
```diff
@@ -218,6 +218,29 @@ def tagged
     end
   end
 
+  def for_domain
```
pushcx: Nice way to present these. Do we want to drop the `domain:` operator or make it use the Domain model as well? I guess the latter? If I pulled stats on usage they'd be totally smashed by the aggressive scraping the other day, so we don't have clear signal that it's useful, but it seems worth supporting.

app/models/domain.rb
hmadison: We can remove the `:destroy` and suppress the cop inline. That feels like the cleanest situation.

.travis.yml
hmadison: Side effect of me testing out switching various Ubuntu versions to get Mariadb to work. I'll revert it.

app/controllers/home_controller.rb
hmadison: I think it may be worth having a meta thread about search at some point to discuss this with the community as a whole. Leaving the current functionality in place doesn't make the situation any worse than it is right now.

cnst:
I believe I was referred to this bug report from https://twitter.com/pushcx/status/1212089737969684485, which kind of doesn't make any sense, because I'm experiencing multi-minute delays with many page loads (I don't have happy-eyeballs, and you must be doing `block drop` instead of the more civilised `block return` that would make my setup automatically go to IPv6).

Why don't you just implement a functionality limiting search from within nginx?  Or at the very least `block return` so that actual folks don't have to wait several minutes for their connections to timeout?

pushcx:
> Why don't you just implement a functionality limiting search from within nginx?

This was also one of our mitigations. We discussed it in chat at the time and it's recorded in the ansible repo. LMK if there's anything else I can catch you up on; Friday's original offer to whitelist false positive IPs is still open. We'll look into updating the block as time permits.

app/controllers/home_controller.rb
pushcx: Yeah. Can you update `Search#with_stories_in_domain` to use the new model?

app/models/domain.rb
```diff
@@ -0,0 +1,3 @@
+class Domain < ApplicationRecord
+  has_many :stories # rubocop:disable Rails/HasManyOrHasOneDependent
+end
```
pushcx: validate presence of fields

config/routes.rb
```diff
@@ -65,6 +65,9 @@
   get "/t/:tag" => "home#tagged", :as => "tag"
   get "/t/:tag/page/:page" => "home#tagged"
 
+  get "/domain/:fqdn" => "home#for_domain", :as => "for_domain"
```
pushcx: Been thinking about it. Because we do normalization like strip off `www` and don't include the trailing `.` this isn't quite an FQDN. Rather than get technically nitpicky with PQDN, let's just call the field `domain`, similar to how `Tag` has `tag`.

db/migrate/20191227234358_add_domain_to_stories.rb
```diff
@@ -0,0 +1,31 @@
...
+        ph.dog po.st prettylinkpro.com q.gs qr.ae qr.net research.eligrey.com scrnch.me s.id sptfy.com
+        t.co tinyarrows.com tiny.cc tinyurl.com tny.im tr.im tweez.md twitthis.com u.bb u.to v.gd
+        vzturl.com wp.me ➡.ws ✩.ws x.co yep.it yourls.org zip.net
+      }.each { |d| Domain.create!(fqdn: d, is_tracker: true) }
+
+    Story.find_in_batches do |s|
+      match = u.match(URL_RE)
+      next unless match
+
+      name = match ? match[:domain].sub(/^www\d*\./, '') : nil
```
pushcx: Why duplicate the method instead of call it? We don't try to keep migrations working to be run from scratch to create a database.

hmadison: We have a linter which prevents rails migrations from autoloading anything.

app/controllers/home_controller.rb
```diff
@@ -218,6 +218,29 @@ def tagged
...
+    @stories, @show_more = get_from_cache(domain: @domain.domain) do
+      paginate @domain.stories
+    end
+
+    @heading = params[:domain]
+    @title = "Stories submitted for #{domain.domain}."
+    @cur_url = for_domain_url(params[:domain])
+
+    @rss_link = {
+      title: "RSS 2.0 - For Domain #{tags_with_description_for_rss(@tags)}",
```
pushcx: Need to be updated to print the domain, also link on next line.

Please add a test for this endpoint - doesn't have to do anything clever, just add one Story with url `example.com` and title "Domain Example" and see that this page includes the title.

app/models/domain.rb
pushcx: also needs a presence validation for the `is_tracker` boolean - nil is not ok

.travis.yml
```diff
@@ -1,10 +1,11 @@
 ---
 language: ruby
+dist: xenial
 services:
   - mysql
 env:
   global:
-    - DATABASE_URL=mysql2://root:@localhost/lobsters_dev
+    - DATABASE_URL=mysql2:///lobsters_test
```
hmadison: @pushcx In response to your comment on IRC, I left this change in the .travis.yml to mirror the authentication mechanism [we use in our ansible playbooks](https://github.com/lobsters/lobsters-ansible/commit/e3c374fd14d9fd0cc8fdf14ecd7dbfa4842b7845)

spec/rails_helper.rb
```diff
@@ -25,6 +25,7 @@
   config.use_transactional_fixtures = true
 
   config.before(:suite) do
+    DatabaseCleaner.allow_remote_database_url = ENV['TRAVIS'].present?
```
hmadison: Same deal as the travis change, this lets DB cleaner run against local socket auth during test.

db/migrate/20191227234358_add_domain_to_stories.rb
hmadison: To clarify: the linter is https://github.com/testdouble/good-migrations

pushcx: Thanks for that link. Forgot about it. I'll have to review it for removal as part of our "no, you don't create your db by running every migration" strategy.

Issue #801, opened 2020-01-04
URL dupe checker needs to ignore anchors


pushcx:
Today someone submitted https://artagnon.com/articles/leancoq and someone else submitted 
https://artagnon.com/articles/leancoq#main a few minutes later. The dupe checker should've caught and prevented this by ignoring any anchors (`#main`) at the end of the URL.

arandomandy:
I had a look at this bug. Where do you consider best implementing a fix? Before or after it hits the back end? A few lines of js may do the job. What do you think?

```
checkStoryDuplicate: function(form) {
  var params = $(form).serializeArray();
  var url = params[2].value.replace(/(#.*)/g, '');
  params[2].value = url;
  $.post("/stories/check_url_dupe", params, function(formErrorsHtml) {
    $(form).find(".form_errors_header").html(formErrorsHtml);
  });
}
```

pushcx:
Backend. JavaScript is optional.

arandomandy:
I came up with [this fix](https://github.com/lobsters/lobsters/pull/802/files). I haven't written any tests yet but I plan on doing so soon.

I'd like to point out that I'm not sure about a) RLIKE since it is MySQL-dependent. Maybe someone knows a better way to build the query? b) is there a way to optimize the query?

@pushcx @hmadison

hmadison:
I feel like that it would make more sense to strip out anchors when we submit stories and have a migration go back through the `stories` table and strip out any anchors in the `url` column.

alanpost:
> I feel like that it would make more sense to strip out anchors when we submit stories and have a migration go back through the stories table and strip out any anchors in the url column.

I have submitted at least one story where I used an anchor to better draw attention to the start of teh content (https://lobste.rs/s/csj1wf).  I think it is all right to have an anchor in the story URL, it should just be ignored for purposes of duplicate testing.

pushcx:
...and doubled slashes, which happened a minute ago. I feel safe assuming this is just about always a typo that the site will ignore. eg `https://example.com//foo/bar` or `https://example.com/foo//bar` are both dupes for `https://example.com/foo/bar`

NattyNarwhal:
Anchors might be significant though. Example cases:

* Some blogs do not have separate permalinks for articles. They rely on anchors to focus on the linked articles.
* Some (probably poorly implemented) SPAs carry navigationally significant data in the anchor.

pushcx:
[...]

PR #802, opened 2020-01-06
Add checks to URL dupe checker

arandomandy:
This PR includes two features:

1. It makes the URL dupe checker ignore URLs with anchors, so that scenario A is covered
2. It makes the URL dupe checker consider scenario B, it does so by appending a # to the URL and searching for a pattern where it matches the URL plus any characters after the #

Scenario A: Alice submits the following URL https://artagnon.com/articles/leancoq, three minutes later Bob tries submitting https://artagnon.com/articles/leancoq#main

Scenario B: Alice submits the following URL https://artagnon.com/articles/leancoq#main, three minutes later Bob tries submitting https://artagnon.com/articles/leancoq


app/models/story.rb
```diff
@@ -229,10 +230,16 @@ def self.find_similar_by_url(url)
     end
     urls = urls2.uniq
 
+    # trailing pound
+    urls.each do |u|
+      urls_with_trailing_pound.push u + "#"
```
pushcx: Is this just for the case where the url already in the database is `https://example.com/foo#`, right? If so, you can remove it - we only have 29 of them. Let's just write a migration to remove them.

app/models/story.rb
```diff
@@ -229,10 +230,16 @@ def self.find_similar_by_url(url)
...
+    end
+
     # if a previous submission was moderated, return it to block it from being
     # submitted again
     Story
       .where(:url => urls)
+      .or(Story.where("url RLIKE ?", urls_with_trailing_pound.join(".|")))
```
pushcx: What is the `.` for in this `join`?

app/models/story.rb
arandomandy: No, this is instead to take care of the scenario described below. It does so by appending a `#` to the URL and searching for a pattern where it matches the URL plus any characters after the `#`, thus the ugly and expensive `RLIKE`.

Alice submits the following URL https://artagnon.com/articles/leancoq#main, three minutes later Bob tries submitting https://artagnon.com/articles/leanco

app/models/story.rb
arandomandy: Short answer: 

Pattern matching: `.` matches any character.

Long answer:

So, let's say Alice submits the following URL https://artagnon.com/articles/leancoq#main, three minutes later Bob tries submitting https://artagnon.com/articles/leancoq

How do I know if https://artagnon.com/articles/leancoq has been submitted with an anchor before, i.e. https://artagnon.com/articles/leancoq#main? I first [added a `#` to all URLs](https://github.com/lobsters/lobsters/pull/802/files#diff-8a4ad5774f65a8a217b216fa2038a5bcR235), then I try to find any URL that matches the following pattern; https://artagnon.com/articles/leancoq#anycharacter

I'm able to do so by appending a `.` at the end of each URL.

As I pointed out [here](https://github.com/lobsters/lobsters/issues/801#issuecomment-571098708) I have my doubts about this approach.

pushcx:
Thanks for doing this, I'm glad to start catching these at submission.

PR #805, opened 2020-01-10
Feat: add RSS feed for /top submissions

mbologna:
Adds a new route for returning the RSS feed for /top submissions

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -234,7 +234,15 @@ class HomeController < ApplicationController
       @heading = @title = "Top Stories of the Past " << length[:intv]
     end
 
-    render :action => "index"
+    @rss_link ||= {
+      :title => "RSS 2.0 - " + @heading,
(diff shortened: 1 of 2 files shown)


pushcx:
We [chatted about this](https://lobste.rs/chat) yesterday. It's still a little weird to me to have an RSS feed for data that's not ordered by time and how there's potential to miss items as new ones arrive, but a user wanting to see just the best stuff is expressly OK with missing some things.

NightMachinary:
@mbologna @pushcx Can this be further refined to accept a `min-score` parameter? I want to get the “must-read” links, and no others.

mbologna:
> @mbologna @pushcx Can this be further refined to accept a `min-score` parameter? I want to get the “must-read” links, and no others.

That would be possible, yes, but you need to either fiddle with `stories` in https://github.com/lobsters/lobsters/blob/master/app/controllers/home_controller.rb#L69-L71 or even better filter at the `get_from_cache` level.

PR #811, opened 2020-01-22
Change domain constraint to allow for dots

hmadison:
Resolves the two issues noticed during the deploy this morning:

* Adds a second test which runs through Journey 
* Updates the constraint on the route to allow for `.`s in domains to propagate to the `domain` parameter.


pushcx:
Thanks for patching this up. Deploying now...

PR #814, opened 2020-01-25
Fix grammar in about page

WesleyAC:


pushcx:
Thanks for catching this inconsistency, and for sending in a patch.

Issue #637, opened 2019-02-01
Dark theme / night mod for PC


vladimircicovic:
Could you please add night mode / dark theme for lobster site so ppl who reading could use it. 
Thank you.

pushcx:
You can accomplish this with a user style.

This is also a dupe of #404.

pushcx:
I saw a [dark mode screenshot](https://twitter.com/lattera/status/1100837450400829440) by @lattera go by on Twitter, he suggested the [Dark Reader](https://darkreader.org/) browser extension. Hope this works out for you.

adriangoransson:
Would it be of interest to add a dark mode using the CSS [`prefers-color-scheme` feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)?

Pros:
* No plugin necessary.
* [Pretty good support](https://caniuse.com/#feat=prefers-color-scheme) and **works on mobile**.
* Toggle is on system level and needs no custom UI or per user storage.

I see no explicit downsides except that it requires some more CSS. And the obligatory disagreements on how a dark theme should look.

magikid:
I like the `prefers-color-scheme` option but I agree with @pushcx in #404 that this should probably be in a meta thread on lobsters.

adriangoransson:
@magikid I agree, but I haven't registered yet. 😀 Would you mind sending me an invite?

soc:
I apologize for resurrecting this thread.

I'd be happy to provide a PR implementing a dark mode based on `prefers-color-scheme`, if this doesn't go against https://github.com/lobsters/lobsters/issues/404#issuecomment-333977723.¹

The good thing about `prefers-color-scheme` is that the concern raised in the comment regarding

> A single toggle to a dark theme seems OKish if someone can suggest a good place in the UI for it

is not an issue, as `prefers-color-scheme` does not require any toggle or UI change to work.

---

¹ There are two ways I'm able to read @pushcx comment, either as "I'll consider implementing this if there is overwhelming demand" or as "I'll accept an implementation if there is overwhelming demand". If it's the former, I'm happy to provide a PR.

pushcx:
Sounds like a great approach that doesn't make the site any more complicated. I'd merge a PR for it.

soc:
@pushcx Cool, thanks!

soc:
Is there a way to get a Lobste.rs instance running without the database setup mentioned in the README?

pushcx:
[...]

PR #823, opened 2020-02-13
Add Lobsters dark mode

soc:
dark-mode

See #637.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -8,6 +8,98 @@
 
 /* generics */
 
+:root {
+    --color: #333333;
+    --color-background: #fefefe;
+    --color-selection: HighlightText;
+    --color-selection-background: Highlight;
+
+    --color-link: #2562dc;
+    --color-link-visited: #7395d9;
+
+    --color-form: #555555;
+    --color-form-focus: #303030;
+    --color-form-border: #cccccc;
+    --color-form-border-focus: #888888;
+    --color-form-background: #ffffff;
+
+    --color-button: #333333;
+    --color-button-border: #cccccc;
+    --color-button-hover: #333333;
+    --color-button-background: #fafafa;
+    --color-button-background-hover: #e6e6e6;
+
+    --color-table-background-header: #eaeaea;
+    --color-table-background-odd: #f8f8f8;
+    --color-table-background-even: #f5f5f5;
+
+    --color-tag: #555555;
+    --color-tag-background: #fffcd7;
+    --color-tag-border: #d5d458;
+    --color-tag-media: #555555;
+    --color-tag-media-background: #ddebf9;
+    --color-tag-media-border: #b2ccf0;
+
+    --color-header: #777777;
+    --color-header-lobsters: #333333;
+    --color-header-title: #333333;
+
+    --color-markdown_help-background: #f4f4f4;
+    --color-markdown_help-border: #e0e0e0;
+
+    --color-downvote_why: #555555;
+    --color-downvote_why-background: #ffffff;
+    --color-downvote_why-background-hover: #eeeeee;
+    --color-downvote_why-background-cancel: #eeeeee;
+}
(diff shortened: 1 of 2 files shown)


soc:
I couldn't verify whether the missing L in the red square was caused by these changes or is just a quirk in Firefox caused by using the Style Editor in combination with the Dark Website Forcer extension to develop this.

The color scheme are intentionally chosen very conservatively; the colors of tags and hats have been preserved.

danielcompton:
I’m on board with this approach, but the contrast seems very low with these colours. I had previously started this, I’ll try and pull out my branch and the colours I used.

soc:
> I’m on board with this approach, but the contrast seems very low with these colours.

The reduction of the contrast in the dark scheme is intentional, I believe it's part of the reason of having one. I'd propose negotiating the colors after we have dark mode and received some feedback.

> Another option is to add a CSS Postprocessor like PostCSS to add the non-var version of the light mode colours before the var version, e.g. https://github.com/MadLittleMods/postcss-css-variables. I’m not sure if the added complexity is worth it though, it would significantly complicate the build.

@pushcx Mentioned in the issue that it's not required.

soc:
I pushed some refinements and better variable names.

soc:
@pushcx Is there anything you'd like me to do, or is this good to go?

pushcx:
Could you catch me up on what's happening between this PR and #847? I see your name on both, and if you and @soc have collaborated, can we combine these into one PR?

soc:
From my point of view (which is just my personal opinion) I believe that dark modes are meant to be less eye-straining than the "light" mode – which is why the color theme I picked has reduced contrast.

I have been using roughly the same color scheme on my personal website for years and have received zero complaints about it being _too_ low contrast.

This is not to say that there won't be any complaints – but I think it would make sense to either pick @danielcompton's or my scheme and adjust it based on the specific criticism received after it has been introduced – instead of trying to find a compromise that makes no one happy in the end.

pushcx:
I'm more concerned with making no one happy at the beginning - er, well, cutting down the chaos, really. Generally we avoid sinking into endless bikeshedding by not doing a lot of cosmetic changes, so a big visible change is going to prompt a round of distress at the change and likely heated arguments on tiny tweaks, or impractical schemes to make everything configurable that turn out to make everything expensive to maintain. If it's possible to do one round of that instead of two, that'd be a lot less stressful.

So I wondered if there's a fundamental difference between the two approaches worth talking about once up-front. Let's see what Daniel has to say about how he picked that set of colors. If these come down to personal preference, well, that's sometimes how it goes and I'll figure out how to cut the baby in half.

soc:
I defined over 30 CSS variables that can be used to adjust both the light and the dark theme, so I think we are pretty close to making everything configurable already. (In the sense of people being able to drop their preferred CSS into their Stylus extension.) :-)

danielcompton:
I approached my design with these goals:

* make the dark mode quite dark so the screen isn't too bright at night
* have a high contrast ratio for accessibility
* not change the feel too much from the light mode Lobsters theme

I used the light mode theme colours as the base and then inverted them or lightened them to fit the dark mode context.

GarbageHamburger:
As mentioned on the IRC, this could also be implemented as an alternative stylesheet: https://developer.mozilla.org/en-US/docs/Web/CSS/Alternative_style_sheets

Note that as far as I know this only works for Firefox. However it's a nice alternative that does not require any use of Javascript to achieve.

pushcx:
@GarbageHamburger This PR also doesn't require JS.

@soc I hope you are prepared for folks to show up to help paint this bikeshed, because I'm going to merge this and push it live. Thanks for contributing it. :D

pushcx:
Firefox has announced UI for dev support: https://twitter.com/FirefoxDevTools/status/1359896019757113344

I'm unaware of the state of tooling around css variables, but this may be worth picking back up after Firefox 87.

PR #825, opened 2020-02-17
update ruby version to 2.6.5

favadi:
Rationale: Ruby 2.3.x is quite old and couldn't compile cleanly on latest Ubuntu LTS version (18.04) due to incompatible with libssl-dev1.1 (https://github.com/rvm/rvm/issues/4357).
Proposal: update ruby version to 2.6.5.
Test suite passed and manually testing shows nothing broken.


thomasdziedzic:
Note that this isn't as simple as just changing the .ruby-version. You need to upgrade ubuntu to 18.04 for this to be available. https://github.com/lobsters/lobsters-ansible/issues/22

favadi:
> we will only get ruby 2.5.

Ruby 2.5 should be fine IMO. So the reason we haven't update ruby  version is the production server is still running Ubuntu 16.04 and we haven't got around to update to Ubuntu 18.04 yet?

thomasdziedzic:
> Ruby 2.5 should be fine IMO. So the reason we haven't update ruby version is the production server is still running Ubuntu 16.04 and we haven't got around to update to Ubuntu 18.04 yet?

Ya

roryokane:
After the recent downtime to upgrade Lobsters, I noticed there’s a new [`ruby27` branch](https://github.com/lobsters/lobsters/tree/ruby27) in this repo that includes commit 87f2404c08b49bd3746c7cf1c96ccdce5d5c8436, which upgrades to Ruby 2.7.0. There’s no PR for that branch yet though.

favadi:
I'm going to close this as it seems that we are targeting ruby2.7.

PR #826, opened 2020-02-17
passing Github access token in header instead of query parameter

### LINKED ISSUE #819: Update to new GitHub API
Got a note from GitHub that we need to update how we authenticate to match user accounts.

```
Hello there!

On February 8th, 2020 at 06:38 (UTC) your application (Lobsters) used an access token (with the User-Agent
Mozilla/5.0 (compatible)) as part of a query parameter to access an endpoint through the GitHub API.

https://api.github.com/user

Please use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated
and will be removed July 1st, 2020.

Depending on your API usage, we'll be sending you this email reminder once every 3 days.

Visit
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-
using-query-parameters for more information.

Thanks,
The GitHub Team
```

favadi:
https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

Fixes #819

### DIFF
--- a/extras/github.rb
+++ b/extras/github.rb
@@ -26,7 +26,8 @@ class Github
     tok = ps["access_token"].first
 
     if tok.present?
-      res = s.fetch("https://api.github.com/user?access_token=#{tok}").body
+      headers = { "Authorization" => "token #{tok}" }
+      res = s.fetch("https://api.github.com/user", :get, nil, nil, headers).body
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for figuring this out!

PR #827, opened 2020-02-18
add one_foreign_key_present validator to moderation model

### LINKED ISSUE #820: Moderations: validate exactly one fk is present
The `Moderation` model needs a validation that it's only pointing at a single object. Right now its validations don't express that it's for a story, a comment, a user, *or* a domain, so it would be possible to have both a `story_id` and a `user_id`, which is not coherent.

favadi:
Fixes #820

### DIFF
--- a/app/models/moderation.rb
+++ b/app/models/moderation.rb
@@ -23,6 +23,7 @@ class Moderation < ApplicationRecord
   }
 
   validates :action, :reason, length: { maximum: 16_777_215 }
+  validate :one_foreign_key_present
 
   after_create :send_message_to_moderated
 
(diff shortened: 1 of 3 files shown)


app/models/moderation.rb
```diff
@@ -77,4 +78,12 @@ def send_message_to_moderated
 
     m.save
   end
+
+protected
+
+  def one_foreign_key_present
+    fields = [comment, domain, story, tag, user]
+    fk_count = fields.inject(0) {|result, fk| fk.nil? ? result : result + 1 }
```
pushcx: Using the model attribute like this instead of the `_id` column will cause Rails to query for and instantiate the related object(s) unnecessarily. 

You can also express this more directly with `.compact.length`.

favadi: > Using the model attribute like this instead of the _id column will cause Rails to query for and instantiate the related object(s) unnecessarily.

I changed everything to _id.

> You can also express this more directly with .compact.length.

I didn't know of this, thanks.

pushcx:
Nicely done, favadi!

PR #831, opened 2020-03-02
Mark story links as user generated content

hmadison:
We aleady add rel=nofollow to links in our comments. This
does the same thing for stories following the newest Google recomendations.

https://webmasters.googleblog.com/2019/09/evolving-nofollow-new-ways-to-identify.html


pushcx:
Seems like a decent signal to send. I'll update the rest.

PR #833, opened 2020-03-03
Fix PDF title PR and add test case for PDF titles

### LINKED ISSUE #832: pdf title PR broke on merge
Can reproduce in localhost. Need to determine cause.

NattyNarwhal:
When it was introduced to the master branch, a sanity check on
response size was added. However, it did a non-sensical comparison
on string vs. number; this fixes it so it compares the string
length now.

Also add a test to ensure that this continues to work... hopefully.

Fixes #832


pushcx:
Thanks for figuring out my dumb bug.

PR #837, opened 2020-03-08
fix fake_data failed to run due to DST adjustment

favadi:
The default application time zone, Central Time (US & Canada), got an DST adjustment today (March 08 2020) and causes following problems:

    irb(main):003:0> User::NEW_USER_DAYS.days.ago.utc
    => 2019-12-29 11:51:16 UTC
    irb(main):004:0> User::NEW_USER_DAYS.days.ago(Time.current.utc)
    => 2019-12-29 10:51:24 UTC

The commit changes timezone on development, test environment to UTC to avoid DST.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -59,6 +59,8 @@ Rails.application.configure do
     Bullet.console = true
     Bullet.rails_logger = true
   end
+
+  config.time_zone = 'UTC'
 end
 
 Keybase.DOMAIN = Rails.application.domain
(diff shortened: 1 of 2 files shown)


pushcx:
I think this was fixed in #842 - can you confirm? If so, great to have the diagnosis of what went wrong with this test.

I've always thought of the central time zone as a bit of charming character that I'd prefer not to revert, so if it's fixed mayyyybe we can fix this one the wrong way and move on.

favadi:
> I think this was fixed in #842 - can you confirm?

My guess is yes, closing.

PR #840, opened 2020-03-12
Log error when creating/sending an invitation fails

### LINKED ISSUE #839: Log rescued errors
Today in the chat room someone needed help with being unable to send invitation emails from their instance of the codebase. We had to add a `raise` to the `rescue` block on line 54 of `app/controllers/invitations_controller.rb` to get the real error message from their email server.

This site, and probably many places where we `rescue`, should call `logger.error` to provide more info to site owners in the site logs.

arandomandy:
Resolves #839


pushcx:
Cool, thanks for adding this, it'll help future debuggers.

PR #844, opened 2020-03-25
Highlight comments for showing user

danielcompton:
When reading comments by a particular user, it can be difficult to find where their comments start, as comment threads are shown from the root, not from where the user commented.

This commit adds a highlight on comments by the @showing_user's comments to make them easier to distinguish.

Adding the `showing-user-comment` class to the `details` comment div looks like this:

Screenshot of Safari (25-03-20, 4-14-47 PM)

Adding the `showing-user-comment` class to the `comment` class (which is where the existing `:target` based selector applies) looks like this:

Screenshot of Safari (25-03-20, 4-14-59 PM)

I don't really mind which, I thought the former looked better, but either is fine.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -481,6 +481,12 @@ li div.details {
 	border-radius: 20px;
 }
 
+.showing-user-comment .time_ago {
+	background-color: #fffcd7;
+	border-radius: 5px;
+	padding: 3px;
+}
+
 li .link {
 	font-weight: bold;
 	vertical-align: middle;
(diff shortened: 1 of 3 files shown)


pushcx:
I can see how this addresses an inconvenience, but it seems like an overwhelming amount of color for our really muted site. Do you see any other approaches?

danielcompton:
How about this?

danielcompton:
Actually, this should be on their name, not on the timestamp.

pushcx:
After seeing how the highlight looks on the permalink, I agree that it'd be perfect on the username. It's clear and easy to pick out without being a distraction.

pushcx:
Hi @danielcompton - looking forward to the username highlight. You can ignore the Travis build failure here, it was a config issue on their end; I expect the build to go green on your next commit.

danielcompton:
Will try to get back to this soon, on a new laptop so need to get setup, etc.

pushcx:
I tweaked and landed this PR in 08765a2. Thanks for implementing this!

PR #845, opened 2020-03-29
Replace link to IRC channel in 500 error message

amir:
In case of 500 errors it's very likely that the link to `/chat` would also lead to a 500 error. This PR changes the link directly to the IRC channel.

### DIFF
--- a/public/500.html
+++ b/public/500.html
@@ -6,6 +6,6 @@
 <body>
   <h1>Lobsters HTTP 500 status: server error</h1>
   <p>The page you requested cannot be displayed.</p>
-  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in <a href="/chat">chat</a>.</p>
+  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in the <a href="irc://irc.freenode.net/lobsters">IRC channel</a> (also accessible via <a href="https://webchat.freenode.net/#lobsters">webchat</a>).</p>
 </body>
 </html>


alanpost:
Fantastic catch, @amir.  Thank you.

Along with a link to the channel using the IRC protocol, would you also add an https link?:

https://webchat.freenode.net/#lobsters

amir:
@alanpost thanks for the review. I just added the webchat link.

public/500.html
```diff
@@ -6,6 +6,6 @@
 <body>
   <h1>Lobsters HTTP 500 status: server error</h1>
   <p>The page you requested cannot be displayed.</p>
-  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in <a href="irc://irc.freenode.net/lobsters">IRC channel</a>.</p>
+  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in the <a href="irc://irc.freenode.net/lobsters">IRC channel</a> (also accessible via <a href="https://webchat.freenode.net/#lobsters">web<a>).</p>
```
alanpost: Will you say "webchat" rather than "web" in the anchor text?  I like being a titch more precise here given the circumstances (site failure) and attendant stress that might cause a reader.

alanpost:
@amir, I was unclear about one thing: would you squash these two commits together as one?

amir:
@alanpost squashed and updated the anchor text.

public/500.html
```diff
@@ -6,6 +6,6 @@
 <body>
   <h1>Lobsters HTTP 500 status: server error</h1>
   <p>The page you requested cannot be displayed.</p>
-  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in <a href="/chat">chat</a>.</p>
+  <p>The site operators have been notified of this error.  You're welcome to and encouraged to also report it in the <a href="irc://irc.freenode.net/lobsters">IRC channel</a> (also accessible via <a href="https://webchat.freenode.net/#lobsters">webchat<a>).</p>
```
alanpost: I just noticed the closing anchor on the webchat a tag is not a closing tag, as it is missing a forward-slash (/):

<a href="">anchor text</a>

amir:
fixed, sorry! never though it'll take me four commits to update a link :)

alanpost:
Thank you @amir.  I'm really glad you spotted this one today during our outage.

PR #847, opened 2020-03-30
Alternate dark mode based on #823

danielcompton:
This is an alternate set of colours for dark mode, based on @soc's PR in
#823. Some of these dark styles were derived from the webkit.org blog,
which has a nice set of dark mode colours.

This isn't quite finished, and still needs a mobile styled theme.

Screen Shot 2020-03-30 at 2 06 32 PM

Screenshot of Safari (30-03-20, 2-07-30 PM)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -8,6 +8,98 @@
 
 /* generics */
 
+:root {
+    --color: #333333;
+    --color-background: #fefefe;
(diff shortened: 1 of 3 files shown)


pushcx:
There wasn't any discussion after the call on #823 so I went ahead and merged that. We can pick this or other colorschemes up in a `meta`-tagged ~flamewar~ enlightened discussion of differing opinions based on available evidence.

PR #848, opened 2020-04-06
Allow users to link Freenode accounts to Lobste.rs accounts

hmadison:
Mirroring our integrations for Twitter, Github and Keybase, I've added one that allows a user to claim a Freenode account by using the taxonomy features of NickServ. 

The scope of the changes is:
- Added a class to connect to Freenode and read the taxonomy information directly. The alternative to this would be to add a HTTP endpoint to Mockturtle and use that to query the information we need. I'm hesitant to do that because it couples us to the bot which I don't want to do without first having the appropriate discussion.
- Moved the `External Accounts` links out of the general user settings `form_with` tag and into its own section on the page. You can't nest two `form_with` tags and we can keep the same visual styling using `label_tag` to replace our usage of `f.label ...`.

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -256,6 +256,30 @@ class SettingsController < ApplicationController
     return redirect_to "/settings"
   end
 
+  def freenode_callback
+    reader = FreenodeTaxonomyReader.new
+    taxonomy = reader.for_user(params[:user][:freenode_username])
+
+    if taxonomy['LOBSTERS'] == @user.username
+      @user.freenode_username = params[:user][:freenode_username]
+      @user.save!
+      flash[:success] = "Your account has been linked to Freenoe user #{@user.freenode_username}."
+    elsif taxonomy['LOBSTERS']
+      flash[:error] = "Expected #{@user.username} in the taxonomy, not #{taxonomy['LOBSTERS']}."
+    else
+      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy."
+    end
+
+    return redirect_to "/settings"
+  end
+
+  def freenode_disconnect
+    @user.freenode_username = nil
+    @user.save!
+    flash[:success] = "Your Freenode association has been removed."
+    return redirect_to "/settings"
+  end
+
 private
 
   def user_params
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -263,7 +263,7 @@ class SettingsController < ApplicationController
     if taxonomy['LOBSTERS'] == @user.username
       @user.freenode_username = params[:user][:freenode_username]
       @user.save!
-      flash[:success] = "Your account has been linked to Freenoe user #{@user.freenode_username}."
+      flash[:success] = "Your account has been linked to freenode user #{@user.freenode_username}."
     elsif taxonomy['LOBSTERS']
       flash[:error] = "Expected #{@user.username} in the taxonomy, not #{taxonomy['LOBSTERS']}."
     else
@@ -276,7 +276,7 @@ class SettingsController < ApplicationController
   def freenode_disconnect
     @user.freenode_username = nil
     @user.save!
-    flash[:success] = "Your Freenode association has been removed."
+    flash[:success] = "Your freenode association has been removed."
     return redirect_to "/settings"
   end
 
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -265,9 +265,9 @@ class SettingsController < ApplicationController
       @user.save!
       flash[:success] = "Your account has been linked to freenode user #{@user.freenode_username}."
     elsif taxonomy['LOBSTERS']
-      flash[:error] = "Checked on your username, but Freenode told us it's something other than #{@user.username}."
+      flash[:error] = "Checked on your username, but Freenode told us it's something other than #{@user.username}." # rubocop:disable Layout/LineLength
     else
-      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy. Message NickServ `SET PROPERTY LOBSTERS #{@user.username}` to set it."
+      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy. Message NickServ `SET PROPERTY LOBSTERS #{@user.username}` to set it." # rubocop:disable Layout/LineLength
     end
 
     return redirect_to "/settings"
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -265,9 +265,9 @@ class SettingsController < ApplicationController
       @user.save!
       flash[:success] = "Your account has been linked to freenode user #{@user.freenode_username}."
     elsif taxonomy['LOBSTERS']
-      flash[:error] = "Expected #{@user.username} in the taxonomy, not #{taxonomy['LOBSTERS']}."
+      flash[:error] = "Checked on your username, but Freenode told us it's something other than #{@user.username}."
     else
-      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy."
+      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy. Message NickServ `SET PROPERTY LOBSTERS #{@user.username}` to set it."
     end
 
     return redirect_to "/settings"
@@ -277,7 +277,7 @@ class SettingsController < ApplicationController
     @user.freenode_username = nil
     @user.save!
     flash[:success] = "Your freenode association has been removed."
-    return redirect_to "/settings"
+    return redirect_to user_path(@user)
   end
 
 private
--- a/app/views/settings/index.html.erb
+++ b/app/views/settings/index.html.erb
@@ -72,86 +72,6 @@
     </p>
 
     <br>
-
-    <div class="legend">
-      External Accounts
-    </div>
-
-    <div class="boxline">
-      <%= f.label :gravatar,
-        raw("<a href=\"https://gravatar.com/\">Gravatar</a>:"),
-        :class => "required" %>
-      <span>
-        Your avatar will be cached from the Gravatar icon for your e-mail
-        address above.
-        (<%= link_to "Expire cache", "/avatars/expire", :method => :post %>)
-      </span>
-    </div>
-
-    <% if Pushover.enabled? %>
-      <div class="boxline">
-        <%= f.label :pushover_user_key,
-          raw("<a href=\"https://pushover.net/\">Pushover</a>:"),
-          :class => "required" %>
-        <span>
-        <%= link_to((@edit_user.pushover_user_key.present??
-          "Manage Pushover Subscription" : "Subscribe With Pushover"),
-          "/settings/pushover_auth", :class => "pushover_button",
-          :method => :post) %>
-        <span class="hint">
-          For optional comment and message notifications below
-        </span>
-        </span>
-      </div>
-    <% end %>
-
-    <% if Github.enabled? %>
-      <div class="boxline">
-        <span>
-        <%= f.label :github_username,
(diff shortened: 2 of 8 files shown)


app/controllers/settings_controller.rb
```diff
@@ -256,6 +256,30 @@ def twitter_disconnect
...
   end
 
+  def freenode_callback
+    reader = FreenodeTaxonomyReader.new
+    taxonomy = reader.for_user(params[:user][:freenode_username])
+
+    if taxonomy['LOBSTERS'] == @user.username
+      @user.freenode_username = params[:user][:freenode_username]
+      @user.save!
+      flash[:success] = "Your account has been linked to Freenoe user #{@user.freenode_username}."
```
dlowe-net, line 266: *Freenode

app/views/users/show.html.erb
```diff
@@ -150,6 +150,16 @@
     <br>
   <% end %>
 
+  <% if @showing_user.freenode_username.present? %>
+    <label class="required">Freenode:</label>
+
+    <span class="d">
+      <a href="irc://chat.freenode.net:667/<%= h(@showing_user.freenode_username) %>"
```
dlowe-net, line 157: 6667

app/controllers/settings_controller.rb
AbstractBeliefs: (but with a lowercase f)

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,49 @@
...
+    taxonomy_lines.map do |line|
+      line
+        .split(':')[2,3]
+        .map { |t| t.strip }
+    end.to_h
+  end
+
+  private
+
+  def taxonomy_for(username)
```
ilbelkyr, line 29: For the record, I dislike how this function fundamentally works; it doesn't understand the protocol, instead basically scraping the lines it receives for the data it wants. This makes it somewhat fragile.

Additionally, there is no error handling; if things break (e.g. it's hitting some connection rate limit, services are temporarily down for maintenance, etc), it doesn't seem to distinguish that from simply getting an empty taxonomy output.

app/views/users/show.html.erb
```diff
@@ -150,6 +150,15 @@
     <br>
   <% end %>
 
+  <% if @showing_user.freenode_username.present? %>
+    <label class="required">freenode:</label>
+
+    <span class="d">
+      <a>@<%= h(@showing_user.freenode_username) %></a>
```
pushcx, line 157: seems weird to have an `@` here, they're not an irc op

db/schema.rb
```diff
@@ -320,7 +319,6 @@
   add_foreign_key "invitations", "users", column: "new_user_id", name: "invitations_new_user_id_fk"
   add_foreign_key "invitations", "users", name: "invitations_user_id_fk"
   add_foreign_key "messages", "hats", name: "messages_hat_id_fk"
-  add_foreign_key "messages", "users", column: "author_user_id", name: "messages_ibfk_1"
```
pushcx, line 323: weird that these changes snuck into the pr, please rm

extras/freenode_taxonomy_reader.rb
pushcx: Howdy, can you point to a doc about the protocol and format for these messages? I don't know if we want to actually break out a proper parser but I'd like to catch error messages and such if possible.

app/controllers/settings_controller.rb
```diff
@@ -256,6 +256,30 @@ def twitter_disconnect
...
+  def freenode_callback
+    reader = FreenodeTaxonomyReader.new
+    taxonomy = reader.for_user(params[:user][:freenode_username])
+
+    if taxonomy['LOBSTERS'] == @user.username
+      @user.freenode_username = params[:user][:freenode_username]
+      @user.save!
+      flash[:success] = "Your account has been linked to freenode user #{@user.freenode_username}."
+    elsif taxonomy['LOBSTERS']
+      flash[:error] = "Expected #{@user.username} in the taxonomy, not #{taxonomy['LOBSTERS']}."
```
pushcx, line 268: This is not an especially useful error mesage for a user to read. And until I know more about where these taxonomies get filled in from, I'm a little hesitant to drop potential user input into it.

Maybe a more useful error would be something like "Checked on your username, but Freenode told us it's something other than #{@user.username}" because that also handles the case of someone deliberately inputting a different person's username.

app/controllers/settings_controller.rb
```diff
@@ -256,6 +256,30 @@ def twitter_disconnect
...
+    else
+      flash[:error] = "The LOBSTERS key was not in #{params[:user][:freenode_username]}'s taxonomy."
```
pushcx, line 270: This error should tell/link them to how to put the freenode username in their taxonomy.

app/controllers/settings_controller.rb
```diff
@@ -256,6 +256,30 @@ def twitter_disconnect
...
+    end
+
+    return redirect_to "/settings"
+  end
+
+  def freenode_disconnect
+    @user.freenode_username = nil
+    @user.save!
+    flash[:success] = "Your freenode association has been removed."
+    return redirect_to "/settings"
```
pushcx, line 280: Let's save them a click and link to their profile, because they're going to immediately go click on their profile to see it gone. :D

extras/freenode_taxonomy_reader.rb
AbstractBeliefs: You get a 32-wide, space padded key field, followed by a `: ` and the value field, with a header and footer NOTICE that can be used to track the state of your query.

`:` is illegal in key names, but can be included in values.

```
<< PRIVMSG nickserv :taxonomy kline
>> :NickServ!NickServ@services. NOTICE kline :Taxonomy for kline:
>> :NickServ!NickServ@services. NOTICE kline :abc                             : def
>> :NickServ!NickServ@services. NOTICE kline :foo                             : bar:baz
>> :NickServ!NickServ@services. NOTICE kline :End of kline taxonomy.
```

app/views/settings/index.html.erb
```diff
@@ -299,6 +219,104 @@
...
+  </div>
+  <% end %>
+
+  <div class="boxline">
+    <%= label_tag(:freenode_username, raw('<a href="https://freenode.net">Freenode</a>:'), class: "required") %>
+    <span>
+      <% if @edit_user.freenode_username.present? %>
+        <strong><%= @edit_user.freenode_username %></strong>
+        (<%= link_to "Disconnect", "/settings/freenode_disconnect",
+             :method => :post %>)
```
pushcx, line 307: I think this trigger's Rails's overly-clever UJS stuff to decorate the link. Can we avoid that? I'm trying to avoid new hard dependencies on js (with the eventual goal of making js optional even for logged-in users). OK if this is a button.

pushcx: OK, discussed in stream, will leave for now. We can make a later PR use buttons and css abuse them into prettiness somehow later.

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,59 @@
...
+require 'securerandom'
+require 'timeout'
+
+class FreenodeTaxonomyReader
+  NICKSERV = 'NickServ!NickServ@services.'
+  END_MESSGE_REGEX = /End of (.+) taxonomy|(.+) is not registered/
+
+  def initialize(
+        socket_provider: ->() { ssl_socket },
+        username_provider: ->() { "lobsters-#{SecureRandom.alphanumeric(8)}" }
```
pushcx, line 11: Grab sitename from `Rails.application.name` (`config/application.rb`).

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  private
+
+  def taxonomy_for(username)
+    taxonomy_lines = []
+    s = @socket_provider.call
+
+    Timeout.timeout(30) do
+      s.write("NICK #{@nick}\r\n")
+      s.write("USER #{@nick} * * :Lobsters\r\n")
```
pushcx, line 35: `Rails.application.name` again

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,59 @@
...
+      s.write("PRIVMSG NickServ :TAXONOMY #{username}\r\n")
+
+      while line = s.gets
```
pushcx, line 38: Let's see what comes back about format, but if we don't decide to write a bigger parser let's have a max line count of 100 to avoid people stuffing war and peace into their taxonomy and hanging an app worker.

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,59 @@
...
+        username_provider: ->() { "lobsters-#{SecureRandom.alphanumeric(8)}" }
+      )
+    @socket_provider = socket_provider
+    @nick = username_provider.call
+  end
+
+  def for_user(username)
+    _head, *taxonomy_lines = taxonomy_for(username)
+
+    taxonomy_lines.map do |line|
```
pushcx, line 20: I see how this splits apart the levels of abstraction - raw irc data from a proper hash. But given that we care about exactly one line, let's skip it and return/yield that one line if present and ignore everything else. Like the issue with someone throwing thousands of lines in their taxonomy, this also helps us if someone has a single really long line, we don't need to allocate state for it.

extras/freenode_taxonomy_reader.rb
```diff
@@ -0,0 +1,59 @@
...
+    s = @socket_provider.call
+
+    Timeout.timeout(30) do
+      s.write("NICK #{@nick}\r\n")
+      s.write("USER #{@nick} * * :Lobsters\r\n")
+      s.write("PRIVMSG NickServ :TAXONOMY #{username}\r\n")
+
+      while line = s.gets
+        next unless line.include?(":#{NICKSERV}")
+        break if line.match(END_MESSGE_REGEX)
```
pushcx, line 40: typo, `MESSGE`

extras/freenode_taxonomy_reader.rb
ilbelkyr: For the protocol in general, RFC1459 has [the detailed syntax](https://tools.ietf.org/html/rfc1459#section-2.3.1). You might get away with some clever use of a string split function for this; it's just the last parameter which makes it a bit more annoying (since it has to allow spaces).

If there's an issue that gets your entire connection rejected, you should receive an `ERROR` message with a parameter that tells you the issue (which might not be useful to display to users, but might be worth logging or something to be able to figure things out if they go wrong).

You might also want to listen to any "numeric", i.e. messages with a three-digit command, in the 400-599 range; those tend to be (usually non-fatal to your connection) errors, such as "no such nick" errors that would imply NickServ is down for maintenance. (IRC being a mess, there are error numerics outside that range, though none I see that'd apply in your use case – with our server software, anyway.)

For the actual taxonomy output, @AbstractBeliefs has you covered. [Note you may also receive an error – as a `NOTICE` – if the user isn't registered](https://github.com/atheme/atheme/blob/f8cd25ff8e35d8fde3d136e1fb5421d0068aa465/modules/nickserv/taxonomy.c#L51).

I really need to make a decent API for this sort of thing at some point …

extras/freenode_taxonomy_reader.rb
ilbelkyr: While freenode does limit the number of metadata entries per account to something reasonable, that's probably a decent idea nonetheless.

app/controllers/settings_controller.rb
hmadison: Should we update all of the service integrations to do that too?

pushcx: After this is merged we can make an issue for it, no need to do the others now.

extras/freenode_taxonomy_reader.rb
pushcx: This isn't resolved, `taxonomy_lines` returns an array when we only need it to return a string for the single username line.

pushcx:
One small code change, one irritating little conflict (always with `schema.rb`...), and a handful of `rubocop` gripes.

I'm sorry I let this hang out so long, but I look forward to merging it.

PR #862, opened 2020-05-31
add simplecov gem

gabrielbaldao:


pushcx:
Hi Gabriel,

Thanks for opening a PR. Could you say some more about the benefit here? Right now I've been weighing the value of adding more tests by seeing what breaks, or just general experience with test ROI. I've never seen a strong benefit to having 100% test coverage by default, which simplecov encourages.

gabrielbaldao:
Hi @pushcx, sorry, i just forgot about my description,
So, using the simplecov helps me to see where needs more attention.
Because in some moment, you can write some code and this code can broken another code, so this gem can help to watch about how many tests you have been write.
I know about 100% in simplecov not is really 100%  coverage but already help, because in this project just have 50%.

pushcx:
After thinking about it a while, I'd like to merge this. It would be helpful to have a guide to important places we're lacking tests. Like: we added a test that the front page loads only fairly recently, and it turns out that it's pretty important that it load.

For anyone seeing this in the future: I don't like adding tests only to move up the percentage, which having simplecov encourages. As I'm programming and merging PRs today I dealt with a couple tests that haven't caught a bug in years but needed maintenance. So let's continue to make thoughtful guesses of the return on investment that each test will provide.

pushcx:
Thanks for this. I resolved the merge conflict from the Rails 6 upgrade and I'm glad to add this.

Issue #863, opened 2020-06-01
Master CI is failing.


thomasdziedzic:
From https://github.com/lobsters/lobsters/commit/208a4a89459f7999a55b0d3e86a505303c993a96

https://travis-ci.org/github/lobsters/lobsters/builds/693304480

```
  1) keybase proofs new renders the expected kb_username

     Failure/Error: @kb_avatar = Keybase.avatar_url(@kb_username)

     

     OpenSSL::SSL::SSLError:

       SSL_connect returned=1 errno=0 state=error: certificate verify failed (certificate has expired)

     # ./extras/sponge.rb:17:in `start'

     # ./extras/sponge.rb:209:in `block in fetch'

     # ./extras/sponge.rb:205:in `fetch'

     # ./extras/keybase.rb:19:in `avatar_url'

     # ./app/controllers/keybase_proofs_controller.rb:11:in `new'

     # ./spec/requests/keybase_proofs_spec.rb:20:in `block (3 levels) in <top (required)>'
```

thomasdziedzic:
I just tried running the test on my laptop with `bundle exec rspec spec/requests/keybase_proofs_spec.rb` and it failed.

Looks like this might be an error related to running the test on travis ci.

pushcx:
The main bug here is that the test is performing a network request at all, let's fix it at that level. (@NattyNarwhal suggests it may be an intermediate cert expiring.)

PR #865, opened 2020-06-11
fix keybase spec

thomasdziedzic:
Fix https://github.com/lobsters/lobsters/issues/863


pushcx:
Thanks for addressing this. Webmock is a significant new dep, but I can see it giving a lot of value in other tests.

thomasdziedzic:
Thanks for merging this fix in. Looks like master is now succeeding!

Issue #854, opened 2020-05-12
hide email addresses on the hats page when not logged in


zg:
I'm one of the few users with a hat, and I received a spam email because someone scrapped the page.

One way to mitigate this is to require being logged in to view the sensitive contents of how users are validated.

arandomandy:
I'm sorry to hear that, @zg.

This sounds good to me. What do you say, @pushcx ? I suppose my question is whether you think it is useful to display the @domain.ltd? To be honest, I do not but perhaps I misunderstand its purpose. In my mind, if the hat's there, it means someone approved it; that should be enough to prove the veracity of it for others. However, there might be other reasons to include it that I am not aware of.

pushcx:
It's convenient to give a method of contact, as hats are for speaking officially on behalf of a project, so I'd like to keep that display for logged-in users. But yeah, let's not display them publicly.

cnst:
Another option is to use web-pages in place of email addresses.  I believe when the feature was originally implemented by `jcs@`, he was asking for email addresses, because OpenBSD, doesn't have webpages with a catalogue of developers, but in NetBSD we do, and I've added access by ID a decade ago, so, that's a potential improvement to consider (however, it seems like not all developers have bothered to update the official webpage with their name and login).

zg:
> Another option is to use web-pages in place of email addresses. I believe when the feature was originally implemented by `jcs@`, he was asking for email addresses, because OpenBSD, doesn't have webpages with a catalogue of developers, but in NetBSD we do, and I've added access by ID a decade ago, so, that's a potential improvement to consider (however, it seems like not all developers have bothered to update the official webpage with their name and login).

How would that work in the context of someone who has a hat designating that they're an employee for a large corporation, e.g. Apple?

jcs:
This was implemented in #868 and should be closed.

pushcx:
Thanks for catching the stale issue, closing.

PR #868, opened 2020-06-15
Updates hats index view

magikid:
This checks that a user is logged in before showing the hat's link on
the hat index page.  This change is needed because a malicious anonymous
user scraped the page and send a user spam based on the email found on
it.

Issue: #854

### DIFF
--- a/app/models/hat.rb
+++ b/app/models/hat.rb
@@ -36,7 +36,7 @@ class Hat < ApplicationRecord
         "title=\"Granted #{self.created_at.strftime('%Y-%m-%d')}"
 
     if !hl && self.link.present?
-      h << " - #{ERB::Util.html_escape(self.link)}"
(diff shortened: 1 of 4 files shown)


pushcx:
Thanks for taking a look at this, but this fix isn't quite right. This would also stop showing logged-out visitors if hats were doffed, or links to credentials. How about just showing the `@example.com` for email addresses to logged-out visitors?

pushcx:
Ping @magikid - it's been a few weeks, do you want to pick this back up?

magikid:
I kept meaning to take another look at this but just couldn't find the time. I finally did and made the change that you suggested.  Now logged out users will only see the domain from the email address and not the user part.  I hope it's alright but I stripped out the `@` sign and didn't include it in what is shown to logged out users.

pushcx:
Looks great, seems fine to leave off the `@`. I like the clear tests here.

PR #870, opened 2020-06-19
Support token-based access for the RSS feed of upvoted stories

creature:
Lobste.rs has a [page of a user's upvoted stories](https://lobste.rs/upvoted/stories), which includes an RSS feed. Most of Lobste.rs' feeds can authenticate a user via a `?token=` URL parameter, so feed readers can retrieve them without needing a user's session cookie. This was missing for upvoted stories; requests for the RSS feed redirect to the Lobste.rs login page.

This PR fixes this bug, and includes a test case to demonstrate it works as expected.


pushcx:
Thanks for recognizing and fixing this bug, great to have this working consistently (and at all).

Issue #864, opened 2020-06-09
Refactoring: rename Comment scope `for_user` to `accessible_to_user`


roryokane:
As discussed briefly on [pushcx’s stream today](https://www.twitch.tv/videos/645356413) from 1:14:00 to 1:15:25, this `for_user` scope in the `Comment` model:

https://github.com/lobsters/lobsters/blob/ee42f2f308fcc29b3c9f1250a0de58014ccb3567/app/models/comment.rb#L36

would be better if named `accessible_to_user`, given the separation of the in-development `Comment.not_on_story_hidden_by` scope.

All the callers will need to be renamed too. I see only 7 search results for `.for_user` in the codebase.

gabrielbaldao:
can I get this one?

roryokane:
I’m not working on this, so feel free.

gabrielbaldao:
I cannot watch your video, can you look at this PR? I also did a refact in the scope condition, I wanna know if I am forgetting to change something

pushcx:
Completed by @gabrielbaldao's PR. Thanks again!

PR #874, opened 2020-06-30
rename comment scope

gabrielbaldao:
As quoted in #864, i change the `for_user`scope to `accessible_to_user` 
I also change the condition `user && user.is_moderator?` to ` user&.is_moderator?`

### DIFF
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -37,4 +37,18 @@ describe Comment do
     comment = build(:comment, markeddown_comment: "a" * 16_777_216)
     expect(comment).to_not be_valid
   end
+
+  describe ".accessible_to_user" do
+    it "when user is a moderator" do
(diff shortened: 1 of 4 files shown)


spec/models/comment_spec.rb
```diff
@@ -37,4 +37,23 @@
     comment = build(:comment, markeddown_comment: "a" * 16_777_216)
     expect(comment).to_not be_valid
   end
+
+  describe ".accessible_to_user" do
+
+    context "when user is a moderator" do
+      let(:moderator) { create :user, :moderator }
```
pushcx, line 44: Please rewrite these to just be `it` tests, move the creation and such inside. I try to only use `let` in the codebase when there's a very generic object that many tests want to use without caring very much about.

pushcx:
Thanks for matching the test style and, of course, for taking on this code cleanup.

PR #877, opened 2020-07-06
Update the version of Ruby in the Readme

ahuth:
87f2404c08b49bd3746c7cf1c96ccdce5d5c8436 bumped the Ruby version from 2.3.3 to 2.7.0, but it looks like the Readme was never updated.


pushcx:
Yep, I missed updating that. And there's basically a 100% chance it gets forgotten again. How about replacing the exact version with a markdown link to `.ruby-version`, which is what gets used by the app? Little less convenient, but at least it won't be able to go out of date.

Thanks for catching this and contributing.

ahuth:
Good idea. Will do 👍

pushcx:
Perfect, thank you!

PR #878, opened 2020-07-08
Use a details tag for the story guidelines

ahuth:
Use a `<details>` tag for the story guidelines on the story submission page, instead of an anchor tag with a click handler. This improves accessibility, and is related to #755.

Collapsed:
story submission guidelines collapsed

Expanded:
story submission guidelines expanded

I wanted to contribute back to Lobsters (which has been an amazing community). This seemed like a straightforward change to make as my first one to this project.

https://github.com/lobsters/lobsters/issues/755#issuecomment-538805549 rightfully indicates an aversion to bike-shedding, and I think this is a clear accessibility win, because

- Assistive technology will indicate that there's more content, and whether it's expanded or collapsed.
- We avoid an anchor tag, which generally is expected to navigate to another url or another location on this page. That's not really what's happening here.

Also, we end up with (slightly) less JS, which is always nice.

One thing to note is that [IE11 doesn't support the details tag](https://caniuse.com/#feat=details) (although it is supported on all versions of Edge). I haven't been able to find any info on browser support for Lobsters, but if we want to support IE11, these changes won't be a good solution.

### DIFF
--- a/app/views/stories/_form.html.erb
+++ b/app/views/stories/_form.html.erb
@@ -73,54 +73,50 @@ Please don't use this to promote the story, summarize the post, or explain why y
 See the guidelines below for more." %>
     </div>
 
-    <div class="boxline actions markdown_help_toggler">
-      <a href="#" id="story_guidelines_toggler">
-        Story submission guidelines
-      </a>
-      <div id="story_guidelines" style="<%= show_guidelines?? "" : "display: none;" %>">
-        <ul>
+    <%= tag.details class: "boxline actions", open: show_guidelines? ? true : nil do %>
+      <summary>Story submission guidelines</summary>
+      <ul>
 
-          <li><p>
-          Do not editorialize story titles, but when the original story's
-          title has no context or is unclear, please change it.  <strong>Please
-          remove extraneous components from titles such as the name of the
-          site, blog, section, and author.</strong>
-          </p></li>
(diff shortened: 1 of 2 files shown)


app/views/stories/_form.html.erb
```diff
@@ -73,54 +73,50 @@ Please don't use this to promote the story, summarize the post, or explain why y
 See the guidelines below for more." %>
     </div>
 
-    <div class="boxline actions markdown_help_toggler">
-      <a href="#" id="story_guidelines_toggler">
-        Story submission guidelines
-      </a>
-      <div id="story_guidelines" style="<%= show_guidelines?? "" : "display: none;" %>">
```
ahuth, line 80: We could keep this `<div>`, and prevent the re-indentation of the guidelines. This would make `git blame` cleaner. Not sure it's worth the trouble, but I'm open to doing that.

pushcx:
Cool, thanks for this great little PR. There's also `markdown_help_toggler` if you want to do this a second time. :)

And then the only other toggle is the comments toggler which... might get hairy. Maybe that one would be more work than it's worth; `_comment.html.erb` is the most complicated markup on the site.

ahuth:
Thanks! I'll look at markdown_help_toggler, and _comment.html.erb to see if it's worth looking at.

PR #880, opened 2020-07-12
filter out front page posts from 'newest'

AlessandroMinali:
According to:
https://github.com/lobsters/lobsters/blob/86fb86788e3afa4de4b8a10d25fd09785897b0dd/app/views/home/index.html.erb#L1-L5

Currently this is not reflected when visiting https://lobste.rs/newest.

### DIFF
--- a/app/models/story_repository.rb
+++ b/app/models/story_repository.rb
@@ -15,7 +15,8 @@ class StoryRepository
   end
 
   def newest
-    Story.base.filter_tags(@params[:exclude_tags] || []).order(id: :desc)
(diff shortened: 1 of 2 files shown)


pushcx:
`/newest` is support to be everything. This message appears at the top of [/recent](https://lobste.rs/recent), which is the newest stories with front page posts filtered out.

AlessandroMinali:
Okay, but '/recent' also does not filter front page posts. I see dups right now between both, live. Shall I make a PR for that?

pushcx:
Yes, if a story on the home page appears on `/recent`, that's a bug I'd appreciate a patch for. Probably the right place to fix is the `Stories.recent` scope.

PR #881, opened 2020-07-13
Update deprecrated Faker params

AlessandroMinali:
Just a small fix for error encountered when setting up this project.

Faker reference: https://github.com/faker-ruby/faker/blob/3f72057a7522849243c4c7fa8d0527580588cbcf/lib/faker/default/internet.rb#L44-L47

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -11,7 +11,7 @@ class FakeDataGenerator
       name = Faker::Name.name
       password = Faker::Internet.password
       create_args = {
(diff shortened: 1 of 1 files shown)


lib/tasks/fake_data.rake
```diff
@@ -14,7 +14,7 @@ class FakeDataGenerator
         email: Faker::Internet.email(name),
```
pushcx, line 14:
```suggestion
        email: Faker::Internet.email(name: name),
```

pushcx:
Thanks for finding this bug and fixing it!

PR #883, opened 2020-07-13
Build a story object to test valid urls

AlessandroMinali:
Currently `spec/models/story_spec.rb` fails when run by itself.

Fix is to build a real object so validations are performed properly.


pushcx:
Thanks for finding this surprising little bug. I've added this missing nil check as we'd end up getting bitten by the missing check again.

PR #886, opened 2020-07-14
Only deliver reply notifications on reply+mention

### LINKED ISSUE #876: Don't email about a reply and mention for the same comment
If someone has reply and mention notifications turned on, a [response that mentions them](https://lobste.rs/s/orrpdv/lobsters_launched_8_years_ago_today#c_rlvnjv) will generate two emails. Let's skip the mention email in favor of the reply email.

emdemir:
When a reply is made to a parent comment that mentions the parent user,
the parent gets notifications for both the reply and the mention. This
commit makes it so that the reply notification is prioritized, and
mention notifications are only sent when the user didn't directly
reply to the comment.

Fixes #876.


app/models/comment.rb
```diff
@@ -248,8 +248,19 @@ def delete_for_user(user, reason = nil)
...
-  def deliver_mention_notifications
+  def deliver_notifications
+    notified = deliver_reply_notifications
+    deliver_mention_notifications(notified)
+  end
+
+  def deliver_mention_notifications(notified = nil)
+    notified ||= []
+
     self.plaintext_comment.scan(/\B\@([\w\-]+)/).flatten.uniq.each do |mention|
```
pushcx, line 259: Name this array like `to_notify = self.plaint....` and then update the loop to `(to_notify - notified).each do`, then you don't need the imperative if/next.

app/models/comment.rb
```diff
@@ -248,8 +248,19 @@ def delete_for_user(user, reason = nil)
     self.user.refresh_counts!
   end
 
-  def deliver_mention_notifications
+  def deliver_notifications
+    notified = deliver_reply_notifications
+    deliver_mention_notifications(notified)
+  end
+
+  def deliver_mention_notifications(notified = nil)
```
pushcx, line 256: You can set `notified=[]`, Ruby doesn't have the Python gotcha about list defaults in args.

app/models/comment.rb
```diff
@@ -293,10 +304,13 @@ def users_following_thread
   end
 
   def deliver_reply_notifications
+    notified = []
```
pushcx, line 307: Will this list ever be different from `users_following_thread`? If you use that list directly you won't have dupes, either.

spec/models/comment_spec.rb
```diff
@@ -1,5 +1,10 @@
 require "rails_helper"
 
+# TODO: This should probably go somewhere else.
```
pushcx, line 3: `spec/rails_helper.rb`

spec/models/comment_spec.rb
```diff
@@ -37,4 +42,73 @@
     comment = build(:comment, markeddown_comment: "a" * 16_777_216)
     expect(comment).to_not be_valid
   end
+
+  it "sends reply notification" do
+    # The user receiving the notification.
```
pushcx, line 47: How about renaming `u` to `recipient`?

spec/models/comment_spec.rb
```diff
@@ -37,4 +42,73 @@
...
+    u = create(:user)
+    u.settings["email_notifications"] = true
+    u.settings["email_replies"] = true
+
+    # The user sending the notification.
+    u2 = create(:user)
```
pushcx, line 53:
```suggestion
    sender = create(:user)
```

spec/models/comment_spec.rb
```diff
@@ -37,4 +42,73 @@
...
+    # Story under which the comments are posted.
```
pushcx, line 54: You don't need to create `s` explicitly, creating a comment without specifying a story will implicitly create one.

spec/models/comment_spec.rb
```diff
@@ -37,4 +42,73 @@
...
+    s = create(:story)
+
+    # Parent comment.
+    c = build(:comment, story: s, user: u)
```
pushcx, line 58:
```suggestion
    c = create(:comment, user: recipient)
```

Then there's no need for the explicit `c.save` on the next line. Same for next comment.

pushcx (approved):
Hi, thanks for taking this on! I see how your approach works, and I really appreciate that you added tests to make sure this stays right. I've left comments for some changes, which I think are all Ruby/Rails idioms. Thanks for jumping into an unfamiliar environment for this. :)

pushcx (changes requested):

spec/models/comment_spec.rb
emdemir: Wouldn't the parent and child comments need to share a story? Or does that happen automatically when I pass the parent comment?

app/models/comment.rb
emdemir: Yes, because not all people who follow the thread have `email_replies` or `pushover_replies` enabled.

spec/models/comment_spec.rb
emdemir: Save as above, isn't it required to share a story object if you're replying to a comment?

pushcx:
I'm sorry I lost track of this for so long. I've dusted it off by merging in `master`. I'm going to merge this in and immediately land a PR for the linter fixups.

PR #887, opened 2020-07-14
Add URL prefix match to search

### LINKED ISSUE #850: Feature: A search by `url:` option
There's a search by `domain:` option, but nothing more granular. This proposal is for a search by `url:` option that can cover the "I want search for submissions that point to this url" use case.

emdemir:
This commit adds a url: keyword to the search that matches stories and
comments by an exact URL.

Fixes #850.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -75,7 +75,7 @@ class Search
   end
 
   def with_stories_pointing_to_url(base, url)
-    base.where('url = ?', url)
+    base.where('url LIKE ?', "#{url}%")
   end
 
   def with_stories_matching_tags(base, tag_scopes)
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -74,6 +74,10 @@ class Search
     end.where('domains.domain = ?', domain)
   end
 
+  def with_stories_pointing_to_url(base, url)
+    base.where('url = ?', url)
+  end
+
   def with_stories_matching_tags(base, tag_scopes)
     story_ids_matching_tags = with_tags(
       Story.unmerged.where(:is_expired => false), tag_scopes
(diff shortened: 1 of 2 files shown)


rwanyoike:
Thanks. Is it possible for the `url:` query to be flexible (like the `title` search)  so you can search eg.:

```
url:http://example.com/blog => 10 results
url:http://example.com/blog/2020 => 4 results
url:http://example.com/blog/2020/02/ => 2 results
url:http://example.com/blog/2020/02/hello-world.html => 1 result (exact match)
```

and eg.

```
url:http://example.com/blog/2020/02/hello-world.html => same result
url:example.com/blog/2020/02/hello-world.html => same result
```

emdemir:
Sure, I don't know how this impacts search performance though.

app/models/search.rb
```diff
@@ -74,6 +74,10 @@ def with_stories_in_domain(base, domain)
     end.where('domains.domain = ?', domain)
   end
 
+  def with_stories_pointing_to_url(base, url)
+    base.where('url LIKE ?', "#{url}%")
```
pushcx, line 78: Two questions:
1. I don't think this correctly escapes the url. What if the user enters `url:htt%`? I think this would find every story with a url. And permit SQL injection. (Maybe I'm missing it and this class escapes at a different level? There's a lot happening in here.)
2. How does this handle merged stories?

pushcx (changes requested):

app/models/search.rb
emdemir: 1. ActiveRecord uses parameter binding. The data isn't interpolated directly into the string but sent separately.
2. I will have to investigate that.

pushcx:
The search engine has been entirely redesigned to use a parser, so this PR is too different to apply. Thanks for taking a try at it.

PR #890, opened 2020-07-15
Add configurable maximum page limit to search

### LINKED ISSUE #889: Raise max search result pages to 50
No matter how many results are returned in the search results, I can’t seem to go beyond page 5. Is that a feature? I tried looking into the code but couldn’t understand if there’s a limitation.

emdemir:
This commit adds a new constant, SEARCH_MAX_PAGES, which limits the
maximum number of pages that search results can have.

Fixes #889.


app/models/search.rb
```diff
@@ -23,7 +26,12 @@ def initialize
   end
 
   def max_matches
-    100
+    if SEARCH_MAX_PAGES.nil?
+      # A conservative maximum.
+      2 ** 31
+    else
+      SEARCH_MAX_PAGES * self.per_page
```
pushcx, line 33: I like how clear this multiplication is. Let's remove the `if` so it's clearer. Just stick it in your comment, `# Set this to 2 ** 31 for more than enough pages`. If someone's editing this, they know at least a little code, so we don't have to to offer them `nil` instead of directly using this constant.

pushcx (approved):
Thanks for lifting this limitation. One small change before it's ready to ship.

pushcx (changes requested):

pushcx:
This was fixed elsewhere. Thanks for taking a shot at it.

Issue #861, opened 2020-05-30
Run tests against mariadb and postgresql


pushcx:
We run mariadb in production, but many sister sites choose to run postgresql. We pretty regularly let mysql-only quirks sneak into the codebase when it could be fairly cheap to prevent this. I'd also like to have the option to migrate us: we only get burned by a mysqlism about once a year, though not too badly, but some issues are a constant irritation like mariadb not running DDL transformations in a transaction so migration development is an iterative process of repairing a half-transformed database.

So, if anyone would like a project, let's see what it would take to set up our test suite to run twice: once against mariadb/mysql, once against postgresql. Ideally it'd skip a run if `config/database.yml` doesn't set up a db, and give a useful error if neither database is configured. That way contributors don't have to configure both if they don't want to - but we'd update jenkins to always run both, of course.

If this means writing a custom test runner, an optional nicety would be to also run `rubocop` and `ruumba` so I could run one command instead of `rspec && rubocop && ruumba -D -c .ruumba.yml`

PR #891, opened 2020-07-20
Run specs against MySQL and PostgreSQL

Quintasan:
This is the dumbest approach that:

1. Checks if there's a database configuration with a name matching `/test/` and a given adapter.
2. If there's a match then it renames the configuration to `test` instead of whatever it was.
3. Runs the spec suite
4. Reverts the changes made (even if you CTRL+C)
5. Exits cleanly if there's no adapter found of any of the databases

Advantages:

* It is relatively simple
* Doesn't involve monkey-patching
* Doesn't mess with Rails internal connection handling

Disadvantages:

* Relies on `sed` (this can be refactored if portability is a concern)
* Relies on the configuration name to have the word `test` in it

Related to #861

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,4 +63,5 @@ group :test, :development do
   gem "vcr"
   gem "webmock" # used to support vcr
   gem 'simplecov', require: false
+  gem "pg"
 end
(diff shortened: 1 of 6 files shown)


Quintasan:
@pushcx small nudge, let me know if you don't like it, I have another approach if this one feels too complicated

Quintasan:
@pushcx nudge

PR #894, opened 2020-08-03
Fixes bug #893 by not considering tags on expired stories.

### LINKED ISSUE #893: Bug: User model considers expired stories when computing most_common_tag, leading to users with 0 stories and a most common tag
For users with no submitted stories, when viewing their own profile, it claims there is a most commonly used tag. I believe this should not be appearing, since the most common tag of nothing is nothing.

image

When viewing other user profiles who have no submitted stories, there is no listed tag.

image

Because the user helper checks to see if a tag exists before inserting that wording, it looks like this error is generated in this method https://github.com/lobsters/lobsters/blob/083dbb1e67d337b1d0481c294b3f362d0c68c2b5/app/models/user.rb#L494

_Edit: My apologies, when inserting the screenshot on a new ticket, the ticket opened itself before it was complete._

threeplanetssoftware:
The code to compute a user's most common tags to display on their profile page doesn't take into account the state of the story, leading to users being able to have 0 stories, yet a most common tag for those 0 stories, as evidenced in Bug #893. This change edits `lobsters/app/models/user.rb` to have `most_common_story_tag` take into account the `is_expired` nature of the story and ensure the story is still visible. 

https://github.com/threeplanetssoftware/lobsters/blob/87dc77fe7ef3989045fd2fbea96c4a0e38aee052/app/models/user.rb#L494

This pull request should fix #893, unless it is intended behavior for expired posts to be included in these calculations. If that is the case this can be tossed out and the bug report closed as a feature.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -495,7 +495,7 @@ class User < ApplicationRecord
     Tag.active.joins(
       :stories
     ).where(
-      :stories => { :user_id => self.id, is_expired: false }
+      :stories => { :user_id => self.id, :is_expired => false }
     ).group(
       Tag.arel_table[:id]
     ).order(
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -495,7 +495,7 @@ class User < ApplicationRecord
     Tag.active.joins(
       :stories
     ).where(
-      :stories => { :user_id => self.id }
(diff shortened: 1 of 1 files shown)


pushcx:
Seems legit, and rare enough it's not worth a slow test hitting the database. Thanks for this fix!

PR #895, opened 2020-08-03
Fixes bug #830 by adding a check for if a user is new

### LINKED ISSUE #830: users can suggest tags that story submitters are not permitted to use
A new user (an account less than 70 days old) recently submitted story rbbtgu, tagged testing.  The story was an "ask."  New users are not allowed to use the ask tag.  Someone, prior to my deleting the story, suggested it be tagged "ask."  That was appropriate, as such, but that tag should not be available as a suggestion if the user themselves cannot use it.

The story tag suggestion feature should account for new (green) users and not permit the tags ask, meta, or show to be suggested in that case.

threeplanetssoftware:
Edits the `Tag.valid_for?` method to add a check on whether the user is new or not. It draws on `Tag.permit_by_new_users` to disallow any that are not for new users. This pull request also adds in some tests for the Tag object since complexity is being add (before it was just checking if it was a moderated tag, now also new users, in the future ??). This will ensure other tag changes in the future still allow or deny these tags are appropriate.

Fixes #830


app/models/tag.rb
```diff
@@ -50,6 +50,8 @@ def css_class
   def valid_for?(user)
```
pushcx, line 50: We've got some semantic overloading happening. `Tag#valid_for` is also used by `Tag.all_with_filtered_counts_for` (which should probably be done by SQL rather than Ruby, possibly as a separate query as 0b67cd2de932d03027959901a5d240b8dfd53c50 did for `Tag.active.all_with_story_counts_for`). So this change to `valid_for` prohibits new users from filtering out `ask` posts.

I don't want to change the core behavior, the goal of #23 was to make sure users don't filter out the admin-only tags.

Maybe the right solution here is renaming `valid_for?` to something more like `can_be_applied_by?` and creating a similar second method named something like `can_be_filtered_by?` (I'm not in love with these names, top of my head.)

pushcx (changes requested):
I see how this mostly works, but there's a shared dependency here that I think we're going to have to tease apart.

And then, second: how do we communicate this to users? There's two cases, one of which is not apparent because you're not a mod, but I can unfairly toss you in the deep end over because you will be bathed in the warm, deep-end-tossing glow of nostalgia for Walraven:

1. How will suggesting users understand this restriction? Should wanting to suggest a tag that the user isn't able to use just be a flag to moderators that the story should be removed? (Which feels like a hyper-specific flag and hard to explain itself.)
2. New users don't get it. They see the error that they're not allowed to use a tag and most of the time they tag the story as `programming` and submit it anyways. How do we improve that error message so they don't want to do that? How do we acculturate more effectively? (An approach that could work but feels unacceptably 'gotcha' would be to show the tag and, when they attempt to save with it, prevent them from submitting that URL at all by saving it to a new model.) And: can we ModNote that attempt like `Story` calling `ModNote.tattle_on_story_domain`?

threeplanetssoftware:
Thanks for the feedback. I'll take a look and try for a more elegant solution!

threeplanetssoftware:
Just checking in, anything further you need me to do for this?

pushcx:
LTGM :shipit:

Issue #735, opened 2019-09-11
Redesign tag page


pushcx:
Chatting about [the site design](
https://old.reddit.com/r/web_design/comments/d0cvrl/whats_the_name_of_this_designstyle_found_in/) reminded me that our [tags page](https://lobste.rs/tags) is not particularly useful and could stand to be redesigned from the ground up. This is an open-ended issue to make this page more useful to long-time readers, including defining "better".

rvalles:
It would help to be able to add or suggest adding new tags.

pushcx:
Adding and editing has been possible since 62ef8eeab899418f0b6bf7d4632a32513d70d3de.

I think we're good with suggesting being a `meta` thread, it'd be a bunch of work to build a second way of measuring community interest and concerns.

rushsteve1:
Just throwing a few ideas out there to get the discussion going (I don't think this needs a meta thread but I'll move it if needed):
- It's hard to get to the tags page, it's only linked with one-word links on the search and individual tag (`/t/`) pages. Adding a link either at the top by the "Recent" button or at the bottom by the "Hats" buttons would be nice.
- Personally I do like the word-cloud to show the prominence of each tag. Adding an actual number to each might also be nice since relative sizes can be deceiving.
- Maybe add some more basic info, like the number of tags or the most recent tag added?
- Adding the brief description and filtering statistics like the "suggest" page has seems important.
- A search of some kind? Seems unnecessary but it can probably be copied from the "suggest" page.

roryokane:
I very much agree that the tags page should have the description from the “suggest” page, which is also the [Submit Story page](https://lobste.rs/stories/new). For convenience, here’s a screenshot of the tags field in the Submit Story form:

<img width="860" alt="tags list on Submit Story page" src="https://user-images.githubusercontent.com/79168/65858071-2380ed80-e333-11e9-8fa4-aa22035cd195.png">

I remember multiple occasions when I went to the Tags page to look up the description of a tag, couldn’t find it, and then realized that I had to go to Submit Story instead.

pushcx:
#580 has a big description of where a link to /tags would be appropriate (amongst other navigation cleanup).

Maybe there's also opportunity to make #664 more visible.

Big agreement that description would be valuable here. This page should really be working hard to help people understand the content of the site.

[...]

PR #898, opened 2020-08-08
Updates tag controller to use tag name, fixes #892

### LINKED ISSUE #892: Bug: don't leak tag ids in edit URLs
The urls for tag editing look like https://lobste.rs/tags/35/edit instead of having the name of the tag.

It's admin-only, but it's a bad habit and a small fix.

threeplanetssoftware:
Edits the tag controller, view, and routes to use the tag name instead of the tag ID in the URL. Uses the `:tag_name` variable in the params to pass the name, since using `:tag` conflicts with the `tag` object in the controller. Updated the rspec tests to change anywhere the ID was passed to be the tag name instead. 

Fixes #892


pushcx:
Thank you for taking a run at this unexciting cleanup. I'm sorry to say I've given you a merge conflict when I started #735 to redesign the tag page. Do you mind updating the template again?

threeplanetssoftware:
I don't mind at all, that wasn't much of a conflict!

pushcx:
Thanks for solving this.

PR #899, opened 2020-08-09
Wrap Story and Comment creation inside of a transcation.

hmadison:
The new reconciliation process between an object's Vote and its
memoized count requires that MariaDB can see the records in the Vote
table. Currently, we are seeing the vote be created, then the
memoization for new comments and stories to zero.

We now force this behavior to take place inside of a transaction.
This should keep the code working the way we expect.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -48,7 +48,7 @@ class CommentsController < ApplicationController
       end
     end
 
-    if comment.valid? && params[:preview].blank? && comment.save
+    if comment.valid? && params[:preview].blank? && ActiveRecord::Base.transaction { comment.save }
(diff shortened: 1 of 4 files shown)


app/controllers/comments_controller.rb
```diff
@@ -48,7 +48,11 @@ def create
       end
     end
 
-    if comment.valid? && params[:preview].blank? && comment.save
+    was_saved = ActiveRecord::Base.transaction do
```
hmadison, line 51: Rubocop complains locally that putting this inline makes the line too long.

pushcx:
This PR is @hmadison cleaning up after [some big commits I landed](https://lobste.rs/s/wdjrcd/tags_flags) around tags and flags.

Thanks very much for the help.

PR #902, opened 2020-08-10
sitemap.rb: the upvotes column in the stories model is now score.

alanpost:
I'm completely unclear whether this is sufficient to fix sitemap.rb, and idly wonder if there is any testing that could be done that would have caught this mistake.

Please provide feedback on how this fix can be improved.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -2,7 +2,7 @@ SitemapGenerator::Sitemap.default_host = "https://lobste.rs"
 
 check_hourly = 4.days.ago
 check_daily = 2.weeks.ago
-top_score = Story.all.maximum('upvotes')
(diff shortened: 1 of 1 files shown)


pushcx:
Usually this sort of thing would be caught with a smoke test (and boy did the smoke tests I've added over the last three years pay off for the upvotes/downvotes -> score/flags change) to the endpoint. Looking at the [readme](
https://github.com/kjvarga/sitemap_generator) I don't immediately see a way to exercise this config (which is code rather than data) without actually spitting out files on disk. I'm leery of the alternate approach of stubbing out calls into the library (because it exercises so much less code).

So I'm going to punt on testing and take this fix as-is. Thanks for diagnosing and fixing this, Alan. You now officially know Ruby. :D

PR #904, opened 2020-08-11
about.html: use vs. not vs for versus

alanpost:
The other use on this page uses vs. h/t geoffwozniak.

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -22,7 +22,7 @@
 
   <h2 id="tagging">Tags and Topicality</h2>
   <p>
-  When links/stories are submitted, they must be tagged by the submitter from a list of predefined <a href="/tags">tags</a>.
+  When links or stories are submitted, they must be tagged by the submitter from a list of predefined <a href="/tags">tags</a>.
   Users can choose to <a href="/filters">filter</a> out or
   subscribe to all submissions with particular tags (example: <a href="https://lobste.rs/t/programming.rss">programming.rss</a>)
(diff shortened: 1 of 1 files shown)


app/views/home/about.html.erb
```diff
@@ -143,7 +143,7 @@
   </div>
 
   <p>
-  Less than 1% of stories or comments get flagged, and users are not automatically punished by flags.
+  Fewer than 1% of stories or comments get flagged, and users are not automatically punished by flags.
```
alanpost, line 146: I draw a reviewer's attention to la wik's discussion of this topic:

https://en.wikipedia.org/wiki/Fewer_versus_less

pushcx: Yeah, pass on this one, it doesn't read right.

alanpost: Undone.

app/views/home/about.html.erb
```diff
@@ -74,7 +74,7 @@
   </p>
 
   <p>
-  Some things that are off-topic here but popular on larger, similar sites: entrepreneurship, management, news about companies that employ a lot of programmers, investing, world events, anthropology, and defining the single morally correct economic and political system for the entire world when we can't even settle tabs vs. spaces.
+  Some things that are off-topic here, but popular on larger, similar sites: entrepreneurship, management, news about companies that employ a lot of programmers, investing, world events, anthropology, and defining the single morally correct economic and political system for the entire world when we can't even settle tabs vs. spaces.
```
pushcx, line 77: I can see why one might set off this clause, but I prefer not to break up this sentence.

alanpost: It's undone.

Issue #900, opened 2020-08-09
lobsters-cron, lobsters-daily, expire_page_cache should honor /srv/lobste.rs/http/public/maintenance.html


alanpost:
Our nginx configuration has the following stanza:

```
if (-f $document_root/maintenance.html) {
    rewrite  ^(.*)$  /maintenance.html last;
    return 503;
}
```

Which prevents delegation of requests to puma while maintenance is underway.

Our cron scripts should similarly exit and not do work if this file exists. Currently we comment them out in crontab for the same effect, but would be nice if there was a single switch to enter maintenance mode.

Please add a similar check in each of our three cron jobs.

alanpost:
The lobsters-ansible.git end of this change is now on master: https://github.com/lobsters/lobsters-ansible/pull/67

The lobsters.git end of this is PR #905.

alanpost:
We declined to merge PR #905, as letting expire_page_cache run during maintenance causes no harm and might be a help--reduces the risk of stale data when maintenance is lifted.

As the lobsters-ansible.git end of the problem is merged to master, I'll resolve this ticket.

Issue #67, opened 2013-12-10
Sorting looks wrong in search


nexneo:
Stories with negative points shouldn't be on top. https://lobste.rs/search?utf8=✓&q=go&what=stories&order=points

PR #905, opened 2020-08-11
expire_page_cache: exit early during a maintenance window.

alanpost:
during a scheduled maintenance window, don't expire the page
cache. exit without error instead.

### DIFF
--- a/script/expire_page_cache
+++ b/script/expire_page_cache
@@ -1,4 +1,9 @@
 #!/bin/sh
 
+if [ -f /srv/lobste.rs/http/public/maintenance.html ];
+then
+  exit 0
+fi
+
 find /srv/lobste.rs/http/public/cache/ -type f -not -mmin 2 -not -name ".gitkeep" -delete
 find /srv/lobste.rs/http/public/cache/ -type d -empty -delete


alanpost:
This change, along with #67 in lobsters-ansible (https://github.com/lobsters/lobsters-ansible/pull/67) fixes #900.

pushcx:
What's the value to this one? I know caching is incomplete and deactivated, but even working, if we've put the site into maintenance mode it feels like a risk to leave cached files around to start serving for two minutes when we come back from maintenance.

alanpost:
Good question, and I think this patch is borderline--you did in the last maintenance window comment out all of cron, including this script, which I believe happened as simple habit--that if you had asked yourself this question then, you would not have commented out this script and left it to run.

If that strikes you as plausible about the only remaining reason to add this patch is to leave nothing running whatsoever during maintenance, for reasons of predictability or completeness. That is an abstract and therefor spurious reason, compared to the actual scenario you describe above. Unless you're persuaded by it for reasons not coming to me, we should abandon this change, not comment out cron jobs during maintenance, and leave https://github.com/lobsters/lobsters-ansible/pull/67 as the only necessary work to allow for that.

pushcx:
OK, I'll go ahead and close this. My inbox thanks you for reducing the 500s in maintenance windows. :smile:

PR #906, opened 2020-08-11
comments: remove dangling pipe (|) when vote_summary_for_user is empty.

alanpost:
On /comments, when logged in as a moderator and when a comment has
received no upvotes or flags, a trailing pipe (|) appears in the
byline.

Once a comment is upvoted, the spurious pipe (|) disappears.

Do not display this character when the vote summary is the empty
string.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -124,7 +124,9 @@ class="comment <%= comment.current_vote ? (comment.current_vote[:vote] == 1 ?
           <% if comment.flags > 0 &&
           comment.show_score_to_user?(@user) &&
           (comment.user_id == @user.try(:id) || @user.try("is_moderator?")) %>
-            | <%= comment.vote_summary_for_user(@user).downcase %>
+            <% if (s = comment.vote_summary_for_user(@user).downcase).present? %>
+              | <%= s %>
+            <% end %>
           <% elsif comment.current_vote && comment.current_vote[:vote] == -1 %>
             | -1
             <%= Vote::ALL_COMMENT_REASONS[comment.current_vote[:reason]].downcase %>


app/models/comment.rb
```diff
@@ -495,15 +495,20 @@ def vote_summary_for_user(u)
...
         o << " (#{r_users[k].join(', ')})"
       end
       o
-    }.compact.join(", ")
+    }
+    if r
+        "| ".r.compact.join(", ")
+    else
+       ""
+    end
```
alanpost, line 511: There must be a nicer way of saying this in Ruby, guidance appreciated.

app/views/comments/_comment.html.erb
```diff
@@ -124,7 +124,7 @@ class="comment <%= comment.current_vote ? (comment.current_vote[:vote] == 1 ?
           <% if comment.flags > 0 &&
           comment.show_score_to_user?(@user) &&
           (comment.user_id == @user.try(:id) || @user.try("is_moderator?")) %>
-            | <%= comment.vote_summary_for_user(@user).downcase %>
+            <%= comment.vote_summary_for_user(@user).downcase %>
```
alanpost, line 127: In one sense, this patch simply draws attention to the display bug by changing the lines in question. As far as I understand it, the problem was introduced in commit e1537b0 with that work following a score related regression fixed in commit f377fd.

In this case, I'd appreciate a more idiomatic or straightforward fix.

app/models/comment.rb
alanpost: I avoided modifying the model by moving the conditional to the template.

app/views/comments/_comment.html.erb
```diff
@@ -124,7 +124,7 @@ class="comment <%= comment.current_vote ? (comment.current_vote[:vote] == 1 ?
           <% if comment.flags > 0 &&
           comment.show_score_to_user?(@user) &&
           (comment.user_id == @user.try(:id) || @user.try("is_moderator?")) %>
-            | <%= comment.vote_summary_for_user(@user).downcase %>
+            <%= "| #{comment.vote_summary_for_user(@user).downcase}" if comment.vote_summary_for_user(@user) %>
```
alanpost, line 127: Thank you @hmadison for helping me make this patch more concise.

Here the side-effect is that we invoke vote_summary_for_user twice--if the template language supports local variables I can use that to avoid doing that, here. (Or, clearly, one could memoize the function call.)

hmadison, line 127:
```suggestion
<% if (s = comment.vote_summary_for_user(@user).downcase).present? %>
  | <%= s %>
<% end %>
```

Something like this would memoize the call.

alanpost: done.

alanpost: side effect addressed in the latest patch set.

pushcx:
Thanks for diagnosing and solving this, @alanpost and @hmadison!

Issue #724, opened 2019-08-19
Rails 6 Post Upgrade


thomasdziedzic:
Enable the following in `config/initializers/new_framework_defaults_6_0.rb`:

- [x]  # Rails.application.config.action_view.default_enforce_utf8 = false
- [ ]  # Rails.application.config.action_dispatch.use_cookies_with_metadata = true
- [ ]  # Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false
- [x]  # Rails.application.config.active_job.return_false_on_aborted_enqueue = true
- [x]  # Rails.application.config.active_storage.queues.analysis = :active_storage_analysis
- [x]  # Rails.application.config.active_storage.queues.purge    = :active_storage_purge
- [x]  # Rails.application.config.active_storage.replace_on_assign_to_many = true
- [ ]  # Rails.application.config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
- [ ]  # Rails.application.config.active_record.collection_cache_versioning = true

# After upgrades

- [ ] Update https://github.com/lobsters/lobsters/blob/441786e4c1b37ae7e78f6dbd9d6256678b950acb/config/application.rb#L27 to 6.0 and remove upgrade file.

jcs:
Is this still relevant now that the project is on Rails 7?

pushcx:
No, these defaults were all adopted. Thanks for noticing the stale issue.

PR #911, opened 2020-08-19
use Rails 6.0 defaults

thomasdziedzic:
This fixes https://github.com/lobsters/lobsters/issues/724

You can read more about the options and what they do, here: https://medium.com/@dylansreile/rails-6-0-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-586146f371e8

Since lobste.rs has been running for a while on Rails 6.0, I just decided that using the Rails 6.0 defaults would now be fine instead of uncommenting each option 1 by 1.


thomasdziedzic:
@pushcx any chance you could review this? Once this is done, I'll look into doing the rails 6.1 upgrade.

PR #916, opened 2020-08-20
Update fake_data to create test Categories and pass those in to new Tags

### LINKED ISSUE #914: Rake fake_data errors out due to not adding categories to the creation of tags.
After categories were to tags, the `rake fake_data` script needs to be updated to reflect the validation of a tag to belong to a category. Currently test data generation fails because the tag is being created without a category which is invalid. While the category could be hardcoded to the default 'test' that is created, it would be better to generate a series of test categories so the tag page functionality can be adequately tested. 

In addition, the use of `Faker::Lorem.sentence()` and `Faker::Lorem.paragraph()` could be updated to the new format that Faker expects `.sentence(:word_count: x)` and `.paragraph(number: x)` to prevent warnings. This is similar to the change from `.email(name)` to `.email(name: name)` from #881.

threeplanetssoftware:
This fixes #914 by creating test Categories during `rails fake_data` and pulling those randomly in to the creation of new Tags. 

**Note**: This does not use `find_and_create_by!` for Tags since that could attempt to create another tag with the same name if the random Category is different. Instead it checks for an existing tag (regardless of Category) and, if that is not found, then tries to create a new tag. 

This also updates `rails fake_data` to use the newer Faker::Lorem syntax, specifying the parameter name in .sentences() and .paragraphs(). This is a follow-on from the work in #881.

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -1,7 +1,8 @@
 class FakeDataGenerator
-  def initialize(users_count, stories_count)
+  def initialize(users_count, stories_count, categories_count)
     @users_count = users_count
     @stories_count = stories_count
+    @categories_count = categories_count
   end
 
   def generate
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for fixing this, and even updating our deprecated style.

PR #918, opened 2020-08-20
Add category CRU (no D)

### LINKED ISSUE #917: Web-based category editing
We have basic CRUD forms for admins to create and edit tags with updates to the modlog. We could use the exact same functionality for categories. Same logging, same permissions, very similar forms. This should be pretty straightforward to implement given that it basically involves porting over `TagsController` (except `index`) with its views and tests.

(Maybe I can sucker @threeplanetssoftware into this given his familiarity with tag editing code from #898 and categories from #914?)

threeplanetssoftware:
This fixes #917 by adding in basic create and update functionality for categories, it does not add delete as tags also are not deletable and I was mirroring their functionality. The edit link for a tag is the first list item (for an admin) in the tags that belong to that category. The link to create a new category is at the bottom of the tags page next to the link to create a new tag. 

This adds tests for categories, basically mirroring those of tags. 

This cleans up a few, presumed bugs, such as not having a close `<li>` in the list of tags and flipping the ordering of the start of the ordered list and the tag enumeration. Currently a new `<ol>` is created for every tag, with exactly one `<li>`, which seems odd, so I flipped that to be one `<ol>` with all the tags in separate list items. 

This has a migration to add the category_id to moderations and tweaks the moderation page to allow categories to make sense.


pushcx:
This is perfect, thank you. I'm sorry it took me so long to merge.

PR #921, opened 2020-08-26
Rewrite small method `Story#can_be_seen_by_user?`

ron-wolf:
The new method definition should return the same value while being simple, readable, and intuitive w.r.t. the name of the function.


pushcx:
Looks like a nice cleanup of something that's grown a bit hairy over time, thanks!

PR #922, opened 2020-08-26
search.rb: association is :story_text, not :story_texts.

alanpost:


pushcx:
Thanks for working this out from the 500s, deploying.

PR #923, opened 2020-08-27
various cosmetic cleanups to moderation logs

alanpost:
Add some missing spaces between fields in a mod note, be extra clear that a short_id is a story short id, and remove the domain_id from the change report--the url contains the necessary information.


alanpost:
I'm happy to write up SQL to migrate the existing database records. If you want it done with the Ruby ORM though I'll need someone to help with that. Let me know if you'd like either.

pushcx:
I'd love to have a migration cleaning them up, and I'm not picky about the form it takes. Run `rails generate migration name_here` to create the file, and see the recent `db/migrate/20200807021926_memoize_score_not_upvotes.rb` for an example of how to invoke an SQL query.

PR #924, opened 2020-08-28
Update fake_data script to help troubleshoot flags.

threeplanetssoftware:
Creates users that are admins, as well as moderators, flags comments and stories from random users, adds user karma and invite hierarchy. Not wedded to the numbers used (5% chance of a comment or story flag for every comment and story, have to have at least 5 users [2 defaults, 3 others] to start into the user invite tree).


threeplanetssoftware:
This should progress #414 for the following items:

- stories with text and links
- saved stories
- moderations of user (ban, unban, disable + enable invites, grant moderatship)
- moderations of hats (add, delete)

threeplanetssoftware:
After the last commit, Fixes #414.

lib/tasks/fake_data.rake
```diff
@@ -17,15 +18,29 @@ class FakeDataGenerator
...
+        invited_by_user: User.select(&:can_invite?).sample,
       }
-      create_args.merge!(is_moderator: true) if i % 7 == 0
+      create_args.merge!(is_admin: true) if i % 8 == 0
       users << User.create!(create_args)
+      if i % 7 == 0
+        users[i].grant_moderatorship_by_user!(mod)
+      end
+      if i % 3 == 0
+        users[i].enable_invite_by_user!(mod)
```
pushcx, line 32: What effect does this have? By default the new users all have the ability to invite (well, after `NEW_USER_DAYS`). I think this would only affect those with invites disabled, which you do a few lines down?

lib/tasks/fake_data.rake
```diff
@@ -62,6 +80,13 @@ class FakeDataGenerator
         editor: user
     end
 
+    # User-saved stories
+    (@stories_count / 10).times do
+      user = users[Random.rand(@users_count-1)]
+      story = Story.all.sample
```
pushcx, line 86: Why not stick this in a `@stories` like `@categories`? I don't have a strong feeling on trading off a little memory vs speed in a one-off script like this, I'm just distracted by the inconsistency.

pushcx:
I left a couple small questions, and I have one more, now that you've spent some time with it. Do you think it's worth bothering with `users_count, stories_count, categories_count` in the initializer? We have a lot of hardcoded values in here, it doesn't seem like there's much reason to parameterize any of them.

lib/tasks/fake_data.rake
threeplanetssoftware: Honestly, I was just trying to interact more with the database to learn that better. No good reason to do that, I'll shove them into arrays.

lib/tasks/fake_data.rake
threeplanetssoftware: I did this because the accounts that are created should still appear as new since they won't be NEW_USER_DAYS old yet. Therefore `invited_by_user: User.select(&:can_invite?).sample` will would be none of the new accounts and I wanted to have a more realistic looking user tree. 

I was disabling some accounts then to generate good moderation data.

threeplanetssoftware:
> I left a couple small questions, and I have one more, now that you've spent some time with it. Do you think it's worth bothering with `users_count, stories_count, categories_count` in the initializer? We have a lot of hardcoded values in here, it doesn't seem like there's much reason to parameterize any of them.

When I was working through this, I debated taking those out. If nothing else, I think it makes sense to move the hard coded values up to the top as defaults for the initializer. That will make them more obvious to find and they can be overridden if someone wants, but otherwise aren't needed.

lib/tasks/fake_data.rake
threeplanetssoftware: As expected, you were right and this wasn't needed. :smile:

pushcx:
I am working through the PRs forgotten during the pandemic. I've worked out the merge conflicts on this and am glad to commit it.

Thanks for these many improvements you've made.

PR #925, opened 2020-08-28
when there are no votes for a comment, score is 0 not NULL.

alanpost:
A null score, present when there are no votes for a comment,
causes an ActiveRecord::NotNullViolation.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -318,7 +318,7 @@ class Comment < ApplicationRecord
     self.flags += flag_delta
     Comment.connection.execute <<~SQL
       UPDATE comments SET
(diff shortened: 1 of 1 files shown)


alanpost:
It is not clear to me this is the correct fix for this condition, it could be that we actually expect in all cases there to be at least one entry in the vote table when update_score_and_recalculate! is called.

If that's true, the bug here is elsewhere. (We had an exception generated today related to this function) If it is possible, however, this fix prevents setting score to NULL in violation of the database schema.

pushcx:
This is a correct fix. This is a very uncommon state because people rarely unvote their own comments. Thanks for solving this.

PR #928, opened 2020-09-05
Update message reply form to only include wearable hats. 

### LINKED ISSUE #927: Doffed hats are offered when composing or replying to private messages
Though sending with a doffed hat returns an error as expected.

hat

threeplanetssoftware:
Fixes #927 by only allowing wearable hats to populate the message reply form. Copied the syntax from the comment reply box to ensure consistency.

### DIFF
--- a/app/views/messages/_form.html.erb
+++ b/app/views/messages/_form.html.erb
@@ -24,9 +24,9 @@
   <% if @user.wearable_hats.any? %>
     <div class="boxline">
       <%= f.label :hat_id, 'Put on hat:' %>
-      <%= f.select :hat_id, options_for_select( [['','']] +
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for jumping on this bug, and I like the cleanup here. But this loses the `'data-modnote' => h.modlog_use` that informs the frontend js to [check the box to make a mod note](https://github.com/lobsters/lobsters/blob/master/app/assets/javascripts/application.js.erb#L423-L426). This is a pretty dang useful quality-of-life feature for mods. :)

threeplanetssoftware:
So I looked at that little guy and just wasn't sure why it was present. This makes a lot more sense and I have gone back to a simpler fix which only changes the call from 'hats' to 'wearable_hats'. My apologies for the potential detriment to quality-of-life. :)

pushcx:
Looks perfect, thank you!

PR #939, opened 2020-10-21
refact flagged commenters model

gabrielbaldao:
This PR is just to let the DownvotedCommenters clean, creting the query object and remove all sql query from model


pushcx:
Thanks, but I don't see a benefit here. There's low churn on this code, it sees little usage, so it might as well just be written in a single coherent place.

PR #940, opened 2020-10-22
Add X-Is-Author to mailing list headers

GarbageHamburger:
Currently, there is no way to tell whether a story has been authored by a submitter or not from the e-mail that is received when one subscribes to the mailing list. This information can be useful for

1. Marking a story as authored by the submitter in things like mockturtle;
2. Filtering out authored articles at the user's preference.

The `X-Is-Author` header will be `true` when the submitter of the story checks the "Authored" box, and `false` otherwise.


pushcx:
I can see this being handy for anyone building on top of these emails, and it's very cheap to maintain. Thanks for thinking of it.

PR #944, opened 2020-11-06
Allow gemini and gopher homepage URLs

illfygli:
I thought it would be nice to allow Gemini spaces and Gopher holes in the homepage field.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,7 +83,9 @@ class User < ApplicationRecord
             :uniqueness => { :case_sensitive => false }
 
   validates :homepage,
-            :format => { :with => /\Ahttps?:\/\/[^\/\s]+\.[^.\/\s]+(\/.*)?\Z/ },
+            :format => {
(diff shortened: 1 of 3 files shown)


KangOl:
It would be great to have a link to a http-proxy version of theses links.

And maybe also extract canonical link when a proxyed url is submitted (like https://lobste.rs/s/jnwzms/my_first_freebsd_port_castor_gemini)

alternet0:
LGTM

pushcx:
@illfygli Thanks for suggesting this, seems fun.

I agree with @KangOl's suggestion. Could you add a `(<a href="">proxy</a>)` if the protocol is `gemini` or `gopher`? I know very few users will have their browsers set up for those protocols (or want to immediately do so for an unknown link).

illfygli:
I gave it a try. I'm very bad at this. :)

pushcx:
Nope, I have to disagree, you've done nicely and I'm merging this.

PR #949, opened 2020-11-27
Update gitignore to deal with droppings from Vim and RubyMine

NattyNarwhal:


pushcx:
For config that's specific to your personal development env, you can add this to `~/.gitignore` so it'll be ignored from all projects you contribute to.

Or tweak vim: you can `:set backupdir=~/.vim/backup//` to keep them out of all projects.

PR #955, opened 2020-12-29
Find socket file location

Charleshthomasiii:
My socket file was in a different location than /tmp so I added instructions on how to find it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ running tests:
       password: *test_password*
     ```
 
+* Your socket file may be in a different location than '/tmp/mysql.sock'. You can find it with:
+```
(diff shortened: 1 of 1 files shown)


arp242:
`mysqladmin variables | grep ' socket' | tr -s ' '` would be a better command as it won't include some unrelated stuff.

But to be honest I'm not sure if the README needs to be a "MySQL setup tutorial". You can also have the socket disabled and listen only on the network.

pushcx:
@Charleshthomasiii I see the value here, but various distributions can set up mysql in quite different ways that we can't expect to keep up with them. I'm happy to provide rough instructions, but not invite support requests.

PR #956, opened 2020-12-29
Add null check for title match

Charleshthomasiii:
Replicating the error: 
On the submit a story page, type in a title then click on the text box. A TypeError appears in the console because .some is read on the null value returned by  ```title.match(/: | - | – | — | \| | · | • | by /) || title.match(/\([^\)]*\)/g)```. This fix adds a null check before reading .some on the value.

clicking_away
error

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -273,9 +273,13 @@ var _Lobsters = Class.extend({
 
     // common separators or (parens) that don't enclose a 4-digit year
     if (title.match(/: | - | – | — | \| | · | • | by /) ||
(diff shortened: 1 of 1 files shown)


pushcx:
Hey, thanks - this is a great way to report the bug.

I'm not getting how `|| ... != null` predicate isn't a no-op, if `match` returns `null` that would already falsify the `if`. But maybe that question is pointing to a tidier approach that wouldn't require duplicating the two regexps: `(title.match(/\([^\)]*\)/g) || []).some`

pushcx:
This bug disappeared in the big js refactor last year.

PR #957, opened 2021-01-02
Update Rails version description

frenkel:


pushcx:
Thanks for spotting this and offering the small update. I think this is going to be like the Ruby version where we never remember to update it, so we update the readme to point to the `.ruby-version` file that sets it. What do you think about dropping the `6` here?

frenkel:
Sounds like a more future proof plan :-)

frenkel:
How could I obtain an invite to register on lobste.rs by the way? ☺️

pushcx:
Contributing to the codebase is a pretty good way. :D I'll need your email address to send it to - you could dm me on twitter (same username) or email me (my first name @push.cx).

PR #963, opened 2021-01-19
Remember collapsed threads

arp242:
This change remembers collapsed threads, similar to HN. Unlike HN, it
doesn't store this on the server but in the browser's localStorage. This
is done for both logged in and logged out users.

Alternatively, we could use sessionStorage so it's less persistent. Or
add a "remember collapsed threads" setting with the options
"persistent", "session", and "none". I'm not sure if there are any
people who would consider this feature undesirable?

Tested locally in Firefox and Chromium and seems to work well.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -600,7 +600,7 @@ $(document).ready(function() {
         storyID  = $('.story').attr('data-shortid'),
         collapse = JSON.parse(localStorage.getItem(storyID) || '{}');
     if (e.target.checked)
-      collapse[id] = true;
+      collapse[id] = 1;
     else
       delete(collapse[id]);
     localStorage.setItem(storyID, JSON.stringify(collapse));
@@ -612,7 +612,7 @@ $(document).ready(function() {
         collapse = JSON.parse(localStorage.getItem(storyID) || '{}');
 
     for (var k in collapse)
-      $('input#comment_folder_' + k).attr('checked', (collapse[k]))
+      $('input#comment_folder_' + k).attr('checked', !!collapse[k])
   });
 
   $('textarea#comment').keydown(function (e) {
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -596,7 +596,7 @@ $(document).ready(function() {
 
   // Remember story collapses; this is stored in localStorage for every story ID as an object.
   $(document).on('change', '.comment_folder_button', function(e) {
-    var id       = $(e.target).closest('.comments_subtree').find('[data-shortid]:first').attr('data-shortid'),
+    var id       = $(e.target).find('~ div.comment').attr('data-shortid'),
         storyID  = $('.story').attr('data-shortid'),
         collapse = JSON.parse(localStorage.getItem(storyID) || '{}');
     collapse[id] = e.target.checked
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -590,6 +590,24 @@ $(document).ready(function() {
       this.value = 'http://' + this.value;
   });
 
+  // Remember story collapses; this is stored in localStorage for every story ID as an object.
+  $(document).on('change', '.comment_folder_button', function(e) {
+    var id       = $(e.target).closest('.comments_subtree').find('[data-shortid]:first').attr('data-shortid'),
```
pushcx, line 595: Could the `find` here be replaced by looking for `.closest('.comment')`?

pushcx (changes requested):
This looks like a convenient feature, thanks for writing it. Though: it won't work for logged-out users, because we don't serve them this (or any) js.

One concern: the data structure is going to grow until it hits the LocalStorage limit, then probably silently fail to write, so the feature will appear to be permanently broken to its most active users.

Let's reduce that likelihood by not storing comments that are not collapsed (as `collapse[id] = e.target.checked` does), and adding some kind of expiration. I can imagine a couple approaches: we could bin by year at the top level, or store a timestamp of when the story was last viewed and prune when writing, or maybe something else?

app/assets/javascripts/application.js.erb
arp242: The way the HTML is currently layed out, no:

```
<li class="comments_subtree">
	<input id="comment_folder_inhoqk" class="comment_folder_button" type="checkbox">
	<div id="c_inhoqk" data-shortid="inhoqk" class="comment">
	    <label for="comment_folder_inhoqk" class="comment_folder">[-]</label>

	    <div class="details">
		    [..comment body and other buttons..]
	    </div>
</li>
```

`e.target` is that `input`, but I see no reason that can't be moved next to the `label` which would indeed simplify this a bit. I'll try and see if there are any problems with that.

arp242: Actually, moving that won't work due to how the hiding is implemented: it needs to be before the `<div class="comment">` in order to hide that in CSS (no parent selector in CSS).

But it can be simplified to `find('~ div.comment')`, which matches what the CSS does.

arp242:
> One concern: the data structure is going to grow until it hits the LocalStorage limit, then probably silently fail to write, so the feature will appear to be permanently broken to its most active users.

I changed it to not store the expanded threads, and it's stored as an int now (instead of bool).

It's now 12 bytes per collapsed thread, so that'll be 120K for 10k collapsed threads; the limit seems to be around 10M in a quick search so won't be hit until about 850k threads. If you're collapsing 20 threads/day then that's 7.3k/year, and you won't hit the limit for over 100 years.

Aside that most people don't live that long, in reality, people tend to switch browsers, buy a new laptop, etc. which causes this to be cleared "naturally" anyway.

I could add it, but I'm not sure this is really something we need to worry about?

pushcx:
Ah, that's a lot higher than I remember it being. I agree that expiring isn't needed, then. I'll try to experiment with this a bit in the next day or two so I can merge it confidently.

pushcx:
I merged this in 0037cf47 which required substantial rewriting for the big js refactor from last year.

Thanks for writing this feature, it's a very nice quality-of-life improvement.

PR #964, opened 2021-01-19
Clear reply textarea after posting

arp242:
1. Write your comment, post it.
2. Press F5
3. The reply field now contains your comment, since it was submitted and
   hidden via JS, and Firefox will still re-populate it after reloading.

This only happens in Firefox; Chrome doesn't have this behaviour.


pushcx:
Ah, I see the annoyance there, if you're reloading to look for replies. This doesn't seem like it'll introduce a risk that someone loses a comment because it doesn't fire until the server has acknowledged it. Thanks for a nice solution.

Issue #954, opened 2020-12-17
Recognize arxiv url dupes


pushcx:
https://lobste.rs/s/zm1xgo/air_fi_generating_covert_wi_fi_signals

We pretty regularly get people submitting both the arxiv page and the direct link to the pdf, let's recognize them as dupes. This patch will be in `Story.find_similar_by_url`.

prabuw:
Hello - long time fan of the site. 

@pushcx I am happy to pick this issue up and contribute a PR?

pushcx:
You don't need permission, please go right ahead. :)

prabuw:
Cheers - PR raised. 👍🏽

pushcx:
https://lobste.rs/s/kcyg0p/gwp_asan_sampling_based_detection_memory

This story had an alternate arxiv URL merged in. I don't know if it's a new URL structure on their part or a bug on ours that we didn't identify it, but it's the first dupe I'm aware of in the 2.5y since this was merged.

fanf2:
There’s a v2 at the end of the URL that isn’t matched by the arxiv dup detector

pushcx:
Yep, that was it, good eye.

PR #965, opened 2021-01-19
deduplicate arvix page and pdf urls #954

prabuw:
Arvix pages are offered in the format of an HTML page and pdf.
This commit deduplicates them if one of them has already been
submitted as a story.


pushcx:
Thanks for taking this on. Two things:

1. Could you only fire this if the `domain` is `arxiv.org`? No need to check this for any other URLs.
2. Could you add a test, please? There's already a couple in `spec/models/story_spec.rb` you could use as examples.

prabuw:
@pushcx Updated the PR with the check and tests. Let me know if the check is sufficient and if the tests are adequate. 😄

prabuw:
@pushcx I have fetched the latest changes from the master branch and it should be passing all checks now. Can you please have a look at this PR at your convenience? 🙂

Issue #967, opened 2021-01-25
Reduce comment double-posting


pushcx:
Our current database setup is occasionally significantly slower than previous. We've seen a rash of double-posted comments because it looks like nothing's happened.

We're working on replacing the slow database, but in any case, we should prevent this. Off the top of my head, this probably looks like updating `CommentsController.create` to look for a comment from the author in the last ten minutes with identical text. If one is found, silently skip creating a comment and render the existing one. There's a `spec/requests/comments_spec.rb` that a test for this behavior can live in.

meaagan:
I believe this is already done on line 38 of comments_controller.rb.

pushcx:
Unfortunately, no. That code's been present a long time and hasn't fired here. I'd guess it's the `params[:preview].blank?` causing it to get skipped. I don't understand why that clause is there but want to think about it and research a bit before I remove it.

And even if that is changed, it's worth having @arp242's nicer UI design to make it clear that something is happening while waiting for a comment to post.

PR #968, opened 2021-01-25
Disable "Post" button after click

arp242:
This disables the "Post" button after clicking so you can't accidentally
submit it twice.

I also added this to the "reply" button; this loads the HTML from the
server and I've had a few cases where this took a while, and I wasn't
sure if I had clicked it correct so I clicked again and ended up with
two reply fields. There may be other places where this could be added as
well.

It also adds a small loading animation for a bit of feedback.

I also added .fail() for the comment post, since there was no error
before at all.

Partial fix for #967, but it won't cover no-js users.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -205,35 +205,50 @@ var _Lobsters = Class.extend({
       action, { reason: reason });
   },
 
-  postComment: function(form) {
+  postComment: function(btn, form) {
     var params = $(form).serializeArray();
     params.push({"name": "show_tree_lines", "value": "true"});
 
-    $.post($(form).attr("action"), params, function(data) {
-      // Clear form: Firefox will keep form values on reload (e.g. F5), which isn't too useful if
-      // it's already posted.
-      $(form).find('textarea').val('')
+    var done = Lobsters.loading(btn, () => {
+      $.post($(form).attr("action"), params, function(data) {
+        // Clear form: Firefox will keep form values on reload (e.g. F5), which isn't too useful if
+        // it's already posted.
+        $(form).find('textarea').val('')
 
-      var parsedHTML = $.parseHTML(data);
+        var parsedHTML = $.parseHTML(data);
 
-      var isEdit = $(form).hasClass("edit_comment");
-      var hasChildren = $(form).closest(".comments_subtree")
-        .find(".comments_subtree").length;
-      if (isEdit && hasChildren) {
-        $(parsedHTML).find(".comment_parent_tree_line:first")
-          .removeClass("no_children");
-      }
+        var isEdit = $(form).hasClass("edit_comment");
+        var hasChildren = $(form).closest(".comments_subtree")
+          .find(".comments_subtree").length;
+        if (isEdit && hasChildren) {
+          $(parsedHTML).find(".comment_parent_tree_line:first")
+            .removeClass("no_children");
+        }
 
-      if ($(form).find("#parent_comment_short_id").length) {
-        // reply to comment
-        $(form).closest(".comments_subtree")
-          .find(".comment_parent_tree_line:first").removeClass("no_children");
-        $(form).closest(".comment").replaceWith(parsedHTML);
-      } else {
-        // reply to story
-        $(form).parent(".comment").replaceWith(parsedHTML);
-      }
-    });
+        if ($(form).find("#parent_comment_short_id").length) {
+          // reply to comment
+          $(form).closest(".comments_subtree")
+            .find(".comment_parent_tree_line:first").removeClass("no_children");
+          $(form).closest(".comment").replaceWith(parsedHTML);
+        } else {
+          // reply to story
+          $(form).parent(".comment").replaceWith(parsedHTML);
+        }
+      })
+        .always(() => done())
+        .fail(function(req, err) { alert(`could not post comment: ${err}`) })
+    })
+  },
+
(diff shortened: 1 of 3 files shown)


arp242:
Easiest to see the diff by disabling whitspace: https://github.com/lobsters/lobsters/pull/968/files?diff=unified&w=1

And here's what it looks like, the first is with an artificial `sleep 1` in the controller, and the second is without that.


https://user-images.githubusercontent.com/1032692/105657212-d6993100-5efe-11eb-8ebd-2a5071451384.mp4


https://user-images.githubusercontent.com/1032692/105657215-d862f480-5efe-11eb-943d-30f079d010ae.mp4

app/assets/stylesheets/application.css
```diff
@@ -1287,3 +1287,17 @@ div.flash-success h2 {
 	text-decoration: none;
 	vertical-align: middle;
 }
+
+
+/* Loading indicator */
+@keyframes loading {
```
pushcx, line 1293: This is a really nice lightweight technique!

app/views/comments/_commentbox.html.erb
```diff
@@ -31,8 +31,8 @@ data-shortid="<%= comment.short_id if comment.persisted? %>">
         </div>
       <% end %>
 
-      <%= f.submit "#{comment.new_record?? "Post" : "Update"}",
-        :class => "comment-post", :disabled => !@user %>
+      <%= f.button "#{comment.new_record?? "Post" : "Update"}",
```
pushcx, line 34: Why change this? I thought the rails method here is going to make an `<input type="submit">`, right?

app/assets/javascripts/application.js.erb
```diff
@@ -205,35 +205,50 @@ var _Lobsters = Class.extend({
...
+        }
+      })
+        .always(() => done())
+        .fail(function(req, err) { alert(`could not post comment: ${err}`) })
+    })
+  },
+
+  // Prevent a button/link from working while an AJAX request is in progress;
+  // otherwise smashing a "post comment" button will post a comment twice.
+  loading: function(btn, f) {
```
pushcx, line 245: Why require the button as an argument? There's only one submit button in the form and it's very unlikely there'll be a second. Could you just select it out of the form like `$(f).querySelector('button[type="submit"]')`?

app/assets/javascripts/application.js.erb
```diff
@@ -457,15 +472,17 @@ $(document).ready(function() {
...
-    $.get("/comments/" + comment.attr("data-shortid") + "/reply",
-    function(data) {
-      var reply = $($.parseHTML(data));
-      reply.attr("id", "reply_form_" + comment.attr("id"));
-      replies.prepend(reply);
-      var ta = reply.find("textarea");
-      ta.focus().text(sel);
-      autosize(ta);
-    });
+    var done = Lobsters.loading($(this), () => {
```
pushcx, line 475: The arrow function here and on line 251 are our first usage. So my old fart question is how universal browser support is for them now? We don't have any preprocessing step to transpile down, and I really don't want one.

app/assets/javascripts/application.js.erb
```diff
@@ -457,15 +472,17 @@ $(document).ready(function() {
...
+      $.get("/comments/" + comment.attr("data-shortid") + "/reply", function(data) {
```
pushcx, line 476: Instead of `.attr("data-shortid")`, the jquery idiom is `.data("shortid")`.

app/assets/javascripts/application.js.erb
```diff
@@ -457,15 +472,17 @@ $(document).ready(function() {
...
+        done()
```
pushcx, line 477: I'm confused by names here - is this calling the `done` defined on line 212 or 475? Maybe let's give them both more descriptive names like `commentFormSubmitted` to match their behavior.

app/assets/javascripts/application.js.erb
```diff
@@ -205,35 +205,50 @@ var _Lobsters = Class.extend({
...
+          $(form).closest(".comments_subtree")
+            .find(".comment_parent_tree_line:first").removeClass("no_children");
+          $(form).closest(".comment").replaceWith(parsedHTML);
+        } else {
+          // reply to story
+          $(form).parent(".comment").replaceWith(parsedHTML);
+        }
+      })
+        .always(() => done())
+        .fail(function(req, err) { alert(`could not post comment: ${err}`) })
```
pushcx, line 239: Rather than add a modal `alert`, please add like `<div class="error"></div>` to the form and dump the error in there. And have the `submit` button clear that.

app/assets/javascripts/application.js.erb
arp242: I mostly just copied this approach from another application I wrote, where it's more useful. I guess it could be removed here, but to be honest I'd be in favour of just keeping it, as it makes it easier to re-use just in case.

For example, it could be added to e.g. the "reply" button too (I noticed that can sometimes take a second or two for the reply form to load, and a bit of feedback wouldn't hurt, and with this it should be very easy to add), or maybe the "preview" button.

pushcx:
After another spare of duplicate comments, I took a server-side approach in #967 to recognize and silently accept them.

arp242:
It looks like the comments I posted a few months ago are still in "pending" and never posted the review; do'h >_<

Ah well, nice little improvement; thanks!

pushcx:
Ouch, I've done that one before, too. In any case, thanks for taking a crack at this. I'm sorry it sat so long.

Issue #946, opened 2020-11-20
Can't delete messages


pushcx:
When a story is edited from user suggestions, the submitter gets a `Message` notifying them from the "System" account. In the database this works by setting `author_user_id` to `0`.

It's not possible for the recipient to delete this message. They'll get a 500 because the soft deletion throws `ActiveRecord::RecordInvalid "Validation failed: Author must exist"`.

If someone wants to pick this up, you don't need to test all of the setup, just a model test that a message from "System" can have `deleted_by_recipient` set to `true`, as happens in `MessagesController#destroy`.

dracconi:
I have addressed this in #970.

PR #970, opened 2021-02-11
Allow for deletion of System messages (#946)

dracconi:
Makes the `author` of a message optional and delegates the check to a special
function (`is_authored?`) in the `message` model. It checks whether the author
exists, or, if the message is deleted and the author's id is 0.
Includes the tests for that behaviour.

Side note: Haven't tested if it runs.


dracconi:
Eh forgot a colon and broke that... Sorry... I'll resubmit..

dracconi:
After long hours of CI based development, maybe it will _finally_ pass.

pushcx:
Thanks for fixing this bug. I've fixed the conflicts introduced by standardrb and am merging.

pushcx:
I spent a minute tinkering with these test failures and... uh, the code works without it. I'm not sure what changed here to fix it. I'm sorry your PR was wasted.

PR #976, opened 2021-02-25
Deterrent against off topic conversations.

ibsusu:
I like hearing what people have to say when more when they're not talking about moderation.  Having to tell people to not talk about what you're moderating is a pain because it should usually be a given.

大家说的太多不相关的事情。

This code hasn't been tested because I don't know ruby but I hope you can use it.


pushcx:
Thanks for writing this feature, but this hasn't been a significant problem for the site.

PR #979, opened 2021-03-15
ISS-809: merging with discrepancy in ownership

emmahyde:


emmahyde:
Do we care what happens when both users "claim" to be the author? Or is that a non-event? If it is something we're concerned about, I could write a fourth test, i.e. who becomes author when both are listed as author, although as a first time contributor I probably do not have enough domain knowledge to know what the desirable behavior there is.

app/helpers/traffic_helper.rb
```diff
@@ -6,10 +6,16 @@ module TrafficHelper
   PERIOD_LENGTH = 15 # minutes
   CACHE_FOR = 5 # minutes
 
+  # TODO: ISS-809 – temporary measure for migration performance issue.
```
thnukid, line 9: Just a head up, [rubocop fails](https://travis-ci.org/github/lobsters/lobsters/builds/763025683#L529) in CI
```
Offenses:
app/helpers/traffic_helper.rb:9:19: C: Style/AsciiComments: Use only ascii symbols in comments.

  # TODO: ISS-809 – temporary measure for migration performance issue.
                  ^
176 files inspected, 1 offense detected
The command "bundle exec rubocop" exited with 1.
```

app/helpers/traffic_helper.rb
```diff
@@ -6,10 +6,16 @@ module TrafficHelper
...
+  # transactions (cache only) and returns hard coded numbers (:13, :49).
+  DISABLED = true
+
   def self.traffic_range
-    return [0, 2]
+    return [0, 2] if DISABLED
+
     div = PERIOD_LENGTH * 60
-    start_at = 90.days.ago
+    start_at = 90.days.agor
```
thnukid, line 18: Is this a typo?

`agor` => `ago`

spec/models/story_spec.rb
```diff
@@ -337,6 +337,53 @@ def fake_response(content, type, code = '200')
...
+        ).tap(&:save)
+      end
+      let!(:mgd_story) do
+        create(
+          :story,
+          user: mgd_user,
+          title: 'A Tale of Merging',
+          merged_into_story: og_story,
+          user_is_author: false
+        ).tap(&:save)
```
thnukid, line 358: Do you need to call on `.tap(&:save)`?

FactoryBot `create` returns a `saved` instance

```
# Returns a saved User instance
user = create(:user)
```

https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#build-strategies

thnukid:
@emmahyde overall a nice pull request 👏 

I am missing domain knowledge, therefore I have just disabled the rubocop rules that where breaking the CI, and when merged would automatically make the CI green for the other open pull requests.

> even though we know this will be 0 processes.

Indeed the early `return` statements hint at that the class is disabled / not used. I think I have checked the method names and only one method was used, so basically the file could also be removed / commented out (technical debt). 

On a side note: Maybe rename `DISABLED` to `FEATURE_DISABLED`, because I think this is a feature flag, and not some state for instance of the class. What do you think ? :)

emmahyde:
@thnukid, sorry for the late reply, just saw your comments. I will fix the rubocop failures.  And yes, a dumb typo on that `agor` :) will fix in upcoming PR. 
I will double check the factory bot thing, I think I was mixing up a model.create (which does not save) with a factorybot.create (which seems to), so I should be able to remove those taps.
Will change the string to `FEATURE_DISABLED`, I agree that its the better name! It looks like these methods were handicapped when the app migrated to DigitalOcean (circa [this commit](https://github.com/lobsters/lobsters/commit/9c0e2c03e2c14a2fba39298e76258cc460f6f003)), so I am hesitant to remove them entirely, as it seems like there is some intent to restore them eventually. The flag might be OK for now, unless there is absolutely no intention to get them working again?

pushcx:
I'm sorry I didn't get to this for so long, and that the bug hadn't been properly closed by the time you looked at it. I appreciate you attempting a fix for this issue.

PR #981, opened 2021-03-20
fix rake fake_data

dorianmariefr:
was broken because user accounts were not old enough to post new urls

also simplify code a bit

👋


pushcx:
Thanks for fixing this!

dorianmariefr:
Thanks for merging

Issue #938, opened 2020-10-17
Domain could encompass user part on sites with user part in location


NattyNarwhal:
HN has a feature where their equivalent of the domain feature actually encompasses the actual location part *if* the initial leading part is relevant to scoping, like a user. For example, for a Twitter link:

![Screenshot from 2020-10-17 01-10-25](https://user-images.githubusercontent.com/3161292/96328075-aac00280-1015-11eb-89aa-4180a6542fde.png)

Of course, we shouldn't encourage Twitter links, but being able to do this for say, common sites with ~userdirs or Medium links (well, Medium is greasy at times...), anything where the user part isn't in the domain, but location.

pushcx:
Implemented in 34981c2c692b1c4ea59175e34299171e2d08410d.

PR #982, opened 2021-03-20
add path to twitter and github domains

dorianmariefr:
Screenshot 2021-03-20 at 01 47 31


NattyNarwhal:
Should fix #938. I like this!

PR #987, opened 2021-04-17
Create test data just oustide the new user window

rtpg:
When we set our fake data to `NEW_USER_DAYS.days.ago` then we end up hitting the `is_new?` validation check during fake data generation (especially given that the randomly generated domains tend to hit the error), preventing stories and the like from being created.  Adding an extra day lets these accounts create stories without issue. 

I'm not 100% if this is the right way to resolve this issue, or if I'm doing something wrong, but this feels pretty harmless to me

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -19,7 +19,7 @@ User.create(
     User::MIN_KARMA_TO_SUBMIT_STORIES,
     User::MIN_KARMA_FOR_INVITATION_REQUESTS
   ].max,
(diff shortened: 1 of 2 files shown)


rtpg:
ah seems like this is a bit of a dupe of #981

pushcx:
Thanks for taking a look at this - do you think there's anything left for this PR to do now that I've merged 981?

rtpg:
ah, nope, I'll clsoe this one, and open a new thing if I find an issue again. Thank you!

PR #988, opened 2021-04-19
Add a maxlength to memo field

ysangkok:
So that user doesn't write a long text only to be discarded.


ysangkok:
Closing since it clobbers my PR listing page and I don't think anybody cares about it.

PR #1001, opened 2021-05-23
Configure Rack::Attack to block misbehaving clients

hmadison:
Port the behavior of [this pull request](https://github.com/lobsters/lobsters-ansible/pull/56) over to Rails using [Rack::Attack](https://github.com/rack/rack-attack#installing).


pushcx:
Thanks for finding this and getting it started. I expanded the config a bit and merged this as b8d91ca3.

PR #1007, opened 2021-06-02
Fix: Case bug viewing profile/standing page 

### LINKED ISSUE #1005: Case bug viewing profile/standing page
User names must be unique, ignoring case. UsersController#show (the profile view) works whether or not capitals are used, so if you add `/standing` to get to your UsersController#standing action, it fails the `before_action` that the currently logged-in user is viewing their own page.

The quick fix is to `.lowercase` both sides of this `if`. The thorough fix would be to check case in routes and 301 if it doesn't match the user's username.

rbrishabh:
Fixes #1005 
Hi @pushcx, does this fix work?
Thanks!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -159,7 +159,7 @@ class UsersController < ApplicationController
 private
 
   def only_user_or_moderator
-    if params[:username] == @user.username || @user.is_moderator?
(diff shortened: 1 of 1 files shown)


pushcx:
Yep, thanks for contributing. I'm glad to have the immediate issue addressed.

PR #1008, opened 2021-06-03
Fix wording on IRC nudge

ergl:
The You -> you change is needed since the sentence only starts anew if we allow public invitation requests.


pushcx:
Thanks for catching this bug, merging.

PR #1010, opened 2021-06-13
Fixes involuntary user deletion and comment disown. Closes #1004

### LINKED ISSUE #1004: User deletion disowned comments when the checkbox wasn't set
A user deleted their account [without checking 'disown'](https://lobste.rs/s/zp4ofg/lobster_burntsushi_has_left_site#c_ot2yb7) and all their comments were disowned anyways. My guess without reading the code is that `0` is incorrectly coerced to a truthy value, given we're probably reading this value without using Rails's code for assigning form inputs to model attributes.

briankung:
- [x] Let's add some tests
- [x] fix this by checking for 0 or 1 in the controller
- [x] audit the entire commit to see if we picked up other bugs due to this gotcha

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -10,7 +10,7 @@ class SettingsController < ApplicationController
   end
 
   def delete_account
-    unless params[:user][:i_am_sure].present?
+    unless params[:user][:i_am_sure] == '1'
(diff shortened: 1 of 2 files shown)


pushcx:
Thank you, this is a thorough and clean solution for a very bad bug.

PR #1015, opened 2021-07-11
Correct mismatched username case with a redirect

### LINKED ISSUE #1005: Case bug viewing profile/standing page

kurige:
Fixes #1005

Continuation of https://github.com/lobsters/lobsters/pull/1007, implementing the redirect so that:
* `/u/uSeRnAmE` redirects to `/u/username`
* `/u/uSeRnAmE/standing` redirects to `/u/username/standing`

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,4 +1,5 @@
 class UsersController < ApplicationController
+  before_action :redirect_case_mismatch, :only => [:show, :standing]
   before_action :require_logged_in_moderator,
                 :only => [:enable_invitation, :disable_invitation, :ban, :unban]
   before_action :flag_warning, only: [:show]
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -158,8 +159,24 @@ def standing
...
+        redirect_to(user_standing_path(target.username))
+      else
+        redirect_to(user_path(target.username))
+      end
+    end
+  end
+
   def only_user_or_moderator
-    if params[:username].downcase == @user.username.downcase || @user.is_moderator?
+    if params[:username] == @user.username || @user.is_moderator?
```
pushcx, line 179: Nice catch!

app/controllers/users_controller.rb
```diff
@@ -158,8 +159,24 @@ def standing
 
 private
 
+  # If viewing user at "/u/tEsT" but username is actually "test", then redirect to "/u/test".
+  def redirect_case_mismatch
+    target = User.where(username: params[:username]).first
```
pushcx, line 164: This and `@showing_user = User.where(:username => params[:username]).first` up on line 10 are identical, so we're duplicating the query. Can you update the PR to avoid querying the same record twice?

pushcx:
I'm working down the list of PRs - looks like this just needs a small change to be ready to merge?

pushcx:
Thanks for taking a pass at this. I reused your test in a250c2513c6fe969851e4b615f6ac30ccb493067 to close this out.

PR #1017, opened 2021-08-16
apps/views/users: remove the "naughty box" that users can brigade

Xe:
As the site code works currently, if a user dislikes another user for
some reason they can go around and flag all of their comments and give
the flagged user the "naughty box"[1] for the next month. This will
leave the flagged user mystified as to what they did wrong which can
make them feel gaslit. This "naughty box" has affected other users in
the past (see [2] for a good example).

This does not disable the code that triggers the logic to display the
"naughty box", it merely disables the box itself. I do not feel
comfortable enough with Ruby and Rails to do more than this.

[1]: https://i.imgur.com/uYV1Yp7.png
[2]: https://lobste.rs/s/zp4ofg/lobster_burntsushi_has_left_site

Signed-off-by: Christine Dodrill <me@christine.website>

### DIFF
--- a/app/views/comments/threads.html.erb
+++ b/app/views/comments/threads.html.erb
@@ -1,5 +1,3 @@
-<%= render partial: 'users/flag_warning' if @show_flag_warning && @showing_user == @user %>
-
 <% @threads.each do |thread| %>
 <ol class="comments comments1">
   <% comments_by_parent = thread.group_by(&:parent_comment_id) %>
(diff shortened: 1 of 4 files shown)


pushcx:
I'd take a PR to improve the query to avoid this scenario, or add admin dashboards to highlight this kind of behavior, but the warning has been such a strong positive for the site that disabling it is a non-starter.

PR #1020, opened 2021-09-14
Remove CSS references to "flagger" within "voters"

kconner:
Markup matching those selectors appears not to be emitted since downvote arrows were removed.

"voters" is this whole element:
Screen Shot 2021-09-13 at 11 06 11 PM

"upvoter" is the up arrow. There was once a downvote arrow as well, named "downvoter". It was removed in 9535b05, but its CSS is still here. The term "downvote" appears to have been replaced with "flag" back in 20c1590, so the CSS more recently referred to a "flagger".

This is a side issue discovered while working on #637.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -401,8 +401,7 @@ li.story div.voters div.score {
 	margin-top: 2px;
 }
 
-div.voters .upvoter,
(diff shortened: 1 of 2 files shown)


pushcx:
I really appreciate how thorough you've been in these PRs that you're spotting all this dead code, thank you.

PR #1023, opened 2021-09-14
Reduce unused CSS properties in flash banners

kconner:
Flash banners (of green success, red error, and blue notice types) have yellow default coloring that's always overridden, gradients we never see due to syntax errors, and some properties set twice in one class such that their first value is never used. This change reduces the bulk of that CSS while keeping appearance unchanged.

Before:
before

After:
after

…as captured by Safari Technology Preview web inspector.

This is a side issue discovered while working on #637.


pushcx:
Wow, that's quite a find! It's present in literally the first commit from 9 years ago and has never worked. Thanks for cleaning it up.

PR #1024, opened 2021-09-15
Highlight replies and user settings links in top nav

kconner:
Other top nav links are highlighted when their pages are displayed. These two may have been overlooked.

Replies, before:
replies-before

Replies, after:
replies-after

Settings, after:
settings-after


pushcx:
Thanks for catching this bug, merged.

PR #1025, opened 2021-09-16
Use meta tag style on about page meta tag

kconner:
This tag appears gray normally, but was yellow where specifically included in /about.


pushcx:
Nice little bit of consistency, thanks!

PR #1026, opened 2021-09-16
Dark mode

### LINKED ISSUE #637: Dark theme / night mod for PC

kconner:
Based on @soc's work in #823. Closes #637.

Dark mode is activated by the `prefers-color-scheme: dark` media query, usually when the operating system's UI is set to dark mode. An overriding user setting is out of scope for now.

I tried to minimize the number of distinct values that must be specified for each display mode. The tradeoff is that most foreground text is rendered with alpha blending via `rgba()`. That's about the only way you can get many apparent brightnesses of a base color without having more explicit colors per display mode or adding a dependency like Sass. I'm particularly interested in whether there is a performance impact on your browser.

As for palette choices, I followed my understanding of [dark mode guidance for macOS](https://developer.apple.com/videos/play/wwdc2018/210/?time=345). I went with pure black for the background in order to accommodate users whose eyes have already adjusted to a dark room and then get the most potential contrast per lumen. Foreground colors like blue links and lobster red are a little brighter and paler for contrast. Inset areas like form fields are a bit brighter than the background, which is also true of light mode. By contrast, while buttons darken on hover in light mode, they brighten in dark mode.

#### Samples

light dark

mobile-light mobile-dark

#### Effects on light mode

I did change a few colors in light mode because it made sense at the time:
- Blue background of notice flash banner made paler like error and success cases
- Paginating footer number buttons and markdown help box set to reuse colors from regular form fields and buttons
- Form field placeholders given a little bit more contrast, closer to textarea placeholder contrast

Finally, in pursuit of WCAG AA contrast, I increased contrast on all text I could find that had a contrast ratio beneath 4.5:1. (284d1da)

Before: footer-before vote-count-before

After: footer-after vote-count-after

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -318,21 +318,6 @@
   </p></li>
 
   <li><p>
-    If you prefer to read Lobsters on a dark background instead of a light one, there are a few options. You could install the
-    <a href="https://darkreader.org/">Dark Reader extension</a>;
-    or an extension to add your own CSS to sites like <a href="https://add0n.com/stylus.html">Stylus</a> or <a href="https://github.com/openstyles/stylus/wiki/Usercss">UserCSS</a> (<a href="https://en.wikipedia.org/wiki/Stylish#Privacy_issues">never Stylish</a>). Themes:
-    <a href="https://openusercss.org/theme/5cc5ad28ad27da0c0016cf3c">1</a>
-    <a href="https://openusercss.org/theme/5cc5ab1ead27da0c0016cf3a">2</a>
-    <a href="https://openusercss.org/theme/5cc5aeb3ad27da0c0016cf3d">3</a>
-    <a href="https://openusercss.org/theme/5cc5ac10ad27da0c0016cf3b">4</a>.
-    Discussion:
-    <a href="https://github.com/lobsters/lobsters/issues/404">1</a>
-    <a href="https://github.com/lobsters/lobsters/issues/637">2</a>
-    <a href="https://lobste.rs/s/om839j/dark_light_mode">3</a>
-    <a href="https://lobste.rs/s/iaajwb/dark_theme_for_lobsters">4</a>
-  </p></li>
-
-  <li><p>
     If you prefer the article links or comment links to open a new tab instead of navigating in your current tab you can add
     <a href="https://greasyfork.org/en/scripts/392307-lobste-rs-open-in-new-tab">a client-side script</a> to do so.
     (<a href="https://lobste.rs/s/0g48hl/open_links_articles_new_tab">Discussion</a>)
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -47,10 +47,10 @@
     --color-markdown_help-background: #f4f4f4;
     --color-markdown_help-border: #e0e0e0;
 
-    --color-downvote_why: #555555;
-    --color-downvote_why-background: #ffffff;
-    --color-downvote_why-background-hover: #eeeeee;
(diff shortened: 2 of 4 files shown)


kconner:
I've started to audit this with a color contrast tool, and I found several places in light mode that don't meet WCAG AA's 4.5:1 ratio minimum. Should my objective be to make changes to both light and dark mode? In that case would it be suitable to do that in a second pull request instead, assuming dark mode passes an eyeball check?

danielcompton:
This all looks pretty good to me, thanks for doing the work on this!

My weak suggestion would be to lighten the background slightly in dark mode, perhaps to `13, 17, 23` or something similar? GitHub uses that value, and Webkit.org uses 22, 23, 22 which seems a little bit too light to me.

kconner:
@danielcompton Thanks for that — what is it about GitHub's dark mode background color choice that you like better? What kind of display are you using to evaluate? (OLED, backlit TFT, phone, laptop?)

danielcompton:
> what is it about GitHub's dark mode background color choice that you like better? What kind of display are you using to evaluate?

I'm using an M1 MacBook Air's LCD display for testing. I'm kinda going from my limited graphic design reading which recommends not using an absolute black, e.g. https://graphicdesign.stackexchange.com/questions/25356/why-not-use-pure-black-000-and-pure-white-fff. That may not apply in this context though!

kconner:
I've done the color contrast pass.

I chose new dark mode text colors to match light mode contrast. I believe we are hitting WCAG AA in both light and dark mode everywhere. That involved darkening a few cases of light mode text where it was very faint.

@danielcompton Much appreciated! I agree most with [this response](https://graphicdesign.stackexchange.com/a/25359): The body deserves more contrast than any navigation elements. Light mode upholds that. I've still kept the background as pure black, because with that I was able to hit all the contrast goals and still leave some headroom, which I'm hoping is most comfortable when Lobsters is your only light source. If you still feel strongly about it, please take a look on your phone in the dark?

app/views/home/about.html.erb
```diff
@@ -317,21 +317,6 @@
   about six years after we started.
   </p></li>
 
-  <li><p>
-    If you prefer to read Lobsters on a dark background instead of a light one, there are a few options. You could install the
```
hmadison, line 321: I feel like we need to update this explanation to talk about when and how the site goes into dark mode, and what variables are worth changing to alter the colors.

pushcx: Good thinking. I'll add a description to the trivia section after I merge this PR.

pushcx:
Thanks for this thorough work. I'll post an `announce` story when I deploy this in a minute to try to keep the bikeshedding to one place.

paulirwin:
Hi all, I am a regular reader/lurker of the site but don't have an account so I can't comment there. I agree with other commenters that this is more of a "high contrast" mode rather than what is typically understood to be dark mode, especially for web pages. Compare it to this GitHub page, for example, which has a softer dark background and softer color contrast (or at least that's how my eye sees it). I don't have a problem reading in GitHub's dark mode, and while I appreciate that some people might need a high contrast dark mode, this first version is too much for my eyes. I am hopeful that you all will be able to add a manual override soon, or a softer dark mode, as I don't want to turn off dark mode in my OSes just for this site. Thanks again for the excellent site and all you do, and I look forward to being able to comfortably read it again soon, even if in "light mode".

jungle-boogie:
I agree with @paulirwin. Viewing lobsters on my iphone was a shocking change.

Could the preferences be saved in a cookie? It might be nice to occasionally visit lobsters in dark mode and sometimes in light mode.

PR #1028, opened 2021-09-27
Use white background in rendered QR code

rjc:


rjc:
A quick comment: QR code rendered looks absolutely fine with default lobste.rs config. However, when using a costom _dark_ theme, the QR code generated is impossible to scan with a mobile phone camera/app, hence a request to be explcit.

Vast majority of QR codes are black on white anyway :^)

PR #1033, opened 2021-10-06
Override system light/dark mode per user profile

DanielHeath:
Fairly simplistic approach - moved the css variables into a .css.erb file and wrote the dark mode settings out twice (once for system, once for the override. 

You could avoid sending 20 lines of CSS twice with some JS, but that seems like substantially more overhead.


DanielHeath:
Yeah, per-session would also be pretty easy to do.

pushcx:
I'm trying to sucker @kconner into taking a look to see if they can spot a way out of repeating the CSS, but more likely I'm just working up to biting the bullet on that.

kconner:
@pushcx Sorry about letting your email get repeatedly buried. I don't see an obvious way to DRY this, but my working knowledge of CSS 3 is largely happenstance, and there might be a way.

pushcx:
Thank you for letting me sucker you into taking a look. I will step up my work on accepting the inevitable.

pushcx:
I have accepted that repetition is the only way to implement this, given the constraints. Thank you again for contributing this, and for your patience with my emotional process.

I spent a while tinkering. There's really no way for the server to change something in the HTML that a media query can pick up, CSS can't express an `or` between `prefers-color-scheme` and selector, and javascript is ugh.

RaymiiOrg:
I know this is not techsupport, but after disabling the setting (from system to light), the front-page hasn't updated entirely:

Screenshot_20211202-075322174


Comment pages are fine, except for the hats symbol is also off:

Screenshot_20211202-075423067


A logout gives me the dark site, logging back in gives me the screenshot above. (Does not help).

I'm on Firefox on Android (11), but I do have battery saving (doze) mode on, which enables the dark mode system wide.  Can't turn off just dark mode and still have battery saving...

pushcx:
No apology needed, this is the best place to report an issue with this PR.

@DanielHeath Can you take a look? At a glance I'd guess there's a color or two that didn't survive getting moved?

DanielHeath:
Looking. Apparently the small-screen color scheme was defined separately and uses hard-coded colors.

Issue #331, opened 2016-12-15
Use archive.org instead of archive.is


picsi:
Archive.is is a company without a business model, while archive.org is a non-profit with millions of donators. Needless to say, the Internet Archive will stick around for much much longer, and is thus a more reliable place to store caches.

volth:
Could you change ```archive.is``` to ```archive.today```?
The reason is as the archive caches everything some of its domains are blocked in some countries.
```archive.today``` redirects to a working domain based on the visitor's geolocation while having very little chances to be blocked as it does not host cached pages.

volth:
... and if the url contains ```#``` you should replace it with ```%23```: ```http://archive.is/https://twitter.com/%23!/jhapadon```

PR #1035, opened 2021-10-19
Archive fixes

alternet0:
This seemed like a good first issue to contribute to. Hopefully this can lead to more patches.

1. The archive.today person actually said he prefers people use archive.today instead of the mirror. (see recent posts on https://blog.archive.today/ and https://blog.archive.today/post/664493714826215424/which-archivetoday-mirror-should-i-use-which-one )

2. Added other sites since people were complaining #762 , #331, #795) and to protect in the instance when a site may not archive properly on .today. 

Also visiting the search page does not archive the page, is this an oversight in diff_bot.rb? If you want to save the page there is a different url at archive.md to do so. (I forget what it was but i can find out).

The cache dropdown IS compatible with dark mode ;)


pushcx:
Offering multiple links is a nice way out of this disagreement. Do you see a way to make this work on mobile? There's a 'check box hack' that we use for toggling comment trees without js, maybe you can reuse that?

alternet0:
Have to say, that CSS thing is one of the more clever things I've learnt in programming.
This changes it so it will work on mobile and desktop which your trick.

The padding-top: 10px is to prevent an issue in when trying to close the cached dialogue results in an accidental click on the cached copy

pushcx:
Thanks, I've merged and deployed this!

This PR isn't going to show as merged here on GitHub because I tucked in some minor tweaks and a second copy of the checkbox hack to close the menu.

PR #1040, opened 2021-12-03
Handle prefers-color-scheme in mobile & lobsters-specific CSS

DanielHeath:
Fixes for the dark-mode mobile view regression introduced in https://github.com/lobsters/lobsters/pull/1033


DanielHeath:
Requiring a helper from config/application.rb is slightly unfortunate, but initializers are not guaranteed to be loaded by assets:precompile.

pushcx:
I can't quite follow what's going on in this PR - there's an entire separate copy of the colors? Why can't that get collapsed down into the existing colors?

DanielHeath:
There's three files defining the existing colors. My previous PR only made one of them conditional - this one applies that change to all three.

pushcx:
It's been a minute since you opened this PR, and in that minute I've screwed up the git history a bit; to the point that it's probably easier for you to re-fork than merge.

But to the point of this PR: I don't think we're getting any value from splitting our CSS across multiple files. We have these bugs and surprises, it's just confusing. Do you think this would get simpler if we just merged `color-scheme.css`, `mobile.rss`, and `local/lobsters.css.erb` down to `application.css`? It seems like then we could just have one section for light with a header `/* mobile */` for that section, etc.

DanielHeath:
Merging color-scheme and mobile into application would have been far simpler for me to understand.

I assumed that `local/lobsters` was to make theming sister sites easy? Perhaps it could live separately.

pushcx:
I've looked at a few repos, they tend to make large changes to the CSS and ignore our edits entirely, rather than keep their customizations to `local/` by overriding anything we do (which would be pretty brittle as we make changes).

Could I talk you into a opening a PR to merge them down?

pushcx:
I've been tinkering with CSS lately, so I brought in your fix from this PR. You're right, it was indeed dead simple. Thank you!

PR #1041, opened 2021-12-04
Simple spelling and consistency fixes

alternet0:
1. The 422 error page was updated to use a capital L for Lobsters, just like what the 404 page does.
2. The libera chat link was upgraded to use HTTPS.


pushcx:
Thanks for catching these, I appreciate the polish.

PR #1042, opened 2021-12-04
Github Actions 🤖

### LINKED ISSUE #1022: Replace TravisCI
Travis [potentially leaked all keys](https://travis-ci.community/t/security-bulletin/12081). Our Travis config had zero prod secrets so we're unaffected, but they didn't notify affected users and had to be [publicly shamed](https://twitter.com/peter_szilagyi/status/1437646122751873028) into reporting it at all.

So we need a new build tool to keep master green and know if PRs pass the test suite. Our [current setup](https://github.com/lobsters/lobsters/blob/master/.travis.yml) is pretty small, so hopefully this isn't too much work.

quad:
Closes #1022.


pushcx:
Hey, thanks for jumping on this. Is this working? I don't know GitHub Actions much to evaluate it, so I figured I'd ask for a bit of context before I go read their docs.

quad:
It's working! You can see [a passing check run on my fork](https://github.com/quad/lobsters/runs/4416440282?check_suite_focus=true).

.github/workflows/check.yml
```diff
@@ -0,0 +1,34 @@
+on: [push, pull_request]
+
+jobs:
+  check:
+    environment: production
+    runs-on: ubuntu-latest
+
+    env:
+      DATABASE_URL: mysql2://root:root@localhost/lobsters_dev
```
pushcx, line 9: I've been reading up on GitHub Actions and I don't get where this database is coming from. I expected to see a `services` keyword but somehow you're referring to a db and it's wired up correctly? How does the name `lobsters_dev` get used on the db server, and how does it show up in the rails `database.yml`? And - can this be mariadb to match prod?

quad: > […] I don't get where this database is coming from.

[MySQL is installed by default in the Ubuntu image.](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#mysql) It used to be started automatically, but [that was turned off by Github in 2020](https://github.blog/changelog/2020-02-21-github-actions-breaking-change-ubuntu-virtual-environments-will-no-longer-start-the-mysql-service-automatically/).

> How does the name `lobsters_dev` get used on the db server […]

The name `lobsters_dev` is arbitrary; I used the [same name as from the original `.travis.yml`](https://github.com/lobsters/lobsters/pull/1042/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L7).

> […] and how does it show up in the rails `database.yml`?

[Rails picks up `DATABASE_URL` and it overrides (the non-existent?) `database.yml`.](https://guides.rubyonrails.org/v5.0/configuring.html#configuring-a-database)

> can this be mariadb to match prod?

Surely! However, I don't know off my head the proper way to install MariaDB. (Presumably it's an `apt-get` away … but 🤷 ) I'd have to first investigate to make that happen.

Does that block this PR? The original Travis configuration used MySQL; this PR add GHA to parity. And in doing so, it already revealed a broken spec.

pushcx: Yep, totally reasonable. I'll open an issue for that in a sec.

pushcx:
I merged this to master and filed #1050. There was a bit of chaos due to fixing some git history, so the GitHub UI doesn't show this as merged. And... probably easier to re-fork the repo than try to fix it up on your side, honestly. Sorry for the hassle, and thanks again for doing this.

quad:
Cheers!

PR #1043, opened 2021-12-06
Fix/#1034

lvancraen:
## Notes:
This is my first change in this repo and have only recently begun coding in Rails. Please let me know if I haven't respected the proper style or have done something that shouldn't be done!

I went back and forth between using the `f.submit`, as I've done, and the `link_to`, as was used before in the edit form. I ultimately decided to go against using the `link_to` button due to the fact I couldn't seem to add the form to the body of the POST. `link_to` only allowed me to add the form as params.

Something that wasn't explicitly asked for in the issue but ultimately would have saved me a lot of time while debugging would have been better error validation on the `moderation_reason` field. The JS function that ran silently returns providing no feedback to the user explaining that the field is required. So, I decided to validate the field and return an error if the field isn't populated. Though I realize this wasn't asked for, I think it adds a lot of value. But do let me know if you don't want it and I can bring back the original functionality.

Thanks!


pushcx:
Hey, thanks for taking this on, you nailed it. This won't show as merged on GitHub because I locally added a commit (369d90336553dfd42dee1c3f8905cf585dadb9ac) to update a couple tests, but I've merged and deployed your work.

PR #1046, opened 2021-12-11
Update the Gemfile.lock for the current iteration

### LINKED ISSUE #989: Dependency on mimemagic 0.3.x no longer valid
Any solutions for installing the current version of lobsters? I always hit "Dependency on mimemagic 0.3.5 no longer valid" and installations stop. Can't find a workaround yet.
I know it's  a big problem for a loooot of apps, more here - [https://dev.to/cseeman/what-s-up-with-mimemagic-breaking-everything-he1](https://dev.to/cseeman/what-s-up-with-mimemagic-breaking-everything-he1)

cameronhotchkies:
The referenced version of rails `6.0.3.3` relies on `mimemagic (0.3.5)` which has been pulled from available downloads. This updates the lock file to reference rails `6.0.3.7` which no longer contains mimemagic as a dependency.

Resolves #989

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -109,7 +109,7 @@ GEM
       i18n (>= 1.6, < 2)
     ffi (1.15.4)
     flamegraph (0.9.5)
-    globalid (0.5.2)
+    globalid (1.0.0)
       activesupport (>= 5.0)
     good_migrations (0.1.0)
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for reporting this. As you noticed, we happened to roll in a fix with some other bumps.

PR #1047, opened 2021-12-30
Add additional tracking query parameters 

alternet0:
Sourced from:

https://support.google.com/google-ads/answer/9744275?hl=en

https://support.google.com/optimize/answer/6361119?hl=en#zippy=%2Cexample-parameters-in-redirect-destination-dont-conflict


pushcx:
Thanks for catching these, nice to clear out a few more.

PR #1053, opened 2022-01-24
Fix domain-specific RSS feeds

kemitchell:
Following on from #1052, which fixed `<meta type=rss>` URL generation. The meta tags point to the right URLs for RSS feeds now, but requests for those feeds respond 404.

Opening this as draft for now, so I can get a CI run and see results. My first commit does not include a fix, just an addition to the spec.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -67,7 +67,17 @@ Rails.application.routes.draw do
   get "/t/:tag" => "home#tagged", :as => "tag"
   get "/t/:tag/page/:page" => "home#tagged"
 
-  get "/domain/:name" => "home#for_domain", :as => "domain", :constraints => { name: /[^\/]+/ }
+  # Why are there two routes for `home#for_domain`?
+  #
+  # By default, Rails omits dots from parameters so that it can use extension
+  # to determine format.  So the `:id` in `foo.json` is `foo`, and JSON is the format.
+  # We need to include dots in domain names, since they have them, but also support
+  # the `.rss` suffix for RSS/Atom feeds.  So we define one route just for HTML and
+  # another, with the extension, specifically for RSS.
+  get "/domain/:name" => "home#for_domain", :as => "domain",
+    :constraints => { name: /[^\/]+/ }, :format => false
+  get "/domain/:name.rss" => "home#for_domain", :as => "domain_rss",
+    :constraints => { name: /[^\/]+/ }, :format => 'rss'
   get "/domain/:name/page/:page" => "home#for_domain", :constraints => { name: /[^\/]+/ }
 
   get "/search" => "search#index"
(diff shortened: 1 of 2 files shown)


pushcx:
So, right off, this is a wonderful way to report a bug, thank you.

I spent a minute playing with it this morning. The issue is in `config/routes.rb`:

`  get "/domain/:name" => "home#for_domain", :as => "domain", :constraints => { name: /[^\/]+/ }`

The regexp is overbroad so the `name` param is grabbing the `format` as well. So when `HomeController#for_domain` runs, `params[:name]` is `example.com.rss` instead of `example.com`.

kemitchell:
Relevant Guide section for Rails 6.1: https://guides.rubyonrails.org/v6.1/routing.html#route-globbing-and-wildcard-segments

Elsewhere in the guide:

> By default the :id parameter doesn’t accept dots – this is because the dot is used as a separator for formatted routes. If you need to use a dot within an :id add a constraint which overrides this – for example :id => /[^\/]+/ allows anything except a slash.

That example RE is the one currently used for `name` in the domain route.

Current API docs for `match`, for which `get` is a specialized shorthand: https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Base.html#method-i-match

kemitchell:
I haven't done real Rails work since 2.1. But I had a quick look at the doc over AM coffee number one. I think I found an answer.

The trouble here is that Rails' default is to exclude periods from its regular expression for route string parameters like `:id`. That enables it to detect file-extension-like suffixes like `.json` and treat them as format specifiers instead.

Domain names are one of the exceptional cases where the names of things always contain periods, so there's potential confusion about whether, say, `.com` is the TLD part of the domain or a format extension.

My approach in 9adb1447f37dfc405a9167f4e5742c5f841d51b3 was to specify two routes, one with `.rss` hard-coded into the path string and the other without. Then override Rails' automatic format parsing on both.

I had to make up a new name, `domain_rss`, for the new domain route, to hush an error about having two routes with the same name `domain`. Apart from clearing the error on build, I have no idea what the other consequences of this might be.

kemitchell:
CI is green on my fork. But I'm not sure it ever went red to begin with, even with the new test.

kemitchell:
@pushcx thank for your time and comment. I have to admit I'm not quite sure what comes next, or how to do it.

I am running low on time and motivation. I'd like a working RSS feed to know when my blog pops up on lobste.rs. But it's not a really high personal priority.

Please feel free to close, take over, whatever you think best.

pushcx:
No worries. I added the test in b49b5c54 and puzzled out the regexp fix.

kemitchell:
@pushcx, thanks so much! I just subscribed to https://lobste.rs/domain/writing.kemitchell.com.rss.

Only today, another Lobster took the time to send a private message about posting one of my blog posts. Now I can see them automatically!

Issue #1037, opened 2021-11-07
Mention regex doesn't match "-" which is allowed in username


NattyNarwhal:
[Problem comment](https://lobste.rs/s/kynguh/backups_corruption#c_jnguqt) to an article posted by `c--`:

> I think this post buries the lede - according to *@c*–, a security update from Apple corrupted user files. This sounds like a way bigger deal than backup strategies.

Should be a trivial fix.

fpsvogel:
I found that CommonMarker is turning those repeated hyphens into dashes, before the user mention is turned into a link. #1039 adds:

- A preliminary step of escaping the hyphens, so that CommonMarker does not turn them into a dash.
- A final step of unescaping hyphens (turning them into dashes after all) if the user does not exist. This has to be a separate step because the user check happens after the CommonMarker conversion.
- An expanded test which includes these cases.

This is my very first PR, so let me know if anything is missing or needs to be changed.

pushcx:
Hi @fpsvogel, thanks for taking a stab at this in #1039. Painful experience has made me really suspicious of bugfixes that take the general form of escaping part of data, handing it off to be modified, and then unescaping that part back. There's the possibility that the raw data includes your escape sequence, or that modified data gains it. And a lot of potential for the escaping and unescaping functions to slip out-of-sync, especially in complex input.

I found another, worse case of this behavior: `CommonMarker.render_doc("c-_-a_", :SMART, []).to_html  =>  "<p>c-<em>-a</em></p>\n"` I think what's happening here is CommonMark considers the `-` to be a word boundary, so a valid (terrible) username is taken as the markup for emphasis.

Looking at our `preprocess_test_node` I see it's taking a `CommonMarker::Node`. Does that `Node` contain the input text? Can we peek at following `Node`s to recognize that our username code should drop out an `<em>` node if it appears? I don't know the API here, but I suspect the AST is the right level of abstraction for this tweak.

fpsvogel:
Thanks @pushcx for the feedback. I see what you mean about the risks of escaping and unescaping data. And thanks for pointing out that other edge case—I've incorporated it into my second attempt described below.

[...]

PR #1056, opened 2022-01-28
Include hyphens and underscores in @username links by de-converting …

fpsvogel:
…dashes and \<em> tags within username

This addresses #1037.

On further reflection, I'm not sure this is the best approach—see my comment https://github.com/lobsters/lobsters/issues/1037#issuecomment-1024521680


fpsvogel:
Closing this old PR to de-clutter.

PR #1061, opened 2022-02-09
Redesign nav for usability and consistency

pushcx:
Implementation for #580. Let's try to keep discussion there, though comments on the diff or fiddly implementation issues are best here.

- [x] update links in nav
- [x] update links in user nav
- [x] update links in footer
- [x] move filter to footer; even if not 'correct' it's low-use
- [x] fix mobile width again
- [x] rename legend (is this ever different from `@title`?) Answer: No.
- [x] move all use of `@heading` into ~legend~ `@title`
- [x] `layouts/application.html.erb`: content_for subnav pattern
- [x] subnav: Home Recent Newest Top
- [x] subnav: Saved Upvoted Stories, Comments Hidden (rm layout)
- [x] subnav: About Chat Stats Hats
- [x] subnav: Messages Unread All Comments Stories
- [x] subnav: mod
- [x] subnav: address usage as 'back button'
- [x] ~move `User#unread_replies_count` into `Keystore`~
- [x] ~`User#inbox_count`: `Keystore.where(unread_replies, unread_messages).sum(:value)`~
- [x] cache `User#unread_replies_count` so not every pageload is slow
- [x] move about/chat/stats/privacy actions to own controller
- [x] update messages sublegend
- [x] ~karma log page?~ bigger feature than I want to take on here
- [x] top nav: Active to sort stories by most recent comments
- [x] close story thread after 90d so /active doesn't promote thread necro
- [x] `RepliesController`: add `after_action` to bust cache on `User#unread_replies_count`
- [x] change box/inside layout to flexbox/grid
- [x] change header to flexbox/grid
- [x] change footer to flexbox/grid
- [x] style subnav, underline current top-level nav, `link_to_different_page`
- [x] ~change comments to grid?~ conceptually better but fixing something that isn't yet broken
- [x] don't show `@title` on every page; redundant on single-story view + many others
- [x] replace `@cur_url` with `request.path` or explicitly-named instance vars
- [x] rename `#l_holder` to `#logo`
- [x] `app/views/home/index` is mess, move to a layout and break out views

PR #1068, opened 2022-03-09
fix horizontal alignment of upvote buttons

talklittle:

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -1539,7 +1539,7 @@ div.flash-success h2 {
 	}
 	ol.stories .upvoter,
 	#inside > ol.comments .upvoter {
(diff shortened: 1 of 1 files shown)


pushcx:
Could you post before and after screenshots for desktop and mobile? I spent a bunch of time fiddling with that alignment in #1061 so it's certainly possible I introduced a bug in one layout.

talklittle:
Mobile: No change

Desktop: 300% zoom, before/after = left/right  

lobsters-upvoter_2022-03-09

Similar results on Firefox and Chrome  
Similar results in comments sections

pushcx:
Well, I can't unsee that, I better merge this.

In 5dae2eba and others I tweaked to try to center these under the logo; a later tweak to nav layout must've nudged it a little or something. Thanks for catching and fixing this!

PR #1074, opened 2022-03-29
Fixes comment flag bug, closes #1062

### LINKED ISSUE #1062: Comment flag reasons appear in weird places.
When trying to flag a comment, the flag reason drop down appears not tied to the flag link, but instead to a fixed position from the screen chrome and remains fixed in position, floating when scrolling.

An example trying to flag Janus

briankung:
Closes #1062

Screen Shot 2022-03-29 at 4 32 17 PM

I don't really know what I'm doing. I just switched the position options around in the affected CSS until it looked okay. I would like people to be able to flag my comments more effectively, though, so here's a fix 🎁

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -1018,7 +1018,7 @@ div.comment_text code {
 
 
 #flag_why, .archive-dropdown {
-	position: fixed;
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for fixing this, nice to see a familiar face. :)

PR #1077, opened 2022-03-30
Use MariaDB instead of MySQL when running tests in GitHub Actions

### LINKED ISSUE #1050: GitHub Actions: replace mysql with mariadb
https://github.com/lobsters/lobsters/pull/1042/files#r769709354

As discussed, GitHub actions is running against MySQL instead of the MariaDB we use in prod. Let's update that to match, we've already had a lot of prod pain because of the differences between the two.

Tenzer:
This starts a Docker container running MariaDB instead of using the MySQL database that GitHub Actions comes with. This means we can configure which specific version of MariaDB to run.

I opted for MariaDB 10.5 as that was my best guess as to what version was running in the deployed setup, based on it being the version available for the "mariadb-server" package in Debian stable.

The Docker container is configured to create the "lobsters_dev" database automatically, so it shouldn't be necessary to run a command to explicitly create the database anymore.

Fixes #1050.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -5,8 +5,17 @@ jobs:
     environment: production
     runs-on: ubuntu-latest
 
+    services:
+      mariadb:
+        image: mariadb:10.5
+        env:
+          MARIADB_ROOT_PASSWORD: root
+          MARIADB_DATABASE: lobsters_dev
+        ports:
+          - 3306:3306
+
     env:
-      DATABASE_URL: mysql2://root:root@localhost/lobsters_dev
+      DATABASE_URL: mysql2://root:root@127.0.0.1/lobsters_dev
       RAILS_ENV: test
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -5,22 +5,26 @@ jobs:
...
+        image: mariadb:10.5
+        env:
+          MARIADB_ROOT_PASSWORD: root
+          MARIADB_DATABASE: lobsters_dev
+        ports:
+          - 3306:3306
+
     env:
-      DATABASE_URL: mysql2://root:root@localhost/lobsters_dev
+      DATABASE_URL: mysql2://root:root@127.0.0.1/lobsters_dev
```
jamietanna, line 18: Does this not work for `localhost` still? Interesting!

Tenzer: No, then it connects through a socket, which worked before when MySQL wasn't running in a container, but doesn't when it's run in Docker.

pushcx:
Wonderful, thank you for sorting out this config. We do juuuuust enough direct querying that I knew we'd get bitten eventually.

PR #1078, opened 2022-04-01
Fix #1049 Can't delete system messages

### LINKED ISSUE #1049: Can't delete system messages
When a Story is edited by user suggestions, the submitter gets a Message. That Message gets an `author_user_id` of `0` and appears in the inbox as being from _System_ rather than a named user.

Since 12644b271 (and triggered by a user for the first time last month after >3y!) a bug prevents deleting these Messages, individually or batch-delete. When a recipient deletes a message it toggles the `deleted_by_recipient` boolean and saves, whereupon a callback (arrrgggggh) checks to see if the record should be deleted from the database. But instead an exception is thrown by `.save!` because the object is invalid without a real `author_user_id`.

This needs a fix so users can delete system messages. Maybe the right fix is to make `author_user_id` nullable and migrate to replace the `0`s with `NULL`s? Or maybe `MessagesController#destroy` and `#batch_delete` should just directly delete records when possible and delete the `check_for_both_deleted` magic entirely?

vincentrolfs:
This also fixes a (unknown?) bug where users do not receive a message if a suggestion on one of their stories was auto-accepted due to quorum.

I implemented this in a kind of TDD style. As stated above, at the moment, users don't get sent any message if a suggestion was auto-accepted. This is because the created message has no author, which causes an error when the message is saved. So I created a test for that first, then made the `author` field nullable in messages.

**Summary of changes:**
- Add tests for auto-accepting suggestions on stories.
- Make author_user_id nullable in messages.
- Fix mod note creation in case message has no author.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -6,7 +6,8 @@ class Message < ApplicationRecord
   belongs_to :author,
              :class_name => "User",
              :foreign_key => "author_user_id",
-             :inverse_of => :sent_messages
+             :inverse_of => :sent_messages,
+             :optional => true
   belongs_to :hat,
              :optional => true
 
--- a/app/models/mod_note.rb
+++ b/app/models/mod_note.rb
@@ -36,13 +36,15 @@ class ModNote < ApplicationRecord
   end
 
   def self.create_from_message(message, moderator)
-    user = moderator.id == message.recipient.id ? message.author : message.recipient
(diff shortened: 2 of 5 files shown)


pushcx:
Yep, that's an unknown bug. That certainly used to work.

Thanks for working through this and adding tests. This looks like a perfect solution.

PR #1086, opened 2022-05-25
Fix the fake data bugs & add progress bar

ymeynot45:
The recent merge commit 4be8ba11 broke some things in the `fake_data` task. This PR fixes those errors, and turns the comments into a simple progress display by printing `.`s as it goes.


pushcx:
Thanks for fixing this. This was indeed the result of a bad merge, and the progress dots are a nice little quality-of-life improvement.

PR #1089, opened 2022-05-30
Fix #1087: Add unban reason to fake data

### LINKED ISSUE #1087: `fake_data.rake` calls `unban_by_user` with incorrect parameters
`unban_by_user` takes two parameters:

https://github.com/lobsters/lobsters/blob/b822c2cdc201df2f4977c2f6539338fde5475c28/app/models/user.rb#L551

`fake_data.rake` gives only one:

https://github.com/lobsters/lobsters/blob/36cee94b693e3781ab70e6ca41ad58c454ad84a2/lib/tasks/fake_data.rake#L228

jasmaa:
Generate unban reason for fake data. Was able to successfully run `rails fake_data` afterwards.

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -225,7 +225,7 @@ class FakeDataGenerator
       user.ban_by_user_for_reason!(User.moderators.sample,
                                    Faker::Lorem.sentence(word_count: 5))
       if i.even?
-        user.unban_by_user!(User.moderators.sample)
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for jumping on this one - I also fixed it in 344e9473 and added a test to hopefully prevent it from breaking again in the future.

Your approach with lorem ipsum is nicer, so I'm going to sort out the merge conflict to pull in your change. Thanks!

PR #1096, opened 2022-06-10
Continuing to remove the jQuery #554

ymeynot45:
This is part two of removing jQuery as a part of #554. 

I added a new function `fetchWithCSRF` which is required for the CSRF token as plain JS requires a little more handling than jQuery to validate requests.

This pull request also adds a `removeExtraInput` function that fixes a bug, in which various calls to the backend result in an additional input being added to the site.  It hasn't been a problem as they are `display: none` so no one has noticed.  However, if you make changes in a particular order on a comment it can become hidden until you reload the page. 

I will post another update next week.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -396,18 +396,6 @@ $(document).ready(function() {
     return false;
   });
 
-  $(document).on("click", "button.comment-cancel", function() {
-    var comment = $(this).closest(".comment");
-    var comment_id = comment.attr("data-shortid");
-    if (comment_id != null && comment_id !== '') {
-      $.get("/comments/" + comment_id + "?show_tree_lines=true", function(data) {
-        comment.replaceWith($.parseHTML(data));
-      });
-    } else {
-      comment.remove();
-    }
-  });
-
   $(document).on("click", "a.comment_deletor", function(event) {
     event.preventDefault();
     if (confirm("Are you sure you want to delete this comment?")) {
@@ -574,6 +562,21 @@ onPageLoad(() => {
   on('keydown', 'textarea#comment', (event) => {
     if ((event.metaKey || event.ctrlKey) && event.keyCode == 13) {
       Lobsters.postComment(parentSelector(event.target, 'form'));
+    }
+  });
+
+  on('click', 'button.comment-cancel', (event) => {
+    const comment = (parentSelector(event.target, '.comment'));
+    const commentId = comment.getAttribute('data-shortid');
+    if (commentId !== null && commentId !== '') {
+      fetch('/comments/' + commentId + '?show_tree_lines=true')
+        .then(response => {
+          return response.text().then((text) => {
+            replace(comment, text);
+          });
+        });
+    } else {
+      comment.remove();
     }
   });
 
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -561,7 +561,7 @@ onPageLoad(() => {
         });
     } else {
       comment.remove();
-    }  
+    }
   });
 
   on('click', 'a.comment_editor', (event) => {
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -391,17 +391,6 @@ $(document).ready(function() {
     return false;
   });
 
-  $(document).on("click", "a.comment_undeletor", function(event) {
-    event.preventDefault();
-    if (confirm("Are you sure you want to undelete this comment?")) {
-      var li = $(this).closest(".comment");
-      $.post("/comments/" + $(li).attr("data-shortid") + "/undelete",
-      function(d) {
-        $(li).replaceWith(d);
-      });
-    }
-  });
-
   $(document).on("click", "a.comment_disownor", function() {
     if (confirm("Are you sure you want to disown this comment?")) {
       var li = $(this).closest(".comment");
@@ -501,11 +490,11 @@ const fetchWithCSRF = (url, params) => {
   return fetch(url, params);
 } 
 
-    const removeExtraInput = (targetElement) => {
-      // This deletion will resovle a bug that creates an extra hidden input when rendering the comment elements.
-       const extraInput2 = targetElement.parentElement.querySelector('.comment_folder_button');
-       extraInput2.remove();
-    }
+const removeExtraInput = (targetElement) => {
+  // This deletion will resovle a bug that creates an extra hidden input when rendering the comment elements.
+  const extraInput2 = targetElement.parentElement.querySelector('.comment_folder_button');
+  extraInput2.remove();
+}
 
 onPageLoad(() => {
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -515,16 +465,41 @@ const onPageLoad = (callback) => {
 const replace = (oldElement, newHTMLString) => {
   const placeHolder = document.createElement('div');
   placeHolder.insertAdjacentHTML('afterBegin', newHTMLString);
-  const newElement = placeHolder.firstElementChild;
-  oldElement.replaceWith(newElement);
+  const newElements = placeHolder.childNodes.values();
```
pushcx, line 468: Ah, yeah, when `app/views/comments/_comment.html.erb` renders it's two HTML elements, the `<input>` for the comment folder and then the `<div>` for the comment.

app/assets/javascripts/application.js.erb
```diff
@@ -515,16 +465,41 @@ const onPageLoad = (callback) => {
...
+  oldElement.replaceWith(...newElements);
+}
+
+const fetchWithCSRF = (url, params) => {
+  let csrfToken = document.querySelector('meta[name="csrf-token"]').getAttribute('content');
+  let h = new Headers();
+  h.append('X-CSRF-Token', csrfToken);
+  params = params || {};
+  params['headers'] = h;
```
pushcx, line 477:
```suggestion
  params = params || {};
  params['headers'] = params['headers'] || new Headers();
  params['headers'].append('X-CSRF-Token', csrfToken);
```
This overwrites all headers set by the caller, so I reused your `||` trick from the line above and rearranged to have the function only replace the CSRF header. It's not a bug yet, or even likely to be a bug, but it would be a pretty frustrating one to run down if it happened.

app/assets/javascripts/application.js.erb
```diff
@@ -515,16 +465,41 @@ const onPageLoad = (callback) => {
...
+  return fetch(url, params);
+} 
+
+const removeExtraInput = (targetElement) => {
```
pushcx, line 481: I see how this function works to fix the template issue, but it's really brittle in this form. If the template is fixed to not render an extra `<input>`, it'll remove the working input. If multiple rerenders of the comment happen before it's called, it'll only remove one.

Let's be more specific in the selector: the goal is to find _extra_ inputs, not all:

```
const extraInputs = targetElement.parentElement.querySelectorAll('.comment_folder_button + .comment_folder_button');
for (const i of extraInputs) {
  i.remove();
}
```

app/assets/javascripts/application.js.erb
```diff
@@ -515,16 +465,41 @@ const onPageLoad = (callback) => {
...
+  // Global Functions
   on('click', '.markdown_help_label', (event) => {
     parentSelector(event.target, '.markdown_help_toggler').querySelector('.markdown_help').classList.toggle('display-block');
   });
 
+  // Inbox Related Funtions
+
+  on('change', '#message_hat_id', (event) => {
+    let selectedOption = event.target.options.selectedIndex;
+    if (event.target.options[selectedOption].getAttribute('data-modnote') == 'true') {
```
pushcx, line 498: Replace the `if`: `document.getElementById('message_mod_note').checked = event.target.options[selectedOption].getAttribute('data-modnote') === 'true');` (but see previous comment)

app/assets/javascripts/application.js.erb
```diff
@@ -515,16 +465,41 @@ const onPageLoad = (callback) => {
...
 
+  // Global Functions
   on('click', '.markdown_help_label', (event) => {
     parentSelector(event.target, '.markdown_help_toggler').querySelector('.markdown_help').classList.toggle('display-block');
   });
 
+  // Inbox Related Funtions
+
+  on('change', '#message_hat_id', (event) => {
+    let selectedOption = event.target.options.selectedIndex;
```
pushcx, line 497: Use [`selectedOptions.first`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions) to avoid hassling around with indexes.

app/assets/javascripts/application.js.erb
```diff
@@ -576,13 +551,67 @@ onPageLoad(() => {
...
+        response.text().then(text => replace(comment, text));
+      });  
+    removeExtraInput(comment); 
+  });
+
+  on("click", "a.comment_deletor", (event) => {
+    event.preventDefault();
+    if (confirm("Are you sure you want to delete this comment?")) {
+      const comment = parentSelector(event.target, '.comment');
+      removeExtraInput(comment);
```
pushcx, line 581: I think you have to move this down after the fetched comment is inserted into the page. Same with the other calls to the cleanup function.

app/assets/javascripts/application.js.erb
```diff
@@ -576,13 +551,67 @@ onPageLoad(() => {
...
+  on('click', 'a.comment_moderator', (event) => {
+    const reason = prompt("Moderation reason:");
+    if (reason == null || reason == '') 
+      return false;
+
+    const formData = new FormData();
+    formData.append('reason', reason);
+    const comment = parentSelector(event.target, '.comment');
+    const commentId = comment.getAttribute('data-shortid');
+    fetchWithCSRF('/comments/' + commentId + '/delete', { method: 'post', body: formData })
```
pushcx, line 612: I think this needs a `removeExtraInputs` call as well?

pushcx (changes requested):
Thanks for continuing this big cleanup!

I was pretty surprised to see the harmless bug with the extra `<input>` tags, that's been lurking for many years now. I agree with your approach of removing the extra ones client-side rather than taking on a big refactor of the most important (and complicated) template on the site. Maybe after jQuery is gone. :D

Overall another solid PR, though I had a few more small code tweaks on this one.

app/assets/javascripts/application.js.erb
ymeynot45: Thank you for catching the oversight.

app/assets/javascripts/application.js.erb
ymeynot45: That is a great refactoring.  Thanks for streamlining it.

app/assets/javascripts/application.js.erb
ymeynot45: So I had to use `event.target.selectedOptions[0]` because `.first` is a Ruby function and won't work in JS.

app/assets/javascripts/application.js.erb
ymeynot45: This is a much better function, as it removes the dependency on the error to not BE an error.

app/assets/javascripts/application.js.erb
ymeynot45: I moved the `removeExtraInput()` into the replace function to deal with an async issue and to reduce the number of place it is called.  It works on all of the comment calls except for the post/edit comment. I haven't rewritten that one yet, to remove the jQuery.

app/assets/javascripts/application.js.erb
ymeynot45: This got covered when I moved the `removeExtraInputs` to the replace function.

app/assets/javascripts/application.js.erb
pushcx: ...guess which language I write more of. Thanks for catching that. I see you caught some other small issues in the code suggestions I wrote off the top of my head.

app/assets/javascripts/application.js.erb
```diff
@@ -558,7 +561,7 @@ onPageLoad(() => {
         });
     } else {
       comment.remove();
-    }
+    }
```
pushcx, line 564: whitespace

PR #1098, opened 2022-06-17
Continuing to remove the jQuery #554 part III

ymeynot45:
Part 3 in the removal of jQuery.

I created the new Lobster class to replace the Lobsters class and make the transition easier. I pre-generated the functions that I need to replace and made notes as to which functions rely on other functions.

I identified and labeled a bug in the `checkStoryDuplicate` function.  I won't fix it until I rewrite the function to remove the jQuery. 

Removed jQuery from:

- Posting comments
- Account page Homepage URL entry
- The call to `checkStoryTitle`

Noted problems with this pull request:
- I put a bug into the post comment function when you edit/update a comment.  I added another commit to fix that issue.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -472,12 +472,82 @@ const fetchWithCSRF = (url, params) => {
 } 
 
 const removeExtraInputs = () => {
-  // This deletion will resovle a bug that creates an extra hidden input when rendering the comment elements.
+  // This deletion will resovle a bug that creates an extra hidden input when rendering the comment elements.<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for continuing this, I see how you've sketched out your to-do list in the code here. When you finish replacing the class, please rename the new implementation from `_LobstersFunction` to `_Lobsters` and variable from `Lobster` to `Lobsters`.

One more thing for your to-do list: there are a couple `<script>` tags in `app/views` that I'd like to roll into this so we can have a stricter CSP. The `curUser` in `app/views/layouts/application.html.erb` should move into being an attribute on the `<body>` tag with a name like `data-curUser` or something.

PR #1099, opened 2022-06-24
Replacing 7 functions to further the removal of jQuery (#554)

ymeynot45:
This request includes:

Replacing the:
-  vote()
-  saveStory()
-  hideStory()
-  checkStoryTitle() - Partial replacement due to select2
-  bounceToLogin()
-  checkStoryDuplication()
-  previewComment()

It also:
-  Removes an embedded script tag that is no longer required.
-  Creates a slideDownJS() to replace jQuery slideDown()
-  Updates AutoSize()

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -224,27 +224,6 @@ var _Lobsters = Class.extend({
     });
   },
 
-  checkStoryTitle: function() {
-    var title = $("#story_title").val();
-    if (title === undefined || title === '') {
-      return;
-    }
-    var m;
-    if (m = title.match(/^(show|ask) lobste\.?rs:? (.+)$/i)) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -463,35 +317,84 @@ const removeExtraInputs = () => {
 
 class _LobstersFunction {
   constructor () {
-    this.curUser = 'will';  //done
-  }
+    this.curUser = 'will';  // remove line when done with production
```
pushcx, line 320: As I [mentioned the other day](https://github.com/lobsters/lobsters/pull/1098#issuecomment-1160486634) this should move to a `data` attribute you can read. I've updated the server side (bbd4887) to present this, so please update this PR to read that. But switching over all the references from `Lobsters.curUser` to `Lobster.curUser` can wait for your next PR.

PR #1100, opened 2022-06-28
Rails 7.0

thomasdziedzic:
I couldn't figure out why passing the headers in the `spec/requests/stories_spec.rb` test fails with rails 7.0 but I got them to pass by passing an explicit format.

Clicked around, things seem to work.


pushcx:
I deeply appreciate that you have repeatedly taken on the unexciting work of our Rails updates, thank you again.

The headers thing is really odd and I want to tinker with that. I was just looking at endpoints for #554 and thought all of our client-side uses POST form payloads rather than JSON.

pushcx:
Yeah, it's just that one endpoint is inconsistent. I'll file an issue to make it match.

PR #1103, opened 2022-06-30
workaround for mimemagic yanked gem version

ghaliouss:
Dependency on mimemagic 0.3.x no longer valid


pushcx:
We handled this back in #989, what prompted this PR?

PR #1104, opened 2022-07-04
Ruby 3.0.4

caioalonso:


pushcx:
Thanks for picking this up. From the failed build, it looks like Rubocop needs to be updated as well. Let me see if I can bump that without too much pain...

caioalonso:
@pushcx yeah, I think there will be some pain with the offenses that couldn't be autocorrected, even though I disabled all the new cops.

PR #1108, opened 2022-07-09
Don't confirm email is used on the password reset page.

febuiles:
Currently, a user can find out if someone else is a lobste.rs member by trying to reset their password (supposing they know their email). Depending on the resulting message ("Reset instructions have been sent" vs "User not found") you can deduce who the users are. 

With the changes in this PR the same message will be shown regardless of the email's presence in the database.

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -110,7 +110,8 @@ class LoginController < ApplicationController
     @found_user = User.where("email = ? OR username = ?", params[:email], params[:email]).first
 
     if !@found_user
-      flash.now[:error] = "Unknown e-mail address or username."
+      flash.now[:success] = "Password reset instructions have been " <<
+                            "e-mailed to you if this account exists."
       return forgot_password
     end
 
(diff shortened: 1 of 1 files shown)


pushcx:
Is there much value to this? We get a significantly worse error message for a fairly common occurrence (someone not remembering which email they used) that'll mean I get more support emails. And we'll still leak email use on login and account creation.

febuiles:
@pushcx Thanks for the quick feedback. The idea behind the PR was to improve the users' privacy by a tiny amount, it stops folks from entering a bunch of emails and confirming which ones have a Lobste.rs account. I was not aware that the login page showed a different message based on user existence, but that one could be changed too. Account creation is something I'm not too worried about since you have an invite-based system.

With all that said: I empathize with the volume of support requests, and I think the UX will be shittier after this change. Feel free to close the PR if you don't think it adds any value!

pushcx:
I looked at what emails people are using and there's three buckets (very! rough numbers): around 85% of users use an email that they publish on GitHub or a personal site, around 10% use a site-specific email like `lobsters@example.com` or `username+lobsters@gmail.com`, and 5% use what is apparently a one-off from a privacy-oriented site like tutanota, riseup, or protonmail.

Given that our userbase is generally very privacy-savvy and can use privacy-oriented services to not just prevent this kind of data mining but keep private from site mods, I think this is not a good tradeoff for the site. If we discussed sensitive topics in private I'd see a lot more value in it, but we go the opposite direction in having the [hat system](https://lobste.rs/hats) to publicly verify identities.

Thanks for putting your time and effort into this PR. You prompted a serious public examination of why the site should work the way it does, which is valuable even if it didn't get merged.

Issue #1110, opened 2022-07-17
Message on hat approvals


pushcx:
There's a moderator page for approving hat requests at https://lobste.rs/hats/requests

It has a text box for sending a message on rejecting hat requests, but it would also be really convenient to send a message on approval, too. (Lately, that's me apologizing for the delay in approvals.)

For UI, just have one text box and an approve or reject button. Please style the "reject" button red, like the "Ban" button admins see on user profiles.

As long as you're on the form, please also display the hat request `created_at` using `time_ago_in_words`. It doesn't need its own line, just put it to the right of the username.

ymeynot45:
Ok, I'll start that monday after untangling the vote function.

saksham-jain:
Can I take it and work on this if this is not picked up yet?

pushcx:
Yes - I talked with @ymeynot45 offline and he's going to pass on this issue as it's all Ruby and he's doing javascript.

ymeynot45:
Cool, thanks!
It is always nice to distribute the work. Sorry, I didn't get back to you
sooner.

saksham-jain:
@pushcx Is this fine?
<img width="1440" alt="Screenshot 2022-07-19 at 11 56 04 AM" src="https://user-images.githubusercontent.com/27828987/179680691-442675dd-347e-4868-bd34-16d81cee2e08.png">

saksham-jain:
@pushcx created PR #1111 for this issue. Kindly check once and let me know if any changes are required.

pushcx:
Closing, @saksham-jain completed this in #1111

PR #1111, opened 2022-07-19
Message for hat approvals

saksham-jain:
- Common reason box for hat approval and rejection.
- Change the reason box param name from rejection_comment to reason.
- Change approve_by_user! to approve_by_user_for_reason! with reason param.

### DIFF
--- a/app/controllers/hats_controller.rb
+++ b/app/controllers/hats_controller.rb
@@ -44,8 +44,8 @@ class HatsController < ApplicationController
   def approve_request
     @hat_request = HatRequest.find(params[:id])
     @hat_request.update!(params.require(:hat_request)
-      .permit(:hat, :link))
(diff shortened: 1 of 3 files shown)


saksham-jain:
> Your screenshot looks great, I really appreciate that you took this on. I like that you renamed this function.
> 
> I'm going to merge this because it works correctly, but there's an opportunity for improvement I didn't realize when I filed #1110. `HatsController` doesn't use the [strong params](https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters) feature to protect against abuse. It's using `params` in a safe way now, but it'd be great to use the tool to make sure it stays safe. Would you like to take that, or should I file a separate issue for whoever comes along?

Thanks, @pushcx. Surely I will pick it up and fix it soon.

PR #1113, opened 2022-07-20
Refactoring and removing jQuery from the vote function 

ymeynot45:
The vote function was used for both stories & comments to process both voting them up & flagging them.  This allowed for a lot of bugs and required onClick event functions to be created within the function, instead of as an onPageLoad event.
This pull request is step 1 of a 2 step process to split the voting into separate functions for stories or comments.  It will also reduce the chance of errors by reducing the number of params passed. Instead the functions will internally know if they are handling a story or comment, rather than searching the DOM to determine what type of element it is handling .

This will double some of the code but remove complexity and chance for errors. I can later reduce some of the duplication in the follow up pull request.

This pull request includes:

Functions that have removed jQuery:
- auto adding tags when people post from sites like youtube (Still needs work due to select2)
- flagging stories & comments

Other fixes:
- removing html script tags and changing them to an onPageLoad function
- prevented select2 from running before the fetch request resolved
- a typo in the css
- Split voting and flagging into comments & stories rather than one really large function that does all four. 
    - This removes the 'story'/'comment' param from the function and all of the gates to determine which of the two it is trying to process.  This makes it much easier to fix future bugs.
    - This will fix future bugs related to flagging and an event trigger that currently get created when you click the flag button.
-  Created a function to handle all of the flagging & modal removal.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -185,12 +185,6 @@ class _LobstersFunction {
       }
       flaggingDropDown.append(a);
     });
-
-    on('click', '#flag_dropdown a', (event) => {
-      if (event.target.getAttribute('data') != '')
-        onChooseWhy(event.target.getAttribute('data'));
-      return false;
-    });
   }
 
   checkStoryDuplicate(form) {
@@ -516,6 +510,9 @@ onPageLoad(() => {
   });
 
   on('click', '#flag_dropdown a', (event) => {
+    if (event.target.getAttribute('data') != '') {
+      Lobster.voteStory(parentSelector(event.target, '.story'), -1,  event.target.getAttribute('data'));
+    }
     document.querySelector("#flag_dropdown").remove();
     document.querySelector("#modal_backdrop").remove();
     document.querySelector('.flag-wrapper').remove();
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -130,23 +130,23 @@ class _LobstersFunction {
     document.location = "/login?return=" + encodeURIComponent(document.location);
   }
 
-  modalFlaggingDropDown(flaggedItemType, voterEl) {
+  modalFlaggingDropDown(flaggedItemType, voterEl, reasons) {
     if (!Lobsters.curUser)
       return Lobster.bounceToLogin();
 
-      const li = parentSelector(voterEl, '.story, .comment');
-      if (li.classList.contains('flagged')) {
-        /* already upvoted, neutralize */
-        if (li.classList.contains('story')) {
+    const li = parentSelector(voterEl, '.story, .comment');
+    if (li.classList.contains('flagged')) {
+      /* already upvoted, neutralize */
+      if (li.classList.contains('story')) {
         Lobster.voteStory(voterEl, -1, null);
-        } else {
-          Lobster.voteComment(voterEl, -1, null);
-        }
-        return
+      } else {
+        Lobster.voteComment(voterEl, -1, null);
       }
+      return
+    }
 
     if (document.querySelector("#flag_dropdown") || document.querySelector('#modal_backdrop')) {
-      Lobster.modalRemoval()
+      Lobster.removeFlagModal()
     }
 
     const modalDiv = document.createElement("div");
@@ -155,30 +155,21 @@ class _LobstersFunction {
 
     const flaggingDropDown = document.createElement('div');
     flaggingDropDown.setAttribute('id', 'flag_dropdown');
-    const flaggingWrapper = document.createElement('span');
-    flaggingWrapper.setAttribute('class', 'flag-wrapper')
-    flaggingWrapper.style.position = 'relative';
-    flaggingWrapper.style.maxWidth = '1 px';
-    flaggingWrapper.style.maxHeight = '1 px';
-    flaggingWrapper.appendChild(flaggingDropDown);
-    voterEl.after(flaggingWrapper);
-
-    let reasons;
-    if (flaggedItemType == "comment") {
-      reasons = Lobster.commentFlagReasons;
-    } else {
-      reasons = Lobster.storyFlagReasons;
-    }
+    const flagWrapper = document.createElement('span');
+    flagWrapper.setAttribute('id', 'flag_wrapper')
+    flagWrapper.style.position = 'relative';
+    flagWrapper.style.maxWidth = '1 px';
+    flagWrapper.style.maxHeight = '1 px';
+    flagWrapper.appendChild(flaggingDropDown);
+    voterEl.after(flagWrapper);
 
     Object.keys(reasons).map(function(k, v) {
       let a = document.createElement('a')
       a.textContent = reasons[k]
       a.setAttribute('data', k)
+      a.setAttribute('href', '#')
       if (k === ''){
-        a.setAttribute('href', '#' )
         a.classList.add('cancel-reason')
-      } else {
-        a.setAttribute('href', '#')
       }
       flaggingDropDown.append(a);
     });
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -230,6 +130,60 @@ class _LobstersFunction {
     document.location = "/login?return=" + encodeURIComponent(document.location);
   }
 
+  modalFlaggingDropDown(flaggedItemType, voterEl) {
+    if (!Lobsters.curUser)
+      return Lobster.bounceToLogin();
+
+      const li = parentSelector(voterEl, '.story, .comment');
```
pushcx, line 137: This is indented one level too deeply.

app/assets/javascripts/application.js.erb
```diff
@@ -230,6 +130,60 @@ class _LobstersFunction {
...
+    }
+
+    const modalDiv = document.createElement("div");
+    modalDiv.setAttribute('id', 'modal_backdrop');
+    document.body.appendChild(modalDiv);
+
+    const flaggingDropDown = document.createElement('div');
+    flaggingDropDown.setAttribute('id', 'flag_dropdown');
+    const flaggingWrapper = document.createElement('span');
+    flaggingWrapper.setAttribute('class', 'flag-wrapper')
```
pushcx, line 159: match the name `flaggingWrapper` to the class `flag-wrapper` (so `flagWrapper`) or vice-versa (`flagging-wrapper`)

app/assets/javascripts/application.js.erb
```diff
@@ -230,6 +130,60 @@ class _LobstersFunction {
...
+    } else {
+      reasons = Lobster.storyFlagReasons;
+    }
+
+    Object.keys(reasons).map(function(k, v) {
+      let a = document.createElement('a')
+      a.textContent = reasons[k]
+      a.setAttribute('data', k)
+      if (k === ''){
+        a.setAttribute('href', '#' )
```
pushcx, line 178: This line is the same in both branches of the `if/else`, so move it above the `if`

app/assets/javascripts/application.js.erb
```diff
@@ -319,6 +263,12 @@ class _LobstersFunction {
     fetchWithCSRF("/stories/" + li.getAttribute("data-shortid") + "/" + act, {method: 'post'});
   }
 
+  modalRemoval() {
```
pushcx, line 266: Rename to put the verb first and be explicit about the subject: `removeFlagModal`.

app/assets/javascripts/application.js.erb
```diff
@@ -319,6 +263,12 @@ class _LobstersFunction {
...
+    document.querySelector("#flag_dropdown").remove();
+    document.querySelector("#modal_backdrop").remove();
+    document.querySelector('.flag-wrapper').remove();
```
pushcx, line 269: If there can only be one of these (implied by the use of `querySelector` instead of `querySelectorAll`), use an id rather than a class.

app/assets/javascripts/application.js.erb
```diff
@@ -464,13 +410,81 @@ class _LobstersFunction {
...
+      score++;
+      action = "upvote";
+    } else if (point < 0) {
+      if (li.classList.contains("upvoted")) {
+        /* Removes the upvote point this user already gave the story*/
+        score--;
+      }
+      li.classList.remove("upvoted");
+      li.classList.add("flagged");
+      if (li.parentElement.querySelector('.comment_folder_button')) {
```
pushcx, line 463: Under what conditions would the folder be missing? When I look at `app/views/comments/_comment.html.erb` there's no conditional in there, every comment gets one.

If it's always there, remove this guard. If there's a place it might be missing, edit this + next line to extract the element to a variable rather than select it twice.

app/assets/javascripts/application.js.erb
```diff
@@ -464,13 +410,81 @@ class _LobstersFunction {
...
+        li.parentElement.querySelector('.comment_folder_button').setAttribute("checked", true);
+      };
```
pushcx, line 465:
```suggestion
      }
```

app/assets/javascripts/application.js.erb
```diff
@@ -464,13 +410,81 @@ class _LobstersFunction {
...
+      score--;
+      action = "flag";
+    }
+    if (showScore) {
+      scoreDiv.innerHTML = score;
+    } else {
+      scoreDiv.innerHTML = '~';
     }
+    if (action == "upvote" || action == "unvote") {
+      if (li.querySelector(".reason")) {
```
pushcx, line 476: Similar to the above comment about whether the folder control is present, is the reason ever missing? It's weird to me that this branch of the `if (action...` has to test for it but the one a few lines down is safe to assume it's present.

app/assets/javascripts/application.js.erb
```diff
@@ -464,13 +410,81 @@ class _LobstersFunction {
...
+        li.querySelector(".reason").innerHTML = ""
+      };
```
pushcx, line 478:
```suggestion
      }
```

app/assets/javascripts/application.js.erb
```diff
@@ -488,6 +502,10 @@ onPageLoad(() => {
     parentSelector(event.target, '.markdown_help_toggler').querySelector('.markdown_help').classList.toggle('display-block');
   });
 
+  on('click', '#modal_backdrop', () => {
```
pushcx, line 505: Move this out to the other event handlers, or every time someone brings up the flagging interface you'll be re-binding this, so the event will fire multiple times on close.

app/assets/javascripts/application.js.erb
```diff
@@ -537,6 +566,34 @@ onPageLoad(() => {
...
+      // can't remove the jquery until I replace select 2
+      if ($("#story_url").val().match(new RegExp(u, "i"))) {
+        var ta = $("#story_tags_a").data("select2");
+        if (ta.getVal().indexOf(tag) < 0)
+          ta.addSelectedChoice({ id: tag });
+      }
+    }
+
+    // check for dupe if there's a URL, but not when editing existing
+
```
pushcx, line 589: remove blank line

app/assets/javascripts/application.js.erb
```diff
@@ -230,6 +130,60 @@ class _LobstersFunction {
...
+    const flaggingWrapper = document.createElement('span');
+    flaggingWrapper.setAttribute('class', 'flag-wrapper')
+    flaggingWrapper.style.position = 'relative';
+    flaggingWrapper.style.maxWidth = '1 px';
+    flaggingWrapper.style.maxHeight = '1 px';
+    flaggingWrapper.appendChild(flaggingDropDown);
+    voterEl.after(flaggingWrapper);
+
+    let reasons;
+    if (flaggedItemType == "comment") {
```
pushcx, line 167: Make `reasons` a parameter to this function, so the caller (which you've split into two functions) makes the decision. Not only do you get rid of a conditional, this function becomes purely about presentation.

app/assets/javascripts/application.js.erb
ymeynot45: This is already in in the onPageLoad() section.  So it should only fire the first time someone arrives at the site.  I'm not sure what you want me to do.

app/assets/javascripts/application.js.erb
```diff
@@ -155,30 +155,21 @@ class _LobstersFunction {
...
+    flagWrapper.style.maxHeight = '1 px';
+    flagWrapper.appendChild(flaggingDropDown);
+    voterEl.after(flagWrapper);
 
     Object.keys(reasons).map(function(k, v) {
       let a = document.createElement('a')
       a.textContent = reasons[k]
       a.setAttribute('data', k)
+      a.setAttribute('href', '#')
       if (k === ''){
```
pushcx, line 171: missing space between `)` and `{`

PR #1116, opened 2022-07-27
Fix js bugs in the story functions from removing the jQuery

ymeynot45:
This will fix a few bugs that fell into production during the migration away from jQuery.

- fetching the url title when it is blank
- putting a warning up for a duplicate url over 30 days old
- pressing the preview story button a second time


pushcx:
Thanks for fixing this so quick, and solving more along the way.

PR #1117, opened 2022-08-04
HTML related fixes for the tag list page.

GetToSet:
This PR contains two minor HTML related fixes for the tag list (`/tags`).

* Fix unexpected nested `<ol>` elements for tag list.
  Screenshot 2022-08-04 23 13 31

* Use `<p>` instead of `<div>` for admin actions area at the bottom of the tag list to provide extra vertical margins.
  Screenshot 2022-08-04 23 18 03

### DIFF
--- a/app/views/tags/index.html.erb
+++ b/app/views/tags/index.html.erb
@@ -36,11 +36,9 @@
   <% end %>
 
   <% if is_admin %>
-    <div>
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for catching and fixing this bug, I appreciate it!

PR #1118, opened 2022-08-04
Fix wrong category edit paths for the tag list page

GetToSet:
This PR fixes wrong category edit link paths for the tag list page.


pushcx:
Hey, good catch of an obscure bug, thanks for this fix. Did you catch it because you're setting up a sister site?

GetToSet:
> Hey, good catch of an obscure bug, thanks for this fix. Did you catch it because you're setting up a sister site?

Sure I'm setting up a site using Lobsters to organize my links and keep track of articles for read-later purposes. Although Lobsters is designed for communities, I found it perfectly suitable for pure-personal  sites.

Issue #1080, opened 2022-04-06
Domain: fold tracker bans into regular bans, add UI


pushcx:
For a long time, the site has had code banning link shorteners like `t.co` and `bit.ly` to avoid duplicate submissions and inconvenience marketers who are trying to get good [metrics](https://github.com/lobsters/lobsters/commit/e2ea9decb21de6d375148168e0aa8aeb31be1cdf) on their spam. When we [broke out the `Domain` model](https://github.com/lobsters/lobsters/commit/f718da18205ae9fcabe45903677572651274b6fe) that static list of banned domains became the `is_tracker` attribute on the `Domain` model. This functionality was made redundant by domain [banning](https://github.com/lobsters/lobsters/commit/630ace42f80af5dcd1cb5b7609629cc4c8dd4291), which also has the proper hooks in place to show up in the mod log. As unobjectionable as blocking tracking domains is, I don't like that `is_tracker` isn't visible in the source or UI anywhere.

The two features here are:

1. A migration to remove the `is_tracker` attribute. All the tracker domains should get set as banned with the reason "Used for link shortening and ad tracking." I know there's the `banned_at` timestamp; it's not worth your time to spelunk git history to find a correct timestamp and author for these from the the old `Story.TRACKING_DOMAINS`, you can just call `ban_by_user_for_reason!(User.find_by!(username: 'pushcx', ...`
2. UI for the admin (not mods) to ban domains. We'll need a `DomainsController` with `edit`/`update` actions; look to `TagsController` for most of the setup. (And, yep, domain banning has been rare enough that I've been doing it in the prod rails console.)

fatshinobi:
May I fulfill this task?

pushcx:
Please do! Happy to talk through any of this here or [on IRC](https://lobste.rs/chat).

pushcx:
This should've been closed with #1119, whoops.

PR #1119, opened 2022-08-07
Domains update action and banning logic

fatshinobi:
Domains controller with update action and banning logic.
Removed is_tracker field from Domain model.
According this issue:
https://github.com/lobsters/lobsters/issues/1080

### DIFF
--- a/app/controllers/domains_controller.rb
+++ b/app/controllers/domains_controller.rb
@@ -1,13 +1,13 @@
 class DomainsController < ApplicationController
   before_action :require_logged_in_admin
   before_action :set_domain, only: [:update, :edit, :unban]
+  before_action :set_reason, only: [:update, :unban]
 
   def edit; end
 
   def update
-    banned_reason = params.dig('domain', 'banned_reason')
-    if banned_reason.present?
-      @domain.ban_by_user_for_reason!(@user, banned_reason)
+    if @banned_reason.present?
+      @domain.ban_by_user_for_reason!(@user, @banned_reason)
     else
       flash[:error] = "You must give a reason for the ban"
     end
@@ -16,7 +16,12 @@ class DomainsController < ApplicationController
   end
 
   def unban
-    @domain.unban_by_user!(@user)
+    if @banned_reason.present?
+      @domain.unban_by_user_for_reason!(@user, @banned_reason)
+    else
+      flash[:error] = "You must give a reason for the unban"
+    end
+
     redirect_to edit_domain_path(name: @domain.domain)
   end
 
@@ -25,4 +30,21 @@ private
   def set_domain
     @domain = Domain.find_by(domain: params[:name])
   end
+
+  def set_reason
+    @banned_reason = params.dig('domain', 'banned_reason')
+  end
+
+  def path_of_form(domain)
+    prms = { name: domain.domain }
+    domain.banned_at ? unban_domain_path(prms) : update_domain_path(prms)
+  end
+
+  helper_method :path_of_form
+
+  def caption_of_button(domain)
+    domain.banned_at ? 'Unban' : 'Ban'
+  end
+
+  helper_method :caption_of_button
 end
--- a/app/controllers/domains_controller.rb
+++ b/app/controllers/domains_controller.rb
@@ -1,6 +1,6 @@
 class DomainsController < ApplicationController
   before_action :require_logged_in_admin
-  before_action :set_domain, only: [:update, :edit]
+  before_action :set_domain, only: [:update, :edit, :unban]
 
   def edit; end
 
@@ -12,12 +12,17 @@ class DomainsController < ApplicationController
       flash[:error] = "You must give a reason for the ban"
     end
 
-    redirect_to edit_domain_path(domain_name: @domain.domain)
+    redirect_to edit_domain_path(name: @domain.domain)
+  end
+
+  def unban
+    @domain.unban_by_user!(@user)
+    redirect_to edit_domain_path(name: @domain.domain)
   end
 
 private
 
   def set_domain
-    @domain = Domain.find_by(domain: params[:domain_name])
+    @domain = Domain.find_by(domain: params[:name])
   end
 end
--- a/app/controllers/domains_controller.rb
+++ b/app/controllers/domains_controller.rb
@@ -1,23 +1,10 @@
 class DomainsController < ApplicationController
   before_action :require_logged_in_admin
-  before_action :set_domain, only: [:update, :edit, :ban]
+  before_action :set_domain, only: [:update, :edit]
 
   def edit; end
 
   def update
-    result_domain_name =
-      if @domain.update(domain_params)
-        flash[:success] = "Domain #{@domain.domain} has been updated"
-        @domain.domain
-      else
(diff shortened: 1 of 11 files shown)


config/routes.rb
```diff
@@ -178,6 +178,10 @@
   post "/tags" => "tags#create"
   post "/tags/:tag_name" => "tags#update", :as => "update_tag"
 
+  dm_con = { :domain_name => /[-,0-z\.]+/ }
```
pushcx, line 181: I like that you broke out this constraint to share between the two routes. Can it also be shared with the existing `get "/domain/:name...` route? Maybe not - I've had constant hassle keeping that route happy with Rails wanting to interpret the `.com` as a mime type.

db/migrate/20220806200248_set_is_tracker_as_banned_to_domains.rb
```diff
@@ -0,0 +1,10 @@
+class SetIsTrackerAsBannedToDomains < ActiveRecord::Migration[7.0]
+  def up
+    if ActiveRecord::Base.connection.column_exists?(:domains, :is_tracker)
```
pushcx, line 3: Please remove this `if`. If the migration is run against a db that doesn't have this column, I want it to throw errors and fail.

db/migrate/20220806202153_remove_is_tracker_from_domains.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveIsTrackerFromDomains < ActiveRecord::Migration[7.0]
+  def change
+    if ActiveRecord::Base.connection.column_exists?(:domains, :is_tracker)
+      remove_column :domains, :is_tracker, :boolean
```
pushcx, line 4: Move this `remove_column` up into the prior migration, there isn't a benefit to having two migrations.

spec/features/submit_story_spec.rb
```diff
@@ -134,7 +134,7 @@
   end
 
   scenario "submitting a tracking link" do
```
pushcx, line 136:
```suggestion
  scenario "submitting a banned domain" do
```

Just updating the name to reflect that this test is more general now.

spec/requests/domains_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  before do
+    sign_in user
+    allow_any_instance_of(DomainsController).to receive(:require_logged_in_admin)
+  end
+
+  context 'update' do
+    let(:domain) { create(:domain) }
+
+    it 'buns domain with valid params' do
```
pushcx, line 14:
```suggestion
    it 'bans domain with valid params' do
```

typo

app/controllers/domains_controller.rb
```diff
@@ -0,0 +1,23 @@
+class DomainsController < ApplicationController
+  before_action :require_logged_in_admin
+  before_action :set_domain, only: [:update, :edit]
+
+  def edit; end
+
+  def update
+    banned_reason = params.dig('domain', 'banned_reason')
+    if banned_reason.present?
```
pushcx, line 9: I forgot to mention in the issue, but it should also be possible to unban domains via the edit form. If you want to include that functionality I'd appreciate it, but I see how it'd require touching the model as well, and I can add that to a separate issue after merging this PR if you don't want to take it.

pushcx (changes requested):
Thanks for taking this on! This will be a real improvement to site transparency. There's a couple minor changes and then an omission on my part that you can add or leave for me.

Sorry it took me a little while to review, I was traveling internationally.

config/routes.rb
fatshinobi: It works well for .com domains. This constraint concern for this.
constrains_for_com

app/controllers/domains_controller.rb
fatshinobi: Let me add the 'unban' functionality in this PR, please

db/migrate/20220806200248_set_is_tracker_as_banned_to_domains.rb
fatshinobi: OK

db/migrate/20220806202153_remove_is_tracker_from_domains.rb
fatshinobi: OK

config/routes.rb
pushcx: Yeah - I'm saying that in `config/routes.rb`, there's also lines 72 and 73 with basically this exact same constraint, and it would be good to reuse this so they all match exactly.

app/controllers/domains_controller.rb
pushcx: I'd appreciate it, thank you!

config/routes.rb
fatshinobi: Oh i see now. I will try to implement this constraint to it.

app/views/domains/edit.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="box wide">
+  <h1><%= @domain.domain %></h1>
+  <% unless @domain.banned_at %>
```
pushcx, line 3: You need to remove this `if` or the form can't be used to unban domains.

app/models/domain.rb
```diff
@@ -7,20 +7,13 @@ class Domain < ApplicationRecord
...
 
   validates :domain, presence: true
-  validates :is_tracker, inclusion: { in: [true, false] }
 
   def ban_by_user_for_reason!(banner, reason)
-    self.banned_at = Time.current
-    self.banned_by_user_id = banner.id
-    self.banned_reason = reason
-    self.save!
+    banning_switcher(:on, banner, reason)
```
pushcx, line 12: I really don't like this style of code sharing, where `banning_switcher` toggles between the two. For such sensitive functionality, I would rather have simpler, redundant code, like how `app/models/user.rb` implements `ban_by_user_for_reason!` and `unban_by_user!`. They are explicit and obvious about setting all the state they care about.

app/models/domain.rb
```diff
@@ -7,20 +7,13 @@ class Domain < ApplicationRecord
...
+  end
 
-    m = Moderation.new
-    m.moderator_user_id = banner.id
-    m.domain = self
-    m.action = "Banned"
-    m.reason = reason
-    m.save!
+  def unban_by_user!(banner)
```
pushcx, line 15: This must take and log a `reason`.

pushcx (changes requested):
Thank you for making the changes and adding the ability to unban domains. I don't like the way it's implemented, I like critical code to be very simple, without conditionals. If you change it to work like the `User` model, I think this is ready to merge.

app/models/domain.rb
fatshinobi: OK. I will change it

app/models/domain.rb
fatshinobi: Done

app/models/domain.rb
fatshinobi: Done

app/views/domains/edit.html.erb
fatshinobi: Now we can use the form for this. Check it please

fatshinobi:
Thank you for sharing your experience

PR #1120, opened 2022-08-09
Fix 500 status in /recent.json by adding respond_to (#1114)

### LINKED ISSUE #1114: Status 500 on /recent.json
Url: https://lobste.rs/recent.json.

A 500 status is unexpected, because
`/hottest.json` correctly returns json data, status 200.
`/recent.jsoaasjh` (a non existent route) correctly returns status 404

rinonb:
Fixes #1114 

/recent.json is returning a 500 error because the template is not found.
We fix this by adding multiple response types, just like in other methods is
already done.


pushcx:
Thanks for taking a look at this. Your PR would prevent the 500s, but it's not right for the usage of the site. We don't need json from this endpoint, and we don't present RSS here because the items are frequently reordered rather than being in most-recent order.

I don't know what the config is, but there must be some broader config problem that Rails thinks every endpoint should serve every mime type it knows about.

PR #1121, opened 2022-08-12
Fix incorrect dropdown menu location for the archive button

GetToSet:
This PR fixes the incorrect dropdown menu location for the archive button.
Screenshot 2022-08-12 23 51 03

Both dropdown menus for the flag and dropdown buttons should work well after this fix:
Screenshot 2022-08-12 23 52 37

Screenshot 2022-08-12 23 52 34

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -155,13 +155,7 @@ class _LobstersFunction {
 
     const flaggingDropDown = document.createElement('div');
     flaggingDropDown.setAttribute('id', 'flag_dropdown');
-    const flagWrapper = document.createElement('span');
-    flagWrapper.setAttribute('id', 'flag_wrapper')
-    flagWrapper.style.position = 'relative';
-    flagWrapper.style.maxWidth = '1 px';
-    flagWrapper.style.maxHeight = '1 px';
-    flagWrapper.appendChild(flaggingDropDown);
-    voterEl.after(flagWrapper);
+    voterEl.after(flaggingDropDown);
 
     Object.keys(reasons).map(function(k, v) {
       let a = document.createElement('a')
@@ -258,7 +252,6 @@ class _LobstersFunction {
   removeFlagModal() {
     document.getElementById("flag_dropdown").remove();
     document.getElementById("modal_backdrop").remove();
-    document.querySelector('#flag_wrapper').remove();
   }
 
   postComment(form) {
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -1016,10 +1016,13 @@ div.comment_text code {
 	line-height: 1.2em;
 }
 
+.dropdown_parent {
+	position: relative;
+}
 
 #flag_dropdown, .archive-dropdown {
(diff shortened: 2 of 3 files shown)


ymeynot45:
First, thank you for finding that bug as I introduced it when I refactored the flag dropdown to remove the jQuery. Additionally, it is awesome that you found a solution so fast.

I am just going to give two notes that will get your pull request approved faster.

1) The site is trying to move away from inline styling.  If you want to get ahead of the upcoming change requests, move it to a css file and add classes to the flagging span & the archive span.
2) I would also convert it back to rems.  Building sites for people with vision problems has become a priority and rems [make it easier for people with disabilities to use the site](https://github.com/lobsters/lobsters/issues/1123.). 
The flag dropdown needs -2rem and the archive needs -0.125rem to align them to the previous lobsters px standard.

Once again, thank you for fixing my mistakes.

GetToSet:
> First, thank you for finding that bug as I introduced it when I refactored the flag dropdown to remove the jQuery. Additionally, it is awesome that you found a solution so fast.
> 
> I am just going to give two notes that will get your pull request approved faster.
> 
> 1. The site is trying to move away from inline styling.  If you want to get ahead of the upcoming change requests, move it to a css file and add classes to the flagging span & the archive span.
> 2. I would also convert it back to rems.  Building sites for people with vision problems has become a priority and rems [make it easier for people with disabilities to use the site](https://github.com/lobsters/lobsters/issues/1123).
>    The flag dropdown needs -2rem and the archive needs -0.125rem to align them to the previous lobsters px standard.
> 
> Once again, thank you for fixing my mistakes.

Hi @ymeynot45, I've removed inline styling by adding a new CSS class `dropdown_parent`. This should be a common class for all drop-downs. I've also updated px to rems, and could you provide some more context about why rems are better than px in the context of accessibility?

ymeynot45:
Hey thanks for replying:
So when people with vision problems change the size of the font, so they
can see it, the whole layout goes out the window if you use px. Px is a
static number that doesn't change based on the user settings.  But rems are
based off of the font size the user has set as their default.  So if the
user has the font set to 125% then rems will see that and change everything
to be unscaled to 125%.  This helps keep things inline based on the user
settings.

This is a nice article that covers it.
https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

It will show that in this particular case it isn't that important.  But it
is good to get people in the habit of using rems except where they
shouldn't.  Just like you use "const" instead of "let" except when you
shouldn't. I'll admit that it is a little pedantic.  But that is how
progress is made.

Thank you, once again for replying, your work is greatly appreciated.
- YmeYnot

pushcx:
@GetToSet Thanks for finding this bug and fixing it, getting a PR is basically the best kind of bug report there is. :D

@ymeynot45 Thanks for guiding this PR with your familiarity from our js/css work, I agree with your notes.

I'm really glad to merge this fix, cheers!

ymeynot45:
LOL, I remember when I originally fixed that the first time.  I wonder what
changed that broke it again.
Thank you for catching the new bug.
Just a note, the reason I started using rem, it is better for people with
vision problems.

PR #1124, opened 2022-08-16
Fix duplicate url check bug, for both less & greater than 30 days

ymeynot45:
This is an easy fix just changing a "getAttribute" to a .value as the former only looks at the state in which it was created, not the current state.

I'm sorry that the bug crept back into the code.

I also took the opportunity to clean up some other inconsistencies. 
- When you flag a story it no longer jumps to the top of the page.


pushcx:
Thanks for running down these two bugs. I'm OK with taking them in one PR because you know the js better than anyone at this point and we're a pretty informal repo.

PR #1126, opened 2022-08-25
Fix bug where the comment flag dropdown is out of position.

ymeynot45:
When the last fix for the archive/flag drop downs went through it only fixed the story related drop downs. This patch with make those same changes to the flag drop down in the comments section.


pushcx:
Thanks for the quick fix!

PR #1128, opened 2022-09-03
Fix archive button has no hand cursor on hover

GetToSet:
In #1121 I fixed the issue that the dropdown for the archive button had a wrong position. I still spot a minor issue the archive button has no hand cursor on hover.

This PR is a quick fix to this issue.

This PR also removes the `archive_dismiss` `<label>` and associated styles since it's not needed anymore.


pushcx:
Great, thanks for spotting this confusing little UI issue.

PR #1130, opened 2022-09-12
Fix Archive Dropdown Z-index

ymeynot45:
Hotfix Z index problems with the Archive Dropdown on stories with a negative rating

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -76,10 +76,6 @@ light = <<-LIGHT
 	--color-lobsters-hat-sysop-crown-fill: #ddc7c7;
 	--color-lobsters-hat-sysop-brim-stroke: #bbb2b2;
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.css.erb
```diff
@@ -154,6 +158,10 @@ dark = <<-DARK
 	--color-lobsters-hat-sysop-crown-fill: #3e2f2f;
 	--color-lobsters-hat-sysop-brim-stroke: #2c2222;
 
+  --color-negative-story-rating-1: #666666;
```
pushcx, line 161: Are these extra colors needed? Can we do this with only opacity and clever selectors?

pushcx:
Thanks for this fix! It looks a lot more robust, so hopefully we stop breaking these dropdowns every couple  months. I added a comment to hopefully reduce the amount of breakage (or at least shorten the debugging cycle).

PR #1132, opened 2022-09-30
Clarify when a story/comment is added to the moderation queue

itamarst:


pushcx:
I've been slow merging this because I was letting perfect be the enemy of good. I'm going to merge it because it's an accurate and useful description of the way this works, but I think also I need to improve this queue so it's more of a ticketing/work queue than the current system. Thanks for contributing.

PR #1143, opened 2022-11-29
Fix subnav of Upvoted Comments

PipecraftNet:
The subnav is not showing on the Upvoted Comments page. (Not showing up since https://github.com/lobsters/lobsters/commit/91891dc214b39ad1a673b6fea0a68c6e0f13c221)

refer to https://github.com/lobsters/lobsters/blob/b731305b941e65e7e76185cea3329f7e3fefd2ad/app/controllers/home_controller.rb#L285-L291

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -286,7 +286,7 @@ class CommentsController < ApplicationController
     }
 
     @title = "Upvoted Comments"
-    @saved_subnav = true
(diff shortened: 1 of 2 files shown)


PipecraftNet:
- Before


- After

pushcx:
Ah, thank you, I missed this in that big cleanup. I appreciate you recognizing this bug and fixing it.

PR #1148, opened 2022-12-20
link_post fix

### LINKED ISSUE #1142: Light/dark theme settings don't work
I tried setting my theme setting to dark, but even after clicking "save settings" the setting doesn't stick.

macOS 12.5, Firefox 106.0.5

### LINKED ISSUE #1144: Gravatar cache expiration doesn't work
Clicking `Expire cache` in the Gravatar settings prints "Successfully updated settings" in the flash message, but the old avatar remains.

Perhaps the "fiddly" fix for #1136?

kivikakk:
More details in each commit message, but essentially, fix the fiddly from 2c68883/#1136.

Fixes #1142, fixes #1144.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -489,6 +489,25 @@ onPageLoad(() => {
     }
   });
 
+  on('click', 'a.link_post', (event) => {
+    event.preventDefault();
+
+    const link = event.target;
+
+    const form = document.createElement('form');
+    form.method = 'POST';
+    form.action = link.dataset.url;
+
+    const csrf = document.createElement('input');
+    csrf.type = 'hidden';
+    csrf.name = 'authenticity_token';
+    csrf.value = csrfToken();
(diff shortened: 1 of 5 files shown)


kivikakk:
Looking into the spec failure …

kivikakk:
Addressed; PR should be good to go.

lorthirk:
Any news on this? It would also fix #1152

pushcx:
Thanks for taking this on, sorry for the delay in reviewing.

This would make all the links dependent on js, which not all of our users run. (Yeah, we're an odd userbase that way.) In 2c688838 I was trying to make sure that we still generated an HTML form for them.

pushcx:
It's a bit irksome that the Rails form tag helpers don't catch when they're being nested to create invalid HTML, but at least it's a straightforward improvement to this form to move these tags out of the main settings form.

Thanks again for looking at this and tying the bug reports together, I appreciate it.

Issue #1123, opened 2022-08-15
Lack of semantic structure makes webpage challenging for screen readers


robertmeta:
As a visually impaired user who depends on a screen reader, having semantic elements to interact with on web pages is very important. Without these elements like headers, articles, etc I am forced to cycle through all elements or subsets to use the site.  .

This most critical piece from my perspective is missing headings to jump between on the main site.  Each story title should be a heading, then by using keys to jump to next heading I can quickly read all the headings on the main page.  In terms of value, this would give a ton.  

There are other improvements I would love to see, but all pale in comparison to the above.

robertmeta:
Adding a bit more information by request of @pushcx.

- 101: https://www.w3.org/WAI/fundamentals/accessibility-principles/

A page like that can seem overwhelming -- but just grabbing the lowest hanging fruit and the highest value items can do a huge amount.  As in all things, screen reader users have lots of different workflows, but using headings is very common.  

----

For additional context, when I move around using my screen reader I am moving a screen reader cursor that will read the selected thing, there are mony tools for navigating around available to me, like jump to next/prev of: blockquote, bold, button, checkbox, color change, column, control, different item, font change, frame, heading, heading1/2/3..., image, italic, item of same style, landmark, link, list, plain text, radio group, table, text field, text with different attributes underline, visited  link, and of course a generic next / prev element on page.  

To move around a page properly setup with headers, it is 1 keypress per header and then I can dive in using next element or next link or next plain text to read more. 

After the basic semantic tags are in place, even if not perfect it gives me some tools to build my own browsing flow around, maybe it is 1 header jump, 2 link jumps to get to a story, or 1 cheader then search for the word comment, etc. 

At far beyond the sort of 101, you have tools like discourse and slack that actually make custom bindings that do smart things, like arrow up and down going through messages but skipping all the noise between, this would be adream for reading comments.  But even just having the comments in a unique style from other elements in some form would let me jump between them skipping all other content (all the actions you can take on a comment)  by using jump to next element of attributes style.

pushcx:
Speaking of 101, I found a Chrome extension for highlighting some of these on live pages: https://chrome.google.com/webstore/detail/arc-toolkit/chdkkkccnlfncngelccgbgfmjebmkmce?hl=en

To anyone reading this and considering a contribution: please feel free to jump right in, you don't need to ask for permission, or try to do everything in a single PR.

PR #1149, opened 2022-12-23
Added aria role and levels for stories and comments

robertmeta:
These changes have zero impact on sighted users,
but give screen reader users some core elements
to key off of for navigation, level 1 headings
for stories, level 2 for comment links and level
3 for comments themselves.

The reason the comment role was not used is that
it is not yet supported by most screen readers,
but I plan to add that in another forthcoming
PR.

closes: #1123


pushcx:
Thanks for sorting out this improvement. I’m deploying this now. Please let me know how it works out for you in practice!

robertmeta:
@pushcx working amazing, thank you!

PR #1153, opened 2023-01-14
fix unintentional scrolling when flagging a comment

### LINKED ISSUE #1134: flagging a comment scrolls to the top of the page
Click to flag a comment, click on a reason. The flag is recorded and the comment collapsed, but also the browser scrolls to the top of the page. 

Looking at `application.js.erb` I see:

```
  on('click', '.comment a.flagger', (event) => {
    event.preventDefault();
```

And my first guess for this is that the call to `preventDefault` was missing, so I don't know off the top of my head what's wrong and wanted to file the bug to track it.

Maybe @ymeynot45 wants to take this?

johnpash:
This fixes #1134 

Flagging a comment is a two-step process.

1. Click the "flag" link. This opens a list of reasons for the flag.
2. Click the reason

Step 2 was missing `event.preventDefault()` in the click handler.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -637,6 +637,7 @@ onPageLoad(() => {
   });
 
   on('click', '.comment #flag_dropdown a', (event) => {
+    event.preventDefault();
     if (event.target.getAttribute('data') != '') {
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for figuring out this bug and submitting a fix, great to have this sorted out.

PR #1155, opened 2023-01-18
Fix duplicate password. fixes #1097

### LINKED ISSUE #1097: A css #ID is used twice.
On the account settings page you use `#user_password` twice.
For the New Password input & the Verify Password input on the delete account section.

johnpash:
The settings page contains two forms. Each form contains a `password` field. The Rails `form_with` helper for each form is creating an identical id for each field.

To fix this, I've added a namespace to each form.

In addition I've added the `autocomplete="off"` attribute to a different password field, as this is seemingly best practice to help out password managers. Although I'm not sure that I've run into this particular problem with 1password on any site. So please let me know if you want me to remove this modification.


johnpash:
> Could you share a link for `autocomplete=“off”` being a best practice? It sounds like a nice improvement to roll into this, so I’d like to be able to document it in the commit message.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

If you were to adopt this globally, there are a few other instances where this would need to be added. As it stands, there are 10 password fields in the codebase with 5 containing "autocomplete=off" and the other 5 without. I'm happy to do that in this PR if you want.

pushcx:
Thanks for that link! I'm going to merge this to fix the reused ID.

If you'd like to add `autocomplete=off` to the other password fields and/or the `autocomplete='new-password'` that this article mentions, I'd appreciate a PR for it.

Issue #919, opened 2020-08-21
Clickable (unread) labels


pushcx:
Since 5bd669dab8f15509d50993c9832a31295d75e18e unread comments have a label reading `(unread)`, separate from the `3 hours ago` label. It'd be a nice little use of javascript if clicking an `(unread)` moved you down the page to the next unread comment. (We could do it with html anchors, but I don't want to loop the comments an extra time, comment rendering is expensive enough already.)

johnpash:
I've added the scrolling feature as suggested. My initial attempt was using html anchors but I found that the user-experience wasn't as good as with JS. The anchors worked essentially the same way but it happened to quickly that you could miss what just happened. JS with smooth scrolling gives you time to see what's going on, but not so much that it's annoying.

I've also added "cursor: pointer" to these so that it's discoverable.

pushcx:
Just noticed this feature request didn't get closed when the feature was merged.

pushcx:
Also touched in #1319 just to cross-link.

PR #1156, opened 2023-01-19
Feature: scroll to unread comments

johnpash:
The is a new feature from request #919

When viewing comments you can now click on the "(unread)" label to scroll to the next unread label.

This was done by adding a unique id and a bit of JS to find the next label then scroll.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -736,10 +736,8 @@ onPageLoad(() => {
       }
     })
 
-    var targetIndex = foundIndex + 1
-    if (targetIndex < nodes.length) {
-      var targetY = nodes[targetIndex].getBoundingClientRect().top + window.scrollY
-      window.scrollTo({ top: targetY, behavior: 'smooth' })
-    }
+    var targetIndex = (foundIndex + 1) % nodes.length;
+    var targetY = nodes[targetIndex].getBoundingClientRect().top + window.scrollY
+    window.scrollTo({ top: targetY, behavior: 'smooth' })
   });
 });
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -729,15 +729,10 @@ onPageLoad(() => {
   });
 
   on('click', '.comment_unread', (event) => {
-    var nodes = document.getElementsByClassName('comment_unread')
-    var foundIndex = Array.from(nodes).findIndex(function(n){
-      if (n === event.target) {
-        return true;
-      }
-    })
-
-    var targetIndex = (foundIndex + 1) % nodes.length;
-    var targetY = nodes[targetIndex].getBoundingClientRect().top + window.scrollY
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -727,4 +727,19 @@ onPageLoad(() => {
         response.text().then(text => replace(comment, text));
       });
   });
+
+  on('click', '.comment_unread', (event) => {
+    var nodes = document.getElementsByClassName('comment_unread')
+    var foundIndex = Array.from(nodes).findIndex(function(n){
```
pushcx, line 733: Does this need to add an `id` to the comments? It feels like this could be `.findIndex((n) => { n == event.target })`.

app/assets/javascripts/application.js.erb
```diff
@@ -727,4 +727,19 @@ onPageLoad(() => {
...
+      if (n.id == event.target.id) {
+        return true;
+      }
+    })
+
+    var targetIndex = foundIndex + 1
```
pushcx, line 739: `targetIndex = (foundIndex + 1) % nodes.length;`

Then this wraps around, a nice feature for people who start from the middle.

It also means you can remove the `if` from the next line. (I thought about making it `foundIndex != targetIndex` so that we don’t try to scroll to the current comment when there’s only one unread comment, but if nothing happens it just looks broken - recentering the comment at least acknowledges the click.)

app/assets/stylesheets/application.css.erb
```diff
@@ -1176,6 +1176,7 @@ div.comment_form_container textarea {
 span.comment_unread {
 	color: var(--color-fg-accent);
 	font-weight: 600;
+	cursor: pointer;
```
pushcx, line 1179: Nice catch.

pushcx (changes requested):
Thanks for taking on this feature request! You’ve definitely got a good approach. I’d like a couple small changes before merging.

app/assets/javascripts/application.js.erb
johnpash: You were correct, the id wasn't needed after all.

I've also cleaned up the logic and added the "wrap around" feature that you requested. So now, when you click on the final comment it scrolls to the first one on the page.

app/assets/javascripts/application.js.erb
```diff
@@ -727,4 +727,17 @@ onPageLoad(() => {
         response.text().then(text => replace(comment, text));
       });
   });
+
+  on('click', '.comment_unread', (event) => {
+    var nodes = document.getElementsByClassName('comment_unread')
+    var foundIndex = Array.from(nodes).findIndex(function(n){
+      if (n === event.target) {
```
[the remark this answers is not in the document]
johnpash: Oops. I can't believe I missed this!

I've now made the change.

pushcx:
Thanks for working through this feature, I'll deploy it with a small bugfix in a couple minutes. :D

Issue #1157, opened 2023-01-23
"Suggest" tags unreadably pale in dark mode


AbstractBeliefs:
When the dark colour scheme is selected, new tags being typed in the Suggest page are unreadably pale until they are completed.

Toggling one CSS rule appears to fix this, as shown in the images below. The rule is `color: var(--color-fg);` applied to `input:focus, textarea:focus` (as identified by leon on irc, thanks!), which is located at https://github.com/lobsters/lobsters/blob/master/app/assets/stylesheets/application.css.erb#L378 .

## With CSS rule enabled
![As shown here](https://paste.pulham.info/lobsters-suggest-pale.png)

## With CSS rule disabled
![As shown here](https://paste.pulham.info/lobsters-suggest-dark.png)

pushcx:
Thank you for this clear and detailed bug report. This is a real odd one, it almost looks like the light-mode value for `--palette-fg` is being used.

talklittle:
This was fixed as part of #1145, at least when I test a local development instance. The fix was never deployed to production though.

![lobsters-suggest_2023-02-02](https://user-images.githubusercontent.com/87961/216515579-558c41fe-5c3c-4eb9-8a31-2bbea7015b73.png)

ymeynot45:
@AbstractBeliefs
Can you give me some more details about your setup? I am correcting the issue but I can't replicate your problem. Looking at your screen shots I have a few questions:
1. Why is your tags entry not in dark mode while the rest of your screen is in dark mode. The rule that you change affects the color of the text, while the problem is the color of the entry space as it is white.  It should be black.
2. Is the tags entry white only in the suggest form or is it also white for submit story and edit story?

talklittle:
@ymeynot45 The fix (#1145) was deployed to production at some point after my previous comment. This issue is resolved.

pushcx:
With the merge of #1166, @ymeynot45 has addressed all the details here, so I’m closing.

PR #1166, opened 2023-03-15
correct dark mode tags

ymeynot45:
Issue #1157 
- Created css variables to the TomSelect tags in forms.
- Eliminated the background-image on TomSelect tags in dark mode.
- Added the story_holder div to the suggest form as it got forgotten with the original shift to TomSelect. This should stop the Tags entry from being white in dark mode.


pushcx:
Thanks for taking this, I see how you’ve followed our existing patterns to fix it. I’m going to merge this to get this fix out, but I’m curious: is `#story_holder` still needed? Can it be removed and have the CSS target the `#story_box`?

PR #1169, opened 2023-04-04
Fix spec of bg in mobile comments speech bubble

kasbah:
There's a slight spec of the background color / lack of fill in the mobile comments speech bubble when you have the page at a higher zoom level. The issue is more visible on Firefox but also happens on Chrome. 

before: 
image

after:

image


pushcx:
Thanks for finding and fixing this misbehaving pixel!

PR #1172, opened 2023-04-22
Add reveal for TOTP secret for manual entry

jmcharter:
Small change to show TOTP secret (initially hidden and revealed on click). This allows for manual entry, which is particularly useful if your 2FA is a browser extension such as bitwarden.

Preview:
lobsters_totp

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -101,6 +101,7 @@ class SettingsController < ApplicationController
                        module_size: 5,
                        shape_rendering: "crispEdges")
 
+    @qr_secret = totp.secret
     @qr_svg = "<a href=\"#{totp_url}\">#{qr}</a>"
   end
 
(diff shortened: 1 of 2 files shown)


pushcx:
I can see how this would be handy for folks, thanks for contributing it.

Before I merge - what's the value of requiring a click to reveal the secret? The QR code is already the visible secret, so I it's not protecting against shoulder surfing. But I'm not super-familiar with TOTP so if this is just the standard workflow everyone expects, it makes sense to match it.

jmcharter:
To be totally honest, yes it is more in-keeping with how a lot of other sites present the TOTP secret. I think there is some level of over-shoulder protection as it is at least not displayed immediately and unexpectedly (although I suppose anyone with nefarious intentions could just as easily snap the QR code).

Most places I've seen it (GitHub, AWS, DigitalOcean) either hide it in a similar way to this, or make it totally invisible and add a copy button next to it.

pushcx:
Great!

Issue #1141, opened 2022-11-18
Add profile field for fediverse username


pushcx:
We have a field for Twitter and GitHub usernames. With the current state of Twitter, it seems prudent to add support for Fediverse accounts.

I don't know if Mastodon/etc. does OAuth, but I would like to see some scheme for linking accounts to confirm they're real. Obviously we don't want any permissions to post on someone's behalf or read their timeline or anything, just confirm username.

sknebel:
Mastodon (and I think some other fediverse software) does URL verification based on `rel=me` attributes. 

E.g. if you look at my profile at https://chaos.social/@HeNeArXn, it shows my website with a green checkmark because on my website I have a link `<a href="https://chaos.social/@HeNeArXn" rel="me">` , and Mastodon checks for this and recognizes a matching pair of such claims as confirmation they are indeed linked.

I see that lobsters already also uses `rel=me` on links to e.g. the users homepage, so this mechanism could apply here too - having a matching pair of fields would be green on Mastodon, and if lobsters were to check the reverse would also work. 

This would also be somewhat general and also useful for e.g. the website field in a profile, not specific to Mastodon - so maybe having the option to have multiple, general "this URL is also me" fields would be nicer than a Mastodon/Fediverse specific one?

pushcx:
That sounds like a straightforward and scalable approach. I'm a little concerned that it would force people to link back to their Lobsters profile because I don't think they should be required to advertise their Lobsters profile on some other site to have us to link out. It just feels kind of manipulative, like we're trying to "growth hack".

And to scalability, so far all our integrations have been oauth-based so the amount of site-specific code is pretty minimal and has required darn-near zero maintenance, so it's worked out quite nicely.

sknebel:
> EDIT: and Mastodon instances do indeed use a form of OAuth - given there are many instances they use some sort of dynamic registration, so instead of having to pre-register with the instance as you would register an app with Github or Twitter, this can be done dynamically with a POST request before the login flow.

Here are the relevant documentation bits about that: https://docs.joinmastodon.org/client/token/ for the registration and then https://docs.joinmastodon.org/client/authorized/ for using it/scopes etc

icyphox:
OAuth and `rel="me"` links are largely Mastodon specific. I think the ideal solution to this would (and should) be implementation agnostic.

winks:
[...]

PR #1173, opened 2023-04-25
Add Mastodon connector to settings / profile

jmcharter:
Related to #1141 

Adds a profile connector for Mastodon to settings and public profile.

This is similar to how GitHub and Twitter are currently handled, but with a key difference: There's no central app registration. Each instance requires OAuth app registration. I've handled this by registering with an instance the first time a user connects and storing the client_id and client_secret for future connections.


pushcx:
Thanks for taking on such a big feature! The Twitter endpoint is getting less reliable, so this is pretty timely.

It looks like you’ve made a solid start. I can see you’re still working on this, so please `@mention` me when you’re ready for a review

app/views/settings/mastodon_authentication.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<div class="box wide">
+  <%= form_with url: "/settings/mastodon_auth", method: :get do |f| %>
+    <p>
+    Enter the name of your Mastodon instance without the scheme, 
+    i.e mastodon.social not https://mastodon.social
+    </p>
+
+    <div class="boxline">
+      <%= f.label :mastodon_instance_name, "Mastodon Instance:", :class => "required" %>
```
pushcx, line 13: We have a few `placeholder` attributes in the code, I think this would be a good place for one. With the way you wrote `sanitized_instance_name`, you don’t even need to include the instruction above about not including `https`.

jmcharter: Ah yes, good point! I added `sanitized_instance_name` later. I'll await further review before making any changes, but I've noted this.

app/views/settings/index.html.erb
```diff
@@ -254,6 +254,24 @@
...
+        <%= label_tag :mastodon_username,
+          raw("<a href=\"https://joinmastodon.org/\">Mastodon</a>:"),
+          :class => "required" %>
+        <% if @edit_user.mastodon_username.present? %>
+          Linked to
+          <strong><a href="https://<%= @edit_user.mastodon_instance %>/@<%= h(@edit_user.mastodon_username)
+            %>">@<%= h(@edit_user.mastodon_username) %>@<%=@edit_user.mastodon_instance%></a></strong>
+            <%= link_post "Disconnect", "/settings/mastodon_disconnect" %>
+        <% else %>
+          <a href="/settings/mastodon_authentication">Connect</a>
```
pushcx, line 269: `link_to`

app/views/settings/mastodon_authentication.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :subnav do %>
+  <a href="/settings">Back to Settings</a>
+<% end %>
+
+<div class="box wide">
+  <%= form_with url: "/settings/mastodon_auth", method: :get do |f| %>
```
pushcx, line 6: Use the Rails url generation here and above. Run `rails routes` in your console to see the name for this endpoint, then add `_path` to generate a relative URL.

app/controllers/settings_controller.rb
```diff
@@ -184,6 +184,39 @@ def pushover_callback
...
+      @user.mastodon_oauth_token = tok
+      @user.mastodon_username = username
+      @user.mastodon_instance = params[:instance]
+      @user.save!
+      flash[:success] = "Your account has been linked to Mastodon user #{username}."
+    else
+      return mastodon_disconnect
+    end
+
+    return redirect_to "/settings"
```
pushcx, line 209: `settings_path`, redundant `return`

app/views/settings/mastodon_authentication.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <p>
+    Enter the name of your Mastodon instance without the scheme, 
+    i.e mastodon.social not https://mastodon.social
+    </p>
+
+    <div class="boxline">
+      <%= f.label :mastodon_instance_name, "Mastodon Instance:", :class => "required" %>
+      <%= f.text_field :mastodon_instance_name, :autofocus => true  %>
```
pushcx, line 14: Remove `autofocus`

app/controllers/settings_controller.rb
```diff
@@ -184,6 +184,39 @@ def pushover_callback
...
+    return redirect_to Mastodon.oauth_auth_url(instance_name)
+  end
+
+  def mastodon_callback
+    if !params[:code].present?
+      flash[:error] = "Invalid OAuth state"
+      return redirect_to "/settings"
+    end
+
+    tok, username = Mastodon.token_and_user_from_code(params[:instance], params[:code])
```
pushcx, line 198: I think this creates duplicate `MastodonInstance`s when two users use the same `Instance`. You probably want `MastodonInstance.find_or_create_by`

pushcx (commented):
This is a solid start, I see how you got it working with a minimum of fuss. I left some small notes, but this also needs some high-level rearranging.

The `User` model should `has_one :mastodon_instance` rather than repeating the instance name. The functions in `Mastodon` that take `instance_name` as an argument should move into `MastodonInstance` (perhaps `register_application` should be a factory?).

jmcharter:
@pushcx thanks for taking the time to review this PR. My inexperience with Ruby / Rails is shining through, clearly.
I'll go through the changes you've suggest at some point this weekend and come back to you if I have any questions; otherwise I'll make the the changes and ping you for another review once I'm done.

pushcx:
@jmcharter Thanks for taking this on. I had some free dev time so I decided to get this over the line.

jmcharter:
@pushcx thanks for doing so - I've always meant to come back to this, but you know it is. Cheers!

PR #1179, opened 2023-06-16
Update README.md

sharunkumar:
reddit link updated to a link to webcache of the page.


pushcx:
The google caches are not useful for more than a week or few, would you mind using an archive.org link?

sharunkumar:
> The google caches are not useful for more than a week or few, would you mind using an archive.org link?

done ✌🏻

PR #1181, opened 2023-06-26
Update messaging for Pushover settings

stackptr:
Settings that require a Pushover subscription are rendered before the button to setup the Pushover integration, so the labels "above" and "below" should be swapped.


pushcx:
Thanks for catching this error.

I missed updating this in 9ec9610a57d80321cf0898e1cc2275116fcf19cb

Issue #1180, opened 2023-06-20
warn about capslock on password input


pushcx:
https://defensivecss.dev/articles/intro-defensive-css/

mahadmughal:
Hey @pushcx , are you working on it. If not then can I help in this ?

pushcx:
I posted this because it seemed like a nice idea, I don't have any concrete plans to do it myself. I put the `good first issue` tag on anything that would be easy for someone else to pick up.

Br0k3nh4nd012:
@pushcx Based on the title I can understand that we need to add a text denoting the caps lock status in the password input field on the login page.
I am ready with the code changes. I am creating a [PR](https://github.com/lobsters/lobsters/pull/1183) for the same.
Can you please take a look at it?

Br0k3nh4nd012:
@pushcx Any updates on this?

Br0k3nh4nd012:
@pushcx Anything?

pushcx:
I don't know why I thought this was possible with CSS, but it doesn't seem to be. I definitely didn't paste a relevant link.

PR #1183, opened 2023-07-09
Added Capslock warn message near password field(login page)

Br0k3nh4nd012:
## warn about capslock on password input
1. Added a 'key up' event to the password field.
2. show/hide the capslock warn message based on the 'CapsLock' state.
3. Done a few linting on the same file as well.


pushcx:
We don't serve js to logged out users, so #1180 doesn't seem possible. Sorry for your wasted time.

PR #1190, opened 2023-08-09
Recursive CTE arrange_for_user

pushcx:
Our biggest performance cost is `Comment#arrange_for_user`, which fetches the comments for a story and loops them to build a data structure of parents + children, which `app/views/comments/threads.html.erb` must mirror perfectly to render comments via non-obvious templating. This is pretty slow because it’s a lot of extra allocations and GC on the Ruby app workers, and it’s caused worker exhaustion in prod when a big thread has gotten popular. On a recent with 43 comments, a first page render for a logged-in user takes 500ms! Subsequent renders are ~200ms (probably due to the db having queries freshly cached), which is also very slow for moving <100kb of text.

MariaDB has now supported [recursive CTEs](https://mariadb.com/kb/en/recursive-common-table-expressions-overview/) for a few years, so I decided to experiment with using them to retrieve comments with the db doing the work of sorting and arranging comments. Initial results are very promising - if I dump all comments via `arrange_for_user` (`script/arranged_comment_tree.rb`) it takes 70s; one query to thread the comments across all stories (`script/selected_bytepack_comment_tree.rb`) takes 34s. I expect the perf test of rendering comments will match or improve on this speedup.

This is a messy WIP PR because I wanted somewhere to preserve my various test scripts and notes. Expect the merge commit to include substantial edits.

todo:

  - [x] integration perf test: loop all stories and time rendering them 10x
    - [x] w arrange_for_user
    - [x] w selected
    - [x] w selected_bytepack
  - [x]  better name for ord/ordpath                                              
  - [x] extract exploratory tests in test_bitpacking to test suite
  - [x] update call sites of `arrange_for_user`
  - [x] materialize confidence_order column
  - [x] inline the _comment partial, ~20% perf win
  - [x] materialize similar_stories


app/models/comment.rb
```diff
@@ -199,6 +154,11 @@ def as_json(_options = {})
 
   def assign_initial_confidence
     self.confidence = self.calculated_confidence
+    # 0 is a placeholder, don't have an id before save.
+    # This will move it up so it gets seen + voted on.
+    self.confidence_order =
```
hmadison, line 159: This feels like it could do with more explanation. From my read, we are taking the confidence, normalizing it, then converting it into a big edian 16 bit unsigned integer to use as a sorting key during the CTE.

app/models/comment.rb
```diff
@@ -344,7 +317,8 @@ def update_score_and_recalculate!(score_delta, flag_delta)
       UPDATE comments SET
         score = (select coalesce(sum(vote), 0) from votes where comment_id = comments.id),
         flags = (select count(*) from votes where comment_id = comments.id and vote = -1),
-        confidence = #{self.calculated_confidence}
+        confidence = #{self.calculated_confidence},
+        confidence_order = concat(lpad(char(65536 - floor(((confidence - -0.2) * 65535) / 1.2) using binary), 2, '0'), char(id & 0xff using binary))
```
hmadison, line 321: Is there a way to keep the math in one place to calculate confidence order? Maybe this is a good place to use a UDF to hold the confidence to confidence order calculation and use it here and in `#assign_initial_confidence`?

config/initializers/heinous_inline_partial.rb
```diff
@@ -0,0 +1,61 @@
+# Inlines partials into parent templates for performance.
+# In production & test: once at startup; development: every request
+#
+# Inlining the comments/_comment partial into stories/show (bulk of our traffic!)
+# gives a 25-50% speedup over 'render collection: @comment', and 80-90% speedup over looping to
+# 'render partial: comment'.
+
+# rubocop:disable Style/MutableConstant
+HEINOUS_INLINE_PARTIALS = {
```
hmadison, line 9: I'm wondering if we could go for a `render collection` with some rails cache tweaks to get a "good enough" speedup to skip this. While the performance is great, I'm concerned as to how well its going to upgrade.

pushcx: I share your preference for caching, and the logged-out users get a cache from `StoriesController#get_arranged_comments_from_cache`. But comments include per-user vote arrows, so we'd need to cache per-user, and I'd rather logged-in-users see discussions live. The recursive CTE takes fetching + arranging comments from ~30ms to ~5ms, so there's not a lot of win available on that side - the expensive part is the rendering.

Inlining the partial is roughly a 2x speedup over `render @collection` ([https://medium.com/@coorasse/partial-rendering-performance-in-rails-101fdfb6ffb9](seems a common result)). So, with full awareness that I am doing something kind of awful, I tried to reduce the likelihood of footguns - prevent recursion or chaining, require conspicuous templating literals, emit conspicuous warnings at startup.

--

As long as I am braindumping: the part of the code I am least satisfied with is in `app/views/comment/_thread.html.erb`, which has a lot of clever and unstructured string munging to nest comments properly in a single pass. A comment has to be rendered without knowing if it has children so 1. the `.comment_parent_tree_line` becomes brittle css, and 2. the `<ol>` has to be inferred when rendering the child and `</ol>` when rendering the next comment at equal or lesser depth, or end of the comment tree. And because a comment has to be rendered without knowing its siblings, the `.comment_sibling_tree_line` is also a clever bit of css, though at least it's not brittle. (The brittle parent line could be a short + reliable use of `:has`, which currently is almost universally supported - it's [in testing](https://caniuse.com/css-has) in Firefox.)

All this ugliness could be fixed by adding a `reply_count` [counter cache](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache) to `comments` to track the number of direct replies to each comment. That solves the first two problems of knowing if a comment has children when rendering it are solved, and adding to a hash like `reply_counts[comment.id] = comment.reply_count` as comments are rendered would let children look up `reply_counts[comment.parent_comment_id]` to know if they have siblings.

I think the current solution is acceptable as long as it's not breaking, and I think we'll touch it infrequently enough that this is acceptable. I am significantly dissatisfied that I couldn't find a working HTML validator gem to make sure we are outputing properly nested and closed `<ol>` and `<li>`s; I tested by hand-pasting into the [w3c validator](https://validator.w3.org/#validate_by_input). If this becomes a problem, I'd start to think it was worth the cost of widening `comments` even more.

app/models/comment.rb
pushcx: Rails does not have good support for UDF/stored procedures, so I worry there'd be pain to maintain that integration.

app/models/comment.rb
pushcx: Added in acd4b8fd

Issue #1202, opened 2023-09-14
Deleted or removed stories can't be unsaved


cipriancraciun:
If I access my saved stories, via <https://lobste.rs/saved>, I have a few items grayed-out -- that I've saved previous to what follows -- because afterwards they have been removed, either by the submitter or a moderator.

However, now I can't unsave them from my list, thus getting stuck in there, because there is no unsave button in the interface.

I've manually tried to replicate the `POST` request that should have been made to unsave them (by copying from the browser the request made for unsaving a valid item and replacing the identifier in the URL), and it all worked flawlessly.

This makes me believe that simply allowing the `unsave` button to be shown also for removed stories would solve this.

PR #1204, opened 2023-09-14
Display unsave link for previously saved stories

thiagofportella:
Closes https://github.com/lobsters/lobsters/issues/1202

Screenshots:
Captura de Tela 2023-09-14 às 14 18 31
Removed story previously saved; `unsave` link available.

Captura de Tela 2023-09-14 às 14 18 53
Removed story; `save` link still unavailable.

### DIFF
--- a/spec/features/read_story_spec.rb
+++ b/spec/features/read_story_spec.rb
@@ -59,4 +59,44 @@ RSpec.feature "Reading Stories", type: :feature do
       expect(page).to have_selector("span.merge")
     end
   end
+
+  feature "reading saved stories" do
+    let(:user) { create(:user) }
+
+    before do
+      story.update!(user:, editor: user)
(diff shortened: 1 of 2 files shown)


spec/features/read_story_spec.rb
```diff
@@ -59,4 +59,44 @@
       expect(page).to have_selector("span.merge")
     end
   end
+
+  feature "reading saved stories" do
+    let(:user) { create(:user) }
+
+    before do
+      story.update!(user:, editor: user)
```
pushcx, line 67: Please edit these tests to create new stories in the correct state instead of editing., I find it really hard to follow as the test file gets longer because I have to look up through each `before` and replay them in my head.

So eg `let(:user_edited_story) { create(:story, editor: user, [is_deleted: true]) }` and so on

PR #1205, opened 2023-09-15
Appease standardrb

byroot:
`or` -> `||`


pushcx:
Thanks for catching this. I'm still settling into new workflow so I've let a couple of these silly errors slip in.

PR #1214, opened 2023-10-03
Fix comment failing test

JuanVqz:
This is the error I'm seeing [in the latest commit](https://github.com/lobsters/lobsters/commit/5338e5a4b335b94337636a59b7bbc83529dfce81)
when running it locally.

Seems to be related to a MySQL flag `sql_mode=only_full_group_by` enabled by default.

```ruby
(byebug) Comment.where(user: user).group(:thread_id).order("id desc").limit(20).pluck(:thread_id)
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error:
Expression #1 of ORDER BY clause is not in GROUP BY clause and
contains nonaggregated column 'lobsters_test.comments.id' which is not
functionally dependent on columns in GROUP BY clause;
this is incompatible with sql_mode=only_full_group_by

nil
```

I don't think is a good idea disabling that flag, so, I just updated the query.

Old query
```ruby
(byebug) Comment.where(user: user).group(:thread_id).order("id desc").limit(20)

(byebug) Comment.where(user: user).group(:thread_id).order("id desc").limit(20).to_sql
"SELECT `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 GROUP BY `comments`.`thread_id` ORDER BY id desc LIMIT 20"
```

New query
```ruby
(byebug) Comment.where(user: user).select(:thread_id).order("id desc").limit(20)

(byebug) Comment.where(user: user).select(:thread_id).order("id desc").limit(20).to_sql
"SELECT `comments`.`thread_id` FROM `comments` WHERE `comments`.`user_id` = 1 ORDER BY id desc LIMIT 20"

(byebug) Comment.where(user: user).select(:thread_id).order("id desc").limit(20).pluck(:thread_id)
[1]
```

Links:
  * https://stackoverflow.com/questions/41887460/select-list-is-not-in-group-by-clause-and-contains-nonaggregated-column-inc
  * https://stackoverflow.com/questions/54547279/mysql-contains-nonaggregated-column-error-when-using-group-rails-5-2
  * https://gitlab.com/gitlab-org/gitlab-foss/-/issues/48428


app/models/comment.rb
```diff
@@ -601,7 +601,7 @@ def self.recent_threads(user)
 
     thread_ids = Comment
       .where(user: user)
-      .group(:thread_id)
+      .select(:thread_id)
```
JuanVqz, line 604: One more benefit on this update is that we are only requiring the thread_id attribute instead of the whole model's attributes

pushcx: The `pluck` a few lines down already does this.

pushcx:
I don’t see this test failure in dev or prod. It must have been difficult to track this down to a server setting!

Your proposed fix is semantically different, though. It will find the `thread_id` for the user’s last 20 comments, not the last 20 threads they commented in.

It would be the same to use `.distinct` instead of `.group`. If that fixes the test for you, please update this PR to it.

JuanVqz:
> I don’t see this test failure in dev or prod. It must have been difficult to track this down to a server setting!
> 
> Your proposed fix is semantically different, though. It will find the `thread_id` for the user’s last 20 comments, not the last 20 threads they commented in.
> 
> It would be the same to use `.distinct` instead of `.group`. If that fixes the test for you, please update this PR to it.

Hi @pushcx unfortunately, `.distinct` didn't work out

```ruby
(byebug) Comment.where(user: user).distinct(:thread_id).order("id desc").limit(20).pluck(:thread_id)
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: Expression #1 of ORDER BY clause is not in SELECT list, references column 'lobsters_test.comments.id' which is not in SELECT list; this is incompatible with DISTINCT

nil
```

pushcx:
Oh, I get what’s happening. The `order by` generated uses the `comments.id`, which is not selected. That’s invalid because we’re not grouping on it, and I can see how a `distinct` implies the same requirement.

Does `distinct(:thread_id)` work if you `order(thread_id: desc)`? The `thread_id` is a synthetic incremented id, it doesn’t reset between stories, and it does have an index, so sorting by it should solve the hassle.

pushcx:
I see the change to use `distinct` and `order` you pushed a few minutes ago. Looks like the test suite is happy, and it looks like what I thought might work in my last comment. Are you happy with it? If so, I'll merge your fix.

JuanVqz:
@pushcx if we sort by `thread_id` then we can return to use `group`, agreed?

pushcx:
I think in this query they'll give the same results. Maybe call `.to_sql` on each and then run `explain select...` with those two queries to see if they basically execute the same, or one hits an index and other doesn't? I don't have any other reason to decide between them.

JuanVqz:
Not the same in the result but the group is the one you had originally.

```ruby
(byebug) Comment.where(user: user).group(:thread_id).order(thread_id: :desc).limit(20).to_sql
"SELECT `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 GROUP BY `comments`.`thread_id` ORDER BY `comments`.`thread_id` DESC LIMIT 20"
```


```ruby
(byebug) Comment.where(user: user).distinct(:thread_id).order(thread_id: :desc).limit(20).to_sql
"SELECT DISTINCT `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 ORDER BY `comments`.`thread_id` DESC LIMIT 20"
```

pushcx:
My explanation was a little short. I was trying to say - run `rails db`, then copy and paste those each query with `explain` in front of them, so `explain select distinct comments.*...`. That will show how mariadb understands and optimizes them.

JuanVqz:
@pushcx thank you for your explanation. I updated the PR

PR #1232, opened 2023-12-22
Fix long lines in moderation action description

### LINKED ISSUE #1231: Long lines in moderation action description cause table to extend beyond container
Latest Orion and Safari (so, WebKit) for on macOS 14.
 
Screenshot 2023-12-16 at 1 14 15 PM

The line is causing it is almost certainly the long Google cache URL. It should be either breaking or capping the overflow here.

myhro:
Closes #1231.

Lines containing a long URL, for instance, would cause the text to overflow its content box.

In the original issue [it was suggested][1] to use `word-break: break-word;` but [this property value is considered to be deprecated][2].

The fix was specifically applied to the `Action` column in order to not break the username in the `Moderator` column. 

# Preview

image

P.s.: I'm not a Frontend Developer, so please let me know if this has to be done in a completely different way.

[1]: https://github.com/lobsters/lobsters/issues/1231#issuecomment-1858890718
[2]: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -1304,6 +1304,9 @@ table.data tr.nobottom td {
 	border-bottom: 0px;
 	padding-bottom: 0px;
 }
+table.data tr.nobottom td.wrap {
+	overflow-wrap: anywhere;
+}
 table.data.tall td {
 	vertical-align: top;
 }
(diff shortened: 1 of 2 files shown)


pushcx:
Looks like a solid solution, and straightforward to revise if needed. Thanks for jumping in and fixing this!

PR #1256, opened 2024-02-26
Add domain to RSS author

timhillgit:
When I'm browsing lobste.rs it's helpful to see the domain without hovering over the link. This adds that ability to the RSS field. It saves space by removing the email-address-like part of the username.


pushcx:
The RSS spec says that’s [supposed to be](https://www.rssboard.org/rss-specification#ltauthorgtSubelementOfLtitemgt) a formatted email like “user@example.com (Alice Example)”. But looking in my feed reader that basically never happens (11 out of 1,621 distinct authors) so I think it’s real unlikely anyone will mind that we’re violating the spec for this.

pushcx:
Fixed bug in f381d3b, added a smoke test in d9aba71a.

Issue #1257, opened 2024-02-26
RSS feeds broken again


ssokolow:
My Thunderbird did the "warning triangle and paused feed" thing and, when I investigated, I found that https://lobste.rs/t/rust.rss *is* still correct according to the meta tag embedded in https://lobste.rs/t/rust but returns a blank response.

kevincox:
I am also seeing a 500 for https://lobste.rs/top/rss

```
% curl -i https://lobste.rs/top/rss
HTTP/2 500 
server: nginx
date: Tue, 27 Feb 2024 12:49:21 GMT
content-type: text/html; charset=UTF-8
content-length: 0
x-request-id: f086645f-cb8a-4af8-b457-b99158720d3b
strict-transport-security: max-age=63072000; includeSubDomains; preload
content-security-policy: frame-ancestors 'none'
```

pushcx:
Thanks for reporting this. This was a bug in #1256, which assumed all stories had links. And I assumed rss feeds had a test when I merged it. I fixed the bug in f381d3b, added a smoke test in d9aba71.

PR #1258, opened 2024-02-26
Revert "Add domain to RSS author"

timhillgit:
Reverts lobsters/lobsters#1256

@pushcx `/rss` is currently returning a 500 and without having access to the logs this PR is almost certainly the cause. I clearly don't understand what information is available on the story object like I thought I did, I'll make sure to investigate more thoroughly before resubmitting anything.


pushcx:
I didn't see this before I fixed it this morning - details in #1257. Thanks for jumping in with a patch, I appreciate the effort.

Issue #1252, opened 2024-02-11
Add prosopite


pushcx:
Add [https://github.com/charkost/prosopite](prosopite) to detect 1+n queries.  Configure it to raise in dev/test and use the rails logger in prod (I’ll probably turn this off in prod if nothing shows up in a few weeks).

mikalai-yankouski:
please take a look

pushcx:
Merged #1260 by @mikalai-yankouski to complete this.

byroot:
> (I’ll probably turn this off in prod if nothing shows up in a few weeks).

FYI: It's still active in production, and adds a non-negligible overhead, not sure if it was intended.

pushcx:
We don't have 100% code path test coverage so the production logging is the only way to find some 1 + n queries. I've gone and looked at the log a few times to fix them, but it's been a little while since the last attempt. I figured at some point there will be few enough that I'll set it to throw instead of log in prod, but I haven't gotten there. 

I haven't found a tool or practice that reliably prevents ActiveRecord from producing more of them. Suggestions would be very welcome.

byroot:
I think the vanilla way is to use `strict_loading`: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-strict_loading

pushcx:
I can see how that's a start, but I don't see how to enforce its use. A rubocop rule would have to catch any mention of a model followed by a method that touches the broad and growing AR query API, and then also check the models for `self.` invocations, and I'm sure there's more I'm missing.

byroot:
> I don't see how to enforce its use.

Via default scopes? e.g. to enable globally:

```ruby
class ApplicationRecord
  default_scope { strict_loading }
end
```

But can also be done gradually on a per class basis.

pushcx:
Huh. I'm so used to `default_scope` as a footgun that it didn't occur to me.

I regularly walk associations in the console for development and debugging, so I started writing an initializer to remove the `default_scopes`.*

That worked, but the error message and traceback for `strict_loading` and `StrictLoadingViolationError` are pretty confusing. They don't lead to where it's configured so I can leave an explanation, and the top google hits are not great. A lot of Lobsters contributors are new to Rails and I've seen them get stuck on things like this.

As I was looking around for how to improve that, I ran into an undocumented option `strict_loading_by_default` that I can set per-environment to recreate the Prosopite behavior of logging in prod and raising in dev and test. I'll try setting this up on Monday.


[...]

PR #1260, opened 2024-02-29
Added gem prosopite

mikalai-yankouski:
https://github.com/lobsters/lobsters/issues/1252

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -408,6 +408,7 @@ class Story < ApplicationRecord
       return
     end
 
+    # ignored to manage tags_a for nicer UI and because the n is typically 2-5 tags
     Prosopite.pause
     taggings.each do |t|
       if !t.tag.can_be_applied_by?(u) && t.tag.privileged?
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -408,6 +408,7 @@ class Story < ApplicationRecord
       return
     end
 
+    Prosopite.pause
     taggings.each do |t|
       if !t.tag.can_be_applied_by?(u) && t.tag.privileged?
         raise "#{u.username} does not have permission to use privileged tag #{t.tag.tag}"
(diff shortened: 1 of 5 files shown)


app/models/story.rb
```diff
@@ -408,6 +408,7 @@ def check_tags
       return
     end
 
+    Prosopite.pause
```
mikalai-yankouski, line 411: skipped because `taggings` is not created yet and cannot join `tag` to them

config/initializers/prosopite.rb
```diff
@@ -0,0 +1,5 @@
+if Rails.env.production?
+  Prosopite.custom_logger = Logger.new("log/n_plus_one_detection.log")
```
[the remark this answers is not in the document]
mikalai-yankouski: done

app/models/story.rb
pushcx: I spent a few minutes tinkering with this, the `resume` can move up to line 425, the section `if taggings.reject...` isn't tripping it.

pushcx (changes requested):
There's one small change to minimize how much is ignored, and I'll merge with that.

This is really nice to have, thanks for working through the setup.

app/models/story.rb
pushcx, line 411:
```suggestion
    # ignored to manage tags_a for nicer UI and because the n is typically 2-5 tags
    Prosopite.pause
```

pushcx:
Oh, the build is failing because the new initializer doesn't start with `# typed: false` (see the others for examples), so please add that line.

mikalai-yankouski:
> There's one small change to minimize how much is ignored, and I'll merge with that.
> 
> This is really nice to have, thanks for working through the setup.

you are welcome

pushcx:
Thanks for working through these reviews, I'm very happy to merge this for better protection against perf regressions.

PR #1266, opened 2024-03-27
Fix unnecessary tag wrapping in Firefox.

Nimdoc:
Noticed this small thing while browsing lobste.rs on Firefox on my phone today.

thats-a-wrap


pushcx:
Thanks for spotting and fixing this!

Issue #1281, opened 2024-05-29
Add database_consistency and database_validations gems


pushcx:
https://github.com/djezzzl/database_consistency

https://github.com/toptal/database_validations

Lobsters would benefit from the improved performance + data quality provided by these gems.

These each could be a significant amount of work. It’s OK to work through their addition in several PRs if you prefer.

hmaddocks:
I have experience with `database_consistency` and a bunch of other similar tools. I can start knocking some of these off.
One question; is lobsters OK with foreign key constraints? I know some teams prefer not to use them.

pushcx:
That would be great, thank you for volunteering.

Lobsters [uses FKs now](https://github.com/lobsters/lobsters/blob/99ea2af4014ae78250315d3315969f42459d3317/db/schema.rb#L336-L376) and I'm happy to continue that. (Speaking of `schema.db`, I do slightly prefer using `schema.sql` but there's never been a reason to touch it; you can swap if you it offers value to this project.)

petrenkorf:
@pushcx Looks like the PR above is already merged. May we close this issue?

(Yeah, I still looking for good first issues :D)

hmaddocks:
There are still some improvements to make. I have a local branch with some of them but I got really busy sorry.

Eduardo-Lpz:
As another round, this [PR](https://github.com/lobsters/lobsters/pull/1351) addresses some of the warnings flagged by `active_record doctor`, i will like to solve all of them but probably will be later next week.

PR #1286, opened 2024-06-09
Database consistency fixes

hmaddocks:
First pass at making updates recommended by `database_consistency` and `active_record_doctor` as per issue #1281 
These are the straight forward updates. There's some "non-standard" stuff in the schema that's resulting tricky fixes and  false positives. I think this PR is boarding on too big so I'll tackle the rest in another PR.

Note: There is a possibility that the migration that adds NOT NULL constraints might fail if those columns have NULL in them.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -107,13 +107,15 @@ class Story < ApplicationRecord
       .limit(10)
   }
 
-  validates :title, length: {in: 3..150}
+  validates :title, length: {in: 3..150}, presence: true
   validates :description, length: {maximum: (64 * 1024)}
   validates :url, length: {maximum: 250, allow_nil: true}
   validates :short_id, presence: true, length: {maximum: 6}
   validates :markeddown_description, length: {maximum: 16_777_215, allow_nil: true}
   validates :mastodon_id, length: {maximum: 25, allow_nil: true}
   validates :twitter_id, length: {maximum: 20, allow_nil: true}
+  validates :is_deleted, :is_moderated, :user_is_author, :user_is_following, inclusion: {in: [true, false]}
+  validates :score, :flags, :hotness, :comments_count, presence: true
 
   validates_each :merged_story_id do |record, _attr, value|
     if value.to_i == record.id
@@ -1020,7 +1022,7 @@ class Story < ApplicationRecord
 
   def self.title_maximum_length
     validators_on(:title)
-      .sole { |v| v.is_a? ActiveRecord::Validations::LengthValidator }
+      .find { |v| v.is_a? ActiveRecord::Validations::LengthValidator }
       .options[:maximum]
   end
 
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -116,6 +116,7 @@ class Story < ApplicationRecord
   validates :twitter_id, length: {maximum: 20, allow_nil: true}
   validates :is_deleted, :is_moderated, :user_is_author, :user_is_following, inclusion: {in: [true, false]}
   validates :score, :flags, :hotness, :comments_count, presence: true
+  validates :normalized_url, length: {maximum: 255, allow_nil: true}
 
   validates_each :merged_story_id do |record, _attr, value|
     if value.to_i == record.id
--- a/Gemfile
+++ b/Gemfile
@@ -63,4 +63,6 @@ group :test, :development do
   gem "vcr"
   gem "webmock" # used to support vcr
   gem "simplecov", require: false
+  gem "active_record_doctor"
+  gem "database_consistency"
 end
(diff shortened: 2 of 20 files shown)


pushcx:
Checking for nulls in prod data:

 * domains.domain: 0
 * categories.category: 0
 * users.is_admin: 0
 * users.is_moderator: 0
 * users.pushover_mentions: 0
 * comments.is_deleted: 0
 * comments.is_moderated: 0
 * comments.is_from_email: 0
 * moderations.is_from_suggestions: 0
 * messages.has_been_read: 0
 * messages.deleted_by_author: 0
 * messages.deleted_by_recipient: 0
 * stories.user_is_author: 0
 * invitation_requests.is_verified: 0
 * hats.modlog_use: 0

So there's no hassles there at least.

To make sure I understand: you think this PR is ready to go? I'll clone the db to my local machine and test there before merging.

app/models/story.rb
```diff
@@ -1020,7 +1023,7 @@ def fetched_attributes
 
   def self.title_maximum_length
     validators_on(:title)
-      .sole { |v| v.is_a? ActiveRecord::Validations::LengthValidator }
+      .find { |v| v.is_a? ActiveRecord::Validations::LengthValidator }
```
pushcx, line 1026: Why doesn't `sole` work here anymore? Does ActiveRecord implement the `presence` validator as a second `LengthValidator`?

pushcx:
Found [one story](https://lobste.rs/s/jltytq) with `select length(title)` of 152; I edited the smart quotes to straight ascii quotes and now get 148.

app/models/story.rb
hmaddocks: `sole` expects 1 but there are now two validations on `title`, the `length` and now the `presence`. Sole checks the length of the enumerable before it applies the block.

hmaddocks:
Applying new validations on existing data is always a bit dodgy :) Good to see there wasn't much invalid data. 
I forgot to commit the `Gemfile.lock` 
I don't know why the workflow is failing though. I'll take a look a bit later.

app/models/story.rb
pushcx: Oh, thank you. My incorrect mental model was that it was like `select`, returning a single object for which the block returned non-`nil` and raising an exception if there were 0 or more than one items. ([Array#sole doc](https://edgeapi.rubyonrails.org/classes/Enumerable.html#method-i-sole))

The block I wrote is basically a no-op. It sounds like the right code would be `.find { |v| v.is_a? ActiveRecord::Validations::LengthValidator }.sole.options[:maxiumum]` - does that look right to you?

hmaddocks: It's a bit weird. I don't know why it takes a block. Selecting the items and then asserting that there is only one makes sense. Thinking about it a bit more maybe `select` would be better than `find` but the spec passes...

```
irb(main):001> [1,2,3].sole {_1.even?}
/Users/henry/.gem/ruby/3.3.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/enumerable.rb:210:in `sole': multiple items found (Enumerable::SoleItemExpectedError)
irb(main):002> [1,2,3].select {_1.even?}
=> [2]
irb(main):003> [1,2,3].select {_1.even?}.sole
=> 2
irb(main):004> 
```

hmaddocks:
Hi, are you waiting on anything from me to merge this?

app/models/story.rb
pushcx: I get it, it doesn't take a block, but Ruby functions accept a block by default, and the language doesn't enforce they'll get called because there are lots of useful functions that only conditionally call their blocks.

```
irb(main):001> def foo
irb(main):002*   puts 'foo'
irb(main):003*   end
=> :foo
irb(main):004> foo { puts 'bar' }
foo
=> nil
```

```
irb(main):001> [1,2,3].sole {puts 'bar'}
[path elided]/enumerable.rb:210:in `sole': multiple items found (Enumerable::SoleItemExpectedError)
```
Note that neither example prints `bar`.

pushcx:
The workflow failure is the style linter. If you don't want to integrate standardrb into your editor, you can run `bundle exec standardrb --fix` and all of these are autofixable.

I was waiting for you to say you think it's ready to merge. I pulled the code and ran `bundle exec database_consistency`, it looks like there's a couple dozen failures. Do you want to do those in this PR, or leave for later? I'm asking because once we do, I can add that command to the workflow so it catches the introduction of new issues.

pushcx:
Great, merging now. I'll skip the todo file if you think you'll get to these anytime soon. Thanks for picking up this unglamorous work, I appreciate the help.

PR #1290, opened 2024-06-20
Make the application compatible with `--enable-frozen-string-literal`

casperisfine:
As you may know, Ruby 4 is likely to make frozen string literals the default, in that context I wanted to see how much work would be needed for a relatively classic app.

Turns out it wasn't much work for lobsters.

I'm not particularly asking for this to be merged, just submitting it in case you'd want this change. It would allow you to run ruby with `--enable-frozen-string-literal` in production, which should reduce the GC pressure a bit (but don't expect crazy numbers).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -34,6 +34,10 @@ jobs:
         run: |
           bundle exec rspec
           bundle exec rspec spec/slow/*_spec.rb
+        env:
+          # Enabling frozen string literals in rspec only because brakeman
+          # isn't compatible yet: https://github.com/presidentbeef/brakeman/pull/1855
+          RUBYOPT: "--enable-frozen-string-literal --debug-frozen-string-literal"
       - name: Run linters
         run: |
           bundle exec standardrb
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,7 +18,7 @@ module ApplicationHelper
   end
 
   def errors_for(object)
-    html = ""
+    html = +""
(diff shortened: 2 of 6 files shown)


.github/workflows/check.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         run: |
           bundle exec rspec
           bundle exec rspec spec/slow/*_spec.rb
+        env:
+          RUBYOPT: "--enable-frozen-string-literal --debug-frozen-string-literal"
```
casperisfine, line 38: NB: only enabling it for `rspec` because `brakeman` isn't yet compatible.

pushcx: Could you add a comment about this, maybe link to brakeman's tracking issue? I'm never going to remember why there's this distinction and #1286 has some upcoming edits to this file.

pushcx:
Thanks! This seems like a great bit of preparation.

Because this is one of those changes that's going to change global behavior and we don't have anything like 100% test coverage, I'm going to wait a minute to merge this until I can deploy it and keep an eye on exceptions for a few hours.

.github/workflows/check.yml
casperisfine: I haven't opened an issue on brakeman yet, they have a couple dependencies that need a fix first. But I'll open it.

casperisfine: Comment added.

pushcx:
The build failure is not relevant; I had to hack out code in prod for shopify/ruby#556.

PR #1291, opened 2024-06-29
Fix mismatched tags on search page

KingSkyrock:
`tag:meta` in the search page is inconsistent:
image

Replaced a mismatched `<code>` tag to fix the inconsistency.


pushcx:
Thanks for this fix! How’d you happen to find it?

KingSkyrock:
> Thanks for this fix! How’d you happen to find it?

Just happened to notice that `tag:meta` was missing the background and that these links in the bottom right were monospace when they shouldn't be.
image

pushcx:
Good eye, thanks for running that down.

PR #1292, opened 2024-07-06
Dump heap after boot and after bloat

byroot:
Ref: https://github.com/Shopify/ruby/issues/556#issuecomment-2211754838

Note: This is totally untested

`out_of_band` is invoke when a request finishes and no other thread is busy. So theorically it may never be invoked if the worker is under constant load, but I'm hoping it will eventually.

The idea here is to wait for the number of live object to go over some threshold we found in the GC stqats we dumped previously, and when it's reached we dump the heap a second time.

By comparing it with the heap that was dumped after boot, we should be able to figure out what extra objects are being retained after the worker ages.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -37,15 +37,40 @@ workers ENV.fetch("PUMA_WORKERS") { 3 }
 worker_boot_timeout 180
 
 # https://github.com/Shopify/ruby/issues/556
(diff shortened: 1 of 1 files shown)


pushcx:
My schedule's been a bit overwhelming but I'm testing this in prod now.

PR #1293, opened 2024-07-13
Removed old error status message

duck-of-death:
Howdy, I removed an old error/info message as seen on the 504.html page


pushcx:
Looks great, thanks for cleaning this up.

(This came up in the chat room; it's been outdated since the search engine rebuild.)

PR #1294, opened 2024-07-17
Fixed text highlight color

shurj0:
System colors are deprecated in CSS3, and as such the Highlight and HighlightText keywords don't function properly on Chrome. 

https://www.w3.org/TR/CSS2/ui.html#system-colors

Removing these lines makes browsers just use the defaults, which looks more similar (albeit not the same) across the board. Don't know if you want to hard code in a selection color value, some users might want to define their own.

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -181,10 +181,6 @@ html.color-scheme-dark {
 /* overall page */
 
 html {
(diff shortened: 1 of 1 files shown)


pushcx:
Ah dang. Not a lot of value to picking a custom color, let's leave it to the user agent.

Thanks for figuring out this odd bug!

PR #1302, opened 2024-08-15
Provide an example database.yml setup powered by a container

nogweii:
Making the local development setup a bit easier. Not sure if this is how you want to set it up, or how it might interfere with the deployment process.


pushcx:
Sure, I think it's an improvement to have a clear example, especially for folks who aren't so familiar with Rails conventions. 

The config file needs to be named `config/database.yml.sample`. The path you have now is in `.gitignore` so that devs can have their local config.

pushcx:
Thanks for helping improve the experience for new contributors!

PR #1305, opened 2024-08-18
Remove redundant limit from #pick call

the-spectator:
`#pick` is a short-hand for `relation.limit(1).pluck(*column_names).first`, hence the `limit` clause is redundant here.


pushcx:
Ah, yup. Thanks! I'm curious, how'd you happen to catch this?

the-spectator:
Was just reading through code. 😀

Issue #1238, opened 2024-01-05
mariadb deadlocks


pushcx:
https://github.com/lobsters/lobsters/issues/1612

I wanted to cross-post this bug here because I can't just move it. It's in the Rails code, but all the tech details are already over there.

This deadlock happens because comment (and, much more rarely, story) creation is locking tables in a different order than vote creation (slash updating?). Maybe because of assigning the user's initial upvote or some memoized value like score/karma/votes/number of comments/etc.

The job here is to vote/unvote stories and comments in dev, and read the rails server log for the order that tables are touched in. Then same for story and comment posting. I am almost certain that will uncover tables accessed in a different order. Just posting those notes as a comment here would be very helpful!

If that is indeed the cause of the deadlock, then we have to shlep through the controllers and callbacks to figure out how to reorder, simplify, or break apart the transactions. Probably nothing even needs to be in a transaction, we're just getting it from Rails by default. But since I touched all the score code in 20c15907 I think voting should always work by doing `score += 1` or `score = sum(votes)` rather than an unsafe `select ... add 1 in ruby ... update`, so a transaction is unneeded.

pushcx:
This got much worse in the last couple days because I had the page cache disabled to correctly show the banner pointing UK users to the [call for help with their OSA](https://lobste.rs/s/ukosa1). Instead of being a once-a-month bug it was once or twice a day.

Maybe the most straightforward fix to this would be to move `Story#calculated_hotness` and `Comment#calculated_confidence` into idempotent database queries, so we'd have these code paths insert/update/delete votes and then call `story.calculate_confidence!` which would issue a minimal `update`. That would eliminate the need to lock tables data is read and written back in a transaction.

pushcx:
Nope, got a deadlock on a new comment this morning:

```
An ActiveRecord::Deadlocked occurred in comments#create:

  Trilogy::ProtocolError: 1213: Deadlock found when trying to get lock; try restarting transaction (trilogy_read_row)
  app/controllers/comments_controller.rb:50:in 'CommentsController#create'
```

pushcx:
by requests, all hits from a couple seconds before the exception through the deadlock

```
{"method":"GET","path":"/~Vanishftw","format":"*/*","controller":"UsersController","action":"show","status":404,"allocations":2385,"duration":10.78,"view":3.95,"db":3.12,"timestamp":"2025-02-1
[...]

PR #1308, opened 2024-08-22
Precompute Comment scores on initial insertion

byroot:
I was looking at https://github.com/lobsters/lobsters/issues/1238 and the comment insertion transaction is indeed very meaty, over 22 queries.

Looking into how to merge or eliminate some of them, I noticed that the comment is quickly updated right after insertion, from `update_score_and_recalculate!`, and that it's quite a big query for something that always amount to the same result:

  - A new comment only has the author's vote.
  - It also has no flags.
  - And the confiderence is directly derived from the votes.

As such we can skip all this and just insert default values.

This reduce the number of queries in that transaction from 22 to 16.

The eliminated queries however are likely not the one causing contention, It's way more likely to be the one that update the story, but it may still help a bit by making the transaction shorter.


spec/models/comment_spec.rb
```diff
@@ -151,6 +151,28 @@
...
+
+    it "uses precomputed score on initial insertion" do
+      comment = Comment.build(user: author, story: story, comment: "cool story!")
+      comment.assign_initial_confidence
+
+      expect {
+        expect {
+          expect {
+            expect {
+              comment.update_score_and_recalculate! 0, 0
```
byroot, line 166: The idea here is to trigger the DB side computation, and ensure the result is the same than what we precomputed.

pushcx:
This is a wonderful improvement, thanks for spotting the opportunity. It looks like there's a noisy test or two but I really look forward to merging this.

byroot:
Right, I didn't run the full suite locally, and didn't spot these. I updated them. Please be careful though, I don't know the codebase much, my assumptions may be plain wrong.

byroot:
Alright, I'm not super proud of the fix, but tests are now green. Now it's brakeman complaining because a new version was released or something.

PR #1310, opened 2024-08-23
Replace mysql2 with trilogy

byroot:
Not sure if that's something you care about.

Trilogy is very marginally more performant, but more importantly has no dependencies so avoid the painful installation issues of `mysql2`.

It will become the default adapter for MySQL/MariaDB in Rails 8.


nogweii:
@byroot could you update the `database.yml.sample` since it still references the mysql driver?

byroot:
For people running into https://github.com/cainlevy/scenic-mysql_adapter/pull/2#issuecomment-2311158831

```
BaseConnectionError: trilogy_auth_recv: caching_sha2_password requires either TCP with TLS or a unix socket: TRILOGY_UNSUPPORTED (Trilogy::BaseConnectionError)
```

A short term workaround is: https://stackoverflow.com/a/49935803, I'll see if we can improve trilogy to handle that.

pushcx:
Deploying revealed something I should've realized from @nogweii's comment, the prod `config/database.yml` needed to be updated. I spent a minute confused by seeing `LoadError: Error loading the 'mysql2' Active Record adapter. Missing a gem it depends on? mysql2 is not part of the bundle. Add it to your Gemfile. (LoadError)` which was Rails trying to be helpful and suggest adding back the outdated adapter it saw in `database.yml`. The coffee's still kicking in but I thought the bundle was in some kind of inconsistent state.

I know the list of [sister sites](https://github.com/lobsters/lobsters/wiki) is out-of-date (I just saw an Armenian instance somewhere recently). I have been reluctant to promise support to sister sites, but I hate not having a way to give them a heads up before they have a deploy break because they need to update this config file.

byroot:
Yeah, it's entirely possible to leave `mysql2` in the Gemfile for a while if you want to not break them.

pushcx:
But there's no way to tell them it's there. Ideally they'd skim every commit as they merge but that's not a realistic expectation. For most admins it is their first time using Rails.

PR #1312, opened 2024-08-26
Stop loading rubocop in production workers

byroot:
As noticed in heap dumps.


pushcx:
Thanks for discovering this!

PR #1314, opened 2024-08-31
Upgrade to Rails 7.2

byroot:
3 gems had compatiblity issues and had to be upgraded:

  - factory_bot
  - activerecord-typedstore
  - database_cleaner-active_record

Other than that, nothing to report the suite passed immediately.

NB: `application.rb` still uses `config.load_defaults 7.1`, as it's best practice to first deploy the new version with the old default. Even though IIRC it's not strictly necessary for that specific upgrade.


config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,71 @@
+# typed: false
```
byroot, line 1: BTW I'm intrigued by this. Are you using sorbet somehow? I see `standard-sorbet` in the Gemfile, but no actual sorbet.

pushcx: It has been on my maybe-one-day fun project list, but no, not yet.

Issue #1311, opened 2024-08-26
Upgrade to Rails 7.2


pushcx:
https://rubyonrails.org/2024/8/22/Rails-7-2-1-has-been-released

petrenkorf:
@pushcx Is this issue already taken by someone? Otherwise, I'm willing to take this as it is a "good first issue"

pushcx:
This was implemented in #1314, which I just caught up with and merged. But I hope to see your name on another 'good first issue'. :D

PR #1316, opened 2024-08-31
Enable Rails 7.2 defaults

byroot:
Followup on https://github.com/lobsters/lobsters/pull/1314, should ideally be merged only once the actual upgrade is deployed successfully.

PR #1317, opened 2024-09-04
Update to Rails7.2

cpgo:
Saw this issue https://github.com/lobsters/lobsters/issues/1311 and was curious how hard the upgrade was gonna be.
Just updated `database_cleaner` and `activerecord-typedstore`
Tests passed locally but haven't tried the application yet


MatheusRich:
One thing to be aware of is that 7.2 has a new default config that requires users to have modern browsers. Take this into consideration.

pushcx:
@MatheusRich Could you elaborate on that? I don't spot anything with a skim of [the release notes](https://guides.rubyonrails.org/7_2_release_notes.html).

pushcx:
This test failure is irrelevant; I'll push a fix in a few minutes and re-run.

MatheusRich:
@pushcx sure thing. It's [this section](https://edgeguides.rubyonrails.org/7_2_release_notes.html#add-browser-version-guard-by-default). Here's what they say:

> This means that all other unknown browsers, as well as agents that aren't reporting a user-agent header, will be allowed access.
> 
> A browser that's blocked will by default be served the file in public/406-unsupported-browser.html with a HTTP status code of "406 Not Acceptable".

And in the code example:

```ruby
class ApplicationController < ActionController::Base
  # Allow only browsers natively supporting webp images, web push, badges, import maps, CSS nesting + :has
  allow_browser versions: :modern
end
```

But fully reading the text I see:

> **Newly generated applications have this guard set in `ApplicationController`.**


So this only applies to newly generated apps. My bad!!!

pushcx:
[Background](https://github.com/rails/rails/pull/50467).

That is fundamentally opposed to the way this codebase works, progressive enhancement, and the longevity of websites. I would never enable or support it.

MatheusRich:
Yes, that why I wanted to give a heads up. I thought that was enabled by default on 7.2

cpgo:
Not that I took a better look at the open PR's I believe we should close this one in favor of [#1316](https://github.com/lobsters/lobsters/pull/1316) as it seems more complete.

PR #1319, opened 2024-09-06
Accessibility improvements for "unread" feature

illfygli:
I couldn't activate the "unread" link, that scrolls to the first unread comment, because it lacked a role. Now it has a role.

I also took the opportunity to check the "reduced motion" preference to determine scrolling behavior.


pushcx:
Thank you, these are excellent improvements. I'm deploying it now.

(The build failure is not relevant; I've fixed it.)

illfygli:
Thanks! Works for me now on the live site. :partying_face:

PR #1320, opened 2024-09-06
Avoid continiously instantiating HTMLEntities

byroot:
While profiling the lobsters benchmark in the yjit-bench suite I noticed a weird 0.8% of time spent in `instance_eval`.

Tracking it down, it's coming from the htmlentities gem, when calling `HTMLEntities.new` it generates code on the fly based on the options.

~~As such you aren't meant to instantiate a new one every time, but have a single global object somewhere.~~ Actually seems like you are, that gems just bad.

Otherwise this constant code generation is wasting performance for now reason, and potentially use YJIT executable memory if called enough.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -484,7 +484,7 @@ class Story < ApplicationRecord
       s = s.to_s[0, chars].gsub(/ [^ ]*\z/, "")
     end
 
-    HTMLEntities.new.decode(s.to_s)
+    HtmlEncoder.encode(s.to_s)
   end
 
   def domain_search_url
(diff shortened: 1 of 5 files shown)


byroot:
Another possibility would be to just use `CGI.escape_html`, but it doesn't encode as much as `HTMLEntities` with `:decimal` format.

casperisfine:
The smoking gun:

Capture d’écran 2024-09-06 à 17 21 46

pushcx:
So, the Right Fix for RSS generation would probably be to move over to the rss gem or the xml builder. It's a little odd to generate RSS by erb.I don't think we get any benefit from this custom approach and I'm leery of having so many `raw` calls hanging around.

The call from the Story model is mandatory because we're consuming others' HTML and don't want to store entity-encoded, so I'm going to merge this down.

You noticed we had a lot of these HTMLEntity allocations. RSS generation is a significant part of site traffic, so that's probably why we generate and free a ton. I would not be at all surprised if it was a major contributor to the memory fragmentation issue.

pushcx:
OK, I did the Right Fix in 629df028. So now there's only a single call site for this singleton.

Issue #986, opened 2021-04-04
An option to load avatars from GitHub


tie:
It’d be nice to load avatars from linked GitHub account (e.g. https://github.com/tie.png?size=256 where size stands for max size in pixels).

Looks like we’d need to fetch image for self.github_username in User’s fetched_avatar. We’d also need an option to opt-in into using GitHub as avatar source.

https://github.com/lobsters/lobsters/blob/9c0e2c03e2c14a2fba39298e76258cc460f6f003/app/models/user.rb#L334-L351

See also https://github.com/lobsters/lobsters/issues/547

pushcx:
This could be a really nice alternative to Gravatar, given our userbase. Twitter, too.

So:the user would need to link their GitHub or Twitter account, then a `<select>` for which service to pull from. Unlinking an account resets to Gravatar.

toastal:
I would prefer [Libravatar](https://www.libravatar.org/) (failed [merge request](https://github.com/lobsters/lobsters/pull/835)) as it is decentralized and open source. I do not think it's good to promote closed-source, data-tracking, privacy-issue-laden entities like GitHub, Twitter, Gravatar. It's one thing to allow them as options, but there currently isn't an option that users can control.

PR #1326, opened 2024-09-20
Port story and comment replies to Action Mailbox

### LINKED ISSUE #1285: Replace parse_inbound_email with Action Mailbox
The mailing list most receives incoming email to the [parse_inbound_email](https://github.com/lobsters/lobsters/blob/master/script/parse_inbound_mail) script, which has unfortunately been kind of janky. Most recently we had #1253, but there have been others. I remember it's been significant debugging every time there have been major hosting changes.

Years after the script was developed, Rails 6 (I think?) added [Action Mailbox](https://guides.rubyonrails.org/action_mailbox_basics.html) to better address this area. Let's update to be more Railsy and use Action Mailbox.

gmemstr:
Some finer points to clean up (and tests to write), but functionality for replying to stories and comments ported from old scripts to Action Mailbox. It's been tested as much as I can with my understanding of the feature and RoR.

I'm unsure whether Active Storage *is* something we can enable, but I haven't found a way to disable it with Action Mailbox so keeping it around for now.

Closes #1285


pushcx:
I'd like to move the ActiveStorage setup forward for #986 because of some unrelated outside events. You've basically done all the setup in this PR. I'm sorry for the chore, but could I ask you to cherry-pick that code into a new branch and PR? It's OK if it's not complete; I'd just like to have your name on the commit.

gmemstr:
Rebased onto master with new active storage config.

pushcx:
Again, sorry for the churn here. Thanks for helping me move avatars forward. I've started that code and hope to finish it on-stream in a few hours.

gmemstr:
Technically yes but not really sure what else needs doing.

app/mailboxes/inbox_mailbox.rb
```diff
@@ -0,0 +1,64 @@
...
+  def process
+    c = Comment.new(user: sending_user, comment: mail.decoded, is_from_email: true)
+    if parent.is_a?(Comment)
+      c.story_id = parent.story_id
+      c.parent_comment_id = parent.id
+    else
+      c.story_id = parent.id
+    end
+
+    # TODO: Figure out what to do here.
```
pushcx, line 15: What's this `TODO` about? Maybe this wants to just call `c.save!` so we get an exception in prod if a mail fails to parse?

It would be really nice if a failed mail could hang around in queue for a later bugfix/retry, but I'm hoping that a feature Action Mailbox has; it's probably not worth custom code.

app/mailboxes/inbox_mailbox.rb
```diff
@@ -0,0 +1,64 @@
...
+    elsif (m = irt.match(/^story\.([^\.]+)\.\d+@/))
+      @parent = Story.where(short_id: m[1]).first
+    end
+
+    @parent
+  end
+
+  def user_token
+    # RoR "to" gives us an array. Joining here might not be the best?
+    mail.to.join(" ").to_s.gsub(/^#{Rails.application.shortname}-/, "").gsub(/@.*/, "")
```
pushcx, line 52:
```suggestion
    mail.to.join(" ").to_s.gsub(/^#{Rails.application.shortname}-/, "").gsub(/@.*/, "")
    mail.to.select {|e| e =~ /^#{Rails.application-shortname}\-/ }[/\-([^@]*)@/, 1].to_s
```

Not sure about the `\` on the hyphen though.

pushcx (changes requested):
I see you've met our new stale housekeeping bot... sorry, I didn't realize this was waiting on me! I've got one small question and one small fix.

app/mailboxes/inbox_mailbox.rb
pushcx: This code change is so you can rm the comment. Your regexp didn't quite work; the ^ would match start-of-line so joining the `to` addresses with a space meant it would only look for a token in the first email in the `To` line. I tweaked it to be a little more explicit that we're trying to find an address that looks like our reply-to address and then pull the token out of it.

pushcx (approved):
This looks great, thank you for keeping up as `master` kept moving. I'm merging this down now and I'll activate it by directing postfix to use it, probably here on this holiday weekend.

PR #1327, opened 2024-09-21
Improve setup

edwardloveall:
Hi. I was setting up the codebase this morning and struggled a bit as I don't have a lot of mysql experience. I took some steps to get it working, and then updated some README/code:

* Add instructions to start the `mysql` server
* Add instructions to set the root password to match `database.yml.sample`
* Use the `db:setup` command in `bin/setup`. It does the same thing and will be a little faster since it doesn't have to load the rails app for three commands.

Totally open to feedback if there's a more idiomatic way to set up local mariadb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,17 +31,14 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
   On MacOS you can [install with brew](https://mariadb.com/kb/en/installing-mariadb-on-macos-using-homebrew/).
   On Windows there's an [installer](https://mariadb.org/download/?t=mariadb&p=mariadb&r=11.5.2&os=Linux&cpu=x86_64&pkg=tar_gz&i=systemd&mirror=starburst_stlouis).
 
-* Start mysql and set the `root` user password
+* Start the mariadb server, open the console, and set the `root` user password (type `ctrl-d` to exit afterwards)
 
 ```sh
-# Start the server
 mysql.server start
-# open a mysql console
-mysql
+mariadb
 ```
 
 ```sql
-FLUSH PRIVILEGES;
 ALTER USER 'root'@'localhost' IDENTIFIED BY 'localdev';
 ```
 
--- a/README.md
+++ b/README.md
@@ -31,12 +31,9 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
   On MacOS you can [install with brew](https://mariadb.com/kb/en/installing-mariadb-on-macos-using-homebrew/).
   On Windows there's an [installer](https://mariadb.org/download/?t=mariadb&p=mariadb&r=11.5.2&os=Linux&cpu=x86_64&pkg=tar_gz&i=systemd&mirror=starburst_stlouis).
 
-* Start the mariadb server, open the console, and set the `root` user password (type `ctrl-d` to exit afterwards)
+* Start the mariadb server using one of the [methods mentioned in the mariadb knowledge base](https://mariadb.com/kb/en/starting-and-stopping-mariadb-automatically/).
 
-```sh
-mysql.server start
-mariadb
-```
+* Open the console using `mariadb`, and set the `root` user password (type `ctrl-d` to exit afterwards)
 
 ```sql
 ALTER USER 'root'@'localhost' IDENTIFIED BY 'localdev';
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,6 +31,20 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
   On MacOS you can [install with brew](https://mariadb.com/kb/en/installing-mariadb-on-macos-using-homebrew/).
   On Windows there's an [installer](https://mariadb.org/download/?t=mariadb&p=mariadb&r=11.5.2&os=Linux&cpu=x86_64&pkg=tar_gz&i=systemd&mirror=starburst_stlouis).
 
+* Start mysql and set the `root` user password
```
pushcx, line 34:
```suggestion
* Start the mariadb server, open the console, and set the `root` user password (type `ctrl-d` to exit afterwards)
```

README.md
```diff
@@ -31,6 +31,20 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
...
+
+```sh
+# Start the server
```
pushcx, line 37: Please remove these comments, they're redundant with the instruction above.

README.md
```diff
@@ -31,6 +31,20 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
...
+mysql.server start
```
pushcx, line 38: This should be mariadb, not mysql. I'm pretty sure the mariadb fork still aliases the mysql commands, but as we've already had folks confused about which to install I'd like to be consistent.

README.md
```diff
@@ -31,6 +31,20 @@ There's an external project [docker-lobsters](https://github.com/utensils/docker
...
+# open a mysql console
+mysql
+```
+
+```sql
+FLUSH PRIVILEGES;
```
pushcx, line 44: I think the `flush` has to come after the `alter`?

pushcx (changes requested):
Thanks for helping improve the first time setup. I have a couple small tweaks and I look forward to merging this soon.

README.md
edwardloveall: Now that I look into it more, the internet disagrees. Some of the answers on [this stackoverflow question](https://stackoverflow.com/questions/36463966) agree with you. I did some experimenting and I didn't actually need this at all. I could change the root pass, see rails fail, change it back to `localdev`, and see rails succeed, all without ever calling `FLUSH PRIVILEGES`.

I'll remove it because it seems not useful, but I'm happy to add it back (at the end) if you'd like.

README.md
edwardloveall: 👍🏻 Consistent make sense.

I'm struggling to figure out how to start mariadb without explicitly using a service manager or unfriendly command. I can run `mariadbd` but that starts a non-deamon process that does not respond to ctrl-c or backgrounding. I'm using macOS/homebrew which has recommendations. But while I imagine that's a relatively common environment, it's not the only one. Listing all the platform specific options does not sound ideal.

Do you have a better suggestion? Maybe the best thing is to link to the [Starting and Stopping MariaDB Server](https://mariadb.com/kb/en/starting-and-stopping-mariadb-automatically/) page?

README.md
pushcx: Good find! Yeah, the docs are clear that it's not needed with `GRANT`. Let's leave it out.

README.md
pushcx: Did `mariadbd --console` not do what you wanted with giving you a server you could stop with ctrl-c?

edwardloveall: It did not. I was not able to ctrl-c or background. Otherwise it's pretty much what I wanted, though.

When I `kill`ed it, the server _did_ gracefully shutdown, which I thought was odd. Maybe it's something with my personal setup? Either way, I think starting it up with some official background process (and then shutting it down manually if you want) is probably the best option. I don't run mysql for anything else so that's why I was trying to do this, first.

pushcx: Ah, sorry it didn't quite work out.

pushcx (approved):

pushcx:
Thanks for contributing back based on your experience. I'm (mumble) years past my first-run experience so I can't see these details very well anymore. You've helped future contributors here.

PR #1329, opened 2024-09-25
Story: description_or_story_text should return plaintext

### LINKED ISSUE #1328: Story previews displaying escaped characters
As seen in the screenshot below, story previews are rendering with the escaped characters rather than being escaped for the browser to render.

May be linked to #1320.

jcs:
PR #1320 changed this from HTMLEntities.new.decode to HtmlEncoder.encode which does the opposite of what it used to do.

Should fix #1328


pushcx:
Yep, looks like an accidental change. Thanks for the fix!

Issue #1282, opened 2024-05-29
Print media query


pushcx:
We don’t have any print-specific CSS now and look pretty good. There’s opportunity for a few tweaks like hiding `comment_folder` and comment forms. I don’t know anything else off the top of my head and would like to keep this pretty light.

shubhamsugara22:
would like to try this one @pushcx

pushcx:
@shubhamsugara22 did everything asked in this issue, so I'm going to close it. That doesn't mean our print media query is 'done' so much as 'done for now' because it's a feature request. If you are reading this in the future and have improvements, please reopen or open a PR.

PR #1332, opened 2024-09-27
Issue #1282 - Updating application.css.erb rule to hide folders

shubhamsugara22:
Issue #1282

Checked the issue & came to following conclusion
Updated the current print-specific CSS
Defined rules to hide the comment_folder and comment forms when a user prints a page.

Looks fine to me for now will update if more fixes are needed

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -1962,3 +1962,9 @@ input[type="submit"].link_post.pushover_button {
 		float: none;
 	}
 }
+@media print {
+	.comment_folder,
+	.comment_form {
+	    display: none;
+	}
+  }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/application.css.erb
```diff
@@ -1962,3 +1962,9 @@ input[type="submit"].link_post.pushover_button {
 		float: none;
 	}
 }
+@media print {
+	.comment_folder,
+	.comment_form {
+	    display: none;
```
pushcx, line 1968: Thanks for taking this on.

I hate to bring up tabs vs spaces but we don't have linting on css. Could you make this single-tab per indent level to match the file?

shubhamsugara22: ok let me change it to tab per ident

pushcx:
Thanks for taking this!

PR #1335, opened 2024-10-01
Fix #1333 Some but not all text changes size when rotating Mobile Safari to landscape

### LINKED ISSUE #1333: Some but not all text changes size when rotating Mobile Safari to landscape
For example, the top navigation and article titles stay the same size, but article tags on the same line enlarge, as do comments.

IMG_4788

Compare to portrait (screenshot padded to keep sizes comparable):
 
IMG_4787

Resizing for landscape is default behavior, and not necessarily what you intended. Details at this [article about text-size-adjust in CSS resets](https://kilianvalkhof.com/2022/css-html/your-css-reset-needs-text-size-adjust-probably/). 

I like the solution offered in that article. When you rotate with that change applied, text would stay the same size and you'd just get more characters per line. I tried it on my own site and it worked fine.

Alternatively, we could keep the default resizing behavior and figure out how to let all text scale up together.

kconner:
Fixes #1333.

Tested with the home list page and a story page:

## Landscape

Before: Landscape exhibits the bug, in which some fonts are the same as portrait and some are enlarged.

list, landscape, before
story, landscape, before

After: Landscape keeps the same font sizes as portrait.

list, landscape, after
story, landscape, after

## Portrait

Portrait font sizes were unaffected by the change.

Before:

list, portrait, before
story, portrait, before

After:

list, portrait, after
story, portrait, after

Curiously, I didn't reproduce today the enlarged tags on the story page that I saw when I wrote up the issue.

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -223,9 +223,14 @@ html {
 
 /* generics */
 
-/* force a vertical scrollbar to avoid page-shifting */
 html {
+	/* force a vertical scrollbar to avoid page-shifting */
 	overflow-y: scroll;
+
+	/* avoid automatic resizing of some text elements on phones in landscape */
+	-moz-text-size-adjust: none;
+	-webkit-text-size-adjust: none;
+	text-size-adjust: none;
 }
 
 body, textarea, input, button {


pushcx:
Thanks for running down this very odd inconsistency and fixing it!

PR #1337, opened 2024-10-01
Change "archive" link to "caches"

pronoiac:
This might be more clear.
There's a [discussion thread](https://lobste.rs/s/nsdzkf/suggestion_change_archive_link_text)

Perhaps "mirrors" or "archives" would be better, but I figured the "obvious improvement" comment was a sign to go ahead rather than do some bikeshedding.


pushcx:
Thanks for painting the bikeshed, this is an improvement to fit in with the verbs on this line.

Issue #1321, opened 2024-09-06
Preview function loses story tags


epidemian:
This is something i noticed when using the Preview function. It seems to be a little bug.

**Steps to reproduce:**

1. Open Submit form and fill in URL, title, and some tags
2. Click Preview

**Expected behavior:**

The story preview is shown and all fields remain the same.

**Actual behavior:**

The story preview is shown, but the tags are no longer in the form. They also don't show in the story preview. 

I don't know if submitting the story after this preview will still submit the previously selected tags; i don't want to spam the site by trying this. But it think they might be lost, as trying to Preview the story once again after step 2 triggers a validation error about the story requiring at least one tag.

Here's a video demonstrating this:

https://github.com/user-attachments/assets/8b760b1b-b418-464f-ad03-d224597c8042

epidemian:
BTW, if this is a valid bug, i can take a stab at trying to fix it in the upcoming days :)

pushcx:
This is a dupe of #1264 and I'd love a PR for it.

PR #1342, opened 2024-10-11
Remember tags when previewing a story

### LINKED ISSUE #1264: Previewing a submission removes tags
In at least Safari in iOS 17.3, filling in story details and selecting tags, then tapping Preview generates a preview with no tags, and the tags are removed from the form field.

IMG_0531
IMG_0532

epidemian:
This is a late PR after the issue submitted on #1321.

The changes are described on the commit messages. Basically:

* The first commit makes the story preview action not forget the story tags by making `Story#tags_a` work on preview stories (i.e. unsaved). It fixes #1264.
* The second commit makes the story preview show the submitted tags.

I'm not too keen on the approach taken on the 2nd commit. I wish `Story#tags` would Just Work™ on unsaved stories, but alas, it doesn't. Suggestions are welcomed!

The two commits are independent from each other. I left them this way if you prefer to apply the first one without the questionable approach of the second one :)

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -48,7 +48,7 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
     <% end %>
     <% if story.can_be_seen_by_user?(@user) %>
       <span class="tags">
-        <% story.tags.each do |tag| %>
+        <% (story.previewing ? story.preview_tags : story.tags).each do |tag| %>
           <%= tag_link(tag) %>
         <% end %>
       </span>
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -727,13 +727,13 @@ class Story < ApplicationRecord
   end
 
   def tags_a
-    @_tags_a ||= taggings
-      .includes(:tag)
-      .reject(&:marked_for_destruction?)
-      .map { |t| t.tag.tag }
+    @_tags_a ||= Tag
+      .where(id: taggings.reject(&:marked_for_destruction?).map { |t| t.tag_id })
+      .pluck(:tag)
   end
 
   def tags_a=(new_tag_names_a)
+    @_tags_a = nil
     taggings.each do |tagging|
       if !new_tag_names_a.include?(tagging.tag.tag)
         tagging.mark_for_destruction
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -749,6 +749,10 @@ class Story < ApplicationRecord
     end
   end
 
+  def preview_tags
+    Tag.where(id: taggings.map { |t| t.tag_id })
+  end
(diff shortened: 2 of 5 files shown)


app/views/stories/_listdetail.html.erb
```diff
@@ -48,7 +48,7 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
     <% end %>
     <% if story.can_be_seen_by_user?(@user) %>
       <span class="tags">
-        <% story.tags.each do |tag| %>
+        <% (story.previewing ? story.preview_tags : story.tags).each do |tag| %>
```
epidemian, line 51: Oh, btw, with this change, starting the server also makes the files `app/views/home/index.html.erb`  and `app/views/search/index.html.erb` change to have this same thing changed. IDK why that happens. Is that OK? Should i commit those changes too?

pushcx: This is `config/initializers/heinous_inline_partial` doing its work. Please do commit those changes so the codebase is in a consistent state with what the server will run with.

epidemian: LOL, that's a poetic name. I've amended those changes onto their appropriate commit!

epidemian: BTW, i also get this change on  `app/views/home/index.html.erb` when rendering `/stories/new`:

```diff
diff --git a/app/views/home/index.html.erb b/app/views/home/index.html.erb
index 4193138f..633531b8 100644
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -183,7 +183,7 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
           <span> | </span>
           <span class="dropdown_parent">
             <input id="archive_<%= story.short_id %>" class="archive_button" type="checkbox">
-            <label for="archive_<%= story.short_id %>">caches</label>
+            <label for="archive_<%= story.short_id %>">archive</label>
             <div class="archive-dropdown">
               <a href="<%= story.archiveorg_url %>">Archive.org</a>
               <a href="<%= story.archivetoday_url %>">Archive.today</a>
```

But it also happens on `master` so i guess there's some inconsistency on these partial inlines. I'm excluding that change from this PR as it's unrelated.

pushcx: Yes,  e3456c07 missed that update, I'll get it. Living up to its name!

pushcx:
Thanks for taking this on. There's a lot of layers of abstraction at work in this bug! And it's a pretty core site feature, so I would rather bring in a slightly-odd fix than leave it broken for longer.

epidemian:
Glad to be of some help :) 
Thanks for maintaining Lobsters!

PR #1343, opened 2024-10-12
Issue #1338 Partial view disappears on click

### LINKED ISSUE #1338: Warning box dissapears on click
When I want to submit a story from a banned website, a warning pops up. The problem is that the warning disapears on click, preventing me to select the link containing the explanation for banning a domain:

Screenshot 2024-10-02 at 23 21 49

Eduardo-Lpz:
Fix #1338 

Stories form is validating URL duplicity in advance of form submit and due the event used results in an view override when the form have other errors.

**Current behavior:**
onfocusout event is called for URL input, this result in a request (`/stories/check_url_dupe`) that overrides the current form errors, if any.
[Exhibit A](https://www.loom.com/share/c7b54a3cb59f4d0caeecb547bcec39c2?sid=0e1b918d-3e5c-48ad-adac-9d3cf85c82f1)

New behavior:
complement `url_dupe` request with `.valid?` method to avoid loss previous error messages if any
[Exhibit B](https://www.loom.com/share/3bd0562ae78b4c95995c6561e8665feb?sid=c87e7d5f-9a63-446c-9285-1e1665694601)

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -415,6 +415,7 @@ class StoriesController < ApplicationController
     @story = Story.new(user: @user)
     @story.attributes = story_params
     @story.already_posted_recently?
+    @story.valid?
 
     respond_to do |format|
       linking_comments = Link.recently_linked_from_comments(@story.url)


pushcx:
It looks like you've correctly diagnosed the interaction that's causing the form warnings/errors to disappear, but your fix deletes some important code for setting tags based on the URL (lines 575-587).

I also don't want to remove this fetch call. When someone is starting to submit a story, I want them to have immediate feedback that the URL has been seen before (or is banned) because it's a bit of a chore to select tags and decide if and what to write in the story text. If we only validate server-side in `#create` the user's effort is wasted, which is kind of discouraging.

Can you see another approach to keep the fast feedback? If we fetch the errors when the field is unfocused, how is it getting focused to run again?

Eduardo-Lpz:
@pushcx 
Since there is a need for the `check_url_dupe` request to be executed, i think the best/easy approach is to complement the request with the `valid?` method (and any other validation i'm not aware of)...

A more complicated approach will be to dinamically inject the `check_url_dupe` result in the errors messages if present.

Or having the URL errors in another section but (since the request also return `linking_comments`) this may result weird/confusing looking when the request is triggered after a submit fail.

Screenshot 2024-10-14 at 9 39 32 p m

pushcx:
> A more complicated approach will be to dinamically inject the check_url_dupe result in the errors messages if present.

Maybe the answer is here? Is there a reason the errors don't include the duplicate? I looked at the Story model and from a quick read I would expect all of the validations to run every time.

Eduardo-Lpz:
Almost December and probability I will be out for a while 🎄 , since this will need a different approach and i have not work in this for a month, i prefer to close it to avoid someone else think it is actively being tackled and stop some possible contribution.

PR #1345, opened 2024-10-14
Issue #1331 Implement story disowning

### LINKED ISSUE #1331: Story disowning
Users can [disown comments](https://github.com/lobsters/lobsters/blob/f512ae4187be05098978c28055a076db3604df31/app/controllers/comments_controller.rb#L175) individually or entirely as they delete their account, but only disown all stories by deleting their account.

For this feature, add a 'disown' link that functions similarly to comment disowning. It should only appear on the individual story page (not every place the story is listed) and not while the edit window is open, similar to comments.

Eduardo-Lpz:
Fix #1331 

Allow stories to be disowned individually

Screenshot 2024-10-13 at 11 53 44 p m

Screenshot 2024-10-13 at 11 54 06 p m

Validate:

- [x] add disown route
- [x] add disown button only in story page
- [x] add specs

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -198,6 +198,10 @@ class Story < ApplicationRecord
     check_tags
   end
 
+  def is_disownable_by_user?(user)
+    user && user.id == user_id
+  end
+
   def accepting_comments?
     !is_gone? &&
       !previewing &&
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -136,6 +136,7 @@ class Story < ApplicationRecord
 
   COMMENTABLE_DAYS = 90
   FLAGGABLE_DAYS = 14
+  DELETEABLE_DAYS = FLAGGABLE_DAYS * 2
 
   # the lowest a score can go
   FLAGGABLE_MIN_SCORE = -5
@@ -196,10 +197,6 @@ class Story < ApplicationRecord
     end
 
     check_tags
-  end
-
-  def is_disownable_by_user?(user)
-    user && user.id == user_id
   end
 
   def accepting_comments?
@@ -575,6 +572,10 @@ class Story < ApplicationRecord
 
   def hider_count
     @hider_count ||= HiddenStory.where(story_id: id).count
+  end
+
+  def disownable_by_user?(user)
+    user && user.id == user_id && created_at < DELETEABLE_DAYS.days.ago
   end
 
   def is_flaggable?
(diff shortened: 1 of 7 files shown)


app/models/story.rb
```diff
@@ -198,6 +198,10 @@ class Story < ApplicationRecord
     check_tags
   end
 
+  def is_disownable_by_user?(user)
+    user && user.id == user_id
```
Eduardo-Lpz, line 202: does this need validate `created_at` as in [comment model](https://github.com/lobsters/lobsters/blob/master/app/models/comment.rb#L425)?

pushcx: Yes, please add a `DELETEABLE_DAYS` constant to the Story model that is twice `FLAGGABLE_DAYS` like Comment.

app/views/stories/_listdetail.html.erb
```diff
@@ -153,6 +153,10 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
             <span> | </span><%= link_to "hide", story_hide_path(story.short_id),
               :class => "hider" %>
           <% end %>
+          <% if story.is_disownable_by_user?(@user) %>
+          |
+          <a tabindex="0" class="story-disowner" href="#">disown</a>
```
pushcx, line 158:
```suggestion
          <a class="story-disowner" href="#">disown</a>
```

This is unneeded. I see where you copied it from, I'll clean up the other templates.

spec/features/story_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    scenario "shows disown in story page" do
+      visit "/s/#{story.short_id}"
+      expect(page).to have_link("disown")
+    end
+
+    scenario "does not show disown in active stories page" do
+      visit "/active"
+      expect(page).not_to have_link("disown")
```
pushcx, line 33: How does this pass? I don't see the call to `if single_story` in the template that I would've expected.

pushcx:
Thanks for taking on this missing feature! Couple small things and I look forward to merging it in.

app/models/story.rb
Eduardo-Lpz: it is now added.

spec/features/story_spec.rb
Eduardo-Lpz: because the only story wasn't associated to the current user, even though it should not shown the `disowned` link due the current route, it was not a proper way to validate it.

I modified the test stories to have one being candidate to `disowning` (if we were on the correct path)

Eduardo-Lpz:
@pushcx 
any other feedback for this PR? 🤔

pushcx:
Looks great, thanks for working through these changes. I'm glad to merge this!

Lonami:
Thank you @Eduardo-Lpz! Looks like I can postpone learning Ruby for a while longer :)

PR #1347, opened 2024-10-15
Issue #1344 Consider - and _ a collision in usernames

### LINKED ISSUE #1344: Consider - and _ a collision in usernames
It would be good to consider usernames to collide if they swap `-` and `_` as those characters are similar enough to be very confusing. Also to redirect URLs if those are incorrect, same as we do with case now.

Eduardo-Lpz:
Fix #1344 

Added underscore and dash collide validation in `User` model.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -101,7 +101,7 @@ class User < ApplicationRecord
     format: {with: /\A#{VALID_USERNAME}\z/o},
     length: {maximum: 50},
     uniqueness: {case_sensitive: false}
-
+  validate :underscores_and_dashes_in_username
   validates :password_reset_token,
     length: {maximum: 75}
   validates :session_token,
@@ -132,8 +132,6 @@ class User < ApplicationRecord
       record.errors.add(attr, "is not permitted")
     end
   end
-
-  validate :underscores_and_dashes_in_username
 
   scope :active, -> { where(banned_at: nil, deleted_at: nil) }
   scope :moderators, -> {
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -173,6 +175,14 @@ class User < ApplicationRecord
   # minimum number of submitted stories before checking self promotion
   MIN_STORIES_CHECK_SELF_PROMOTION = 2
 
+  def underscores_and_dashes_in_username
+    username = self.username.split(/_|-/)
+    return if username.length < 2
+
+    collisions = User.where('username REGEXP ?', username.join("[-_]"))
```
pushcx, line 182: Clever approach!

To build on it, you could combine your `split` to express this as `username.gsub(/_|-/, "[-_]")`.

app/models/user.rb
```diff
@@ -173,6 +175,14 @@ class User < ApplicationRecord
...
+    errors.add(:username, "is too similar to other usernames") if collisions.any?
```
pushcx, line 183:
```suggestion
    errors.add(:username, "is already in use (perhaps swapping _ and -)") if collisions.any?
```

Trying to be explicit about what "similarity" means.

spec/models/user_spec.rb
```diff
@@ -14,6 +14,16 @@
 
     create(:user, username: "newbie")
     expect { create(:user, username: "newbie") }.to raise_error
+
+    create(:user, username: "underscores_and-dashes")
+    expect { create(:user, username: "underscores-and_dashes") }.to raise_error
+    expect { create(:user, username: "underscores-and-dashes") }.to raise_error
+    expect { create(:user, username: "underscores_and_dashes") }.to raise_error
+
+    create(:user, username: "case_insensitive")
```
pushcx, line 23: Thanks for adding these missing tests.

spec/models/user_spec.rb
```diff
@@ -14,6 +14,16 @@
 
     create(:user, username: "newbie")
     expect { create(:user, username: "newbie") }.to raise_error
+
+    create(:user, username: "underscores_and-dashes")
```
pushcx, line 18: I'd prefer to test validations more explicitly, like the later tests in this file do.

But this comment does not block merging. I see how you're repeating the existing pattern so I don't want to add work unnecessarily. If you'd like to change the tests to match that style I'd appreciate it, but you can leave them alone and I'll update them.

pushcx (changes requested):
Thanks for taking this on. I've appreciated your many recent contributions.

You have a good strategy here for finding duplicates! I have a suggestion for simplifying it by using on the String API, and then some minor tidying. I look forward to merging this fix.

app/models/user.rb
Eduardo-Lpz: In fact, I was hoping for this suggestion, so I could have an example message from you. 😅

pushcx:
Thanks for improving this!

PR #1348, opened 2024-10-16
Issue #1334 Fix validation for resubmitted story form

### LINKED ISSUE #1334: "suggest" form shows "why resubmit" explaination prompt
using the "suggest" link on https://lobste.rs/s/geuifh/secure_py_simplify_http_security_headers the form contains the prompt for "why do you think this is worth resubmitting". Given that this is not a publicly visible field, is it intentional that it's shown to other users and not just in the original submission?

screenshot showing form described above

Eduardo-Lpz:
Fix #1334 

Currently for an already submitted story when a different user tries to do a suggestion (click on `suggest` link) it show the `is_resubmit?` form as well as the `similar_stories.any?` form, this forms are intended when a user is submitting a story.

Current behavior:
- test user with original story:
Screenshot 2024-10-15 at 7 17 48 p m
- renna_collins with a resubmitted story 
Screenshot 2024-10-15 at 7 19 21 p m
Screenshot 2024-10-15 at 7 20 38 p m
- will_shannon clicking "suggest" (different user)
Screenshot 2024-10-15 at 7 22 00 p m
Screenshot 2024-10-15 at 7 22 16 p m

New behavior:
- will_shannon clicking "suggest" (different user)
Screenshot 2024-10-15 at 7 30 52 p m
Screenshot 2024-10-15 at 7 24 36 p m
- long_armstrong clicking "suggest" (different user)
 Screenshot 2024-10-15 at 7 30 05 p m
Screenshot 2024-10-15 at 7 26 40 p m

### DIFF
--- a/app/views/stories/_form_errors.html.erb
+++ b/app/views/stories/_form_errors.html.erb
@@ -1,10 +1,10 @@
-<%# locals: (story:, f: nil, linking_comments: []) -%>
+<%# locals: (story:, f: nil, linking_comments: [], suggesting: false) -%>
 <div>
   <div class="form_errors_header">
     <% if story.errors.any? %>
       <%= errors_for story %>
     <% end %>
-    <% if story.is_resubmit? %>
+    <% if story.is_resubmit? && !suggesting %>
       <div class="flash-notice">
         <p>
           This story has been submitted before. You can resubmit it with a comment to start a fresh discussion.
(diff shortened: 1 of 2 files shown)


app/views/stories/_form_errors.html.erb
```diff
@@ -24,7 +24,7 @@
         <%= text_area_tag "comment", @comment&.comment, :rows => 5 %>
         <br><br>
       </div>
-    <% elsif story.similar_stories.any? %>
+    <% elsif story.similar_stories.any? && story.is_saved_by_cur_user %>
```
pushcx, line 27: Why call this method?

Eduardo-Lpz: That CR was fast 😅 
was my first attempt trying to solve the issue before actually  went to the partial and notice the flag.

Eduardo-Lpz: @pushcx 
btw, not totally sure the images represent well enough the scenario for the current issue... any feedback is appreciated.

pushcx:
Thanks for this fix. Looking at your development process, it’s definitely time to split out these forms.

PR #1350, opened 2024-10-16
Issue #1339 Utils.normalize(url) mishandles www10.org 

### LINKED ISSUE #1339: `Utils.normalize(url)` mishandles `www10.org`
Part of normalizing URLs tries to remove `www` from hostnames, including with digits. It's too aggressive and stripping a real hostname off the TLD. So we have a Domain named `org` instead of `www10.org`: https://lobste.rs/domains/org/

Ryzen114:
Fix #1339 

The url normalizing methods ( Utils.Normalize and Story.set_domain_and_origin ) have been updated in order to be more flexible with domains names as www10.org .

Current behavior : 
www10.org is displayed as org in a story

New behavior :
www10.org is displayed and classic urls are still displayed correctly for instance www.google.com is displayed google.com

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -930,7 +930,7 @@ class Story < ApplicationRecord
   end
 
   def set_domain_and_origin(domain_name)
-    domain_name&.sub!(/^www\d*\./, "")
+    domain_name&.sub!(/\Awww\d*\.(.+?\..+)/, '\1') # remove www\d* from domain if the url is not like www10.org
     if domain_name.present?
       self.domain = Domain.where(domain: domain_name).first_or_initialize
       self.origin = domain&.origin(url)
--- a/extras/utils.rb
+++ b/extras/utils.rb
@@ -19,7 +19,7 @@ class Utils
     url.slice! %r{/Default\.aspx$}
 
     url.slice! %r{https?://} # consider http and https the same
-    url.slice! %r{^(www\d*\.)} # remove www\d* from domain
+    url.sub!(/\Awww\d*\.(.+?\..+)/, '\1') # remove www\d* from domain if the url is not like www10.org
 
     url, *args = url.split(/[&\?]/) # trivia: ?a=1?c=2 is a valid uri
     url ||= "" # if original url was just '#', ''.split made url nil
--- a/spec/extras/normalize_url_spec.rb
+++ b/spec/extras/normalize_url_spec.rb
@@ -21,6 +21,8 @@ describe "normalizing urls" do
     "https://e.com?b=2&a=1" => "e.com?a=1&b=2", # sort query args
     "https://e.com?a=1?b=2" => "e.com?a=1&b=2", # normalize ? to &
     "https://e.com?c=3&a=1?b=2&" => "e.com?a=1&b=2&c=3", # combined; trailing &
+    "https://www10.org" => "www10.org", # keep www10. since org alone doesn't fully describe the domain
+    "https://www10.example.org" => "example.org", # remove www10.
 
     "https://www.arxiv.org" => "arxiv.org",
     "https://arxiv.org/abs/1234.12345" => "arxiv.org/abs/1234.12345",
--- a/spec/models/story_spec.rb
+++ b/spec/models/story_spec.rb
@@ -97,7 +97,9 @@ describe Story do
       "http://example.com:8000" => "example.com",
       "http://example.com:8000/" => "example.com",
       "http://www3.example.com/goose" => "example.com",
-      "http://flub.example.com" => "flub.example.com"
+      "http://flub.example.com" => "flub.example.com",
+      "http://www10.org" => "www10.org",
+      "http://www10.example.org" => "example.org"
     }.each_pair do |url, domain|
       story.url = url
       expect(story.domain.domain).to eq(domain)


app/views/home/index.html.erb
```diff
@@ -183,7 +183,7 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
           <span> | </span>
           <span class="dropdown_parent">
             <input id="archive_<%= story.short_id %>" class="archive_button" type="checkbox">
-            <label for="archive_<%= story.short_id %>">caches</label>
+            <label for="archive_<%= story.short_id %>">archive</label>
```
pushcx, line 186: This is an unrelated change, please merge master into your branch to remove it.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.2].define(version: 2024_10_10_165823) do
-  create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+  create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
```
pushcx, line 14: These changes to `schema.rb` are unrelated and something odd about your development setup. Could you paste the output of `mariadb --version`? We see this occasionally and I'd like to understand what's different so I can gave better guidance to future contributors or improve our setup instructions in the README.

extras/utils.rb
```diff
@@ -19,7 +19,9 @@ def self.normalize(url)
     url.slice! %r{/Default\.aspx$}
 
     url.slice! %r{https?://} # consider http and https the same
-    url.slice! %r{^(www\d*\.)} # remove www\d* from domain
+    if url.split(".").size > 2 # only  removes www\d* if the url is not like www10.org, issue #1339
```
pushcx, line 22: You don't need to include the issue number in a comment, and your "if the url is not like www10.org" can go on the existing comment.

Please add two tests to `spec/extras/normalize_url_spec.rb`, one for `www10.org` and one for `www10.example.org`.

app/models/story.rb
```diff
@@ -925,7 +925,10 @@ def send_referrer?
   end
 
   def set_domain_and_origin(domain_name)
-    domain_name&.sub!(/^www\d*\./, "")
+    if domain_name.present? && domain_name.split(".").size > 2 # only removes www\d* if the url is not like www10.org, issue #1339
+      domain_name&.sub!(/^www\d*\./, "")
```
pushcx, line 929: If the regexp were `domain_name&.sub!(/\Awww\d*\.(.+\.)/, "\1")` you wouldn't need the `if` wrapping this, it would express that we're only matching `www\d+` if there isn't another `.` segment. And it could use a test or two, that's a hairy bit of code!

extras/utils.rb
```diff
@@ -19,7 +19,9 @@ def self.normalize(url)
...
+      url.slice! %r{^(www\d*\.)} # remove www\d* from domain
```
pushcx, line 23: It would be nice to apply the same trick here from Story model so we don't have to `split` and have an extra `if`.

pushcx (changes requested):
Thanks for taking this on, it's a fiddly bit of editing. I didn't expect you'd have to edit two places, but I see how the Story model and `normalize` are a little redundant, and I'm happy to keep them specialized.

Couple small changes and some tests, we'll get this merged soon.

db/schema.rb
Ryzen114: Actually, I don't know why this has been changed. Here is the output of mariadb --version : mariadb  Ver 15.1 Distrib 10.3.39-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2 
I don't know if it's related but i'm using a WSL 2 virtual linux environment to run the project.

app/models/story.rb
Ryzen114: Thanks for the suggestion ! This is actually my very first contribution and i'm still new to Ruby and rails (I'm a French student learning software developpment). So any advice or suggestions are really helpful to me! 

The contribution is part of a team project for a course where we contribute to an open-source project. We are a group of six students from the French engineering school IMT Atlantique, and we chose to contribute to Lobsters. We'll do our best to help improve Lobsters for the next few months!

If you have any suggestions on how we could further contribute or specific areas that need attention, we would be glad to hear them. Currently, we're focusing on several issues tagged "good first issue," but we're open to other contributions as well.

Ryzen114:
It appears that several files have been updated (such as schema.rb and index.html.erb) but i don't know why... This has been done automatically. I'm looking to how i can discard these unwanted changes.

Ryzen114:
I believe everything should be correct now. Apologies for the confusion I created with all these unwanted changes, i will be more carefull next time.

Ryzen114:
I noticed that something were wrong with the regex `domain_name&.sub!(/\Awww\d*\.(.+\.)/, "\1")`, it didn't worked with www10.example.org the output was org.
It should be fixed now with `domain_name&.sub!(/\Awww\d*\.(.+?\..+)/, '\1') `

db/schema.rb
pushcx: That's really odd. I spent a few minutes searching and the difference between the two is an (outdated?) minor performance improvement. I'd like to keep the value we have in production in this file. Sorry for the hassle, but please do revert these collation changes.

app/models/story.rb
pushcx: Well, I appreciate you helping out, and let me know if you need any paperwork signed off on.

pushcx (approved):

pushcx:
I'll merge this as soon as the linter error is handled.

When I'm developing, I run `bundle exec rspec && bundle exec standardrb --fix && bundle exec brakeman -q` to basically run the full build in one command.

db/schema.rb
pushcx: @chamlis figured it out:

> it seems upstream MariaDB defaults to latin1_swedish_ci, but Debian/Ubuntu patch it to default to utf8mb4_general_ci and Arch patches it to default to utf8mb4_unicode_ci

So _I'm_ the one introducing the incorrect value when I'm editing on my local dev, which runs Arch. I'll fix it on `master`.

pushcx:
Thank you for taking on this bug and getting it solved!

PR #1351, opened 2024-10-17
Database consistency fixes

Eduardo-Lpz:
From issue #1281 , this is a small contribution to fix database issues as flagged by `active_record_doctor`

- [x] [unindexed_foreign_keys issues](https://github.com/lobsters/lobsters/commit/cc95f959f26ef46682eeec32b28cd89759e832c7)
- [x] [fix missing_foreign_keys issues](https://github.com/lobsters/lobsters/commit/8942b00b01e58e1d39c5e33e534fac3099a9edd3)
- [x] [fix missing_presence_validation issues](https://github.com/lobsters/lobsters/commit/44ee9e1b0df5492c87d65f8fafbc9d1206f81a02)
- [x] [fix incorrect_length_validation issues](https://github.com/lobsters/lobsters/commit/28e3f267c1bccf91ac9467f4c5cbac4314ffa2b2)

Note:
There are still some warnings to tackle for the remaining detectors, i want to tackle that next week.

### DIFF
--- a/.active_record_doctor.rb
+++ b/.active_record_doctor.rb
@@ -3,20 +3,30 @@
 ActiveRecordDoctor.configure do
   detector :unindexed_foreign_keys,
     ignore_columns: [
-      "hats.short_id", # It's a unique key, but not a foreign key.
-      "stories.twitter_id", # It's not a foreign key.
-      "mastodon_apps.client_id" # It's not a foreign key.
+      "hats.short_id",
+      "stories.twitter_id",
+      "mastodon_apps.client_id"
     ]
 
   detector :missing_foreign_keys,
     ignore_columns: [
-      "messages.short_id", # It's a unique key, but not a foreign key.
-      "mastodon_apps.client_id", # It's not a foreign key.
-      "stories.short_id", # It's a unique key, but not a foreign key.
-      "stories.twitter_id", # It's not a foreign key.
-      "stories.mastodon_id", # Not sure if it should be a foreign key.
-      "hats.short_id", # It's a unique key, but not a foreign key.
-      "comments.short_id", # It's a unique key, but not a foreign key.
-      "comments.thread_id", # TODO: Can point to a comment or a Keystore.
+      "messages.short_id",
+      "mastodon_apps.client_id",
+      "stories.short_id",
+      "stories.twitter_id",
+      "stories.mastodon_id", # TODO: Not sure if it should be a foreign key.
+      "hats.short_id",
+      "comments.short_id",
+      "comments.thread_id" # TODO: Can be a comment or a Keystore.
+    ]
+
+  detector :missing_presence_validation,
+    ignore_models: [
+      "ReplyingComment", # This is a view, not a real table.
+      # ActiveStorage tables
+      "ActiveStorage::Blob", "ActiveStorage::Attachment", "ActiveStorage::VariantRecord"
+    ],
+    ignore_attributes: [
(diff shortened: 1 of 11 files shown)


.active_record_doctor.rb
```diff
@@ -0,0 +1,49 @@
...
+  detector :incorrect_length_validation,
+    ignore_models: [
+      "ReplyingComment", # This is a view, not a real table.
+      # ActiveStorage tables
+      "ActiveStorage::Blob", "ActiveStorage::Attachment", "ActiveStorage::VariantRecord"
+    ],
+    ignore_attributes: [
+      "Vote.reason", # It's using `is` constraint.
+      "User.password_digest",
+      "Story.description", # TODO: Is there a reason why it validates for 65_536 instead of 65_535?
```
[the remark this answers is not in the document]
Eduardo-Lpz: got it 👌

.active_record_doctor.rb
```diff
@@ -0,0 +1,49 @@
...
+      # not a foreign key
+      "messages.short_id",
+      "mastodon_apps.client_id",
+      "stories.short_id",
+      "stories.twitter_id",
+      "hats.short_id",
+      "comments.short_id",
+
+      "stories.mastodon_id", # TODO: Not sure if it should be a foreign key.
+      "comments.thread_id" # TODO: Can be a comment or a Keystore.
```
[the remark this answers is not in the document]
Eduardo-Lpz: interesting 🤔

db/migrate/20241017011246_active_record_doctor_incorrect_length_validation.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def down
+    change_column :links, :url, :string
+    change_column :invitation_requests, :memo, :string
+    change_column :invitations, :memo, :string
+    change_column :hat_requests, :comment, :text, size: :medium
+    change_column :hats, :short_id, :string
+    change_column :categories, :category, :string
+    change_column :stories, :description, :text
```
pushcx, line 20: I'm really puzzled why you removed the `size: :medium` from `text` columns on `invitations`, `messages`, etc. Is there a value to that? I like to have the database backstopping that sort of validation with its explicit size, even if the default for a `text` column is `medium`.

Eduardo-Lpz: Hi, it was a mistake. 😞 
Since it is the `down` method for the migration, it should restore to the previous exact status for a rollback.

db/schema.rb
```diff
@@ -95,7 +95,7 @@
     t.bigint "user_id", null: false, unsigned: true
     t.string "hat", null: false
     t.string "link", null: false
-    t.text "comment", size: :medium, null: false
+    t.text "comment", null: false
```
Eduardo-Lpz, line 98: this kind of chages happen because:
`t.text "comment", size: :medium, null: false` will result in a `mediumtext` column with a [default size](https://mariadb.com/kb/en/mediumtext/) of (2**24 - 1)

Screenshot 2024-10-22 at 10 16 53 p m

since the model is validating for (2**16 - 1) we should have defined the column as:
`t.text "comment", null: false` to have a [proper value](https://mariadb.com/kb/en/text/), this is why the migration define the change as:
`change_column :hat_requests, :comment, :text, limit: 65_535`

Screenshot 2024-10-22 at 10 20 09 p m

pushcx (approved):

PR #1354, opened 2024-10-18
Issue #1353 Fix unread counts for message destroy

### LINKED ISSUE #1353: Removed moderated comments remain in unread counter
My red "unread replies" counter next to "Inbox" at the top right has been stuck at a minimum of 1; I believe this is because I had an unread reply that got removed here: https://lobste.rs/s/1roeql/stallman_report#c_pjhutv

Clicking on the 1 leads me to `/inbox` which redirects to `/messages` when no unread replies can be found.

New unread replies can increase this number above 1, (in which case `/inbox` works for the new replies) but as far as I can tell I'm unable to get it below 1.

Eduardo-Lpz:
Fix #1353 

I was able to reproduce the issue but just for `messages` and the only way i was able to reproduce it was deleting the record through the console/database.

My hunch is there was a directly delete of the record for some reason, or for another strange reason the `after_save :update_unread_counts` hook was not triggered.

- added `before_destroy` to make sure recipient gets `unread_messages_count` updated

Screenshot 2024-10-17 at 11 08 27 p m
Screenshot 2024-10-17 at 11 08 39 p m

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,7 +87,7 @@ GEM
     bcrypt (3.1.20)
     benchmark-perf (0.6.0)
     bigdecimal (3.1.8)
-    brakeman (6.2.1)
+    brakeman (6.2.2)
       racc
     builder (3.3.0)
     byebug (11.1.3)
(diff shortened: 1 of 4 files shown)


pushcx:
Thanks for jumping on this bug and figuring out it was a little different than reported. I'm really glad it was straightforward to fix, and I'm really happy you improved the testing around it.

Issue #1206, opened 2023-09-19
Flag menu displayed in background when submission hidden


FrostKiwi:
Clicking `flag` after hiding a post displays the drop-down menu behind the comment box, making it unreadable.
This issue is the same as https://github.com/lobsters/lobsters/issues/601, but that was already closed with a pull request, so I created a dedicated issue for it.
![image](https://github.com/lobsters/lobsters/assets/60887273/d61b44f8-4073-4edf-9f96-1176be96c411)

pushcx:
This is like the 4th time this bug has happened, so it would be great if whoever fixes it could improve the CSS so it stops happening. Maybe there's a better way to structure the CSS, maybe there's a test we could add, I don't know.

JuanVqz:
Hey @pushcx I have a question on this.
Why are we hiding elements? is it because we are not interested in them? If so, I think the only option that should be visible when is hidden is unhide, once restored yeah you can interact with the item. what do you think?

Also, the hide status is not persistent... when reloading the browser it is shown again, is that expected?

PR #614, opened 2018-11-30
prevent flag menu to inherit opacity

evexoio:
fixes #601

screen shot 2018-11-30 at 2 17 46 pm

PR #1356, opened 2024-10-23
fix issue #1206

Blosterhxh:
When a story is hidden and you click on flag or caches, you can now see the dropdown menu its not hidden behind the textfield anymore

### DIFF
--- a/app/assets/stylesheets/application.css.erb
+++ b/app/assets/stylesheets/application.css.erb
@@ -930,9 +930,11 @@ li .byline a.story_has_suggestions {
 	margin: 1em 0 1em 32px; /* matching div.detail */
 }
 
-li.story.hidden {
-	opacity: 0.25;
+li.story.hidden *:not(div, .dropdown_parent, .dropdown_parent > div *, .comments_label, .link, .tags, img),
+li.story.hidden .score{
+	opacity: 0.5;
 }
+
 ol.show_hidden li.story.hidden {
 	opacity: 1.0 !important;
 }


Blosterhxh:
I dont know why i put the opacity of the hidden story at 0.5 when resolving the issue, i put it back at 0.25 as it was normally

app/assets/stylesheets/application.css.erb
```diff
@@ -930,9 +930,11 @@ li .byline a.story_has_suggestions {
 	margin: 1em 0 1em 32px; /* matching div.detail */
 }
 
-li.story.hidden {
+li.story.hidden *:not(div, .dropdown_parent, .dropdown_parent > div *, .comments_label, .link, .tags, img),
+li.story.hidden .score{
```
pushcx, line 934:
```suggestion
li.story.hidden .score {
```

Sorry there isn’t a linter for the CSS, I couldn’t find a pure-ruby one.

pushcx (commented):
I just have one nitpick style issue.

Blosterhxh:
Thanks for adressing my pull request. Im new to coding, so coud you explain pls why i should have added a space i have no idea why, also i may try to resolve some other issues in the future so it would prevent me to do the same mistake again

pushcx:
To match the existing style. It doesn't change the functionality, but a consistent style means the reader is not distracted by figuring out if a difference is important or not.

pushcx:
I hate to come back on this, but I had a few minutes to test locally and this isn't a working fix. This makes the whole flag menu partially transparent, which is bug #614.

The menu should always be fully opaque even if the story is not.

Blosterhxh:
Ok i'll check that

Blosterhxh:
I tried it and i dont understand where the error is. Personally when i hit flag when the story is hidden i can see the whole dropdown menu and its not opaque, whether i do it on a story alone or on a story in a page list of submissions. 
screen2
screen1

pushcx:
Sorry, I missed your comment and didn't get back to this!

I pulled this locally for testing and works great. I thought this wouldn't work, but I realize that this is creating a second [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) on top of the one that `opacity` creates, and that's applying the `z-index` we want to make sure the flag menu is both visible and on top of any `textarea`.

Issue #1299, opened 2024-08-13
Allow users to export their own data


pushcx:
Per a [request](https://lobste.rs/s/tv11n9/json_representation_user_s_submissions), it would be nice to have an endpoint for a user to export their data like comments, stories, links to upvoted comments/stories, private messages. Though care has to be taken not to export other users' data.

This will need to be a background job as some users have substantial amounts of data. I'm pretty sure I heard Rails 8 is going to ship with SolidQueue so let's install that, store jobs in the db. There are a few things like email notifications that I've wanted to move to jobs for a while.

parisosuch-dev:
@pushcx , I was going to open another `featurereq` about an API. Make requests for data, etc. Will require a lot of overhead but would allow exporting data (through an API).

Rate limiting would need to be necessary to prevent DDOS and not overflow a queue.

pushcx:
It's going to be an HTTPS URL that requires an auth token in a cookie and gives the user a json or zip. What would you want to see different?

xavdid:
> an HTTPS URL

👍

> requires an auth token

Ideally something that doesn't expire (closer to an API key), so I can store it in my env and my script will "just work".

> in a cookie

That's fine enough. For programmatic use I'd usually prefer a header, but as long as it's easy to send using existing tools, it's not a big deal.

> gives the user a json or zip

JSON is strongly preferred, in my case. Paging is also fine if you'd prefer multiple, serial requests rather than a single big one

---

Thanks again for putting the work in for this!

parisosuch-dev:
> It's going to be an HTTPS URL that requires an auth token in a cookie and gives the user a json or zip. What would you want to see different?

Honestly yeah, I was under the assumption it would send an email of a user's exported data. This would work just fine.

petrenkorf:
@pushcx Is this already being actively developed?

pushcx:
Not that I'm aware of.

untitaker:
What would be the benefit of this over extending the existing routes with `.json` counterparts? To me the latter is both more intuitive to use as a developer (just add `.json` to the open webpage) and much more granular in how many rows it fetches from the DB than requesting a full data dump. If it had way more aggressive rate-limiting than the regular web UI, I wouldn't care.

pushcx:
[...]

PR #1358, opened 2024-10-26
add /upvoted/comments.json route

untitaker:
This is useful for me to analyze my own voting behavior.

It would be nice to get /comments.json working too, but I couldn't get
it done with 10 minutes (N+1 queries for some reason), and I don't
need it.


pushcx:
I don't like taking on the maintenance of API endpoints that the frontend doesn't use, and do it under very limited circumstances. I'll run [a query](https://lobste.rs/about#queries) for your data and email it to you, and there's #1299 available for a broader solution.

Issue #1355, opened 2024-10-21
Replace href="#" links


pushcx:
There's a few  `<a href="#"` links in the UI for js to attach handlers to. They're useless without js and fail oddly when our js has an error. Let's make them all work without js by round-tripping to the server.

I'd take PRs to fix these one-at-a-time, you don't have to do them all at once.

Eduardo-Lpz:
Since I'm very aware of one of those blank href, i feel the need to at least tackle that one.

pushcx:
I wrote a more detailed comment explaining the code strategy to take with these links: https://github.com/lobsters/lobsters/pull/1359#pullrequestreview-2414052963

pushcx:
Future contributors: @Eduardo-Lpz's PR #1359 is a great pattern for how all of these should work, and it should be straightforward to adapt the CSS by adding to the selector for the other links. Feel free to add a `.comment-action` class (or something similar, I didn't think hard about this name!) if that feels like a better way to reuse.

pushcx:
I think this is a complete list of remaining links/buttons that don’t work without js:

- [ ] `app/views/stories/_list_detail.html.erb` line 26 `<a class=“upvoter”`
- [ ] `app/views/stories/_list_detail.html.erb` line 141, 146 `<a class=“flagger”`
- [ ] Maybe `app/views/stories/new.html.erb` line 40 `<button class=“story-preview` - Does this work without js?
- [ ] `app/views/comment/_comment.html.erb` line 116, 119, 121 `<a class=“flagger”`
- [ ] `app/views/comment/_comment.html.erb` line 34 `<a class=“upvoter”`
- [ ] `app/views/comment/_comment.html.erb` line 93 `<a class=“comment_editor”`
- [ ] `app/views/comment/_comment.html.erb` line 93 `<a class=“comment_undeletor”`
- [ ] `app/views/comment/_comment.html.erb` line 102 `<a class=“comment_moderator”`
- [ ] `app/views/comment/_comment.html.erb` line 104 `<a class=“comment_deletor”`
- [ ] `app/views/stories/show.html.erb` line 20 `<a class=“story_text_expander”` (maybe a `<details>` tag?)

pushcx:
[...]

PR #1359, opened 2024-10-27
Remove blank href from story disown flow

Eduardo-Lpz:
issue #1355 

- remove previous `story disowner` js
- style `disown` button
- update specs

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -594,6 +594,18 @@ onPageLoad(() => {
     }
   });
 
+  on('submit', 'form.story-disowner-form', (event) => {
+    event.preventDefault();
+
+    if (confirm("Are you sure you want to disown this story?")) {
+      let li = parentSelector(event.target, '.story');
+      fetchWithCSRF(event.target.action, { method: 'post' })
+        .then(response => {
+          response.text().then(text => replace(li, text));
+        });
+    }
+  });
+
   // Comment
 
   // Remember story collapses; this is stored in localStorage for every story ID as an object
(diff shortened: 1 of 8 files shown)


pushcx:
This approach doesn't work at all without Javascript, when the goal is to make the Javascript optional. I don't think the `data: { confirm: ...` will even fire; we removed ujs years ago.

Eduardo-Lpz:
> This approach doesn't work at all without Javascript, when the goal is to make the Javascript optional. I don't think the `data: { confirm: ...` will even fire; we removed ujs years ago.

it actually does
Screenshot 2024-10-31 at 12 05 28 p m

pushcx:
...that's even [my own code](https://github.com/lobsters/lobsters/commit/6261204ad915d2be13f07bfed794fdaf1a06d1e8). Oof, sorry for the confusion.

app/controllers/stories_controller.rb
```diff
@@ -438,13 +438,19 @@ def disown
       return render plain: "can't find story", status: 400
     end
 
-    InactiveUser.disown! story
-    @story = find_story
-    @comments = Comment.story_threads(@story).for_presentation
+    if request.get?
+      InactiveUser.disown! story
```
pushcx, line 442: This introduces a CSRF so any page on the web could include `` so that when the submitter of story `abc123` loads the page, their browser sends a GET request and the story is disowned without their knowledge or action.

Also, it is a bad Rails practice to check the HTTP verb inside an action. The route should recognize the http verb and send to two different actions.

app/assets/javascripts/application.js.erb
```diff
@@ -636,9 +636,11 @@ onPageLoad(() => {
   });
 
   on('click', 'a.story-disowner', (event) => {
+    event.preventDefault();
+
     if (confirm("Are you sure you want to disown this story?")) {
       let li = parentSelector(event.target, '.story');
-      fetchWithCSRF('/stories/' + li.getAttribute('data-shortid') + '/disown', { method: 'post' })
+      fetchWithCSRF(event.target.getAttribute("href"), { method: 'post' })
```
pushcx, line 643: I really like how this change means the js reuses the existing URL generation instead of reimplementing it.

pushcx (changes requested):
Thanks for revising. This is basically what I asked for to allow the user to disown a story without js. But it introduces a security problem because it takes action for the user based on a GET request, which can't be trusted this way. 

Take a look at `CommentsController#create` (well, it calls `render_created_comment`): it always requires a POST, but it checks `request.xhr?` to see if the comment is being created without js (so the whole page should be rendered) or via js (so only the comment should be rendered for the `postComment` js to replace into the page).

Eduardo-Lpz:
@pushcx 
This type of concept are not highly stick to my head since i have not face it commonly, usually was already done by someone else. 😅 

This PR made me refresh/learn concepts and improve a little bit my skills. 👍

pushcx:
Thanks for starting this cleanup! You did a great job of building out a new pattern for fixing all of these `href="#"` links. I hope you'll take on the others, but even if you don't, you did the hard part by writing the first one, thank you.

Issue #1246, opened 2024-02-02
code in extras/ isn't reloading


pushcx:
In `config/application.rb` we configure the autoload path with a line of code straight out of the guide: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#config-autoload-paths

```
    config.autoload_paths << "#{root}/extras"
```

Despite this, files in `extras` are not reloaded on change, making dev tedious. This is in development mode (which does not change the default of `config.enable_reloading = true`).

While this is a good first issue if you are new to the codebase, this is probably not if you are a junior dev. This is maybe a weird bug in Rails or maybe some kind of subtle, accumulated misconfig based on our years of updates from Rails 3.2.2. Or maybe I'm staring through some obvious typo in the word "extras", I don't even know anymore after how many times I've lost time kicking this around.

pushcx:
As feared, fixing this blew up in prod.

`config/initializers/production.rb` wants to set API keys in `Pushover`, `Github`, etc. Those classes live in `extras/`, which now isn't loaded prior. So it threw a `NameError`. I can't tell from [the doc](https://guides.rubyonrails.org/initialization.html) at what point zeitwerk _should_ be up and running, but apparently not yet. (It's not misconfig; I tried loading the `User` model and got the same exception.)

To fix it in prod without a risky refactor to how these constants get set I added a `require` line to `config/initializers/production.rb`... and an apology to sister sites that will need to do the same. See 51420410af63fc3857593aa01c608fa864295dfc.

Maybe all of `config/initalizers/production.rb` can move into `config/environments/production.rb`? The goal is to have a place for sites to put their API keys that isn't committed to git, but the whole thing with class variables is also a kludge (see `.standard.yml` disabling `Naming/VariableName`).

PR #1365, opened 2024-11-06
Fix reloading for extras folder

krydos:
Hi :) 
Fixes https://github.com/lobsters/lobsters/issues/1246

According to documentation, adding `extras` folder to `autoload_path` [here](https://github.com/lobsters/lobsters/issues/1246) is the right way to make every file in it autoloaded.
However it looks like we also manually `require` all files in this folder [here](https://github.com/lobsters/lobsters/blob/master/config/initializers/00_zeitwerk.rb#L14) which probably (I don't actually know) messes up with RoR reload mechanisms. 

Hope this small fix can make hacking on lobsters a more pleasurable experience. 

Let me know what do you think. 

P.S.
Should we add that `lib` folder to `autoload_path` in the `config/application.rb` similarly to how `extras` folder is added? I guess the whole `each` block can be removed then. Apparently there is/was(?) something wrong with how files are autoloaded on prod based on [this](https://github.com/lobsters/lobsters/blob/master/config/initializers/00_zeitwerk.rb#L6-L7) comment. Do you remember any details on it by any chance?

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,7 +42,6 @@ module Lobsters
     config.eager_load_namespaces << I18n
 
     config.autoload_paths.push(
-      "#{root}/app/types",
       "#{root}/extras",
       "#{root}/lib"
     )
--- a/config/application.rb
+++ b/config/application.rb
@@ -43,7 +43,8 @@ module Lobsters
 
     config.autoload_paths.push(
       "#{root}/app/types",
-      "#{root}/extras"
+      "#{root}/extras",
+      "#{root}/lib"
     )
 
     # Raise an exception when using mass assignment with unpermitted attributes
--- a/config/initializers/00_zeitwerk.rb
+++ b/config/initializers/00_zeitwerk.rb
@@ -9,7 +9,7 @@ Rails.autoloaders.main.ignore(Rails.root.join("extras/prohibit*rb"))
 Rails.autoloaders.main.ignore(Rails.root.join("lib/monkey.rb"))
 require Rails.root.join("lib/monkey.rb").to_s
 
-%w[extras lib].each do |dir|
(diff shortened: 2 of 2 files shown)


pushcx:
Thanks for taking on this longtime hassle! I strongly suspect that you're right, our code trying to work around some old change is causing this hassle. Perhaps it's loading the files outside of or before the autoloader so the autoloader deliberately ignores changes in code or simply isn't aware of them.

I'm going to merge and deploy this when I have a chance to babysit production, because this strongly feels like one of those PRs that's going to act differently in dev and prod (even running in the Rails `production` mode locally).

krydos:
Let me know if you have any ideas on how I can replicate the environment that is close to lobste.rs prod. I'd be happy to run some tests. I of course tried `rails s -e production` but I assume it's not enough. 

I've spent quite some time looking into zeitwerk source code trying to solve this issue until one lucky grep helped me to find this `00_zeitwerk.rb` file 😄

pushcx:
At RubyConf last week I spoke with John Hawthorne who's worked on zeitwerk and he confirmed your diagnosis, that requiring these files before zeitwek loads would produce this behavior.

So please do add `lib` to the autoload paths and go ahead and drop this loop that manually requires altogether, and I'll get this merged. Thanks again for figuring this out.

krydos:
wow, thanks for double checking with John Hawthorne, that's pretty cool 😎 

I also went ahead and removed `app/types` folder from autoloading because there is no such folder. And even if we have it, everything in `app/` folder is autoloaded anyway.

pushcx:
The linter failure is me sloppily breaking master and I've fixed it there, so this is set to merge.

Thanks again for investigating this odd behavior. Here goes nothing!

PR #1369, opened 2024-11-08
Replace `.execute` with `.select_all`

### LINKED ISSUE #1367: Replace .execute with .select_all
https://github.com/jbranchaud/til/blob/master/rails/prefer-select-all-over-execute-for-read-queries.md

jasonmokk:
Fixes #1367 

Refactored instances of `.execute` in read-only queries to use `.select_all` instead. There are two more instances of `.exeucte` that are used for UPDATE queries in `app/models/comment.rb` and `app/models/story.rb`, but these were not refactored because `.select_all` is meant only for read-only queries.

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -18,7 +18,7 @@ class StatsController < ApplicationController
       graph_title: "Active users by month",
       scale_y_divisions: 500
     }) {
-      User.connection.execute <<~SQL
+      User.connection.select_all <<~SQL
         SELECT ym, count(distinct user_id)
         FROM (
           SELECT date_format(created_at, '%Y-%m') as ym, user_id FROM stories
--- a/app/helpers/traffic_helper.rb
+++ b/app/helpers/traffic_helper.rb
@@ -11,7 +11,7 @@ module TrafficHelper
   def self.traffic_range
     div = PERIOD_LENGTH * 60
     start_at = 90.days.ago
-    result = ActiveRecord::Base.connection.execute <<-SQL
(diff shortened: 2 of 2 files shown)


pushcx:
Thanks for this!

Did you happen to notice if these have any tests? Hairy queries and low testing often go hand-in-hand, unfortunately. I'm asking to judge how risky the deploy will be.

jasonmokk:
I checked and there aren't any tests covering these queries currently. Should I go ahead and add tests for these?

pushcx:
I'd really appreciate them, if you can. Even a simple smoke test would be an improvement.

The brakeman part of the build is failing because you touched a query that does string interpolation into a SQL query, a practice that often leads to SQL injection attacks. Run `brakeman -I` to update + commit the signature, you can reuse the existing note for that query.

pushcx:
If you don't have the spare time and energy to add tests, I'm happy to merge this and update the issue to note the remaining work. Do you want to close out the `brakeman` warning so I can merge?

jasonmokk:
> If you don't have the spare time and energy to add tests, I'm happy to merge this and update the issue to note the remaining work. Do you want to close out the `brakeman` warning so I can merge?

Yes that'd be great! I've been quite busy as of late and probably won't be able to add the tests for now.

pushcx:
Merged, but as part of updating `master` this was rebased to ee984880 so github doesn't recognize the PR as merged.

Thanks for contributing this improvement and letting me know about your availability! Hope to see you again in the issue tracker when you have a little more free time. :D

Issue #1366, opened 2024-11-08
Upgrade to Rails 8


pushcx:
Todos from [announcement](https://rubyonrails.org/2024/11/7/rails-8-no-paas-required):

 - add Solid Queue and ActiveJob (in prod, store dbs in `/srv/lobste.rs/job`)
 - add + config Solid Cache for views (in prod; `/srv/lobste.rs/cache`; 
 - our `Keystore` does a lot of misc caching, should Solid Cache take over some of these?
 - experiment: add Kamal 2 its Dockerfile for easier dev + prod setup
 - configure Propshaft; remove `nodejs` from ansible
 
Todos from [release notes](https://guides.rubyonrails.org/8_0_release_notes.html):

 - write todos from release notes

rahul1990gupta:
[Rails 8](https://youtu.be/-cEn_83zRFw?si=VuSy0AqsZ0uT-QUF&t=3080) also released [thruster](https://github.com/basecamp/thruster), which replaces the need for [nginx](https://github.com/lobsters/lobsters-ansible/tree/master/roles/nginx) to be used in a rails application. Thruster provides 
- SSL certifications with letsencrypt
- Caching 
- asset compressions 

It uses sensible defaults and nice integrated with rails puma server. Also it is written in go, so concurrency will not be an issue. With this and kamal, lobsters can simplify the  production deplyment

pushcx:
That's a really big, complicated service to reimplement from scratch. The [introduction post](https://dev.37signals.com/thruster-released/) sounds like it's intended for low traffic internal services. Before I would put this into production I would need to build some confidence by seeing experience reports from people who've run it in production at similar scale to Lobsters.

I think thruster is also unlikely to work for us because we use full-page caching (actionpack-page_caching with `caches_page` in controllers) and this doesn't support that. Though that gem is unmaintained and we're probably going to be forced to redo our caching strategy within a year or two anyways when it breaks.

rahul1990gupta:
With [rails 8.1](https://youtu.be/-cEn_83zRFw?si=QB-MM3sgwOcZ1XJh&t=3767) (not yet released), the rails team is working on 
- **Action Notifier**
- **Active Record Search**: This will make full-text search easy. 
- **Action Text with  markdown**: With this we could use action text for rich formatting of comments. This will also have a markdown parser to parse md into html. With this, we could close this dependency on commonMarker #1234.

PR #1377, opened 2024-11-15
Upgrade to Rails 8.0

byroot:
Beside what the number may suggest, for as much as lobsters is concerned, this is a very minor upgrade. Most of Rails 8.0 is just Rails 7.2 with a few new defaults.

The only small issue caused by the upgrade was that `has_secure_password` now adds a methods to generate a password reset token that don't need to be perisisted, and that method did happen to conflict with the existing `password_reset_token` attribute.

There was also nothing in `new_framework_defaults_8_0.rb` that's worth keeping on the 7.2 settings for one deploy, so I just directly flipped the `config.load_defaults 8.0`.


app/models/user.rb
```diff
@@ -56,7 +56,10 @@ class User < ApplicationRecord
     class_name: "Hat",
     inverse_of: :user
 
-  has_secure_password
+  # As of Rails 8.0, `has_secure_password` generates a `password_reset_token`
+  # method that shaddows the explicit `password_reset_token` attribute.
```
pushcx, line 60:
```suggestion
  # method that shadows the explicit `password_reset_token` attribute.
```

pushcx (approved):

pushcx:
Thanks for taking a look at this, I'm glad it wasn't too much hassle.

I filed #1366 about the Rails 8 upgrade and I'm going to keep that open to organize the work around setting up the new features that make up the bulk of Rails 8.

Issue #1380, opened 2024-11-19
Clean up API key config in config/initalizers/production.rb


pushcx:
Following #1246

We have `config/initializers/production.rb` for setting API keys and site settings (see `config/application.rb` for the `class Lobsters` setup) that vary between the sister sites, but Rails style has long-since changed and I think there's a few new styles for better managing those.

To [my comment](https://github.com/lobsters/lobsters/issues/1246#issuecomment-2486766671) we should look at reimplementing this config so that Lobsters and sister sites use a single way that plays well with deploying to a VPS, PaaS, and Kamal, and doesn't end up with `require` statements to kludge around zeitwerk.

It would be especially nice if this cleanup didn't break prod for us or sister sites! And removed the kludge in `config/initializers/00_zeitwerk.rb`.

(Maybe this is the kind of plumbing issue I can nerdsnipe @byroot into?)

byroot:
Could you expand a bit?

Looking at your pointers, I see you are basically forcing eager loading of `app/extra` in production, but I can't understand which files it's for. Is it for some uncommitted files?

krydos:
If I may. 
Here is related PR https://github.com/lobsters/lobsters/pull/1365 that solved one issue and introduced this one. 

There is [config/initializers/production.rb.sample](https://github.com/lobsters/lobsters/blob/master/config/initializers/production.rb.sample) file that you should rename to `config/initializers/production.rb` to see the issue. 
As soon as you start rails with `rails s -e production` it's gonna fail because classes used in that initializer are not loaded yet. These classes are defined in `extras` folder and this folder is added to `config.autoload_paths` (see `config/application.rb`) but the folder's files are loaded AFTER the initializers. 

I thought about moving that initializer into a separate class(es?) and do `ThatSeparateClass.initialize` thing on them in [config.after_initialize](https://github.com/lobsters/lobsters/blob/master/config/application.rb#L63). 
I feel like it may work but I haven't had a chance to try yet. 

I hope it gives a bit more context on the issue.

pushcx:
After sleeping on it a couple days, I'm pretty sure the root cause is our pattern of having utility classes that an initializer sets class variables on. That starts this whole contention of whether the utility class is loaded before the initializer, which was compounded by not using the `autoload_paths` properly.

[...]

PR #1381, opened 2024-11-21
Cleanup lib/monkey.rb

byroot:
Ref: https://github.com/lobsters/lobsters/issues/1380

- I see little point in using a core extension for this.
- It's only used in one place.
- The implementation itself can be simplified a lot.

Not 100% why but fixing this caused the `scripts/*.rb` to start failing to load. But they were using a non-standard way to load the application, so I simply cleaned them up too and the issue disapeared, probably not worth on digging up the root cause.

cc @pushcx 

The only weird zeitwerk workaround left is `extras/prohibit_form_for_and_form_tag.rb`. The funny thing is that unless I'm mistaken, I'm pretty sure it's not loaded by `rubocop/standardrb`, so it's pretty much just dead code.

I'm not super familiar with rubocop/standard, so I'm trying to figure out the right way to do private, in-repo cops, I'll see if I can figure it out before I'm bored.


pushcx:
Thanks for spotting this. `lib/monkey.rb` used to have several monkey patches but we've removed their use over the years. I'm happy to not monkeypatch anything at all.

PR #1382, opened 2024-11-21
Remove useless private Cop

byroot:
Ref: https://github.com/lobsters/lobsters/issues/1380

This cop never did anything because `standardrb` isn't configured to load it, and also because it's not capable of parsing ERB files anyway.

To do that you'd need to bring `erb-lint` or something like that.

I have branch that tries to make it work, but it's a big hassle, so in my humble opinion it's not worth it.

@pushcx your call.


pushcx:
Huh. I thought I'd seen this work, maybe by manually naming it from the command line, but maybe not. The `if Rails.env.production?` at the top is a pretty clear indication that I struggled to get it to work.

It's definitely been broken for a long time, and the move from `form_tag` and `form_for` to `form_with` is far enough in the past that I'm not too worried we're going to reintroduce them. (There _is_ one remaining use of `form_tag` but it's already part of a larger refactoring; see details in #1355.)

Thank you for recognizing this cop. I'm going to go ahead and merge to remove the dead code.

PR #1383, opened 2024-11-22
Update URL when fetching title

yawaramin:
When the user clicks 'Fetch Title', in addition to filling in the title field, also automatically set the value of the URL field. This is primarily because servers often redirect to a canonicalized or secure (HTTPS) version of a URL if someone requests an alternate or insecure (HTTP) version.

With this change, we use the URL that the server prefers for the document.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -185,7 +185,8 @@ class _LobstersFunction {
   }
 
   fetchURLTitle(button) {
-    const targetUrl = qS('#story_url').value;
+    const url_field = qS('#story_url');
+    const targetUrl = url_field.value;
     const title_field = qS('#story_title');
     const formData = new FormData();
     const old_text = button.textContent;
@@ -204,6 +205,7 @@ class _LobstersFunction {
       .then (response => response.json())
       .then (data => {
         title_field.value = data.title
+        url_field.value = data.url
         button.textContent = old_text
       });
     button.removeAttribute("disabled");


app/assets/javascripts/application.js.erb
```diff
@@ -185,7 +185,8 @@ class _LobstersFunction {
   }
 
   fetchURLTitle(button) {
-    const targetUrl = qS('#story_url').value;
+    const url_field = qS('#story_url');
```
yawaramin, line 188: Btw fun fact, IDs that are syntactically valid JavaScript identifiers, automatically become variables pointing to their elements in the scope of the page. So we don't even need to use `qS('#story_url')`, we can access the element directly with `story_url`.

pushcx: I remember being burned by this Back In The Day when we used lots of globals in js, so I am leery of leaning on it now. :)

yawaramin:
Would injecting the message right below the URL input work? Eg:

Screenshot 2024-11-25 at 21 20 30

pushcx:
Sorry I missed this for a minute! I accidentally cleared the 'unread' notification and then didn't check back.

I think right below the URL is the best place for it, yeah. We do a very similar bit of point-in-time help for the `title-reminder` help text, so that'd be a good pattern to continue. 

For a message, how about: `The URL above has been updated to use the [canonical URL](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#canonical) given by the server.` I'd like the message to start with the URL as the object because it's what the user was just thinking about, then explain we're doing what the page's server told us.

pushcx:
@yawaramin Thanks for getting this started. I think it's a great little improvement for the story submission UI.

I saw your PR went stale, so I added a commit with that last bit of UI and merged it. Great to have your contribution here.

yawaramin:
Thanks! Got a little squeezed for time.

PR #1387, opened 2024-11-28
Fix stats exception

### LINKED ISSUE #1386: /stats is throwing an exception
```
  Can not parse time {"ym"=>"2012-06", "count(distinct user_id)"=>5}
  app/controllers/stats_controller.rb:90:in `block in monthly_graph'
  app/controllers/stats_controller.rb:60:in `monthly_graph'
  app/controllers/stats_controller.rb:17:in `index'
  app/controllers/application_controller.rb:198:in `n_plus_one_detection'
```

Probably related to #1369 or the recent bundle update.

ashwin47:
Fixes #1386


pushcx (changes requested):
Thanks for jumping in on this one. I tinkered a bit and learned that this is definitely caused by #1369: `execute` returns an array of arrays and `select_all` returns an array of hashes.

So the fix isn't going to need to check the class of the array elements, it should always be hashes.

I don't like this style of inspecting data in the middle and tweaking it. Just update the consumer to work with the changed data structure.

ashwin47:
`select_all` returns `ActiveRecord::Result`. I have updated the query to convert it to Hash. 

Also, brakeman seems to be giving a false positive warning.

pushcx:
When I said just update the consumer, that's TimeSeries. This query appears in one place and only gets used by one piece of code, in TimeSeries. So if the query produces a different data structure, rather than edit that data structure back to the old one, we can just have TimeSeries know the current one.

Also: the brakeman warning is because you changed the date formatting to an injected value with %. Why did you you change that? It doesn't alter the behavior of the query that I can see.

pushcx (changes requested):
I left a comment but need to need another to explicitly click 'request changes'

pushcx:
I saw you close that and spent some more time on it myself.

I'm sorry my guidance was wrong - in debugging, I didn't see how the one `monthly_graph` function was the problem, not accounting for the difference between the different charts' queries.

ashwin47:
I just didn't want you to be blocked from fixing this as I didn't have enough time to look into it again.

PR #1388, opened 2024-11-28
Add search highlight

### LINKED ISSUE #1373: Comment search: highlight search terms
When someone searches for terms in a comment, those terms could be highlighted with the [mark tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) in the comment text.

Mockup:

Comment text is stored in `comments.comment` as the user wrote, using markdown, and rendered to html in `comments.markeddown_comment`. This is so we don't incur that render cost on every view of a comment, but it means this feature will need to operate on the HTML rather than the markdown. (I suppose it's possible as part of adding Solid Cache in #1366 that we could move this functionality out of the database and into a fragment cache with a long expiry, say 365 days. Though we'd want to add significant testing that changes to our markdown code or commonmark don't lead to comment rendering changing meaningfully over time!)

A corner case: MariaDB's full-text indexes for search ignore punctuation, whether used as part of a sentence or as markdown. So you can search for a quoted phrase like `"sometimes correctly"` to find [this same comment](https://lobste.rs/s/xjs90z/regular_restarts_are_good_actually#c_qy2aox) where it is rendered in HTML as `correctly.</p></blockquote><p>Sometimes`.

ashwin47:
Resolves #1373 

CleanShot 2024-11-28 at 12 37 21@2x

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -151,4 +151,15 @@ module ApplicationHelper
     ago = time_ago_in_words(time)
     content_tag(:span, ago, title: time.strftime("%F %T %z"))
   end
+
+  def highlight_search_terms(text, search_terms)
+    return text if search_terms.blank?
+
+    terms = search_terms.split(/\s+/).map { |t| Regexp.escape(t) }
+    regex = terms.join("|")
+
+    text.gsub(/(#{regex})/i) do |match|
+      "<mark>#{match}</mark>"
+    end
+  end
 end
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -71,4 +71,37 @@ describe ApplicationHelper do
         .to eq([1, "...", 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25])
     end
   end
+
+  describe "#highlight_search_terms" do
+    it "returns original text when search terms are blank" do
+      text = "This is a test comment"
+      expect(helper.highlight_search_terms(text, nil)).to eq(text)
+      expect(helper.highlight_search_terms(text, "")).to eq(text)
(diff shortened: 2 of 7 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -71,4 +71,37 @@
...
+    it "highlights multiple words" do
+      text = "The quick brown fox jumps over the lazy dog"
+      result = helper.highlight_search_terms(text, "quick dog")
+      expect(result).to eq("The <mark>quick</mark> brown fox jumps over the lazy <mark>dog</mark>")
+    end
+
+    it "handles special regex characters in search terms" do
+      text = "Testing (parentheses) and [brackets]"
+      result = helper.highlight_search_terms(text, "(parentheses)")
+      expect(result).to eq("Testing <mark>(parentheses)</mark> and [brackets]")
```
pushcx, line 98: I'm a little puzzled that this works, because the use of `strip_operators` will strip out that punctuation. I don't have a strong preference if this search in your test produces `<mark>(parentheses)</mark>` or `(<mark>parentheses)</mark>)`. The latter makes a little more sense and you may get that result when you make the other change to how `strip_operators` is called.

app/helpers/application_helper.rb
```diff
@@ -151,4 +151,15 @@ def time_ago_in_words_label(time)
     ago = time_ago_in_words(time)
     content_tag(:span, ago, title: time.strftime("%F %T %z"))
   end
+
+  def highlight_search_terms(text, search_terms)
```
pushcx, line 155: I'm not sure here, but maybe pass in the `Search` object? It seems odd to extract terms from `@search` but only pass them into this method. I'm not sure how it'll look after the refactor around `strip_operators` though.

Issue #1378, opened 2024-11-17
‘Relevance’ search ranking of stories not particularly helpful


dpk:
Especially since some PDF documents of academic papers may be hosted at different URLs, I use search for keywords in the title to try to find out if some paper has been posted before. Unfortunately, the ranking options do not make this very easy. The most appropriate one (based on my experience with other search engines) is probably ’Relevance’.

Here’s a real example from today which demonstrates how this ranking could be improved. I read [this](https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf) and thought it would make a good Lobsters post. I searched for [tutorial good lisp style] and these were the results: <https://lobste.rs/search?q=tutorial+good+lisp+style&what=stories&order=relevance>

<img width="1319" alt="Screenshot 2024-11-17 at 13 13 14" src="https://github.com/user-attachments/assets/8488df69-3f78-4de2-914a-3c87b8dd2dbf">

Thinking I had something new, I went to submit it, only to find [the URL itself had been submitted before, very recently](https://lobste.rs/s/yzi1jy/tutorial_on_good_lisp_programming_style). Given all the keywords from my search query appear in the title of the submission, it’s not clear to my why ‘relevance’ doesn’t put this submission even on the first page!

pushcx:
As a pointer, here's where to add the relevance weighting to [story searches](https://github.com/lobsters/lobsters/blob/b537ee6e683d444f90b9707168c11d8783becff4/app/models/search.rb#L215) and [comments](https://github.com/lobsters/lobsters/blob/b537ee6e683d444f90b9707168c11d8783becff4/app/models/search.rb#L323).

PR #1389, opened 2024-11-29
Add relevance weighting to relevance searches #1378

Ryzen114:
This pull request enhances the relevance-based sorting functionality in the perform_story_search and perform_comment_search methods by leveraging FULLTEXT search capabilities.

Key Changes:
Relevance Sorting for Stories:
Added weighted FULLTEXT search across title, description, and body columns in story_texts.
Assigned weights to prioritize results:
title: weight 10
description: weight 3
body: weight 1

Relevance Sorting for Comments:
Implemented FULLTEXT search on the comment column to rank results based on relevance.

Database Migration:
Added FULLTEXT indexes for columns used in the relevance sorting:
title, description, and body in story_texts

Testing: 
Before
Capture d'écran 2024-11-29 110817

After
Capture d'écran 2024-11-29 110916

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -213,7 +213,7 @@ class Search
       # relevance is undefined without search terms so sort by score
       if terms.any?
         terms_sql = <<~SQL.tr("\n", " ")
-        MATCH(comment) AGAINST ('#{terms.map { |s| "+#{s}" }.join(" ")}' IN BOOLEAN MODE) 
+          MATCH(comment) AGAINST ('#{terms.map { |s| "+#{s}" }.join(" ")}' IN BOOLEAN MODE) 
         SQL
         query.order!(Arel.sql(terms_sql + " DESC"))
       else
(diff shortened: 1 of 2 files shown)


app/models/search.rb
```diff
@@ -212,6 +212,9 @@ def perform_comment_search!
     when :relevance
       # relevance is undefined without search terms so sort by score
       if terms.any?
+        terms_sql = <<~SQL.tr("\n", " ")
```
pushcx, line 215: Please rename this variable to `order_sql`. There's a lot going on in this method, I don't want to reuse a variable name and have to keep track of which it is at which line of code.

app/models/search.rb
```diff
@@ -320,6 +323,12 @@ def perform_story_search!
     when :relevance
       # relevance is undefined without search terms so sort by score
       if terms.any?
+        # Assigns different weights: title (10), description (3), body (1) for scoring relevance.
+        terms_sql = <<~SQL.tr("\n", " ")
```
pushcx, line 327: Same rename please.

pushcx (changes requested):
This is great, good choice of weights.

Did you need the three indexes on `story_texts`? We already have one `FULLTEXT` index that covers the same three fields. Does MariaDB not use it if you only search one field for each weighting? In which case, is there still a value to having the one combined index, or should that be dropped?

pushcx:
The build failure is `brakeman`, the security scanner that tries to detect potential SQL injection whenever we do something like combine strings to make a SQL query. It can't know that the `terms` array is all values that have passed through `strip_operators` for sanitization. You can satisfy it by running `brakeman -I` to inspect the new warnings, and add note that it's safe. You can reuse the text of the existing note (visible in [config/brakeman.ignore](https://github.com/lobsters/lobsters/blob/master/config/brakeman.ignore)) of "inputs pass through strip_operators; has some injection tests",

PR #1390, opened 2024-11-30
#fix issue 1298 : hiding merged stories removes the "on + link to story" in the comment title.

Blosterhxh:
Im not sure of my understanding of this issue, yet i tried something. So what it does is that when merged stories are hidden, there are no more comments with the "on + link to story" in the title. However, it doesnt apply immediately after hitting hide, you have to refresh the page manually after that so im not sure its the fix you are looking for.

### DIFF
--- a/app/views/comments/_threads.html.erb
+++ b/app/views/comments/_threads.html.erb
@@ -25,7 +25,7 @@
       force_open = true
       show_tree_lines = true
       # show "on: <title>" for comments from merged stories + on non-story pages
-      show_story ||= story.try(:id) != comment.story_id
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for taking a look at this. In the bug report, when gustafe says they're looking at their comments feed, so that's either `/comments` or `/~gustafe/threads`. I think they expected that when they click "hide" on a story that has others merged into it, or is merged into others, that comments from all stories would disappear. Your PR changes the heading over the comment, when the whole comment should disappear.

I think this is going to have to be fixed by a refactor to the story model, because `hidden_stories` doesn't know anything about story merging and it's a complex join to carry through to every queries that load comments.

PR #1399, opened 2024-12-14
Fix exception raised in StoryUrlsController#all

### LINKED ISSUE #1398: 500 from StoryUrlsController#all
```
An ActionController::ParameterMissing occurred in story_urls#all:

  param is missing or the value is empty or invalid: url
  app/controllers/story_urls_controller.rb:5:in `all'
```

Line 5 requires the `:url` parameter, so that's working correctly, but it needs to return a 400 or a 200 with empty results or something rather than throw an exception and return 500.

kernal053:
Fix #1398

### DIFF
--- a/app/controllers/story_urls_controller.rb
+++ b/app/controllers/story_urls_controller.rb
@@ -9,6 +9,10 @@ class StoryUrlsController < ApplicationController
         render json: Story.find_similar_by_url(url).for_presentation
       }
     end
+  rescue ActionController::ParameterMissing
+    respond_to do |format|
+      format.json { render json: {}, status: :bad_request }
+    end
   end
 
   def latest
--- a/app/controllers/story_urls_controller.rb
+++ b/app/controllers/story_urls_controller.rb
@@ -9,10 +9,6 @@ class StoryUrlsController < ApplicationController
         render json: Story.find_similar_by_url(url).for_presentation
       }
     end
-  rescue ActionController::ParameterMissing
-    respond_to do |format|
-      format.json { render json: {}, status: :bad_request }
-    end
   end
 
   def latest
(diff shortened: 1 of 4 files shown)


app/controllers/story_urls_controller.rb
```diff
@@ -9,6 +9,10 @@ def all
         render json: Story.find_similar_by_url(url).for_presentation
       }
     end
+  rescue ActionController::ParameterMissing
+    respond_to do |format|
+      format.json { render json: {}, status: :bad_request }
+    end
   end
 
   def latest
```
kernal053, line 18: same exception is raised for this action too. What should be the ideal way to handle it ?

pushcx: I spent a minute looking into it on stream today, a viewer pointed out that this should probably happen more universally by catching the `ParameterMissing` exception in middleware. The Strong Parameters docs do recommend using the raise/rescue pattern, and I realized we already do this for the other exception it raises.

If you look at ApplicationController, line 34 rescues `UnpermittedParameters`. Could you move your rescue up alongside that?

Ideally, your `respond_to` block could also return a `plain` version, and the existing rescue for `UnpermittedParameters` could do that, too. But I realize that's extra work and I can do it if you don't want to.

pushcx (changes requested):
Thanks for looking into this. You noted that it's a broader issue and I think there's an opportunity to handle it globally for the app (even as it's odd that Rails doesn't already; but this is probably only triggered by a bot with a really unusual header/url/params combo).

app/controllers/story_urls_controller.rb
kernal053: I have made changes as per above suggestion and updated few specs accordingly. Please do check.

pushcx:
Sorry this took me a minute to get back to, I was sick and then traveling for the holiday.

This looks great, and I really appreciate that you added a solid set of tests. Thanks!

PR #1405, opened 2024-12-22
Add cop to standardrb to enforce Mod::ModeratorController #1393

### LINKED ISSUE #1393: Add a cop to our standardrb to enforce Mod::ModeratorController
All controllers in `app/controllers/mod/` should inherit from `Mod::ModeratorController`. The only way I can think of to enforce this off the top of my head would be to [create a cop](https://docs.rubocop.org/rubocop/development.html) and add it to our standardrb config (I don't know how, but standardrb is a set of rubocop defaults, so it should be possible, right?).

dreig:
All controllers in the `app/controllers/mod/` directory should inherit from `Mod::ModeratorController`
resolves #1393

Note:
The cops are located in the `.lib/custom_cops/` directory, which I've added to the ignore list of `config.autolod_libs`. If my understanding of [config.autoload_libs(ignore:)](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#config-autoload-lib-ignore) is correct, this should prevent this directory from being loaded when the application runs (in any environment).

If your editor is configured with standardrb integration, the code in `.lib/custom_cops/` will load and will run for ruby files in the `app/controllers/mod/` directory.

### DIFF
--- a/.custom_cops.yml
+++ b/.custom_cops.yml
@@ -1,0 +1,11 @@
+require:
+  - ./lib/custom_cops/inherits_moderator_controller.rb
+  # to use additional cops, list individual ruby files here
+  #   (listing directories or glob patterns doesn't seem to work)
+  # ex:
+  # - ./lib/custom_cops/<cop-filename>.rb
+
+CustomCops/InheritsModeratorController:
+  Enabled: true
+  Include:
+    - "app/controllers/mod/**/*.rb"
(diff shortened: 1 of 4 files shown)


pushcx:
Sorry this took me a minute to get to, I was sick and then traveling for the holiday.

This is a totally clear implementation, I’m really happy to merge it so I don’t footgun myself by creating a controller that doesn’t inherit from `ModController` for the authorization. Thank you!

Issue #1279, opened 2024-05-24
Redesign forms and user profile using CSS grid


pushcx:
All of the site forms and the user profile display use a CSS design pattern of having fixed-width (7em) labels that `float: left;`. This is a fairly brittle design and has caused many bugs with content overlapping the labels (#1278), wide content breaking onto new lines, form fields that don't fill the width correctly*, etc.

Since the site was created, CSS has gained a [grid layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) that's the perfect tool for this job. (Well, probably! It's certainly possible there's something even better I'm unaware of.) A grid layout would allow us to directly express this idea that there's a column of labels matched with a column of fields, and then on mobile the labels appear above the fields.

This is a pretty good-sized redesign, if anyone would like to contribute. I guess the first step would be to make a checklist of the site forms and user profile, then work down through it with one or more PRs. Maybe there's the opportunity for more design improvements that are in keeping with the site's pretty plain look, relying on native widgets.

 * My current view in Firefox (desktop with a narrow window to see the mobile layout):

![2024-05-24_16-46](https://github.com/lobsters/lobsters/assets/6262/b111842e-cdd4-48b9-be7c-b8cd143db952)

amu1996:
@pushcx I would like to work on this.

venkat1004:
@pushcx  I would like to work on this.

chamlis:
Work has started here with #1567, but there's a lot of pages to get through, so let's spread everything out and take stock of what we've got.

Forms:
- [x] `categories/_form.html.erb`
- [x] `categories/edit.html.erb`
- [x] `categories/new.html.erb`
- [ ] `comments/_commentbox.html.erb`
- [ ] `comments/_threads.html.erb`
- [ ] `domains/edit.html.erb`
- [ ] `filters/index.html.erb`
- [x] `hat_requests/index.html.erb`
- [x] `hat_requests/new.html.erb`
- [x] `hats/doff.html.erb`
- [x] `hats/edit.html.erb`
- [x] `invitations/build.html.erb`
- [x] `invitations/index.html.erb`
- [x] `login/index.html.erb`
- [x] `login/forgot_password.html.erb`
- [x] `login/set_new_password.html.erb`
- [x] `login/twofa.html.erb`
- [x] `messages/_form.html.erb`
- [x] `messages/index.html.erb`
- [x] `messages/show.html.erb`
[...]

PR #1410, opened 2024-12-26
Sprockets to propshaft

rahul1990gupta:
### Overview
This PR uses propshaft in place of sprockets for asset management and delivery. This is in line with rails 8's recommendation and as outlines in this issue https://github.com/lobsters/lobsters/issues/1366

This PR 
- Removes sprocket and uglifier. Adds importmap-rails and propshaft gems for javascript  and css delivery. 
- Changes application css from erb template to vanilla css file. 

### Testing 

https://github.com/user-attachments/assets/f50d6091-31b6-4add-b9e1-63d659419e97

https://github.com/user-attachments/assets/985c1a2e-96bc-4292-85bd-a131f9e18611


pushcx:
I don’t want to split up the css into multiple files. I don’t see any benefit for a site this small.

Also, the `LIGHT` and `DARK` variables are there so users can override their browser’s default, see the references to `prefers_color_scheme`. I only glanced at the diff here but I think your change might break that. I’d love to reduce the duplication of having to print the `dark` colors twice and saw a mention on Bluesky that an expert thought it was possible, but I can’t find it again in the limited time I have this morning. I’ll have to keep looking.

rahul1990gupta:
The reason I suggested breaking up application.css to make the css more modular and with http 2, it is supported without any performance penalty. But i get it that benefits are mostly cosmetic and css  file is not that too large. 

A side question, what do you think about incorporating tailwindcss into the project ? It would be a larger rewrite of the css, but may make the UI more "modern".. maybe ? :D

pushcx:
Tailwind would reintroduce the nodejs dep I want to remove (or [other](https://tailwindcss.com/blog/standalone-cli) build complexity). It would add a new, big dependency to maintain and for potential contributors to learn. If you want a big css project, #1279 has one. But let's keep it separate from this PR!

pushcx:
I can't find the original article on reducing duplication in the dark and light versions of the CSS but [this article on light-dark](https://12daysofweb.dev/2024/css-light-dark/) has a design pattern that looks great. Probably also worth putting in a different PR than this build update!

rahul1990gupta:
Light-dark css property is great. Maybe will take it up in PR.

rahul1990gupta:
I was looking at the autosize.js javascript library used in the project and was wondering if is possible to replace these legacy libraries with something simple in modern html and js features. 

Will something like [this](https://codepen.io/rahul1990gupta/pen/PwYJdzw) capture all autosize.js functionalies.

pushcx:
So, I don’t think I’ve looked at that autosize js for years, I’m glad you took a fresh look at it. Reviewing, I think you’re right, we could drop that dependency and have just a little custom code for it.

It looks like you’ll need to add `word-wrap: break-word;` to the css (see `autosize.js` line 209). Otherwise a lot of the existing js works around layout issues we don’t have or events/lifecycle hooks we don’t need. Two things I don’t fully understand: 1. is there special behavior when the page resizes up or down? does the current library change autosize when the textarea is taller than the visible window and that resizes? 2. In `resize()`, what’s the purpose of the `scrollTop` code? Is it making sure it doesn’t feel like the textarea’s scroll position changes as the `textarea` changes size?

If you want to continue removing the library, please open a separate PR for removing it. I really look forward to it!

app/javascript/application.js
```diff
@@ -93,11 +95,11 @@ class _LobstersFunction {
     this.curUser = null;
 
     this.storyFlagReasons = ({
-      <%= Vote::STORY_REASONS.map{|k,v| "'#{k}': '#{v}'" }.join(", ") %>
+      'O': 'Off-topic', 'A': 'Already Posted', 'B': 'Broken Link', 'S': 'Spam', '': 'Cancel'
```
pushcx, line 98: This (and `commentFlagReasons`) exist to avoid duplicating flags between the Vote model and here. Can you find another way to remove this duplication? Is there any negative to doing this with ERB?

vendor/javascript/.keep
pushcx, line 1: I hate the Rails pluralization rules, but let's follow then and call this `vendor/javascripts`.

rahul1990gupta: This was auto generated by `rails importmap:install` command. Changing this may cause other problems.

pushcx: Well, my hate for the Rails pluralization rules is how I can never remember the arbitrary switching between singular and plural, so it's fitting that I'm wrong about this one. If this directory is what the generator gave you, that's what I want. Thanks for explaining.

app/javascript/application.js
pushcx: Ah, thanks for explaining.

It was already a project to remove inline scripts from the site (#696 #697 #779) as a long-term security improvement, so I don't want to add one back.

One option would be to have the js be the authoritative list of flag reasons and have Rails parse it out at startup. Maybe it's because I'm mostly a backend developer, but that feels backwards and I don't want to do that.

I guess we only add and remove flags very infrequently. Please add a comment to this js explaining the reasons are duplicated in the Vote model, and add the reverse comment to Vote model.

rahul1990gupta: Added the comments.

pushcx: I talked this through on stream and someone pointed out that I misunderstood your suggestion. An easy way to remove this duplication would be to have an invisible html element with the flags in json and have the js load them. Now that I understand that, I think that's the best idea. Sorry I had you write comments.

We don't need to make it a standalone helper please add a `meta` tag in the `head`.

So I think that'll be `<meta name="storyFlags" content="<%=  Vote::STORY_REASONS.to_json ">` and something very similar for `commentFlags`.

rahul1990gupta: I spent close to 2 hours making test work, but rspec is new to me. Will spend some more time later.

pushcx (changes requested):
There's a bug with light and dark mode, if you work through those combinations of settings I'm ready to merge this.

app/assets/stylesheets/application.css
```diff
@@ -85,9 +77,11 @@ light = <<-LIGHT
 	--color-mobile-story-comments-bg: #e8e8e8;
 	--color-mobile-story-comments-bubble-fill: #ccc;
 	--color-mobile-story-comments-bubble-fill-zero: #d8d8d8;
-LIGHT
+}
 
-dark = <<-DARK
+@media (prefers-color-scheme: dark) {
+  html.color-scheme-system,
```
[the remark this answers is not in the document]
rahul1990gupta: - [ ] browser set to light mode, no user setting (easy to test logged out) -> see light
- [ ] browser set to dark mode, no user setting (easy to test logged out) -> see dark
- [ ] browser set to light, user setting is 'system' -> see light
- [ ] browser set to dark, user setting is 'system' -> see dark
- [ ] browser set to light, user setting is 'dark' -> see dark
- [ ] browser set to dark, user setting is 'dark' -> see dark
- [ ] browser set to light, user setting is 'light' -> see light
- [ ] browser set to dark, user setting is 'light' -> see light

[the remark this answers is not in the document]
rahul1990gupta: my bad. it misses a case.

rahul1990gupta:
Testing after the commit.

- [ ] browser set to light mode, no user setting (easy to test logged out) -> see light
- [ ] browser set to dark mode, no user setting (easy to test logged out) -> see dark
- [ ] browser set to light, user setting is 'system' -> see light
- [ ] browser set to dark, user setting is 'system' -> see dark
- [ ] browser set to light, user setting is 'dark' -> see dark
- [ ] browser set to dark, user setting is 'dark' -> see dark
- [ ] browser set to light, user setting is 'light' -> see light
- [ ] browser set to dark, user setting is 'light' -> see light

app/assets/stylesheets/application.css
```diff
@@ -163,20 +156,87 @@ dark = <<-DARK
...
-		<%= dark %>
-	}
+  }
 }
+
+
 html.color-scheme-dark {
-	<%= dark %>
-}
+	/* main monochrome */
```
rahul1990gupta, line 164: This is duplicated. I couldn't find a good way to reuse the variables and their values. 

On a side note, It will be nice to deprecate the user setting of light/dark mode and use the system settings, as most operating systems and browsers support it now. This will simplify the codebase as well.

pushcx: Did you look at [light-dark](https://12daysofweb.dev/2024/css-light-dark/)? I linked it earlier, maybe it would work for removing the duplication.

I don’t want to remove the user setting. I can’t claim to understand it, but people are feel very strongly about having it per-site.

pushcx: Yeah, I spent a minute reading the MDN page. In the [example section](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark#example) under the heading "CSS" they show moving all the variables into `:root`. Then an example of making a section of the page light mode, but rather than having a `.light` class to force the color scheme, we'd have `html.color-scheme-dark { color-scheme: dark; } html.color-scheme-light { color-scheme: light; }`.

I think that would solve this duplication.

rahul1990gupta:
Testing after the commit.

- [X] browser set to light mode, no user setting (easy to test logged out) -> see light
- [X] browser set to dark mode, no user setting (easy to test logged out) -> see dark
- [X] browser set to light, user setting is 'system' -> see light
- [X] browser set to dark, user setting is 'system' -> see dark
- [X] browser set to light, user setting is 'dark' -> see dark
- [X] browser set to dark, user setting is 'dark' -> see dark
- [X] browser set to light, user setting is 'light' -> see light
- [X] browser set to dark, user setting is 'light' -> see light

pushcx:
That's a really big refactor with light-dark, thanks for working through it and testing it against all those conditions.

This is ready to merge. My experience with the Rails asset pipeline is that every time I deploy a significant change, it breaks, but only in production. So I will wait to merge and deploy until this weekend. But I wanted to write to thank you for completing this project.

pushcx (approved):

app/assets/stylesheets/application.css
```diff
@@ -82,16 +73,21 @@
...
 	--color-mobile-story-border: light-dark(#e0e0e0, #1c1c1c);
 	--color-mobile-story-liner-bg: light-dark(#fbfbfb, #0c0c0c);
 	--color-mobile-story-comments-bg: light-dark(#e8e8e8, #101010);
 	--color-mobile-story-comments-bubble-fill: light-dark(#ccc, #282828);
 	--color-mobile-story-comments-bubble-fill-zero: light-dark(#d8d8d8, #1c1c1c);
-
 }
 
+/*
+By default, in supporting browsers, the color returned by the light-dark() color function depends
```
pushcx, line 85: Great comment explaining this functionality, thanks.

rahul1990gupta:
Tested it again after resolving the recent merge conflict. It is good to go. I will be checking my email this weekend regularly(morning/evening), in case deployment fails.

Let me know if anything else is needed.

pushcx:
Here goes nothing. 😄

pushcx:
Wow, I’m stunned. It deployed clean and Just Worked. Impressive work! Thank you again, Rahul.

PR #1411, opened 2024-12-29
Attempt to fix #1402 (Exception about duplicate Origin records in check_url_dupe)

### LINKED ISSUE #1402: Ironic exception about duplicate records in check_url_dupe
```
An ActiveRecord::RecordNotUnique occurred in stories#check_url_dupe:

  Trilogy::ProtocolError: 1062: Duplicate entry 'medium.com/leboncoin-tech-blog' for key 'index_origins_on_identifier' (trilogy_read_row)
  app/models/domain.rb:77:in `find_or_create_origin'
  app/models/story.rb:950:in `set_domain_and_origin'
  app/models/story.rb:992:in `url='
  app/controllers/stories_controller.rb:417:in `check_url_dupe'
  app/controllers/application_controller.rb:198:in `n_plus_one_detection'
```

The Origin objects are implicitly created in `find_or_create_origin` so perhaos it's failing to find an existing origin, or perhaps there's a race condition because the frontend is firing `check_url_dupe` twice. Either way this is an uncommon intermittent error.

dreig:
I could trigger the exception described in #1402 from the browser (locally) by sending several requests to `stories#check_url_dupe`.

I wrote a simple spec to reproduce the error by calling `find_or_create_origin` from two separate Threads. (But I don't know how to integrate the spec into the test suite, because it relies on the `self.use_transactional_tests = false` config which interferes with the rest of the suite. And I don't think having this spec in the suite is even worth it, beyond fixing the current issue.
I suggest running this spec by itself `rspec spec/models/origin_spec.rb`, but disregard the commit if you decide to merge the PR)

The fix was to use `find_or_create_by!` which will attempt to create the record if it doesn't already exist, but if that fails with `ActiveRecord::RecordNotUnique` error, it will assume it encountered a race condition and will try to find the record once more.
https://api.rubyonrails.org/v8.0.1/classes/ActiveRecord/Relation.html#method-i-find_or_create_by

(I set the PR as Draft because I only want to contribute the one commit)


pushcx:
Ah, yeah, I can see how this should fix the problem. Thank you. You cut through the extra complexity here, broke the cycle between whether Domains create Origins or vice-versa.

If you want to rm the test I'll merge this PR.

pushcx:
The failing build is unrelated; brakeman is configured to fail when there's a new version available. I've already updated master so it's safe to merge.

Thanks again for figuring this out.

PR #1415, opened 2025-01-04
Configures ActiveJob with solid_queue gem

rahul1990gupta:
## Overview 
This PR closely follows the recommendation for setting up active job from the rails official video: https://www.youtube.com/watch?v=ReyKfb12EVU

This PR 
- Install foreman for running the web server as well as job server. 
- Installs solid_queue for sqlite3-based active job backend. 
- Installs mission_control for monitoring jobs in production. 
- Adds admin controller to access the jobs UI via basic authentication. 
- Remove `TimeAgoInWords` from ApplicationHelper because it is already provided by rails and it was having conflict with mission_controls method calling rails's `time_ago_in_words`  which has different signature than the one defined here.  https://github.com/rails/mission_control-jobs/blob/ca09138a06d0322a4a39a63ebf253ba0eb56bd7c/app/helpers/mission_control/jobs/dates_helper.rb#L17
- Adds a Daily Maintenance Job as a first example of an Active Job.

##  Testing 
```
3.3.1 :001 > DailyMaintenanceJob.perform_later
  TRANSACTION (0.0ms)  BEGIN immediate TRANSACTION
  SolidQueue::Job Create (0.4ms)  INSERT INTO "solid_queue_jobs" ("queue_name", "class_name", "arguments", "priority", "active_job_id", "scheduled_at", "finished_at", "concurrency_key", "created_at", "updated_at") VALUES ('default', 'DailyMaintenanceJob', '{"job_class":"DailyMaintenanceJob","job_id":"92cc254c-a881-454c-a8e8-0f99094c242d","provider_job_id":null,"queue_name":"default","priority":null,"arguments":[],"executions":0,"exception_executions":{},"locale":"en","timezone":"Central Time (US \u0026 Canada)","enqueued_at":"2025-01-04T09:45:16.261443263Z","scheduled_at":"2025-01-04T09:45:16.260483255Z"}', 0, '92cc254c-a881-454c-a8e8-0f99094c242d', '2025-01-04 09:45:16.260483', NULL, NULL, '2025-01-04 09:45:16.272599', '2025-01-04 09:45:16.272599') RETURNING "id"
  TRANSACTION (0.8ms)  SAVEPOINT active_record_1
  SolidQueue::Job Load (0.1ms)  SELECT "solid_queue_jobs".* FROM "solid_queue_jobs" WHERE "solid_queue_jobs"."id" = 2 LIMIT 1
  SolidQueue::ReadyExecution Create (0.1ms)  INSERT INTO "solid_queue_ready_executions" ("job_id", "queue_name", "priority", "created_at") VALUES (2, 'default', 0, '2025-01-04 09:45:16.295201') RETURNING "id"
  TRANSACTION (0.0ms)  RELEASE SAVEPOINT active_record_1
  TRANSACTION (0.1ms)  COMMIT TRANSACTION
Enqueued DailyMaintenanceJob (Job ID: 92cc254c-a881-454c-a8e8-0f99094c242d) to SolidQueue(default)
↳ (Lobsters):1:in `<top (required)>'
 => 
#<DailyMaintenanceJob:0x0000708e9dce4f90
 @_halted_callback_hook_called=nil,
 @arguments=[],
 @exception_executions={},
 @executions=0,
 @job_id="92cc254c-a881-454c-a8e8-0f99094c242d",
 @priority=nil,
 @provider_job_id=2,
 @queue_name="default",
 @scheduled_at=2025-01-04 03:45:16.260483255 CST -06:00,
 @successfully_enqueued=true,
 @timezone="Central Time (US & Canada)"> 
```
image

## Deployment playbook 
`rails db:prepare` will generate the required database and corresponding tables.

## Next Steps.
This will lay the foundation for https://github.com/lobsters/lobsters/issues/1299 and other async jobs.


app/jobs/daily_maintenance_job.rb
```diff
@@ -0,0 +1,8 @@
+class DailyMaintenanceJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    ReadRibbon.expire_old_ribbons!
+    CommentStat.daily_fill!
```
pushcx, line 6: Please split this into two named jobs. The overhead of setting up a cron job led to putting these in the same script, but they'll be better off independent of each other.

config/routes.rb
```diff
@@ -263,6 +263,9 @@
     resources :reparents, only: [:new, :create]
   end
 
+  # only admin should be able to access it.. right ?
```
pushcx, line 266: Let's allow any moderator to use it. It'd be good to give them insight into the kinds of partial outages that people might ask about when jobs fail.

pushcx:
I know you haven't requested a review yet, but I am excited to see you starting this. I appreciate you taking on all these features related to Rails 8!

The Rails implementation of `time_ago_in_words` doesn't produce the same output as our helper and I don't want to change that very familiar UI, so please rename our helper to `how_long_ago`.

config/routes.rb
rahul1990gupta: Right now, I have configured `basic_http_authentication` with credentials stored in Rails credentials. Integrating with our native authentication workflow may require some more work.

pushcx: I found [a tutorial](https://kyrylo.org/rails/2024/08/07/how-to-add-mission-control-to-your-rails-app-and-protect-it-with-basic-auth.html) for basic auth. The interesting thing was setting `config.mission_control.jobs.base_controller_class` - if you set that to our `Mod::ModeratorController` (from `app/controllers/mod/moderator_controller.rb`, be careful, there are several controllers with very similar names) would it use our authentication?

rahul1990gupta:
That would be very convenient for the moderators. But, it will require more
work. For example, modcontroller inherits from applicationcontroller which
has n+1 query check middleware, that will cause problem with
mission-control gem.

My guess is if we make a new jobmodcontroller class inheriting from
actioncontroller::Base, maybe we can make it work.

config/routes.rb
rahul1990gupta: That would be very convenient for the moderators. But, it will require more
work. For example, ModController inherits from ApplicationController which
has n+1 query check middleware, that will cause problem with
mission-control gem.

My guess is if we make a new JobModcontroller class inheriting from
Actioncontroller::Base, maybe we can make it work.

rahul1990gupta: Something like this is working. 
https://github.com/rahul1990gupta/lobsters/pull/2

rahul1990gupta: another option is to extract the authentication module into a concern and use it for `ApplicationController` and `JobsModController`
https://github.com/rahul1990gupta/lobsters/pull/3

pushcx: A Concern sounds like the exact right tool for the job, please extract one to share authentication between the mission control and mod controllers. Thanks for figuring out our options here.

app/jobs/expire_old_ribbons_job.rb
```diff
@@ -0,0 +1,7 @@
+class ExpireOldRibbonsJob < ApplicationJob
+  queue_as :default
```
caius, line 2: We should probably use queues based on SLOs, https://alexis.bernard.io/blog/2023-10-15-background-job-queues-and-priorities-may-be-the-wrong-path.html is a half decent summary of doing so (and links off to Ben Sheldon of GoodJob maintainer, and Nate Berkopec of Speedshop … notoriety?)

pushcx: I've seen this in production at scale and agree.

rahul1990gupta:
What do you think about using rails's `time_ago_in_words` function here ? 
This is what it [outputs](https://github.com/rails/rails/blob/7acfb75a6074b608c0ac629dfbae0560b3c3321e/actionview/lib/action_view/helpers/date_helper.rb#L57-L68) 

```
      #   distance_of_time_in_words(from_time, from_time + 50.minutes)                                # => about 1 hour
      #   distance_of_time_in_words(from_time, 50.minutes.from_now)                                   # => about 1 hour
      #   distance_of_time_in_words(from_time, from_time + 15.seconds)                                # => less than a minute
      #   distance_of_time_in_words(from_time, from_time + 15.seconds, include_seconds: true)         # => less than 20 seconds
      #   distance_of_time_in_words(from_time, 3.years.from_now)                                      # => about 3 years
      #   distance_of_time_in_words(from_time, from_time + 60.hours)                                  # => 3 days
      #   distance_of_time_in_words(from_time, from_time + 45.seconds, include_seconds: true)         # => less than a minute
      #   distance_of_time_in_words(from_time, from_time - 45.seconds, include_seconds: true)         # => less than a minute
      #   distance_of_time_in_words(from_time, 76.seconds.from_now)                                   # => 1 minute
      #   distance_of_time_in_words(from_time, from_time + 1.year + 3.days)                           # => about 1 year
      #   distance_of_time_in_words(from_time, from_time + 3.years + 6.months)                        # => over 3 years
      #   distance_of_time_in_words(from_time, from_time + 4.years + 9.days + 30.minutes + 5.seconds) # => about 4 years
```
I know it looks more verbose than our implementation of  "how_long_ago",  but it can be customized using engligh locale file. here is an example from official [reference](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L63-L101)

```
  datetime:
    distance_in_words:
      about_x_hours:
        one: about %{count} hour
        other: about %{count} hours
      about_x_months:
        one: about %{count} month
        other: about %{count} months
      about_x_years:
        one: about %{count} year
        other: about %{count} years
      almost_x_years:
        one: almost %{count} year
        other: almost %{count} years
      half_a_minute: half a minute
      less_than_x_seconds:
        one: less than %{count} second
        other: less than %{count} seconds
```

Another reason for using rails's method is their support of localization. and that is important for us because lobsters codebase is used to power a live [french](https://www.journalduhacker.net/) hacking website and [many more](https://github.com/lobsters/lobsters/wiki#sister-sites). And staying close to rails's feature will take it other mods to launch their local version.


```
      def distance_of_time_in_words(from_time, to_time = 0, options = {})
        options = {
          scope: :'datetime.distance_in_words'
        }.merge!(options)

        from_time = normalize_distance_of_time_argument_to_time(from_time)
        to_time = normalize_distance_of_time_argument_to_time(to_time)
        from_time, to_time = to_time, from_time if from_time > to_time
        distance_in_minutes = ((to_time - from_time) / 60.0).round
        distance_in_seconds = (to_time - from_time).round

        I18n.with_options locale: options[:locale], scope: options[:scope] do |locale|
          case distance_in_minutes
          when 0..1
            return distance_in_minutes == 0 ?
                   locale.t(:less_than_x_minutes, count: 1) :
                   locale.t(:x_minutes, count: distance_in_minutes) unless options[:include_seconds]

````

rahul1990gupta:
This PR is getting big and reaching out of its initial intended scope. So, I have taken out the two unrelated components 
- `time_ago_in_words` upgrade/refactor https://github.com/lobsters/lobsters/pull/1423
- `authenticator` concern https://github.com/lobsters/lobsters/pull/1424

pushcx (changes requested):
I took another read through this PR, I think it's almost ready to merge. There's finishing the rename to `how_long_ago` and then only very small changes.

pushcx:
Great, thank you. This PR is also ready to merge, and I'm also going to hold it for the weekend so that I can deploy while the site is less busy.

rahul1990gupta:
Tested it again after resolving the recent merge conflict. It is good to go. I will be checking my email this weekend regularly(morning/evening), in case deployment fails. 

Let me know if anything else is needed.

pushcx:
Ahah, what funny bad luck in alphabetization that this conflicted with two different PRs. I think I fixed it.

PR #1417, opened 2025-01-06
Enables deployment through kamal

rahul1990gupta:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+# Use the official Ruby image as a base image
+FROM ruby:3.3
+
+# Install dependencies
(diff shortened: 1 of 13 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Use the official Ruby image as a base image
+FROM ruby:3.3
```
pushcx, line 2: Can this import from our `.ruby-version` somehow? The file is brand new and they're already out of sync. :)

rahul1990gupta:
Creates dev setup with docker in this PR here. https://github.com/rahul1990gupta/lobsters/pull/4
This is to be opened after [this](https://github.com/lobsters/lobsters/pull/1410) gets  merged.

PR #1418, opened 2025-01-06
Upgrade to Ruby 3.4.1

### LINKED ISSUE #1409: Merry Christmas, let’s update to Ruby 3.4
https://lobste.rs/s/leizx1/ruby_3_4_0_released

This is as simple as bumping `.ruby-version`, running the build, and handling any issues that come up.

caius:
## What?

- [x] Upgrade to Ruby 3.4.1

## Why?

Fixes #1409, upgrades to latest stable ruby. https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/ is the full set of changes. (There was an issue with the 3.4.0 release version tagging, so we got 3.4.1 pretty quickly. No code changes between the two.)

<https://blog.sinjakli.co.uk/2025/01/01/ruby-3-4-highlights/> lists some of the highlights it includes.


caius:
Pushed a new commit so it picks up the changes in the earlier commits. GitHub really doesn't like stacked branches 😭

caius:
@pushcx this is ready to review I think. Do we have a staging site, or just do the best we can with local/CI and then 🤞🏻 for prod?

pushcx:
There isn't a staging setup, but I for riskier deploys I first run them in my linux dev env in production mode and then deploy them on weekends when traffic is low.

This feels like a low risk commit, so I'm going to push it tomorrow morning with the UK OSA stuff.

PR #1421, opened 2025-01-07
Update standardrb for Ruby 3.4

caius:
## What?

- [x] Updated all standard* gems
- [x] Stop requiring standard* gems in the app
- [x] Fix "issue" to reference `/dev/null` with ruby constant

## Why?

Updates all the standard gems (and dependencies) we have included so they know how to handle Ruby 3.4 and suggest 3.4+ "fixes" we should adhere to.

We also don't need to issue `require "standard"`, etc when the app boots - we don't make use of the gems in the app itself. Setting them to `require: false` in the Gemfile achieves this. They are all still available to call as cli tools, so `bundle exec standardrb` works identically.


pushcx:
Thanks for spotting this. Unfortunately, the GitHub PR workflow is undergoing some changes so I think you'll have to click 'ready for review' for the workflow to run.

caius:
Rebased after #1418 landed too. All seems good thankfully 😁

PR #1422, opened 2025-01-07
Enable YJIT in the test suite

caius:
## What?

- [x] Enable YJIT in the test suite via Envariable

## Why?

We might see a speedup in tests if hot code is JIT'd. Costs us very little to figure that out by opening a PR and letting the specs run.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,6 +23,7 @@ jobs:
     env:
       DATABASE_URL: trilogy://root:root@127.0.0.1/lobsters_dev
       RAILS_ENV: test
+      RUBY_YJIT_ENABLE: "1"
 
     steps:
       - uses: actions/checkout@v3


caius:
@pushcx not sure if we have YJIT enabled in production or not already? Might be an option, there's a tradeoff of memory vs speed with it though so it's not a free win. 🤷🏻

caius:
Not much difference, might be worth exploring after #1418 lands for Ruby 3.4. There were changes to the YJIT in that version.

pushcx:
Heh, it's certainly on in prod. Lobsters was [a benchmark for YJIT](https://lobste.rs/s/m9ivcf/we_turned_lobste_rs_into_rails_benchmark). When we deployed to prod we had a whole long issue with a [memory leak](https://github.com/Shopify/ruby/issues/556) that byroot eventually [diagnosed as memory fragmentation](https://push.cx/stream/2024-08-26-heap-dump-pair).

Happy to turn this on in test. Thanks!

PR #1423, opened 2025-01-07
Replaces custom implementation of time_ago_in_words with rails

rahul1990gupta:
## Overview 
lobsters use a custom implementation of `time_ago_in_words` which allows more control over the output of the method. However, we could achieve the same thing configuring the localization file as done in the PR. 

This has two advantages 
- Less custom code to maintain 
- Better support for lobsters [sister sites](https://github.com/lobsters/lobsters/wiki#sister-sites)

See this PR for more context https://github.com/lobsters/lobsters/pull/1415

## Testing 
```
3.3.1 :003 > include ActionView::Helpers::DateHelper
 => Object 
3.3.1 :004 > time_ago_in_words(1.seconds.ago)
 => "less than a minute ago" 
3.3.1 :005 > time_ago_in_words(1.seconds.ago, include_seconds:true)
 => "just now" 
3.3.1 :006 > 
3.3.1 :007 > time_ago_in_words(50.seconds.ago, include_seconds:true)
 => "less than a minute ago" 
3.3.1 :008 > time_ago_in_words(30.seconds.ago, include_seconds:true)
 => "less than a minute ago" 
3.3.1 :009 > time_ago_in_words(1.minutes.ago, include_seconds:true)
 => "1 minute ago" 
3.3.1 :010 > time_ago_in_words(10.minutes.ago, include_seconds:true)
 => "10 minutes ago" 
3.3.1 :011 > time_ago_in_words(1.hour.ago, include_seconds:true)
 => "1 hour ago" 
3.3.1 :012 > 
```


app/models/mod_note.rb
```diff
@@ -1,7 +1,7 @@
 # typed: false
 
 class ModNote < ApplicationRecord
-  extend TimeAgoInWords
+  extend ActionView::Helpers::DateHelper
```
rahul1990gupta, line 4: It fields weird to extend dateHelper to ModNote. But had to do this to make the tests pass. This is because most of the methods are class methods. could they be made regular methods ?

pushcx:
I don't want to make this change. We don't directly use the I18n framework and I'd rather not start. Also, I think it's a lot simpler to have our own helper that produces the exact strings we want rather than use a Rails helper that produces messages we don't want and then indirectly rewrite them to what we want. It's hard to read a template and understand what's going to appear on the page when the documented output is rewritten without explicitly calling another function.

PR #1424, opened 2025-01-07
Extract authentication concerns

rahul1990gupta:
## Overview 
This PR
- Extracts authentication logic  into a separate concern.
-  Change `brakeman.ignore` to make brakeman tests pass

### DIFF
--- a/app/views/stories/_form_errors.html.erb
+++ b/app/views/stories/_form_errors.html.erb
@@ -37,7 +37,7 @@
             <li>
               Comment <%= link_to link.from_comment.short_id, link.from_comment.url %>
               <%= time_ago_in_words link.from_comment.updated_at %>:
-              <%= raw excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
+              <%= sanitize excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
             </li>
           <% end %>
         </ul>
--- a/app/views/stories/_form_errors.html.erb
+++ b/app/views/stories/_form_errors.html.erb
@@ -37,7 +37,7 @@
             <li>
               Comment <%= link_to link.from_comment.short_id, link.from_comment.url %>
               <%= time_ago_in_words link.from_comment.updated_at %>:
-              <%= sanitize excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
+              <%= raw excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
             </li>
           <% end %>
         </ul>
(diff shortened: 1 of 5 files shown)


app/views/stories/_form_errors.html.erb
```diff
@@ -37,7 +37,7 @@
             <li>
               Comment <%= link_to link.from_comment.short_id, link.from_comment.url %>
               <%= time_ago_in_words link.from_comment.updated_at %>:
-              <%= raw excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
+              <%= sanitize excerpt_fragment_around_link(link.from_comment.markeddown_comment, story.url) %>
```
pushcx, line 40: I think this breaks the link. Maybe a better improvement is for this `excerpt_fragment_around_link` helper to call `.html_safe`.

app/views/stories/show.html.erb
```diff
@@ -11,7 +11,7 @@
 <div class="story_content">
   <% if @story.markeddown_description.present? %>
     <div class="story_text">
-      <%= raw @story.markeddown_description %>
+      <%= sanitize @story.markeddown_description %>
```
pushcx, line 14: The `markeddown_` fields are safe and contain raw html.

pushcx:
@rahul1990gupta I didn't click to create a proper review, but it sounds like brakeman gave you some spurious warnings. I know it can be pretty intimidating to work with, but you didn't need to make any changes for these.

If you don't want to fight with it, just revert the html changes I noted and I can take over this PR.

rahul1990gupta:
reverted the changes made to view html erb files. As expected, it made brakeman unhappy.

pushcx:
I've checked out your branch and run `brakeman -q` to reproduce. Those warnings are not a problem for your PR.

We have these couple parts of the code that look risky to brakeman. In `config/brakeman.ignore` you can see two entries for the fingerprints `6042237b560b898fc4f87d6f95bb5cfc5c3c5e09f1a264a9afdd5ab8f087d2a9` and `a7a1ca7467523961f138efb54cd55ccab09710504b57cdfffb7d6294c5d771b9`  and see that they match the warnings you are seeing now, and that those fingerprints are listed under `Obsolete Ignore Entries`. `brakeman` follows values through all the function calls and reports where they might be printed for cross-site scripting (XSS) attacks in the view. Because you moved some of these functions brakeman doesn't recognize that we "ignored" them to mark them as safe, and it presents the warning again.

To resolve the issue, run `brakeman -I` and follow its prompts to ignore the two warnings with a note, and you can reuse the notes from `config/brakeman.ignore` for these two warnings.

pushcx:
Thanks for finishing this refactoring!

PR #1428, opened 2025-01-09
fix /stats styles to support dark mode

### LINKED ISSUE #1420: Update /stats so graphs support dark mode
https://lobste.rs/stats

The graphs assume light mode. It'd be great to not have a big flash of white.

pitr:
fixes #1420 by replacing hardcoded values with `var()` and adding a few more overrides like axis label colours

**Light mode** (no change)
Screenshot 2025-01-09 at 21 06 20

**Dark mode** (actually dark)
Screenshot 2025-01-09 at 21 06 01

### DIFF
--- a/app/views/stats/index.html.erb
+++ b/app/views/stats/index.html.erb
@@ -30,12 +30,17 @@ Want more info?
 </p>
 
 <style type="text/css">
-.graphBackground { fill: #ffffff; }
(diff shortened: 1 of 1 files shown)


pushcx:
Great, this is a very easy merge, especially because you took the time to include screenshots. Very clear reuse of our existing colors, thanks for figuring out how to do that.

Issue #1349, opened 2024-10-16
Split story forms: submit, resubmit, suggest, edit, mod edit


pushcx:
Bug #1334 noted that resubmission UI was appearing when suggesting changes to a story, and the fix in #1348 went through a few attempts to get a fix. There have been a few other bugs recently (#1338, #1321) and new functionality (bc2452c4). At this point the `stories/_form.html.erb` and `stories/_form_errors.html.erb` templates are littered with booleans for the many contexts those elements are reused in, and StoriesController is both wide (lots of actions) and deep (actions have code that would be better in the Story model). Having specialized forms with specific UI is worth the duplication in html.

Contributors: this is a fair amount of code to unravel, so consider doing one PR per form.

 - [ ] Move suggested edits to a new `SuggestionsController`
 - [ ] Move mod editing to its own controller and form, which will show suggestions. Add a table of `Moderation` records that have edited this Story or are about this user. Copy the table of `ModNote` records about this user from `users/show.html.erb`. In `_list_detail`, add a second link `mod` to link to this - so a moderator would see the same `edit` link as users on their own submission, and a `mod` link on all stories.
 - [ ] Split resubmit out into its own controller and form, so `StoriesController#create` would recognize that it’s a resubmit and send the user to `ResubmitStoriesController#new`(with all their data!), which would then have an explanation (not shoved into flash), the list of previous submissions, and the comment form.
 - [ ] Staying in `StoriesController`, split story editing out into its own form. Show the user title and tag suggestions, as well as how much time is left in `MAX_EDIT_MINS` (don’t make this a live countdown, just a static value when the page is loaded). Remove the `_form_errors` partial by moving those into the `new`/`edit` forms. Move the `similar_stories` display out of flash (#1338) and only into the `new` form.

PSKK-Git:
@pushcx please assign this to me

pushcx:
We don't assign issues, you can just pick part of it to work on.

[...]

PR #1414, opened 2025-01-04
refactors suggestion controller out of story controller

rahul1990gupta:
## Overview 
The goal is to make story controller light. This tackles part 1. 

## Testing 

https://github.com/user-attachments/assets/37631b41-1cf8-4bdd-80aa-a5e22fb99a04

```
3.3.1 :002 > SuggestedTagging.last 
  SuggestedTagging Load (0.6ms)  SELECT `suggested_taggings`.* FROM `suggested_taggings` ORDER BY `suggested_taggings`.`id` DESC LIMIT 1
 => #<SuggestedTagging:0x00007445723c7090 id: 4, story_id: 77, tag_id: 59, user_id: 55> 
3.3.1 :003 > SuggestedTitle.last 
  SuggestedTitle Load (0.3ms)  SELECT `suggested_titles`.* FROM `suggested_titles` ORDER BY `suggested_titles`.`id` DESC LIMIT 1
 => #<SuggestedTitle:0x00007445718b7b50 id: 3, story_id: 77, user_id: 55, title: "Ex veniam vitae 2"> 
```

Issue #1259, opened 2024-02-29
Consistently highlight moderator/admin UI 


pushcx:
All mod UI needs to have some kind of unambiguously distinct and consistent visual styling. Maybe a 45 degree red/white striped background or underline? It’d be good to have a second, similar style for admin-only actions.

The moderator/admin UI is all inline to the Lobsters app, see the many `.is_moderator?` and `.is_admin?` calls in `app/views` and note that many call model methods that also use those methods. There’s some ambiguous UI like the edit link in `stories/_listdetail.html.erb` or flag summaries that can be present because the mod submitted the story or because they are a mod.

pushcx:
I want to highlight the [specific comment](https://github.com/lobsters/lobsters/pull/1431#discussion_r1910607589) where I mentioned this issue because in my head it has become a bigger project because of small confusions over the last year:

> I am slowly working towards making it so mods always see the same UI as users, then also additional, clearly marked (#1259) mod UI. There has been minor but avoidable confusion about things because mods can't remember exactly what users see for scores, vote summaries, etc.

pushcx:
In office hours @chamlis joked about putting mod ui into comic sans and that's no-kidding a pretty good idea. Easy to implement, doesn't change layout the way a border would, and calls back to an [existing easter egg](https://github.com/lobsters/lobsters/blob/dd14f3ac16ba97ed88d5c6f234a56c49935284d4/app/assets/stylesheets/application.css#L140).

PR #1431, opened 2025-01-10
Stories controller split for mod

kernal053:
#1349 2nd Task : Spliting mod editing into it's own ModerateStoriesController.

### DIFF
--- a/app/controllers/moderate_stories_controller.rb
+++ b/app/controllers/moderate_stories_controller.rb
@@ -1,0 +1,43 @@
+class ModerateStoriesController < ApplicationController
+  before_action :require_logged_in_moderator
+  before_action :find_user_story!
+  before_action :show_title_h1
+
+  def edit
+    @title = "Edit Story"
+
+    if @story.merged_into_story
+      @story.merge_story_short_id = @story.merged_into_story.short_id
+      User.update_counters @story.user_id, karma: (@story.votes.count * -2)
+    end
+  end
+
+  def update
+    @story.is_deleted = false
+    @story.editor = @user
+    @story.attributes = story_params
+
+    if @story.save
+      redirect_to @story.comments_path
+    else
+      render action: "edit"
+    end
+  end
+
+  private
+
+  def story_params
+    params.require(:story).permit(
+      :title, :url, :description, :moderation_reason,
+      :merge_story_short_id, :is_unavailable, :user_is_author, :user_is_following,
+      tags_a: []
+    )
+  end
+
+  def find_user_story!
+    @story = Story.where(short_id: params[:story_id] || params[:id]).first
+    if !@story
+      raise ActiveRecord::RecordNotFound
+    end
+  end
+end
--- a/app/views/moderate_stories/edit.html.erb
+++ b/app/views/moderate_stories/edit.html.erb
@@ -1,0 +1,70 @@
+<div class="box" id="story_box">
+	<div id="story_holder">
+    <%= form_with model: @story do |f| %>
+      <%= render :partial => "stories/form", :locals => { :story => @story,
+        :f => f } %>
+
+      <br />
+      <div class="box">
+        <div class="boxline">
+          <%= f.label :merge_story_short_id, "Merge Into:",
+            :class => "required" %>
+          <%= f.text_field :merge_story_short_id, :autocomplete => "off",
+            :placeholder => "Short id of story into which this story " <<
+            "be merged" %>
+        </div>
+        <div class="boxline">
+          <%= f.label :unavailable_at, "Unavailable:",
+            :class => "required" %>
+          <%= f.check_box :is_unavailable %>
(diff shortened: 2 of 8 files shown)


app/controllers/moderate_stories_controller.rb
```diff
@@ -0,0 +1,43 @@
+class ModerateStoriesController < ApplicationController
+  before_action :require_logged_in_moderator
+  before_action :find_user_story!
+  before_action :show_title_h1
+
+  def edit
+    @title = "Edit Story"
+
+    if @story.merged_into_story
```
pushcx, line 9: This code about story merging can be removed from the `StoriesController#edit` action, only mods are allowed to edit this field.

app/views/moderate_stories/edit.html.erb
```diff
@@ -0,0 +1,70 @@
+<div class="box" id="story_box">
+	<div id="story_holder">
```
pushcx, line 2:
```suggestion
  <div id="story_holder">
```

app/views/stories/_listdetail.html.erb
```diff
@@ -129,8 +129,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
 
         <% if story.is_editable_by_user?(@user) %>
           <span> | </span>
-          <a href="<%= edit_story_path(story.short_id) %>" class="<%=
-            story.has_suggestions? ? "story_has_suggestions" : "" %>">edit</a>
+          <a href="<%= @user.is_moderator? ? edit_moderate_story_path(story.short_id) : edit_story_path(story.short_id) %>"
```
pushcx, line 132: So I know this must sound odd, but I would like mods to see two edit links. So this should always be the `edit_story_path` link. Below, you added a `if @user.try(:is_moderator?)` to link to the mod notes for the user (handy!). Please add a second `edit` link leading to the new moderator story editing controller.

I am slowly working towards making it so mods always see the same UI as users, then also additional, clearly marked (#1259) mod UI. There has been minor but avoidable confusion about things because mods can't remember exactly what users see for scores, vote summaries, etc.

app/controllers/moderate_stories_controller.rb
```diff
@@ -0,0 +1,43 @@
+class ModerateStoriesController < ApplicationController
```
pushcx, line 1:
```suggestion
class Mod::StoriesController < Mod::ModeratorController
```

We recently added an `app/controllers/mod/` directory. Please move it in there and have it inherit from `Mod::ModeratorController` (you can then remove the explicit call to `require_logged_in_moderator`, the parent controller handles it).

Sorry for the chore of renaming the file, test, and route.

app/controllers/moderate_stories_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def story_params
+    params.require(:story).permit(
+      :title, :url, :description, :moderation_reason,
+      :merge_story_short_id, :is_unavailable, :user_is_author, :user_is_following,
+      tags_a: []
+    )
+  end
+
+  def find_user_story!
```
pushcx, line 37: In #1414, @rahul1990gupta extracted a `StoryFinder` concern, please use that.

spec/requests/moderate_stories_spec.rb
```diff
@@ -0,0 +1,7 @@
+require "rails_helper"
+
+RSpec.describe "ModerateStories", type: :request do
```
pushcx, line 3: Because moderator functionality is so much less used, this can be one test. Load the form for a story, change the title, submit the form, then assert that it was redirected to the story and that the new title appears. We'll only add more tests if things become unreliable, but this is a clear refactor of very-well-used code so I don't worry that it needs a lot of tests.

pushcx (changes requested):
Thanks for taking on this refactor, you're off to a very good start.

app/controllers/concerns/story_finder.rb
```diff
@@ -2,7 +2,7 @@ module StoryFinder
   extend ActiveSupport::Concern
 
   def find_story
-    story = Story.find_by(short_id: params[:story_id])
+    story = Story.find_by(short_id: params[:story_id] || params[:id])
```
pushcx, line 5: Good way to handle this code reuse.

pushcx (changes requested):
Great, thanks for integrating all those changes. Just one more small change.

In `app/controllers/stories_controller.rb`, `story_params` near the bottom, it no longer needs to check if the user is a moderator and allow them extra fields that you have removed from the edit form.

It's really nice how much simpler you've made this code. There's much less mental state to juggle now that it's clear that one form gets users and one form gets mods.

pushcx:
Thanks for advancing this big refactor. It’s great to get to check off another big piece of functionality! I especially appreciate you working through the chore of renaming because I forgot to update that description.

PR #1432, opened 2025-01-12
Fix 500 in moderations pagination

### LINKED ISSUE #1425: Vulnerability scanner prompted a routing 500
A vulnerability scanner shoved garbage at the site and got a 500 (and an IP ban). It looks like this is a bug in our pagination code. I should mention I’m rewriting pagination in #394 (delayed by an [urgent problem](https://lobste.rs/s/ukosa1)) so a quick fix is appropriate here.

```
An ActionView::Template::Error occurred in moderations#index:

No route matches {:action=>"u99p5", :begindate=>"pc8g8", :c=>"ft9i6", :categoryid=>"uluq7", :code=>"c7t07", :controller=>"moderations", :csrf_token=>"piei1",
:enddate=>"hpd91", :immagine=>"tpnc2", :item=>"kqyg6", :key=>"ar3i4", :keyword=>"r2tr0", :list_type=>"gcoj3", :page=>2, :q=>"s2lm0", :region=>"qbnv3",
:terms=>"z8yw2", :type=>"b1wa1", :url=>"fzg24", :view=>"kxmp2", :what=>{"users"=>"exvx9"}}
  app/views/moderations/index.html.erb:24
  app/controllers/application_controller.rb:204:in `n_plus_one_detection'
```

So the url was (and I’m changing this to localhost to minimize prod 500s about it…): 

`http://localhost:3000/moderations?action=u99p5&begindate=pc8g8&c=ft9i6&categoryid=uluq7&code=c7t07&csrf_token=piei1&enddate=hpd91&immagine=tpnc2&item=kqyg6&key=ar3i4&ke
yword=r2tr0&list_type=gcoj3&page=f9n16&q=s2lm0&region=qbnv3&terms=z8yw2&type=b1wa1&url=fzg24&view=kxmp2&what%5Busers%5D=exvx9`

caius:
## What?

- [x] Add spec to reproduce (`params[:action]` being passed in)
- [x] Lean on `ActionController::Parameters` in the controller to whitelist what we accept/forward in links
- [x] Rewrite a `where()` to stop Brakeman whinging about it

## Why?

Fixes #1425 

The params in the moderations index are used to generate the pagination links so we keep the same filters on the next/previous page. Currently on master we're forwarding *all* `params` whether they're used by the page or not. This breaks if `:action` (or `:controller`) are injected as parameters, due to Rails internals. As a rule of thumb we shouldn't be generating links containing injected content from users/previous URLs, but generating a link only with the whitelisted params we expect to be in the URL.

Brakeman was upset with our construction of SQL conditional from a string, even though we controlled the input from hash keys. Rewrite it slightly to generate the same SQL (minus some superfluous brackets) but using the power of the DSL to escape anything untoward if it somehow made it in there. Removes a brakeman ignore.

(This doesn't currently appear to be a security issue, but it would've been a useful place for injecting things into links on the page if required to pair with another attack later.)

### DIFF
--- a/app/controllers/moderations_controller.rb
+++ b/app/controllers/moderations_controller.rb
@@ -9,15 +9,15 @@ class ModerationsController < ApplicationController
     @title = "Moderation Log"
     @moderators = ["(All)", "(Users)"] + User.moderators.map(&:username)
 
-    @moderator = params.fetch("moderator", "(All)")
+    @moderator = moderation_params.fetch("moderator", "(All)")
     @what = {
-      stories: params.dig(:what, :stories),
-      comments: params.dig(:what, :comments),
-      tags: params.dig(:what, :tags),
-      users: params.dig(:what, :users),
-      domains: params.dig(:what, :domains),
-      origins: params.dig(:what, :origins),
-      categories: params.dig(:what, :categories)
+      stories: moderation_params.dig(:what, :stories),
+      comments: moderation_params.dig(:what, :comments),
+      tags: moderation_params.dig(:what, :tags),
+      users: moderation_params.dig(:what, :users),
+      domains: moderation_params.dig(:what, :domains),
+      origins: moderation_params.dig(:what, :origins),
+      categories: moderation_params.dig(:what, :categories)
     }
     @what.transform_values! { true } if @what.values.none?
 
@@ -48,7 +48,7 @@ class ModerationsController < ApplicationController
 
     # paginate
     @pages = (@moderations.count / ENTRIES_PER_PAGE).ceil
-    @page = params[:page].to_i
+    @page = moderation_params[:page].to_i
     if @page == 0
       @page = 1
     elsif @page < 0 || @page > (2**32) || @page > @pages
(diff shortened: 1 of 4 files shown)


caius:
If you had a further 500 error on this page just about now, it was me. Mixed up live and localhost in testing, apologies

app/controllers/moderations_controller.rb
```diff
@@ -43,12 +43,12 @@ def index
     # filter based on type of thing moderated
     @what.each do |type, checked|
       next if checked
-      @moderations = @moderations.where("`moderations`.`#{type.to_s.singularize}_id` is null")
+      @moderations = @moderations.where("#{type.to_s.singularize}_id": nil)
```
caius, line 46: This is what squishes the brakeman warning

pushcx: Thanks for highlighting this, made it real clear what's happening.

pushcx (approved):
This is a great fix, thanks! Very clear what's happening and I can see the next scanners getting the door shut in their face with 400s.

I was surprised to see the `request.query_parameters.merge`, it slightly raises the hair on the back of my neck to use user input in a template like that, so I'm especially glad to have it removed.

PR #1436, opened 2025-01-17
Split undelete story controller

kernal053:
#1349 3rd task

### DIFF
--- a/app/controllers/mod/stories_controller.rb
+++ b/app/controllers/mod/stories_controller.rb
@@ -37,6 +37,25 @@ class Mod::StoriesController < Mod::ModeratorController
     redirect_to @story.comments_path
   end
 
+  def destroy
+    update_story_attributes
+
+    if @story.user_id != @user.id && @story.moderation_reason.blank?
+      @story.errors.add(:moderation_reason, message: "is required")
+      return render action: "edit"
+    end
+
+    @story.is_deleted = true
+    @story.editor = @user
+
+    if @story.save
+      Keystore.increment_value_for("user:#{@story.user.id}:stories_deleted")
+      Mastodon.delete_post(@story)
+    end
+
+    redirect_to @story.comments_path
+  end
+
   private
 
   def update_story_attributes
--- a/app/controllers/mod/stories_controller.rb
+++ b/app/controllers/mod/stories_controller.rb
@@ -16,7 +16,7 @@ class Mod::StoriesController < Mod::ModeratorController
   def update
     @story.is_deleted = false
     @story.editor = @user
-    @story.attributes = story_params
+    update_story_attributes
 
     if @story.save
       redirect_to @story.comments_path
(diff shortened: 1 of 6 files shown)


app/views/mod/stories/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="box" id="story_box">
   <div id="story_holder">
-    <%= form_with model: @story do |f| %>
+    <%= form_with model: [:mod, @story] do |f| %>
```
kernal053, line 3: Apologies, I had not properly updated after switching to namespaced mod directory for stories.

app/views/mod/stories/edit.html.erb
```diff
@@ -38,17 +38,17 @@
...
           <% elsif !@story.is_gone? %>
             &nbsp; | &nbsp;
             <% if @story.user_id != @user.id %>
-              <%= f.submit "Delete", formaction: story_destroy_path(@story.short_id),
+              <%= f.submit "Delete", formaction: mod_story_destroy_path(@story.short_id),
                 :class => "deletion", :data => { :confirm => "Delete this story?" }  %>
             <% else %>
-              <%= f.submit "Delete", formaction: story_destroy_path(@story.short_id),
+              <%= f.submit "Delete", formaction: mod_story_destroy_path(@story.short_id),
                 :class => "deletion", :data => { :confirm => "Delete this story?" } %>
```
kernal053, line 52: `if` condition seems redundant to me here.

app/views/mod/stories/edit.html.erb
pushcx: No worries, I saw it in prod and fixed it. This fix of yours is better than mine so please keep it when you resolve the merge conflict. I used `url` to target the controller and wasn't aware you could pass a model like that.

app/views/mod/stories/edit.html.erb
pushcx: Yes, like your previous PR for edit links, I'd like for mods to always see two links, one if they are the user and one if they are a moderator. You might be interested to see the commit where I touched up some small things: https://github.com/lobsters/lobsters/commit/9a180400dbe7112efdb047c04488a6a2bea0e2ca

app/controllers/mod/stories_controller.rb
```diff
@@ -16,7 +16,7 @@ def edit
   def update
     @story.is_deleted = false
     @story.editor = @user
-    @story.attributes = story_params
+    update_story_attributes
```
pushcx, line 19: I would rather repeat the standard Rails idiom you removed than wrap up that single line of code in a function.

config/routes.rb
```diff
@@ -264,7 +264,10 @@
 
   namespace :mod do
     resources :reparents, only: [:new, :create]
-    resources :stories, only: [:edit, :update]
+    resources :stories, only: [:edit, :update] do
+      patch "undelete"
+      patch "destroy"
```
pushcx, line 269: This is one of the standard crud routes, you can add `:destroy` to the `only:` list above rather than specify it.

pushcx (changes requested):
Real great to have you continue on from splitting out mod story editing.

Like that PR, please add request tests for the mod path, one for destroy and one for undelete.

config/routes.rb
kernal053: Yes, my first thought also involved using standard crud routes. But I'm unable to override with delete and post request within same forms. Also, we are essentially marking stories as delete rather than removing it from database so I thought going ahead with current patch request seems correct.

pushcx:
Looks great, thanks for working through this. I’ll be very glad to not be doing it through the console any more. :)

PR #1437, opened 2025-01-22
moves "with recursive CTE" to activerecord

rahul1990gupta:
## Overview 
As we prepare to [migrate to postgresql](https://github.com/lobsters/lobsters/issues/539), we will be moving raw sql to active record queries slowly and carefully. This PR moves recursive CTEs to newly introduced `with_recursive` function. 
- [X] Get the right syntax for recursive query
- [X] pushes the `reply_count` and `depth` [into the database](https://github.com/lobsters/lobsters/issues/1438) 
- [ ] Get feature/column parity for recursive CTEs

## Testing 
TBD

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -28,6 +28,8 @@ class CommentsController < ApplicationController
       # includes parent story_id to ensure this comment's story_id matches
       comment.parent_comment =
         Comment.find_by(story_id: story.id, short_id: params[:parent_comment_short_id])
+      comment.depth = comment.parent_comment.depth + 1
+
       if !comment.parent_comment
         return render json: {error: "invalid parent comment", status: 400}
       end
@@ -50,6 +52,11 @@ class CommentsController < ApplicationController
     if comment.valid? && params[:preview].blank? && comment.save
       comment.current_vote = {vote: 1}
       render_created_comment(comment)
+
+      if comment.parent_comment
+        old_reply_count = comment.parent_comment.reply_count
+        comment.parent_comment.update_column(reply_count: old_reply_count + 1)
+      end
     else
       comment.score = 1
       comment.current_vote = {vote: 1}
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -52,11 +52,6 @@ class CommentsController < ApplicationController
     if comment.valid? && params[:preview].blank? && comment.save
       comment.current_vote = {vote: 1}
       render_created_comment(comment)
-
-      if comment.parent_comment
-        old_reply_count = comment.parent_comment.reply_count
-        comment.parent_comment.update_column(reply_count: old_reply_count + 1)
-      end
     else
       comment.score = 1
       comment.current_vote = {vote: 1}
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -131,7 +131,7 @@ class CommentsController < ApplicationController
     if request.xhr?
       render partial: "commentbox", locals: {comment: comment, story: story}
     else
-      parents = comment.parents.with_thread_attributes.for_presentation
+      parents = comment.parents.for_presentation
 
       parent_ids = parents.map(&:id)
       @votes = Vote.comment_votes_by_user_for_comment_ids_hash(@user&.id, parent_ids)
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -10,6 +10,7 @@ class Comment < ApplicationRecord
   belongs_to :parent_comment,
     class_name: "Comment",
     inverse_of: false,
+    counter_cache: :reply_count,
     optional: true
   has_one :moderation,
     class_name: "Moderation",
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -503,31 +503,13 @@ class Comment < ApplicationRecord
   def parents
     return Comment.none if parent_comment_id.nil?
 
-    # starts from parent_comment_id so it works on new records
-    @parents ||= Comment
-      .joins(<<~SQL
-        inner join (
-          with recursive parents as (
-            select
-              id target_id,
-              id,
-              parent_comment_id,
-              0 as depth,
-              (select count(*) from comments where parent_comment_id = id) as reply_count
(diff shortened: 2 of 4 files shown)


app/models/comment.rb
```diff
@@ -647,83 +627,30 @@ def self.recent_threads(user)
...
-        comments_recursive.depth as depth,
-        comments_recursive.reply_count as reply_count
-      ')
+    # Going from top to botton to find all children in the thread
+    Comment.with_recursive(
+      comments_recursive: [
+        Comment.where(thread_id: thread_ids, parent_comment_id: nil),
+        Comment.joins("JOIN comments_recursive ON comments_recursive.id = comments.parent_comment_id")
+      ]
+    ).from("comments_recursive AS comments").order(thread_id: :desc)
```
rahul1990gupta, line 636: It will be nice to simplify the definition of `confidence_order_path` from `confidence_order`.

app/models/comment.rb
```diff
@@ -647,83 +627,30 @@ def self.recent_threads(user)
...
-              parent_comment_id is null
-          union all
-          select
-            c.id,
-            discussion.depth + 1,
-            (select count(*) from comments where parent_comment_id = c.id),
-            cast(concat(
-              left(discussion.confidence_order_path, 3 * (depth + 1)),
-              c.confidence_order
-            ) as char(#{Comment::COP_LENGTH}) character set binary)
```
rahul1990gupta, line 671: It will be nice to simplify the definition of confidence_order_path from confidence_order.

rahul1990gupta:
## Current State 
`reply_count` and `depth` are very critical [attributes](https://github.com/lobsters/lobsters/blob/master/app/models/comment.rb#L28-L29) of the  `Comment` model. Depth is used to render threaded comments with indentation, while reply_count of a comment is aggregated to the story level to show total comment count on the home page. 

Each time someone render the home/story page, the function to traverse the comment tree is called; depth and reply_count for the associated comments are computed. There are two problems with this approach 
- **Inefficient**: This is very inefficient to compute. It is possible to do caching, but expiry time would be in minutes to make sure users get correct data on an active discussion. 
- **Duplicated and complex logic to compute reply_count and depth**: In the comment model, depth and reply_count is computed three tiimes. In [parents](https://github.com/lobsters/lobsters/blob/master/app/models/comment.rb#L507-L530), [Comment.recent_threads](https://github.com/lobsters/lobsters/blob/master/app/models/comment.rb/#L651-L682) and [Comment.story_threads](https://github.com/lobsters/lobsters/blob/master/app/models/comment.rb#L693-L727)

## Proposed state 
- Add `reply_count` and `depth` to the `comments` table with default 0. 
- Update the parent comment's reply_coount when [creating a comment](https://github.com/lobsters/lobsters/blob/master/app/controllers/comments_controller.rb#L22-L25). 
- Update the depth of the comment during creation.

## Benefits
- efficient and optimized code
- clean implementation while rendering 

@pushcx This is my naive understanding of the situation at hand. Maybe I missed some context here. Please suggest what is the best way to proceed here.

app/controllers/comments_controller.rb
```diff
@@ -50,6 +52,11 @@ def create
     if comment.valid? && params[:preview].blank? && comment.save
       comment.current_vote = {vote: 1}
       render_created_comment(comment)
+
+      if comment.parent_comment
+        old_reply_count = comment.parent_comment.reply_count
+        comment.parent_comment.update_column(reply_count: old_reply_count + 1)
```
pushcx, line 58: I'd like to use [counter_cache](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-async_count) for this if possible. If not, I would rather see something like `comment.parent_comment.update_column(reply_count: comment.parent_comment.replies.not.deleted.count))` (I'm guessing at the scope name for non-deleted comments)  to have the database count rather than try to keep `+ 1` and `- 1` updates correct.

app/models/comment.rb
pushcx: Yes! This is a [big evil performance hack](https://push.cx/stream/2024-09-05-you-re-allowed-to-be-evil-in-performance-code) for our biggest hot path feature because MariaDB didn't allow me to use arrays in an RCTE.

app/controllers/comments_controller.rb
```diff
@@ -28,6 +28,8 @@ def create
       # includes parent story_id to ensure this comment's story_id matches
       comment.parent_comment =
         Comment.find_by(story_id: story.id, short_id: params[:parent_comment_short_id])
+      comment.depth = comment.parent_comment.depth + 1
```
pushcx, line 31: I'd also rather call `update_column` (or a similar method) with the `.parents.count` to make the database count. I know this is very simple and correct now, but I do have some feature ideas around being able to move comments between parents.

pushcx (changes requested):
Thanks for taking this on. I'm really impressed by how much shorter these queries get. It is definitely a good idea to materialize `depth` and `reply_count`. There's a Rails feature I think we can use to make this even better, and for consistency I want to do counting in the database rather than round-trip to the app server and risk race conditions or bugs.

app/controllers/comments_controller.rb
pushcx: Discussed on stream. @rahul1990gupta noted that my comment is incorrect for this context - the comment is a new record that is not yet persisted to the database. So we can't call `update_column` or similar.

What we really want to say is that the value of depth is an SQL expression so that the database is always consistent, but ActiveRecord doesn't allow us to express that. A race condition is only theoretical, especially because I haven't actually written a feature that might change a parent's depth at the time someone is posting a comment.

I'll write a checker to watch for validity problems and we can address this when those come up. No change to this code.

pushcx:
This is blocked on #1451 right?

rahul1990gupta:
Our end state is to use `with recursive .... depth first search` which is not supported by `ActiveRecord.with_recursive`, because `depth first search` is not supported by a lot of databases. 

Closing this PR. Will pick up Materializing `reply_count` and `depth` down the line. Now, focusing on `migration to postgres`.

pushcx:
I pulled your materialization work into c72b59f932a684feb8681b9d1576e05fcd5bb216 and finished it.

PR #1441, opened 2025-01-23
Replaces SQL string with activerecord friendly query

rahul1990gupta:
Replaces SQL string with activerecord friendly query 

Reference 
https://github.com/lobsters/lobsters/issues/539

## Testing 
- [ ] Benchmark before and after analysis for the `active` page
- [ ] All big Activerrecord queries have similar SQL before and after.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -37,7 +37,7 @@ class UsersController < ApplicationController
 
     if params[:by].to_s == "karma"
       content = Rails.cache.fetch("users_by_karma_#{newest_user}", expires_in: (60 * 60 * 24)) {
-        @users = User.select(*attrs).order("karma DESC, id ASC").to_a
+        @users = User.select(*attrs).order(karma: desc, id: asc).to_a
         @user_count = @users.length
         @title << " By Karma"
         render_to_string action: "list", layout: nil
@@ -45,17 +45,17 @@ class UsersController < ApplicationController
       render html: content.html_safe, layout: "application"
     elsif params[:moderators]
       @users = User.select(*attrs).where("is_admin = ? OR is_moderator = ?", true, true)
-        .order("id ASC").to_a
+        .order(id: :asc).to_a
       @user_count = @users.length
       @title = "Moderators and Administrators"
       render action: "list"
     else
       # Mod::ReparentsController#create knows this key
       content = Rails.cache.fetch("users_tree_#{newest_user}", expires_in: 12.hours) {
-        users = User.select(*attrs).order("id DESC").to_a
+        users = User.select(*attrs).order(id: :desc).to_a
         @user_count = users.length
         @users_by_parent = users.group_by(&:invited_by_user_id)
-        @newest = User.select(*attrs).order("id DESC").limit(10)
+        @newest = User.select(*attrs).order(id: :desc).limit(10)
         render_to_string action: "tree", layout: nil
       }
       render html: content.html_safe, layout: "application"
@@ -148,7 +148,7 @@ class UsersController < ApplicationController
       .where("
         comments.flags > 0 and
         comments.created_at >= now() - interval #{@interval[:dur]} #{@interval[:intv]}")
-      .order("id DESC")
+      .order(id: :desc)
       .includes(:user, :hat, story: :user)
       .joins(:story)
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -145,12 +145,11 @@ class UsersController < ApplicationController
     @lookup = rows.to_h
 
     @flagged_comments = @showing_user.comments
-      .where("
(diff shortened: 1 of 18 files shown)


app/controllers/users_controller.rb
```diff
@@ -37,25 +37,25 @@ def tree
...
       content = Rails.cache.fetch("users_by_karma_#{newest_user}", expires_in: (60 * 60 * 24)) {
-        @users = User.select(*attrs).order("karma DESC, id ASC").to_a
+        @users = User.select(*attrs).order(karma: desc, id: asc).to_a
         @user_count = @users.length
         @title << " By Karma"
         render_to_string action: "list", layout: nil
       }
       render html: content.html_safe, layout: "application"
     elsif params[:moderators]
       @users = User.select(*attrs).where("is_admin = ? OR is_moderator = ?", true, true)
```
pushcx, line 47: This can use `.or()`

app/controllers/users_controller.rb
```diff
@@ -145,12 +145,11 @@ def standing
     @lookup = rows.to_h
 
     @flagged_comments = @showing_user.comments
-      .where("
-        comments.flags > 0 and
-        comments.created_at >= now() - interval #{@interval[:dur]} #{@interval[:intv]}")
-      .order("id DESC")
-      .includes(:user, :hat, story: :user)
+      .where("comments.flags > ?", 0)
```
pushcx, line 148:
```suggestion
      .where(comments: { flags: 0.. })
```

app/controllers/users_controller.rb
```diff
@@ -145,12 +145,11 @@ def standing
...
+      .where("comments.created_at >= ?", "now() - interval #{@interval[:dur]} #{@interval[:intv]}")
```
pushcx, line 149: It would be nice to move this into AR but I don't know if it's expressive enough for us to do it.

app/models/story_repository.rb
```diff
@@ -37,7 +37,7 @@ def active
           comments.story_id = stories.id and
           comments.created_at >= date_sub(now(), interval 3 day)
       ) as latest_comment_id')
-      .order("latest_comment_id desc")
+      .order(latest_comment_id: :desc)
```
pushcx, line 40: `/active` is slow because of this subquery. Like you noticed with `reply_count` and `depth` in #1437, this should be a materialized column for performance and simplicity. Add a column with an index and use `touch: :last_comment_at` on the association and Rails will maintain the value.

Sorry to add a database refactoring, but it'll be both a performance and code cleanup improvement.

app/models/user.rb
```diff
@@ -428,11 +428,19 @@ def disable_2fa!
   # ensures some users talk to a mod before reactivating
   def good_riddance?
     return if is_banned? # https://www.youtube.com/watch?v=UcZzlPGnKdU
+
+    recent_comments_count = comments
+      .where("created_at >= ?", 30.days.ago)
```
pushcx, line 433:
```suggestion
      .where(created_at: 30.days.ago..)
```

(This might need an extra set of parenthesis around the `30.days.ago`.)

Same edit a couple lines down.

app/models/user.rb
```diff
@@ -631,6 +639,6 @@ def votes_for_others
       .includes(comment: :user, story: :user)
       .where("(votes.comment_id is not null and comments.user_id <> votes.user_id) OR " \
```
pushcx, line 640: `.or()`

pushcx (changes requested):
Thanks for taking on this very unglamorous maintenance work. You've found several places where we do some sloppy things, so I'm glad to have the opportunity to clean them up.

We talked about this on [today's stream](https://push.cx/stream/2025-01-23); the value to using `.or` is that sometimes when composing scopes, using `.joins`, or using the different functions to avoid 1 + N queries, ActiveRecord writes queries that alias tables. Using strings in `where` instead of hashes means it doesn't know a table is being referenced and the developer gets a surprising error.

app/models/user.rb
rahul1990gupta: One way to do this will be 
```
+    votes
+    .left_outer_joins(:story, :comment)
+    .includes(comment: :user, story: :user)
+    .where.not(votes: { comment_id: nil })
+    .where.not("comments.user_id = votes.user_id")
+    .or(
+      votes.where(comment_id: nil)
+           .where("stories.user_id <> votes.user_id")
+    )
+    .order(id: :desc) 
```
But this will change the underlying sql. This will require more wrestling with SQL to get it right.

pushcx: Could you paste both sql queries? I can't guess what the difference is?

pushcx (commented):
I wanted to get this merged before it goes stale or I introduce a bunch of merge conflicts. There's a handful more things we could wrap ActiveRecord around, but I want to make sure this work doesn't get lost.

PR #1442, opened 2025-01-23
Add instrumentation of the GVL

byroot:
In response to https://byroot.github.io/ruby/performance/2025/01/23/the-mythical-io-bound-rails-app.html @pushcx kindly offered to instrument the actual IO ration on lobster production environment.

This PR does exactly that using the `gvl_timing` gem, inside a custom middleware mounted about as high as possible in the stack.

On my local environment it produces the following logs:

```
[GVL] io_percent=9.1, run_ms=30.57, idle_ms=2.95, stall_ms=0.03, gc_ms=0.0, path=/active
```

Perhaps you'd prefer if these were integrated in the main logs? I'm not familiar with your current log infra, but if you point me to it, I can probably hook them off.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -102,6 +102,9 @@ Rails.application.configure do
   # why help timing attacks?
   config.middleware.delete(Rack::Runtime)
 
+  require File.expand_path("../../extras/gvl_instrumentation", __FILE__)
+  config.middleware.insert 0, GvlInstrumentation, File.open(Rails.root.join("/srv/lobste.rs/log/production-gvl.log"), "a+")
+
   # Enable DNS rebinding protection and other `Host` header attacks.
   # config.hosts = [
   #   "example.com",     # Allow requests from example.com
--- a/Gemfile
+++ b/Gemfile
@@ -41,6 +41,7 @@ gem "memory_profiler"
 gem "rack-mini-profiler"
 gem "stackprof"
 gem "prosopite"
+gem "gvl_timing"
 
 gem "builder" # for rss
 gem "oauth" # for linking accounts
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -141,6 +141,7 @@ GEM
     flamegraph (0.9.5)
     globalid (1.2.1)
       activesupport (>= 6.1)
+    gvl_timing (0.2.0)
     hashdiff (1.1.2)
     hashery (2.1.2)
     hashie (5.0.0)
@@ -429,6 +430,7 @@ DEPENDENCIES
   factory_bot_rails
   faker
   flamegraph
+  gvl_timing
   htmlentities
   json
   letter_opener
--- a/extras/gvl_instrumentation.rb
+++ b/extras/gvl_instrumentation.rb
@@ -1,0 +1,37 @@
+require "gvl_timing"
+
+class GvlInstrumentation
+  def initialize(app, log_file)
+    @app = app
+    @log_file = log_file
+    @log_file.sync = true
+  end
+
+  def call(env)
+    response = nil
+
+    before_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    before_gc_time = GC.total_time
+    timer = GVLTiming.measure do
+      response = @app.call(env)
+    end
+    total_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) - before_time
+    gc_time = GC.total_time - before_gc_time
+
+    data = {
+      gc_ms: (gc_time / 1_000_000.0).round(2),
+      run_ms: (timer.cpu_duration * 1_000.0).round(2),
+      idle_ms: (timer.idle_duration * 1_000.0).round(2),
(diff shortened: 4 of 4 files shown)


lib/middleware/gvl_instrumentation.rb
```diff
@@ -0,0 +1,31 @@
...
+    gc_time = GC.total_time - before_gc_time
+
+    gc_ms = (gc_time / 1_000_000.0).round(2)
+    run_ms = (timer.cpu_duration * 1_000.0).round(2)
+    idle_ms = (timer.idle_duration * 1_000.0).round(2)
+    stall_ms = (timer.stalled_duration * 1_000.0).round(2)
+    io_percent = (timer.idle_duration / total_time * 100.0).round(1)
+    Rails.logger.info(
+      "[GVL] io_percent=#{io_percent}, run_ms=#{run_ms}, idle_ms=#{idle_ms}, stall_ms=#{stall_ms}, " \
+      "gc_ms=#{gc_ms}, path=#{env["PATH_INFO"]}"
```
byroot, line 26: Not sure if `PATH_INFO` is the best, as it will have a fairly high cardinality. Perhaps `controller/action` would be better?

pushcx:
Lobsters logs to files. I’d appreciate it if this logged to its own file rather than the main log; lots of things get handled out of it with grep and jq.

config/application.rb
```diff
@@ -20,6 +20,8 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+require File.expand_path("../../extras/gvl_instrumentation", __FILE__)
```
pushcx, line 23: Shouldn’t this be one level of `../`, not two?

lib/middleware/gvl_instrumentation.rb
pushcx: You probably want controller/action. Spiders regularly map all of the site and we have a few hundred thousand URLs. You’d gather hit frequency, but basically everything is log normal on recency like you’d assume.

config/application.rb
byroot: No, because it's relative to `__FILE__`, not `__dir__`.

lib/middleware/gvl_instrumentation.rb
byroot: > Maybe you want http verb too?

Good call. Even though it can be inferred from the action, it's easier that way.

> but I could check logs if it matters to you.

Nah, it's fine. I'm interested in requests that are actually processed by Ruby.

config/environments/production.rb
```diff
@@ -102,6 +102,9 @@
   # why help timing attacks?
   config.middleware.delete(Rack::Runtime)
 
+  require File.expand_path("../../extras/gvl_instrumentation", __FILE__)
+  config.middleware.insert 0, GvlInstrumentation, File.open(Rails.root.join("/srv/lobste.rs/log/production-gvl.log"), "a+")
```
byroot, line 106: I moved the log to its own file. But I don't know if the app has the permission to create it?

pushcx: That'll be fine.

pushcx:
I'm deploying now. Will a couple days of logs work for your purposes?

config/application.rb
pushcx: `/srv/lobste.rs/.rbenv/versions/3.4.1/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': cannot load such file -- /srv/lobste.rs/http/config/extras/gvl_instrumentation (LoadError)`

You moved this into `config/environments/production.rb` so I had to add an extra `../`.

pushcx:
OK. It's collecting in prod now, here's the last couple lines off it:

```
{"gc_ms":13.79,"run_ms":32.6,"idle_ms":2.28,"stall_ms":0.01,"io_percent":6.6,"method":"GET","action":"home#index"}
{"gc_ms":0.0,"run_ms":79.67,"idle_ms":3.29,"stall_ms":0.05,"io_percent":3.2,"method":"GET","action":"home#index"}
{"gc_ms":0.0,"run_ms":101.26,"idle_ms":12.3,"stall_ms":0.02,"io_percent":10.6,"method":"GET","action":"stories#show"}
{"gc_ms":0.0,"run_ms":25.99,"idle_ms":0.04,"stall_ms":0.4,"io_percent":0.1,"method":"HEAD"}
```

config/application.rb
byroot: Argh, so sorry about that -_-. That's why I added it to every evens at first, so it would be tested, but I moved it to producton at the end because the path was production specific...

byroot:
Nice!

byroot:
Unfortunately the link was a 404 😢

pushcx:
Whoops, I put them in the cache dir without thinking, so the link stopped working after 2 minutes. Good now.

byroot:
I got it!

If you have no objections I'll share that data in a google sheet or something in case others want to look at it.

pushcx:
I don't mind at all, I sent the link privately because I don't want to support it indefinitely.

PR #1444, opened 2025-01-25
remove raw_sql used to generate graph data in stats page

kernal053:
#539 Replace raw SQL in StatsController with ActiveRecord


pushcx:
Thanks for fixing this up. It might be a little slower but we cache these anyways, so it's worth cutting out the raw sql.

I'll handle the brakeman warnings.

PR #1447, opened 2025-01-25
Fix #1446 (broken flag reasons)

### LINKED ISSUE #1446: flag dropdown broken
Around 20:35 CET on 2025-01-25 I encountered a message about  new deploy.

When I visited the site afterwards, the following behavior was observed when clicking on the "flag" link:

Image

Behavior is the same on Chrome and Firefox on macOS.

sknebel:
Values are being passed as JSON string, so parse the string to JSON

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -94,9 +94,9 @@ class _LobstersFunction {
   constructor (username) {
     this.curUser = null;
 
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for jumping on this, deploying now.

PR #1448, opened 2025-01-26
reorganize color variables (fix #1445)

### LINKED ISSUE #1445: Bylines and other secondary is now black on white
Looks this way in Firefox and Safari:

Image

Image

May be a regression from #1410; cc @rahul1990gupta

clemeth:
- Since the `light-dark` function can only take color values, the usage is corrected by applying opacity through an inner `rgba` call.
- Use `color-mix` to apply opacity instead of `rgba` where practical.
- Redundant `--palette-*` variables are removed.

This should fix #1445.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -48,7 +48,6 @@
 	/* explicit view colors */
 	--color-box-bg: light-dark(#fff, #141414);
 	--color-box-bg-shaded: light-dark(#eee, #202020);
-	--color-box-bg-shaded-rgb: light-dark(rgb(238, 238, 238), rgb(32, 32, 32));
 	--color-box-border: light-dark(#ccc, #303030);
 	--color-box-border-focus: light-dark(#808080, #808080);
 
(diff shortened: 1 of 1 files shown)


Summertime:
for reference, color-mix has been stable for over a year, is in firefox ESR as well

https://caniuse.com/mdn-css_types_color_color-mix

clemeth:
@Summertime Yes, that was my thinking too.

In a thread on [Lobsters](https://lobste.rs/s/o2hhjc/what_happened_text_colour#c_vm7x8q) some users are reporting black triangles in Chromium on Linux. I experience the same in qutebrowser. Removing `light-dark` fixes it, so it seems support is lacking there as well.

app/assets/stylesheets/application.css
```diff
@@ -4,40 +4,50 @@
...
-	--opacity-contrast-10: light-dark(84%, 87%);
-	--opacity-contrast-7-5: light-dark(75%, 77%);
-	--opacity-contrast-6: light-dark(67%, 69%);
-	--opacity-contrast-5: light-dark(63%, 64%);
-	--opacity-contrast-4-5: light-dark(59%, 61%); /* WCAG AA minimum 4.5:1 text contrast */
-
-	--opacity-gradient-lit: light-dark(0%, 13%);
-	--opacity-gradient-shadowed: light-dark(13%, 3%);
+	/** main monochrome and variants **/
+	--base-bg-light: 254 254 254;
```
pushcx, line 8: Why put this in a variable instead of directly in the `--color-bg` a few lines down? I don't see that it's reused.

app/assets/stylesheets/application.css
```diff
@@ -74,9 +84,7 @@
 	--color-lobsters-hat-sysop-brim-stroke: light-dark(#bbb2b2, #2c2222);
 
 	/* mobile */
-	--color-mobile-story-border: light-dark(#e0e0e0, #1c1c1c);
```
pushcx, line 77: Good catch of dead code.

PR #1449, opened 2025-01-26
fix typo in #1448

clemeth:
I replaced the wrong comma in two lines in #1448.


pushcx:
Thanks for catching it. The typo I just put in the merge commit is not a gag; GitHub generates bad messages and I typod.

Issue #1404, opened 2024-12-20
<hr> style too light on firefox


andychu:
I came here to complain about the <hr> style and saw the shout-out as the latest commit :)

https://github.com/lobsters/lobsters/commit/96cf0b32ee81bb1bd7e6e9cb8c7030631145f902

I agree that the previous `<hr>` was too dark, but now it's too light:

It's almost invisible in Firefox:

![Screenshot from 2024-12-20 11-14-44](https://github.com/user-attachments/assets/c7fdb6f6-c815-4413-9c5e-597007f92238)

Barely visible in Chrome:

![Screenshot from 2024-12-20 11-16-03](https://github.com/user-attachments/assets/76114485-987d-49df-8157-76bc6ff1f865)

I think the issue is that it's both dotted and light color -- it should be dotted XOR light color !

----

IMO this is a good middle ground:

![Screenshot from 2024-12-20 11-18-01](https://github.com/user-attachments/assets/a0b53cfb-8fe9-4dc2-a648-466b24543921)

```
web/manual.css:hr {
  border: none;
  height: 1px;
  background-color: #BBB;
```

pushcx:
I copied the style from the vertical parenting lines. I agree that it could be a touch darker. Could you PR your middle ground?

fedemp:
Would you close this? It seems #1450 fixed this, as tested in production.

andychu:
Yes, thanks pushx for accepting this fix!

PR #1450, opened 2025-01-27
CSS: Make <hr> lighter than the default.

andychu:
But don't make it both dotted and light.

Related to

- commit 96cf0b3
- issue #1404

I remembered this due to recent CSS changes :-)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -44,6 +44,7 @@
 	--color-fg-affirmative: light-dark(rgb(0, 128, 0), rgb(42, 180, 42) );
 	--color-fg-author: light-dark(rgb(96, 129, 189), rgb(80, 112, 177));
 	--color-fg-shape: light-dark(rgb(187, 187, 187), rgb(80, 80, 80));
+	--color-hr: light-dark(rgb(187, 187, 187), rgb(80, 80, 80));
 
 	/* explicit view colors */
 	--color-box-bg: light-dark(#fff, #141414);
@@ -226,7 +227,7 @@ hr {
 	/* make it lighter than the browser's default */
 	border: none;
 	height: 1px;
-	background-color: #BBB;
+	background-color: var(--color-hr);
 }
 pre {
 	overflow-x: auto;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/application.css
```diff
@@ -223,8 +223,10 @@ div.shorten_first_p p:first-child {
 }
 
 hr {
+	/* make it lighter than the browser's default */
 	border: none;
-	border-top: 1px dotted var(--color-fg-shape);
+	height: 1px;
+	background-color: #BBB;
```
pushcx, line 229: Please use one of the existing custom properties for colors or create one. That'll make sure it supports both light and dark mode, and doesn't accidentally get edited in a way that loses contrast.

pushcx (changes requested):
Thanks for remembering this. I'm happy with the style you've chosen but have a minor css style issue for long-term maintainability.

andychu:
Hm I am having trouble testing this, on my machine I still see bug #1445 , where everything has high contrast

Even though I have this commit at HEAD

https://github.com/clemeth/lobsters/commit/f00807e5ea2473b1d74dfeb84cb0fa4bbdc4e0cd

Well I will at least do what I think is right, and then we'll see

andychu:
OK I tested it like this, and it seems to work:

image


```
 body {
   /* --color-hr: light-dark(rgb(187, 187, 187), rgb(80, 80, 80)); */
   --color-hr: rgb(187, 187, 187);
 }
 
 hr {
   /* make it lighter than the browser's default */
   border: none;
   height: 1px;
   background-color: var(--color-hr);
 }

```

pushcx (approved):
Looks great, thank you.

pushcx:
Thanks, looks good.

I don't know what's going on with dark mode. If you see it in Chrome, there's something about the bug we're not understanding, because mostly people use Chrome and don't see the bug (self included, when I test).

pushcx:
Thanks, looks good.

I don't know what's going on with dark mode. If you see it in Chrome, there's something about the bug we're not understanding, because mostly people use Chrome and don't see the bug (self included, when I test).

PR #1453, opened 2025-01-29
Conditionally import CSS color schemes

### LINKED ISSUE #1445: Bylines and other secondary is now black on white

clemeth:
This should fix #1445 for everyone, including those who have outdated browsers that don't support the `light-dark` function (like me).

The solution is to split the color schemes into different files and conditionally import them based on the user setting, and if no setting is present fall back on the `prefers-color-scheme` media feature in CSS.

This avoids code duplication and offers a clear and maintainable structure IMO.


app/assets/stylesheets/application.css
```diff
@@ -227,7 +146,7 @@ hr {
 	/* make it lighter than the browser's default */
 	border: none;
 	height: 1px;
-	background-color: var(--color-hr);
+	background-color: var(--color-fg-shape);
```
clemeth, line 149: I changed to this to use an existing variable and removed the other one, since they had the same value anyway.

PR #1454, opened 2025-01-30
Color schemes with increased contrast

clemeth:
This PR builds on #1453 and adds a user setting that enables increased contrast in the current color scheme.

This functionality was recently brought up after a bug (#1445) turned the text black on white (and vice versa in dark mode), and multiple users in [the thread about it](https://lobste.rs/s/o2hhjc/what_happened_text_colour) expressed support for the inadvertent change.

Feedback on the suggested colors and nuances is more than welcome.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -20,9 +20,9 @@
 
 	--base-shadow: 0 0 0;
 	--color-shadow: rgb(var(--base-shadow));
-	--color-shadow-80: rgb(var(--base-shadow) / 80%);
-	--color-shadow-25: rgb(var(--base-shadow) / 25%);
-	--color-shadow-10: rgb(var(--base-shadow) / 10%);
+	--color-shadow-80: rgb(var(--base-shadow) / var(--opacity-shadow-80));
+	--color-shadow-25: rgb(var(--base-shadow) / var(--opacity-shadow-25));
+	--color-shadow-10: rgb(var(--base-shadow) / var(--opacity-shadow-10));
 }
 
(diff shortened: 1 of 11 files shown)


NattyNarwhal:
Could this use `prefers-contrast`?

clemeth:
Added support for `prefers-contrast` and grouped the color settings together.

Tip: It is possible to test the `prefers-contrast` media feature in Chromium DevTools as follows: "Customize and control DevTools" (vertical ellipsis symbol) > "More tools" > "Rendering" > "Emulate CSS media feature prefers-contrast"

clemeth:
I realize the settings don't fully make sense right now. Currently, the `prefers-contrast` browser setting is ignored if the color scheme is set in the user settings. I think that's surprising behavior and would like to change it.

I guess we actually need a third setting for contrast, like there is for color scheme, i.e.: `{system,normal,high}`. I find the [`prefers-contrast` spec](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast) unclear, and think the most reasonable thing to do is to increase contrast if it is set to `more` and otherwise assume that the user wants the lowest contrast setting. Do you have any thoughts, @pushcx?

To summarize, I suggest these options on the settings page:
```
Color Scheme: [x] System [ ] Light  [ ] Dark
Contrast:     [x] System [ ] Normal [ ] High
```

pushcx:
I think that that parallel design of the setting is perfect.

pushcx (approved):
Looks great, thanks for working through my comments. I'll ping folks on the thread.

PR #1455, opened 2025-01-31
fix count of pages in ModerationsController #1433

### LINKED ISSUE #1433: Moderations pagination never shows the final page of results
**Steps to reproduce**

1. Empty your local moderation log
2. Insert 51 - 99 moderation records
3. Load http://localhost:3000/moderations

**Expected result**

Latest 50 moderation entries listed, with "Page 2" link at bottom of page to view remaining ones

**Actual result**

Latest 50 moderation entries listed, but no link to page 2 appears.

Adding `page=2` to the params results in a routing error "page out of bounds".

**Notes**

This is reproducible with any number of pages too, so it's a permanent off-by-one error in the logic. eg, having 149 moderation entries only results in being able to view 149-99, 99-49. The final 49 entries that should be on page 3 are invisible.

I suspect our maths is wrong in the `ModerationsController#index` action and we're working out the max page number to be 1 less than it should be, failing to account for the final (incomplete) page of entries.

dreig:
this should fix #1433 
I put the calculation in a helper, just so I could write an easy test for it.
There are several other controllers that use pagination (any that define a `*_PER_PAGE` constant) but none count the total number of pages, so don't seem to be affected by this bug.

### DIFF
--- a/app/controllers/moderations_controller.rb
+++ b/app/controllers/moderations_controller.rb
@@ -47,11 +47,10 @@ class ModerationsController < ApplicationController
     end
 
     # paginate
-    @pages = (@moderations.count / ENTRIES_PER_PAGE).ceil
-    @page = moderation_params[:page].to_i
-    if @page == 0
-      @page = 1
-    elsif @page < 0 || @page > (2**32) || @page > @pages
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks for checking for other instances of the bug and fixing it.

PR #1457, opened 2025-02-04
Respect `prefers-contrast` when color scheme is set

clemeth:
Fixes a bug from #1454 (merged as draft) where the `prefers-contrast` media feature would be ignored if the color scheme was set in the user settings.


pushcx:
Thanks for diagnosing this. In all the noise of renaming CSS files I didn't notice the change to the user setting. I'm running a migration to fix any broken users now.

Issue #1458, opened 2025-02-06
Migrations are broken


pushcx:
I feel like I'm reading over some obvious typo or misconfig here, so I'm posting in the hopes someone spots it.

Build is failing: https://github.com/lobsters/lobsters/actions/runs/13169060376/job/36755813576

I can't run tests because it complains that `db/migrate/20250203223624_count_merged_stories.rb` hasn't run. The `stories.stories_count` column that migration creates exists, as does a `20250203223624` in `schema_migrations`.

If I run `db:migrate` I get this failure:

```
$ r db:migrate
W, [2025-02-05T18:21:46.655004 #1098250]  WARN -- : heinous_inline_partial initialized, found: {"app/views/comments/_threads.html.erb" => "app/views/comments/_comment.html.erb", "app/views/home/index.html.erb" => "app/views/stories/_listdetail.html.erb", "app/views/search/index.html.erb" => "app/views/stories/_listdetail.html.erb"}
bin/rails aborted!
ActiveRecord::StatementInvalid: SQLite3::SQLException: near "SHOW": syntax error: (ActiveRecord::StatementInvalid)
SHOW FULL TABLES WHERE table_type = "VIEW"
^


Caused by:
SQLite3::SQLException: near "SHOW": syntax error: (SQLite3::SQLException)
SHOW FULL TABLES WHERE table_type = "VIEW"
^

Tasks: TOP => db:schema:dump
(See full trace by running task with --trace)
```

Which is weird, it looks like it's trying to run mariadb commands against sqlite, though I have my dev and test envs configured to use mariabdb.

```
---
trilogy: &trilogy
  adapter: trilogy
  encoding: utf8mb4
  host: 127.0.0.1
  port: 3306
  pool: 5
  socket: /run/mysqld/mysqld.sock
  username: pushcx
  password: [redacted]

sqlite3: &sqlite3
  adapter: sqlite3
  timeout: 1000

development:
  primary:
    <<: *trilogy
    database: lobsters
  cache:
    <<: *sqlite3
    database: tmp/cache/cache.sqlite3
    migrations_paths: db/cache_migrate

test:
  primary:
    <<: *trilogy
    database: lobsters_test
  cache:
    <<: *sqlite3
    database: tmp/cache/cache.sqlite3
    migrations_paths: db/cache_migrate

production:
  primary:
    <<: *trilogy
    database: lobsters_test
  cache:
    <<: *sqlite3
    database: tmp/cache/cache.sqlite3
[...]

Issue #1076, opened 2022-03-30
Redesign Notifications


pushcx:
Notifications on Lobsters have slowly grown into a fairly hairy muddle. There is inconsistent UI with odd limitations based on vintage. There's a good-sized project here if someone would like to take it on. In particular, this could make a nice portfolio piece for a junior dev.

Things people get notified about

 * Top-level comments to stories they've submitted
 * Direct to their comments (not grandchild replies)
 * `@username` mentions
 * Private messages

Perhaps also users who are particularly interested in a story or response would like to be able to subscribe to it the same way the author can.

Channels people get notified via:

 * Onsite [`/replies`](https://lobste.rs/replies)
 * Onsite [`/messages`](https://lobste.rs/messages)
 * E-mail
 * Pushover
 * [Your Threads](https://lobste.rs/threads)

There's no real rhyme or reason to which things generate which notifications. Except for the distinction between comments and PMs that shows up in `/replies` and `/comments`, maybe all channels should be available for all things?

And speaking of that exception, maybe we should unify the interface for new PMs and new
replies. I'm a fan of the way Reddit presents them together and nudged this direction a
little with the new header's singular Inbox link (#1061).

Additionally, `/replies` works as a view via the `ReplyingComments` model. While an excellent feature with [nice limitations](https://github.com/lobsters/lobsters/blob/e49e1672d7675dd4e1f193bcf3cb1f17939464d4/db/views/replying_comments_v09.sql#L31-L52) to turn down flame wars, it's the most expensive query the site regularly runs (and addressing that with caching [prompted several bugs](https://github.com/lobsters/lobsters/issues/1065)). It would be best if the view's performance could be improved (see `EXPLAIN` below), but maybe this needs to be entirely rebuilt with more of a materialized design? (A hassle there is that some of the conditions that exclude a comment like flags can come well after the comment was posted.)

```
[...]

PR #1459, opened 2025-02-06
Proxy Scenic to correct adapter with multiple databases

caius:
## What?

- [x] Add proxy object to swap between MySQL and SQLite adapters for Scenic gem

## Why?

#1458 exposes the fact we can't run `bin/rails db:schema:dump` successfully on master. We can dump the primary database (MySQL), but we run into an error dumping the cache database (SQLite). This is because the Scenic MySQL adapter runs queries against the SQLite database it doesn't understand so errors out.

Attempt to patch this quickly by adding a proxy object that picks between both depending on what the current connection to a database is.

This works for `bin/rails db:schema:dump`, but I've not tested the app otherwise. It may well break in other places!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem "mission_control-jobs"
 # rails
 gem "scenic"
 gem "scenic-mysql_adapter"
+gem "scenic_sqlite_adapter"
 gem "activerecord-typedstore"
 gem "propshaft"
 gem "importmap-rails", "~> 2.0"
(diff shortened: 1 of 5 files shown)


pushcx:
Scenic is aware of the bug: https://github.com/scenic-views/scenic/issues/399 https://github.com/scenic-views/scenic/issues/402 (possibly more, I haven't searched exhaustively.)

Your workaround is the best of the ones I've seen, so I'm going to merge it and then bump on #1076 in priority. It's time to stop using views and drop scenic, it keeps generating these odd issues for us.

PR #1462, opened 2025-02-06
Close top-level comment tags

tomcat584:
When serializing a thread, the top-level comment `<li>` tags were accidentally left unclosed (such that there was no matching `</li>` tag), unlike the child comments. But HTML5 does not require that `<li>` tags be explicitly closed, and will use heuristics to infer where a missing `</li>` tag probably should be inserted. Such heuristics are sometimes wrong and lead to strange UI issues with nested comments on pages with a thread view (like on stories and user profiles). To fix this issue, we insert the missing `</li>` tags.

We can verify the fix by comparing the resultant HTML output before/after, using an unclosed tags checker such as [this one](https://www.aliciaramirez.com/closing-tags-checker/).

Fun fact, I only noticed this problem is because I have been writing a scraper from scratch, and HTML parsing comes with its own surprises...

### DIFF
--- a/app/views/comments/_threads.html.erb
+++ b/app/views/comments/_threads.html.erb
@@ -15,6 +15,7 @@
           </li>
         </ol>
       <% end %>
+      </li>
     <% end %>
     <% previous_depth = comment.depth %>
     <% top_comment_depth ||= comment.depth %>
@@ -216,4 +217,5 @@
     </li>
   </ol>
 <% end %>
+</li>
 </ol> <%# matches very first comment %>


pushcx:
That can't have been easy to find. I knew there was some mismatch hanging around but couldn't spot it in all the noise.

You know there's json endpoints, right? You don't have to parse my HTML crimes. :) #mismat

tomcat584:
@pushcx It definitely wasn't obvious, I just happened to use tools that are pedantic about unclosed tags and make me look for them. And when the tag checker insisted that *only* the `<li class="comment_subtree">` tags are unclosed, the task was just to figure out where it might be logical to add `</li>` tags to this ERB template if they're missing. Pretty much only in the chunks of code near the start & the end of the file, as they have to do with opening/closing trees of replies. Then a few cycles of trial/error/rationalization to come up with the patch.

> You know there's json endpoints, right? You don't have to parse my HTML crimes. :)

Huh, I'm only just hearing about this. There's a Lobsters API? How can I use it?

Even if you put it that way, speed hacks are a forgivable crime. :) And the overall code quality is excellent.

PR #1469, opened 2025-02-08
use explicit callback instead of `touch: :last_reply_at` as to not change `parent_comment.updated_at`

### LINKED ISSUE #1467: Editing a comment changes the mod time of the whole chain
When I edited [this comment](https://lobste.rs/s/f5ziu7/comments_on_shared_unix_hosting_vs_cloud#c_sxdaki), all the parent comments were also marked as edited at the same time.  
It seemed to also happen for at least the previous comment in the chain as I noticed my first comment sais "edited 58 minutes ago" when I re-opened the thread, which was the time  dijit's comment also showed, and I knew I posted mine quite earlier than that.

~I haven't checked if it happens for other threads.~
It seems like it's happening with all threads on all posts ATM.

dreig:
Using touch with `belongs_to :parent_comment` will automatically modify parent_comment's `updated_at`, causing #1467.
Use an explicit `after_create` callback instead.
fix #1467.

There is another use of [belongs_to :story, touch: :last_comment_at](https://github.com/lobsters/lobsters/blob/cb650ed40fad0d4ea8df24b3ced2dcb66e90b1a2/app/models/comment.rb#L5) Is that (modifying `updated_at` on story) an issue?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,6 +70,7 @@ group :test, :development do
   gem "standard-performance"
   gem "standard-rails"
   gem "super_diff"
+  gem "timecop"
   gem "faker"
   gem "byebug"
   gem "rb-readline"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -409,6 +409,7 @@ GEM
       patience_diff
     svg-graph (2.2.2)
     thor (1.3.2)
+    timecop (0.9.10)
     timeout (0.4.3)
     trilogy (2.9.0)
     ttfunk (1.8.0)
@@ -499,6 +500,7 @@ DEPENDENCIES
   standard-rails
   super_diff
   svg-graph
+  timecop
   trilogy
   vcr
   webmock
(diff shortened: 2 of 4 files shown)


pushcx:
@dreig Yes, that other one's also a bug. I didn't realize `touch` came with this footgun. 

I think rather than reimplementing this functionality I'd rather add an explicit `last_edited_at` to Story and Comment. I can't guess how many other Rails features also (or will also...) bump `updated_at`.

Thanks very much for figuring out what was happening here. Do you want to write a migration and update the `edit` actions and views, or I can take it?

I've got an hour before a social commitment and I'd like to get this squared away so I have less data to slog through logs to fix.

dreig:
@pushcx sure, go ahead and take it.

dreig:
I agree with your suggestion to use a custom field for this logic, instead of relying on `updated_at`.
But I worry that if you try to implement it using `touch` you may run into the issue of cascading updates, where changes to `:last_edited_at` get propagated up the `parent_comment` chain. (I'm not sure of this, you should verify whether that happens)

pushcx:
Thank you for the warning, I'll watch out for it.
Rails has leaned hard into its idea of nested fragment caching, so I think that's why `updated_at` now has this helpful behavior.

Issue #1466, opened 2025-02-07
Newly added comments are off-axis


shrik450:
Note the "test comment" is just slightly off to the right of the sibling comment:

<img width="845" alt="Image" src="https://github.com/user-attachments/assets/1845a1fc-5371-48a4-9c86-62acef216bc5" />

This disappears when the page is reloaded:

<img width="857" alt="Image" src="https://github.com/user-attachments/assets/fb20b428-5308-47b0-9e44-ec65aed7d5a0" />

NattyNarwhal:
These misaligned comments also appear in i.e. `/threads`.

NattyNarwhal:
![Image](https://github.com/user-attachments/assets/72dfdb99-7ed6-4bd5-9b81-37501aa46f1d)

Screenshot to clarify; I note it's not just new replies, but sometimes existing children.

LnMrx:
REally ?

Issue #1236, opened 2024-01-02
Threading bug on the /~user/threads page


jmiven:
On [my threads page](https://lobste.rs/~jmiven/threads), a [3 days old comment on a story](https://lobste.rs/s/kfu8jg/inheritance_ruby_pictures#c_fman15) is threaded under a [more recent comment on another story](https://lobste.rs/s/oa4kgv/yet_another_static_site_generator_for#c_7u0r8y).

<img width="846" alt="threading_bug" src="https://github.com/lobsters/lobsters/assets/6455809/374b0a55-df28-4687-94b0-e7339e468c3d">

pushcx:
Well, this is the result of my big clever redesign of the threading code in #1190. I saw this bug and hoped I could override the `parent_depth` test in `app/views/comments/_threads.html.erb` by comparing `thread_id`, but with @lonjil’s report of #1251 it’s clear this is a more general problem. In that one, like here, the comments are in the wrong place because the parent comments have identical votes (~half of comments have one or two upvotes and zero flags) and the low byte of their `id` is identical. (We get ~200 comments per day, note that in both reports’ screenshots the parents are posted 2 or 3 days apart.)

This is going to be some amount of redesign to the `confidence_order` field and `Comment#update_score_and_recalculate!` for a reliable fix. I had considered that siblings might swap order a bit (see [comment](https://github.com/lobsters/lobsters/blob/df2ac93fe313a7d2dee30da17fe4374e3db9630d/app/models/comment.rb#L366-L377)) but I didn’t realize that I was setting up a 1 in 256 chance that they’d _swap children_ because of it.

An obvious fix would be to increase the comment id component of `confidence_order` to two bytes. It would seem that it reduces the chances from 1 in 256 to 1 in 65,536. However! We get ~200 comments per day and stories accept comments for 90 days (`Story::COMMENTABLE_DAYS`), so we only see ~18,000 comment IDs in the period available for a collision.

That works in practice but not in theory. I don’t love that it would set up an ever rarer bug if we grow. I have a hunch there’s a better fix available that I’m not immediately seeing. Maybe it’ll come to me, or maybe someone else has an idea for a fix?

epidemian:
[...]

PR #1470, opened 2025-02-09
Close the top-level comment tags

tomcat582:
The `<li class="comment_subtree">` tag of the first comment in each comment tree was accidentally left unclosed: it doesn't have a matching  `</li>` tag, but all its child comments do. We should fix this because when a `<li>` element has a missing end tag, HTML5 infers where to place the end tag using heuristics. The heuristics may fail at times, leading to subtle UI bugs on pages that display nested trees of comments (such as stories and user profiles).

We can use a [closing tag checker](https://www.aliciaramirez.com/closing-tags-checker/) to verify that this commit closes these open tags.

(I think github deleted the old pull request because my throwaway email made me look like a bot account. Sorry about the confusion.)

### DIFF
--- a/app/views/comments/_threads.html.erb
+++ b/app/views/comments/_threads.html.erb
@@ -15,6 +15,7 @@
           </li>
         </ol>
       <% end %>
+      </li>
     <% end %>
     <% previous_depth = comment.depth %>
     <% top_comment_depth ||= comment.depth %>
@@ -216,4 +217,5 @@
     </li>
   </ol>
 <% end %>
+</li>
 </ol> <%# matches very first comment %>


pushcx:
I've asked GitHub support to restore it.

What's with the brand new throwaway accounts? I thought it was odd yesterday that someone apparently opened a new account to submit a patch, now two?

tomcat582:
It's just a personal quirk, I'd prefer not to have a github account, but I thought it'd be a waste not to contribute this fix when I already found it.

pushcx:
I spent a few minutes validating pages and it looks like this is a good fix. I will reiterate what I said on your previous PR, that I have no idea how you tracked down where to insert those `</li>` given the giant string soup we have going on in erb, but thank you. :)

tomcat582:
Thank you.

I think this merge may have affected issue https://github.com/lobsters/lobsters/issues/1236

pushcx:
Did you mean to link #1466 ? I could see both affected, really.

tomcat582:
In #1236 @epidemian reports stray `<li>` markers on some threads pages, so hopefully closing the remaining tags fixes that bug at least? As for #1466, I can reproduce it locally by creating comments so it's not fixed, but I'll have a look at it later today.

Issue #1456, opened 2025-02-03
UI improvements for merged stories


pushcx:
I've been working through improving the merged stories feature on [office hours streams](https://push.cx/stream).
This is my to-do list; I'm working from some high-level UI improvements to make it less confusing down through the database improvements to fix the bug factory (latest being #1298) and then write up a public doc of this unique feature.

 - [ ] refresh display at top of single-story page
   - [ ] display count of stories when there's a merge
   - [ ] put submitter of first story alongside title like others
   - [ ] remove redundant merge icons next to titles
 - [ ] split the tree of comments by which merged story they're on w _listdetail header
 - [ ] on a merged story, no top-level comment box, one in a <detail> for each
 - [ ] cross-link stories - footer w backref links to stories and comments linking here
 - [ ] cross-link comments - details/metadata w backref links to comments linking here
 - [ ]   don't cross-link if source comment has a negative score or mod soft-deletes the link
 - [ ] refactor the db into headline and stories?
 - [ ]   stories.stories_count will need a migration to recount; it's # stories merged in, not total # (so: off by one)
 - [ ] story submit: submitter can choose to merge into existing story (and is encouraged to when correct!)
 - [ ] refactor user suggestions into 'draft posts' rather than per-field tables
 - [ ] user suggestions can promote top-level comment with unsubmitted link to Story
 - [ ] user suggestions can merge and unmerge two stories
 - [ ] vague/misdesign: hidden stories or hidden links; same for save, upvote
 - [ ] write up a AboutController page explaining merging

dgilmanAIDENTIFIED:
My two cents but I think you can omit the merge icon/count on the front page list of stories. There's nothing you can do with that information at that location so removing it makes the UI a bit cleaner.

pushcx:
Related: #740

pushcx:
I cleaned up the display of the story titles and metadata on the single-story view in b64cd14e and 426bce96, now all the stories are listed consistently and the metadata for the top headline doesn't appear below all the other headlines.

Before:
![Image](https://github.com/user-attachments/assets/95e789dd-8157-42e3-b1fe-30ddcadc7126)


After:

![Image](https://github.com/user-attachments/assets/6a5dff39-58fa-41db-a81c-ee6ec4c34033)

pushcx:
#1502 was caused by this work, just linking context.

dzwdz:
[...]

PR #1474, opened 2025-02-13
wip on /cabinet view of partials

pushcx:
Just wanted to add visibility here, this is a [Storybook-like](https://5f27fec87f827f00226fd51a-mkzlawwyms.chromatic.com/?path=/story/components-button-iconbutton--variants&globals=colorScheme:dark) view of our templates in different states.

Our templates are reused in several different states, some of which are different enough they should probably be different partials... but then without something like this it's very hard to make sure they're getting revised in sync.

The partial I'm listing here is setup for the `listdetail` improvements planned in #1456. I started shotgunning conditionals around `single_story` and `stories_count > 0` throughout the partial and I realized I had very low odds of producing reasonable output unless I could see all the variants alongside each other.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ gem "stackprof"
 gem "prosopite"
 
 gem "builder" # for rss
+gem "faker" # for factory data and /cabinet
 gem "oauth" # for linking accounts
 gem "mail" # for parsing incoming mail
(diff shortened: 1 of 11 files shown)


pushcx:
I rebased this onto prod in a45014fd

Issue #1430, opened 2025-01-10
Top-level comment box sometimes gets borked when clicking its buttons on a merged(?) story


strugee:
STR:

1. On an Android phone, go to https://lobste.rs/s/gpoygp/google_linux_foundation_launch
2. Paste "This is your regularly scheduled reminder that the Linux Foundation is not an open source or free software organization. It does not exist to promote knowledge or a public good. It is a trade organization which exists to support the businesses that are a part of it." in the comment box, minus quotes
3. Tap the Preview button (or the Submit button, but not in prod obviously) 
4. Observe that the comment box disappeared, and the comment you just posted or previewed took its place.

I'm on Firefox Nightly 136.0a1, built January 9th 2025. Pixel 9 running GrapheneOS.

Screenshot from directly after I tapped Submit:

![Screenshot_20250109-222628](https://github.com/user-attachments/assets/aa8e6c92-825d-4d43-a2b1-c2abc20b15f6)

PR #1475, opened 2025-02-20
Update the frontend code that renders newly posted comments (fixes #1466)

### LINKED ISSUE #1466: Newly added comments are off-axis

tomcat582:
Background: https://github.com/lobsters/lobsters/issues/1466#issuecomment-2670096125

These commits aim to fix the bug by updating the frontend code to match the current design
- create the reply form inside the list of child comments (this seems to have been the original intention and makes it easier to cleanly remove)
- don't wrap newly posted comments in a `<div>`
- update the styles and add tree lines to parent & child comments when posting a reply

On the backend, I thought we should also modify the comment views so that a comment will only have a list of child comments `<ol class="comments">` if it has replies. This is not absolutely necessary but it's consistent with how comment threads are  generated in `_threads.html.erb`.

`_threads.html.erb` will only conditionally insert the `ol.comments` tag:

https://github.com/lobsters/lobsters/blob/2cb0baab355e7fabe26abc2a9a043a1d1a5986bd/app/views/comments/_threads.html.erb#L6-L19

Whereas before this change, several other comment views would always insert the tag, even if it's known to be empty, like when rendering a newly posted reply:

https://github.com/lobsters/lobsters/blob/2cb0baab355e7fabe26abc2a9a043a1d1a5986bd/app/views/comments/_postedreply.html.erb#L2-L6

If we adopt all of these changes, posting a comment should produce a DOM that looks the same as if the page was generated by the backend.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -90,6 +90,17 @@ const removeExtraInputs = () => {
   }
 }
 
+const createChildList = (comment) => {
+  let children = qS(comment, '.comments')
+  if (!children) {
+    children = document.createElement('ol')
+    children.classList.add('comments')
+    // currently this is the layout
+    comment.append(children)
+  }
+  return children
+}
+
 class _LobstersFunction {
   constructor (username) {
     this.curUser = null;
@@ -664,7 +675,9 @@ onPageLoad(() => {
 
     let div = document.createElement('div');
     div.innerHTML = '';
-    comment.lastElementChild.append(div);
+    div.classList.add('reply_form_root')
+    const children = createChildList(comment)
+    children.prepend(div)
 
     fetchWithCSRF('/comments/' + comment.getAttribute('data-shortid') + '/reply')
       .then(response => {
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -27,13 +27,21 @@ const onPageLoad = (callback) => {
   document.addEventListener('DOMContentLoaded', callback);
 };
 
-const parentSelector = (target, selector) => {
+const parentSelectorOrNull = (target, selector) => {
   let parent = target;
   while (!parent.matches(selector)) {
     parent = parent.parentElement;
     if (parent === null) {
-      throw new Error(`Did not match a parent of ${target} with the selector ${selector}`);
+      return null;
     }
+  }
+  return parent;
+};
+
+const parentSelector = (target, selector) => {
+  let parent = parentSelectorOrNull(target, selector)
+  if (parent === null) {
+    throw new Error(`Did not match a parent of ${target} with the selector ${selector}`);
   }
   return parent;
 };
@@ -90,13 +98,12 @@ const removeExtraInputs = () => {
   }
 }
 
-const createChildList = (comment) => {
-  let children = qS(comment, '.comments')
+const createChildList = (parent) => {
+  let children = qS(parent, '.comments')
   if (!children) {
     children = document.createElement('ol')
     children.classList.add('comments')
-    // currently this is the layout
-    comment.append(children)
+    parent.append(children)
   }
   return children
 }
@@ -256,7 +263,57 @@ class _LobstersFunction {
       body: formData
     })
       .then(response => {
-        response.text().then(text => replace(form.parentElement, text));
+        response.text().then(text => {
+
+          // must keep this behavior up to date as the website's templates change:
+          //   app/views/comments/_comment.html.erb
+          //   app/views/comments/_threads.html.erb
+          //   app/views/stories/show.html.erb
+
+          const wrappedComment = document.createElement('p')
+          wrappedComment.innerHTML = text
+          const comment = qS(wrappedComment, '.comments_subtree')
+
+          const replyForm = parentSelectorOrNull(form, '.reply_form_temporary')
+          if (replyForm) {
+            // user submitted from a temporary reply form, so this is a reply to an existing comment
+            const parentComment = parentSelector(replyForm, '.comments_subtree')
+            replyForm.remove()
+
+            const children = createChildList(parentComment)
+            children.prepend(comment)
+
+            /// update styles to exactly reflect what would be generated on the backend
+
+            // comments/_comment.html.erb:
+            //   <%= comment.reply_count.to_i == 0 ? "no_children" : "" %>
+            const parentTreeline = qS(parentComment, '.comment_parent_tree_line')
+            parentTreeline.classList.remove('no_children')
+
+            // comments/_threads.html.erb:
+            //   <% if comment.depth != top_comment_depth %>
+            //     <div class="comment_siblings_tree_line"></div>
+            //   <% end %>
+            const treeline = document.createElement('div')
+            treeline.classList.add('comment_siblings_tree_line')
+            comment.append(treeline)
+
+            // comments/_threads.html.erb:
+            //   <% if !previous_depth || comment.depth > previous_depth %>
+            //     <span class="prior_comment_has_children"></span>
+            const span = document.createElement('span')
+            span.classList.add("prior_comment_has_children")
+            parentComment.insertBefore(span, children)
+
+          } else {
+            // there is no temporary reply form, so user must have created a "top-level" comment
+            parentSelector(form, '.comment_form_container').remove()
+            const storyComments = qS('#story_comments')
+            const comments = createChildList(storyComments)
+            comments.prepend(comment)
+          }
+
+        })
       })
   }
 
@@ -660,8 +717,8 @@ onPageLoad(() => {
     event.preventDefault();
     if (!Lobster.curUser) return Lobster.bounceToLogin();
 
-    const comment = parentSelector(event.target, '.comment');
-    const commentId = comment.getAttribute('id');
+    const commentInfo = parentSelector(event.target, '.comment');
+    const commentId = commentInfo.getAttribute('id');
 
     // guard: don't create multiple reply boxes to one comment
     if (qS('#reply_form_' + commentId)) { return false; }
@@ -675,11 +732,12 @@ onPageLoad(() => {
 
     let div = document.createElement('div');
     div.innerHTML = '';
-    div.classList.add('reply_form_root')
+    div.classList.add('reply_form_temporary')
+    const comment = parentSelector(commentInfo, '.comments_subtree');
     const children = createChildList(comment)
     children.prepend(div)
 
-    fetchWithCSRF('/comments/' + comment.getAttribute('data-shortid') + '/reply')
+    fetchWithCSRF('/comments/' + commentInfo.getAttribute('data-shortid') + '/reply')
       .then(response => {
         response.text().then(text => {
           // guard: don't create multiple reply boxes to one comment
--- a/app/views/comments/_postedreply.html.erb
+++ b/app/views/comments/_postedreply.html.erb
@@ -2,5 +2,4 @@
 <li class="comments_subtree">
   <%= render "comments/comment", :comment => comment,
     :show_tree_lines => true %>
-  <ol class="comments"></ol>
 </li>
--- a/app/views/comments/index.html.erb
+++ b/app/views/comments/index.html.erb
@@ -4,7 +4,6 @@
   <% @comments.each do |comment| %>
     <li><%= render "comments/comment", :comment => comment,
       :show_story => true %>
-      <ol class="comments"></ol>
     </li>
   <% end %>
 </ol>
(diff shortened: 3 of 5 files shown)


app/javascript/application.js
```diff
@@ -649,8 +717,8 @@ onPageLoad(() => {
     event.preventDefault();
     if (!Lobster.curUser) return Lobster.bounceToLogin();
 
-    const comment = parentSelector(event.target, '.comment');
-    const commentId = comment.getAttribute('id');
+    const commentInfo = parentSelector(event.target, '.comment');
```
pushcx, line 720: Why rename this? It is a comment, not some kind of metadata. If you want to use the name `comment` for the new one, how about `parentComment`?

app/javascript/application.js
```diff
@@ -664,9 +732,12 @@ onPageLoad(() => {
 
     let div = document.createElement('div');
     div.innerHTML = '';
-    comment.lastElementChild.append(div);
+    div.classList.add('reply_form_temporary')
+    const comment = parentSelector(commentInfo, '.comments_subtree');
```
pushcx, line 736: I don't understand calling this a `comment`; it's the parent element in the DOM where many comments will be rendered.

app/javascript/application.js
```diff
@@ -245,7 +263,57 @@ class _LobstersFunction {
...
       .then(response => {
-        response.text().then(text => replace(form.parentElement, text));
+        response.text().then(text => {
+
+          // must keep this behavior up to date as the website's templates change:
+          //   app/views/comments/_comment.html.erb
+          //   app/views/comments/_threads.html.erb
+          //   app/views/stories/show.html.erb
+
+          const wrappedComment = document.createElement('p')
```
pushcx, line 273: What is this `p` element added for? I don't think any of the places we render comments have a `p` as a parent tag that this code would be trying to match.

app/javascript/application.js
```diff
@@ -90,6 +98,16 @@ const removeExtraInputs = () => {
   }
 }
 
+const createChildList = (parent) => {
+  let children = qS(parent, '.comments')
+  if (!children) {
```
pushcx, line 103: If some comments don't have a placeholder `<ol class="comments">`, I'd rather fix that inconsistency than remove it from more places and add a branch that rebuilds it. It's one more piece of state to have to keep track of and mutate.

app/javascript/application.js
```diff
@@ -245,7 +263,57 @@ class _LobstersFunction {
...
+
+            const children = createChildList(parentComment)
+            children.prepend(comment)
+
+            /// update styles to exactly reflect what would be generated on the backend
+
+            // comments/_comment.html.erb:
+            //   <%= comment.reply_count.to_i == 0 ? "no_children" : "" %>
+            const parentTreeline = qS(parentComment, '.comment_parent_tree_line')
+            parentTreeline.classList.remove('no_children')
```
pushcx, line 291: Good catch, this has been a minor visual inconsistency forever. :)

app/javascript/application.js
```diff
@@ -245,7 +263,57 @@ class _LobstersFunction {
...
+            // comments/_threads.html.erb:
+            //   <% if !previous_depth || comment.depth > previous_depth %>
+            //     <span class="prior_comment_has_children"></span>
+            const span = document.createElement('span')
+            span.classList.add("prior_comment_has_children")
+            parentComment.insertBefore(span, children)
+
+          } else {
+            // there is no temporary reply form, so user must have created a "top-level" comment
+            parentSelector(form, '.comment_form_container').remove()
```
pushcx, line 310: This `remove` and `prepend` is doing the work of the `replace` call it replaced, right? Is there a benefit to doing it this way instead of keeping that code, at least for this code path?

pushcx (changes requested):
Whew, you really got into the weeds of our fiddly HTML structure for this fix, thank you.

I've got a few questions about why you made some of these changes, mostly trying to make sure that it's clear what's happening in a bunch of code that's very tightly coupled to our DOM complexity. So maybe the best answers for a lot of these would be code comments. :)

app/javascript/application.js
tomcat582: That makes sense, I'll fix it

app/javascript/application.js
tomcat582: After posting a comment, the backend returns a String `text` which contains the HTML code for the reply. But in order to be able to use DOM functions like `prepend`/`append` and modify styles conveniently, I thought we'd need to parse that HTML code to create a HTMLElement object first.

There may be better ways to make the browser parse a String into a HTMLElement, but I thought the easiest way is create a temporary element (`wrappedComment`) and set the text as its `innerHTML`. When we do this the browser will parse the new `innerHTML` to create the HTMLElement DOM objects for us automatically. Afterwards we can just query the children of the temporary element to get our object.

So the element `wrappedComment` is never inserted into the page. It's just a temporary element we create to make the browser parse the String `text` into a HTMLElement object. Only the comment is put in the DOM, later on when we add it to the children list.

(I could add a code comment to explain what's going on here though.)

app/javascript/application.js
tomcat582: The comment form at the top of the page is not in the same DOM subtree as the story's comments. So if we called `replace`, the reply would be in the wrong place. I think `replace` worked in the past because the website was structured differently back then and it made sense, but now people encounter some bugs when creating top-level comments (I think the issue https://github.com/lobsters/lobsters/issues/1430 is related to that).

app/javascript/application.js
tomcat582: I could change it back. I personally thought of the entire `<li class="comment_subtree">` unit as a "comment".

app/javascript/application.js
tomcat582: I personally thought of the entire `<li class="comment_subtree">` unit as a "comment" and I thought of `parentComment` as "the comment to which this one is a reply". So while coding I wished it had been named differently. But I can totally change it back

app/javascript/application.js
tomcat582: What do you think about the new commit? https://github.com/lobsters/lobsters/pull/1475/commits/eef9789cf9001781df5db4739b2a77f1b2a4cd26

app/javascript/application.js
tomcat582: Renamed them `commentSubtree`/`parentSubtree` in the latest commit https://github.com/lobsters/lobsters/pull/1475/commits/097979c3fecf4b3967c9d9f60b440b9335231bfc

app/javascript/application.js
tomcat582: I added comments explaining this in the latest commit

app/javascript/application.js
tomcat582: > (I could add a code comment to explain what's going on here though.)

Added in the latest commit

app/views/comments/_threads.html.erb
```diff
@@ -222,10 +227,15 @@
 
 <% end %>
 
-<%# close tree of replies %>
```
tomcat582, line 216: Fixing a mistake introduced in the earlier commit https://github.com/lobsters/lobsters/pull/1475/commits/ad981a8832e5efe26173bff409ea7183606ea193 :sweat_smile:

app/views/comments/_threads.html.erb
```diff
@@ -222,10 +227,15 @@
...
-<% top_comment_depth && (previous_depth - top_comment_depth + 1).times do %>
-    </li>
-  </ol>
+<% if top_comment_depth %>
```
tomcat582, line 230: Fixing a mistake introduced in the earlier commit https://github.com/lobsters/lobsters/pull/1475/commits/ad981a8832e5efe26173bff409ea7183606ea193 :sweat_smile:

tomcat582:
I'll stop making changes now :) Tried to reduce the maintenance burden as much as possible, let me know which commits we'll keep or delete

app/javascript/application.js
pushcx: Thanks, these comments really helped me understand what's happening, I didn't get it before.

Could we update the page markup to be consistent, move the top-level reply form into the right place in the tree? That also might be good prep for changes I plan to make in #1456, breaking up the threads by which link they're replying to.

app/views/comments/_threads.html.erb
```diff
@@ -1,25 +1,33 @@
...
 <% top_comment_depth = nil %>
+<%# open list of threads %>
+<ol class="comments">
 <% thread.each do |comment| %>
 
-    <% if !previous_depth || comment.depth > previous_depth %>
-      <%# brittle - see li.comments_subtree > .comment:nth-last-child(2) in css %>
-      <span class="prior_comment_has_children"></span>
+    <% if previous_depth %>
+      <% if comment.depth > previous_depth %>
```
pushcx, line 9: Whew, this has gotten complicated, it took me a bit of reading to work out how you'd rearranged the branches and your comments helped a lot.

I've experimented with laying out comments using grid so they'd all be sibling `<li>` on a single `<ol>`. It'll need quite a bit of work to finish, but it would mean we don't have this big hairy code for performantly mapping the linear list of comments into a nested tree structure.

I've been really pleasantly surprised that anyone would tackle all this really hairy coupling on our hot path, so thank you, and if (after I merge this PR) you're interested in burning it all down, I'd be happy to set you up to try it.

app/assets/stylesheets/application.css
```diff
@@ -659,7 +659,8 @@ li .comment_parent_tree_line.score_shown {
 li .comment_parent_tree_line.can_flag {
 	top: 56px;
 }
-li .comment_parent_tree_line.no_children {
+/* hide for comments that don't have children */
+li:not(:has(> .comments > li)) > .comment > .comment_parent_tree_line {
```
pushcx, line 663: That's really nice, `:has` wasn't available when this was first written, but it's great that you saw the opportunity to simplify the amount of state the template has to juggle.

pushcx:
This is great work, thank you. This is the kind of PR where I see four lines of code change and I know it must have taken an hour per line to get right. All this tight coupling between markup and js, and then the complexity of remapping a data structure in a performant way!

I left a comment about [one way this might be further simplified](https://github.com/lobsters/lobsters/pull/1475/files?diff=split&w=1#r1968420112). If you think it can work, let's do it. If you don't think it'll work or is not worth it, I'll go ahead and merge.

You also reminded me of [an experiment](https://github.com/lobsters/lobsters/pull/1475/files#r1968431231) with css grid that might cut down on this complexity; I started it because it would be nice to use [scroll-driven animations](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll-driven_animations) to widen comments on deeply-nested replies. Would you like to try that? At this point you probably understand the comment layout code better than anyone.

pushcx:
I'm sorry it took me a few days to get back to this review, because I really do appreciate and want to merge this work. I try to timebox my Lobsters development to the office hours streams so that it doesn't take over my life.

Especially because you're using a throwaway account, I don't know what your plans are here, if you'd like to keep working on this, to work on a next project, or if you're even seeing my messages.

Please let me know if you'd like to continue, or I don't mind taking it from here. If I haven't heard back on Monday, I'll take over it on the next office hours.

tomcat582:
Sorry about the delay. To me, working on these little pull requests has been a fun way for me to put my mind off my main projects that gave me stress. I tend to work really obsessively on things, so when you suggested another project to potentially rewrite this, I wanted to think that through before promising to continue in light of other tasks I might end up neglecting. I think I will just have to leave the other features to you and ask you to merge this as-is. I'm also totally OK with any future decision to replace the code since I really mostly did this for fun.

And absolutely no worries about following your schedule (IIRC there is a message explaining that when you submit a pull request). I took the opportunity to improve the patch and I'm happier with the end result because of that. I'd also like to thank you since I've had a really positive experience what with you appreciating the work I put in, and I got to learn about Rails for the first time.

app/javascript/application.js
tomcat582: Hi. I wanted to try making it more consistent but moving the comment box to the comments (moving line 48 -> line 66) would position the comment box beneath the "thread summary" (line 53). I wasn't sure whether this is a desirable design change. But yes I think it's possible.

https://github.com/lobsters/lobsters/blob/c8dc8598c25ac6a8d9c539f8609ea35555655354/app/views/stories/show.html.erb#L47-L66

pushcx: OK, I'll look into it again as I work though story merging.

pushcx:
Cool, well, anytime you'd like to take on new projects, I have a long list available. :D

pushcx (approved):

tomcat582:
Great. Thanks :)

pushcx:
Sorry it took me a minute to close out, but thanks again for doing the deep dive here.

Issue #1394, opened 2024-12-09
Port 'Last Read Line' to /comments


pushcx:
In 405156b1 I added a small feature so that [/newest](https://lobste.rs/newest) has a line marking where the logged-in user last loaded `/newest`.

Response has been positive (and it's easy to adblock for the folks who don't like it), so I'd like to copy over the functionality to the equivalent page for comments, [/comments](https://lobste.rs/comments).

There's also a small bug to fix in `last_read_newest`; `HomeController#newest` calls `touch` even if the 'last read line' isn't shown. So the user Alice reads `/newest` and leaves for a few days and 30 stories are submitted, so the line should appear 5 spots down on page 2. Alice loads `/newest` and the line isn't visible because all stories are new. She clicks to page 2 and the line does not appear because loading page 1 already touched the timestamp. Maybe this calls for a helper to share the `@user&.last_read_newest&.after?(story.created_at)` logic from `app/views/home/index.html.erb`, I think if it returned the new value for `already_printed_newest_line`, the controller would know to touch the value and view would know to not print it twice.

pushcx:
I misdesigned this feature trying to fix the bug about updating the 'last read' line in the second paragraph of my opening post on this issue. So I'm reopening this for one more fix.

The issue is that if someone doesn't read for a few days, the 'last read' line will fall back a few pages. This will happen much faster with comments (~250/day) than stories (~20/day). If the user doesn't go through enough pages to find their 'last read' line, it'll _never_ get updated, sinking deeper and deeper into the pages. The feature will effectively stop working for them.

To fix it, we'll have to change the logic again. The 'last read' line should always be updated when the user loads `/newest` or `/comments`. If the 'last read' line appears on that first page, great, everything worked. But if it didn't, we need to update the pagination link to include a `last_read_timestamp` with their old timestamp. Then on page 2 or 3 or 300, if they reach that timestamp, display the line and drop the parameter. If they never hit click back that far, that's fine, loading `/newest` will show it again as expected. So basically there's some extra state to maintain through pagination.

PR #1476, opened 2025-02-22
Fix 'last_read_newest' and 'Last Read' marker 

kernal053:
#1394 Fixed second part of issue related to `last_read_newest` and` Last Read` marker.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -77,6 +77,7 @@ class HomeController < ApplicationController
 
     @title = "Newest Stories"
     @above = {partial: "stories/subnav"}
+    @last_read_story = find_last_read_story
 
     @rss_link = {
       title: "RSS 2.0 - Newest Items",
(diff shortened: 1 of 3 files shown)


pushcx (changes requested):
Thanks for taking on this bug. I just fixed the build, sorry for that distraction. So if you merge in master you should have a green build to work from.

I really like the way your `find_last_read_story` lifts that bit of logic up from the view, good strategy. Can you make it loop `@stories` rather than query the database again? We have the data locally so I don't want the small time cost to query the database again, but also there's a race condition: if the newest story is the last one on a page that the user loads, another user could submit a story in the milliseconds between this loading `@stories` and then calling `Story.find`, so the line would not get updated correctly.

kernal053:
@pushcx I'm actually using already loaded `@stories `collection and not querying database again. I hope I'm not missing anything here.

pushcx:
Oh yep, I didn't see that, I think misread `@stories.find` as `Story.first` because I see `.find` used so infrequently. Thanks for explaining.

Time to merge and deploy!

PR #1480, opened 2025-03-04
Add 'Last Read' marker to /comments 

### LINKED ISSUE #1394: Port 'Last Read Line' to /comments

kernal053:
Fix #1394

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -288,6 +288,7 @@ class CommentsController < ApplicationController
       c.current_vote = @votes[c.id]
       c.vote_summary = summaries[c.id]
     end
+    @last_read_comment = find_last_read_comment
 
     respond_to do |format|
       format.html { render action: "index" }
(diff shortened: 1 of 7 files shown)


pushcx:
I see how this is a really close parallel to the feature on /newest, so I have no suggestions or changes. This looks perfect, thank you for implementing it.

I've realized I misdesigned the feature in #1394 so I'll post there about some follow-up work.

PR #1483, opened 2025-03-06
Add support for OTP recovery codes

### LINKED ISSUE #342: It's easy to lose access to two factor authentication.
When my phone is formatted, stolen, or disposed of without registering TOTP to another phone, I'd be locked out of my account.

lobste.rs needs one of the following.

1. Show TOTP secret and tell users to save the TOTP secret along with password
2. Dispense one-time backup codes to anyone who turns on two factor authentication

Without both options, many users will find themselves locked out of their lobste.rs accounts after changing their phones without registering TOTP to another phone.

hmadison:
Resolves #342 

I've added basic support for generating recovery codes when a user sets up a TOTP device.

### DIFF
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -203,10 +203,11 @@ class LoginController < ApplicationController
       session.delete(:twofa_u)
       redirect_to "/"
     elsif (tmpu = find_twofa_user) && tmpu.authenticate_totp_recovery(params[:totp_code])
-      flash[:error] = "You used a OTP recovery code to log in. Please update your OTP device to ensure continued access to your account."
+      flash[:error] = "You used a OTP recovery code to log in and as such we have disabled MFA. Please add a new OTP device to your ensure continued multifactor authentication for your account."
 
       session[:u] = tmpu.session_token
       session.delete(:twofa_u)
+      tmpu.disable_2fa
 
       redirect_to "/settings/2fa"
     else
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -202,6 +202,13 @@ class LoginController < ApplicationController
       session[:u] = tmpu.session_token
       session.delete(:twofa_u)
       redirect_to "/"
+    elsif (tmpu = find_twofa_user) && tmpu.authenticate_totp_recovery(params[:totp_code])
+      flash[:error] = "You used a OTP recovery code to log in. Please update your OTP device to ensure continued access to your account."
+
+      session[:u] = tmpu.session_token
+      session.delete(:twofa_u)
+      
+      redirect_to "/settings/2fa"
     else
       flash[:error] = "Your TOTP code did not match.  Please try again."
       redirect_to "/login/2fa"
--- a/app/controllers/login_controller.rb
+++ b/app/controllers/login_controller.rb
@@ -207,7 +207,7 @@ class LoginController < ApplicationController
 
       session[:u] = tmpu.session_token
       session.delete(:twofa_u)
-      
+
       redirect_to "/settings/2fa"
     else
       flash[:error] = "Your TOTP code did not match.  Please try again."
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -127,13 +127,14 @@ class SettingsController < ApplicationController
 
     @user.totp_secret = session[:totp_secret]
     if @user.authenticate_totp(params[:totp_code])
+      @user.totp_recovery_codes = Array.new(3).map{ |_| Utils.random_str(20) }
       # re-roll, just in case
       @user.session_token = nil
       @user.save!
 
       session[:u] = @user.session_token
 
-      flash[:success] = "Two-Factor Authentication has been enabled on your account."
+      flash[:success] = "Two-Factor Authentication has been enabled on your account. Your recovery codes are #{@user.totp_recovery_codes.join(', ')}. They can be used to access your account in the event that you loose access to your TOTP device. Store these in a safe place, "
       session.delete(:totp_secret)
       redirect_to "/settings"
     else
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -127,7 +127,7 @@ class SettingsController < ApplicationController
 
     @user.totp_secret = session[:totp_secret]
     if @user.authenticate_totp(params[:totp_code])
-      @user.totp_recovery_codes = Array.new(3).map { |_| Utils.random_str(20) }
+      @user.totp_recovery_codes = Array.new(6) { |_| Utils.random_str(20) }
       # re-roll, just in case
       @user.session_token = nil
       @user.save!
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -134,7 +134,7 @@ class SettingsController < ApplicationController
 
       session[:u] = @user.session_token
 
-      flash[:success] = "Two-Factor Authentication has been enabled on your account. Your recovery codes are #{@user.totp_recovery_codes.join(", ")}. They can be used to access your account in the event that you loose access to your TOTP device. Store these in a safe place, "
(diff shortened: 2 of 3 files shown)


IreneKnapp:
This looks solid. Does it have brute-force protection? That is, is there a layer that will rate-limit attackers? I don't see such a feature in this PR but we might have it in the existing code?

app/controllers/settings_controller.rb
```diff
@@ -127,13 +127,14 @@ def twofa_update
 
     @user.totp_secret = session[:totp_secret]
     if @user.authenticate_totp(params[:totp_code])
+      @user.totp_recovery_codes = Array.new(3).map{ |_| Utils.random_str(20) }
```
IreneKnapp, line 130: good, this seems like adequate entropy to me :)

I think it would be helpful to have something more like 6 codes rather than 3. of course the normal path is to only ever use a single one, then regenerate them all, but I've been in situations where things were too urgent to do that right away, for example because I was dealing with a real-world emergency of some kind that happened to include losing access to stuff as one of its elements. In those times it was reassuring to know I wasn't going to lock myself out of the recovery process by accidentally churning through all of them.

app/controllers/login_controller.rb
```diff
@@ -202,6 +202,13 @@ def twofa_verify
       session[:u] = tmpu.session_token
       session.delete(:twofa_u)
       redirect_to "/"
+    elsif (tmpu = find_twofa_user) && tmpu.authenticate_totp_recovery(params[:totp_code])
+      flash[:error] = "You used a OTP recovery code to log in. Please update your OTP device to ensure continued access to your account."
```
pushcx, line 206: Can we give more-specific instructions? Should using a recovery code disable 2fa so they can re-enable it with a new secret?

app/controllers/settings_controller.rb
```diff
@@ -127,13 +127,14 @@ def twofa_update
...
     if @user.authenticate_totp(params[:totp_code])
+      @user.totp_recovery_codes = Array.new(3).map { |_| Utils.random_str(20) }
       # re-roll, just in case
       @user.session_token = nil
       @user.save!
 
       session[:u] = @user.session_token
 
-      flash[:success] = "Two-Factor Authentication has been enabled on your account."
+      flash[:success] = "Two-Factor Authentication has been enabled on your account. Your recovery codes are #{@user.totp_recovery_codes.join(", ")}. They can be used to access your account in the event that you loose access to your TOTP device. Store these in a safe place, "
```
pushcx, line 137:
```suggestion
      flash[:success] = "Two-Factor Authentication has been enabled on your account. Your recovery codes are #{@user.totp_recovery_codes.join(", ")}. They can be used to access your account in the event that you lose access to your TOTP device. Store these in a safe place, separate from your TOTP device."
```

pushcx (changes requested):
Thanks for taking on such a well-aged feature request. This just needs some small copy tweaks.

The 2fa form needs an explanation that a recovery code can be used in place of the generated TOTP code.

Please add the name of this field to `config/initializers/filter_parameter_logging.rb`. It should never appear in controller params or other things likely to be logged, but better to be thorough.

app/controllers/settings_controller.rb
pushcx: Should we also datestamp them, like `20250310_#{entropy}`? I can see that enabling the flow will regenerate all the existing ones, so it would be nice debugging info for the user who finds old codes, and for the admin (me) if someone tells me their code isn't working.

app/models/user.rb
```diff
@@ -83,6 +83,7 @@ class User < ApplicationRecord
     s.string :mastodon_username
     s.any :keybase_signatures, array: true
     s.string :homepage
+    s.string :totp_recovery_codes, array: true, default: [], null: false
```
pushcx, line 86: I'm glad to have the `null: false`. Can you confirm that we don't need a migration? In a rails console, if you load a user, they'll automatically have an empty array? I don't want `authenticater_totp_recovery` to throw a `NoMethodError` on `nil`. Otherwise this will need a migration.

app/controllers/login_controller.rb
winterqt: > Should using a recovery code disable 2fa so they can re-enable it with a new secret?

I think this is the more ideal solution.

app/controllers/settings_controller.rb
```diff
@@ -127,13 +127,14 @@ def twofa_update
 
     @user.totp_secret = session[:totp_secret]
     if @user.authenticate_totp(params[:totp_code])
+      @user.totp_recovery_codes = Array.new(3).map { |_| Utils.random_str(20) }
```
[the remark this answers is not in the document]
pushcx: I leave this kind of style issue up to the linter. If standardrb is happy with it, I'm happy with it.

app/controllers/login_controller.rb
pushcx: If using a recovery key disables 2fa, and enabling 2fa replaces all the recovery codes, it doesn't seem like there's a benefit to generating more than one. A user could put store their different codes in different backups, but there's no benefit to each having a unique code vs duplicating one code.

This morning GitHub hassled me to view my 16 recovery codes and confirm I had them backed up. They must not disable 2fa on the use of a recovery code, right?

I think my guess about disabling a recovery code was wrong. I don't know what best practices are here and I'm probably not going to reinvent the wheel correctly. @hmadison do you have any kind of guidance we can use here?

pushcx:
I was looking for an answer to [these questions](https://github.com/lobsters/lobsters/pull/1483#discussion_r2027110763) before merging because I don't understand the design. Why have multiple codes if you can only ever use one?

app/models/user.rb
```diff
@@ -231,6 +232,13 @@ def as_json(_options = {})
     h
   end
 
+  def authenticate_totp_recovery(code)
+    return false unless totp_recovery_codes.include?(code)
```
GrayHatter, line 236: This feels like a really bad side channel leakage. Does ruby have constant time compare for `exact` instead of `contains`?

pushcx: [`fixed_length_secure_compare`](https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html)

That'd mean rewriting to `.any?` to run a block which I know does an early exit. I guess that doesn't matter because the attacker would have to know a valid token to see the timing difference.

GrayHatter: *Technically* it still does matter because it creates an oracle, but in this case I'm confident that the info leak is not meaningful.

(Sorry for the duplicate comment, I mistakenly assumed an email reply would reply inline instead of the root thread. *sigh*)

app/controllers/login_controller.rb
hmadison: The best guidance I can find is https://pages.nist.gov/800-63-4/sp800-63b/events/#recovery

GrayHatter: A recovery key isn't the same thing as a 2nd factor "backup code". I'm assuming because you're generating 6, that it's in the group of static (non-time based) 2fa tokens. These can be used when you don't have access to the otp secret transiently (what I assume github is doing), or to provide a 2nd factor when the otp secret is believed to be unrecoverable. If it's exclusively the latter, then you don't need 6, you should only generate one, and those details about account recovery are more relevant. The account recovery case is a lot harder to implement correctly, so my vote would be not modifying the user account at all (other than deleting the used code), and allowing the user to decide what to do next.

pushcx:
Hi stalebot. I'm still puzzled by the security design: what's the best practice for having multiple codes and deleting them on use? Can we get an authoritative answer from an expert source?

I'm nervous about misdesigning something we're asking people save indefinitely; we can't write a Rails migration for a note in the back of someone's filing cabinet.

chamlis:
To add a datapoint, I've been integrating 2FA at work using [Laravel Fortify](https://laravel.com/docs/12.x/fortify), their all-in-one authentication library. That [replaces the individual recovery code on use](https://github.com/laravel/fortify/blob/1.x/src/Http/Controllers/TwoFactorAuthenticatedSessionController.php#L61) with no option to disable this.

PR #1486, opened 2025-03-10
fix indentation in _commentbox.html.erb

roryokane:


pushcx:
Thanks for catching this, happy to have a little style fix.

Context: this came up on [today's stream](https://push.cx/stream/2025-03-10) when Rory mentioned the bug I fixed in 757823ec.

Issue #1488, opened 2025-03-11
Editing a comment is broken


kaycebasques:
https://github.com/user-attachments/assets/c7cafec0-a4d5-4625-86d4-86da4ea7a455

![Image](https://github.com/user-attachments/assets/e2194145-5070-41e1-a237-10bc488b426f)

pushcx:
Most likely a continuation of #1485

roryokane:
## Bug when starting an edit to a comment

This seems likely to be related to https://github.com/lobsters/lobsters/commit/b122213f2b99c922fd86fc50cf5c1a7cb19d0eb4. That change makes the edit page loadable standalone. What I am seeing on desktop Chrome is that after clicking “edit”, the edit form loads, but it includes the header and footer of the page as well.

### Before clicking “edit”

I cropped this screenshot to the bottom of the page. The page looks normal.

> <img width="573.75" alt="normal view of a comment tree" src="https://github.com/user-attachments/assets/b4af8600-20e7-4940-addb-2e5b385fdc9b" />

### After clicking “edit”

The extra header and footer are unexpected.

> <img width="566.25" alt="extra header and footer around comment edit form" src="https://github.com/user-attachments/assets/e29d0630-1da1-4c78-9754-441293014aea" />

## Bug after saving an edit to a comment

I had also previously noticed that after filling out this garbled edit form and clicking Update, the end state is that the editable comment is successfully replaced with a rendered version of the new comment, but it has a copy of the site’s header and footer below it.

> <img width="573.75" alt="edit comment with header and footer below it" src="https://github.com/user-attachments/assets/e8f40097-ea00-4f98-bc01-baa98ff61044" />

In my browser’s dev tools, the extra header and footer are within a `<li class="comments_subtree">` that contains elements as if it were a standalone page. Between the `header` and `footer` of that element tree are these elements:

```html
<div id="inside">
  <ol class="comments comments1">
    <li class="comments_subtree">
      <ol class="comments">
      </ol>
    </li>
  </ol>
</div>
```

Those elements are not visible, which is why only the header and footer appear to have been added.

roryokane:
I noticed another bug that is probably caused by the same issue. After saving an edit to a comment, the edited comment is rendered at the top of the comments section, even though I was editing a comment deep in a subthread under the first comment.

[...]

PR #1489, opened 2025-03-12
Fix comment editing to work with and without JS

### LINKED ISSUE #1487: standalone comment editing doesn't render comment
In b122213f I made it possible to click the `edit` link on comments without javascript and get a standalone comment editing form. (Part of my long-running work to make js optional even for logged-in users.) This form works to edit a comment text, but has two bugs.

1. If you load the form with js enabled (just middle-click an edit link to get a new tab), clicking 'update' does update the comment but renders a page with layout and no comment text. This should redirect back to the single-story view with the comment fragment: `/s/qwertyu#c_asdf12`
2. If you don't have js enabled, saving the comment should also redirect. Looking at `CommentsController#update` it doesn't have a check of `request.xhr?` to redirect if the user doesn't have js enabled.

chamlis:
The first commit here fixes b122213f2b99c922fd86fc50cf5c1a7cb19d0eb4 to correctly render the edit box with JavaScript enabled. Using `fetch` doesn't set the `X-Requested-By` header, so `request.xhr?` will return false. Instead of adding this header manually I've switched to using `fetchWithCSRF` which already added it, for reasons I'll explain at the end.

The second commit fixes #1487 by choosing to either send a redirect or a partial in response to an update request. Note that the fetch API does not allow introspecting on 302 responses, either omitting them entirely or (with `redirect: 'manual'`) redacting them. Therefore, I introduced an `X-Location` header which can be sent in any response to be handled as a redirect by the JavaScript. At present this is sent in another code path, there may be some Rails magic to handle this automatically when a normal redirect is sent from an XHR request.

I introduced handling for this in `fetchWithCSRF`. In sites where I've used progressive enhancement I've often found it useful to have a `fetch` wrapper that "does the right thing" with typical web framework responses.

Please note this is the first Rails (and Ruby!) I've ever written - here be dragons!

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -104,7 +104,7 @@ class CommentsController < ApplicationController
       render partial: "commentbox", layout: false,
         content_type: "text/html", locals: {comment: comment}
     else
-      render "_commentbox", locals: {comment: comment, parents: comment.parents}
+      render "_commentbox", locals: {comment: comment, parents: comment.parents, redirect_on_submit: true}
     end
   end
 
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks for investigating and offering a fix. This looks like a good approach to expressing the desired behavior. I think this should handle it and I don't want to leave comment editing broken, so I'm going to merge it now.

pushcx:
mentioning #1488

PR #1498, opened 2025-03-24
Clarify participation/self-promo text wording

strugee:
It took me literally completely finishing a bug report to realize that this text meant stories marked as "I am the author of the story at this URL", _not_, stories submitted by you (i.e. "_submissions_ authored by you").

So, try and make this text a bit more unambiguous.

This wording probably isn't the best. Feel free to bikeshed and feel free to force-push to my branch. I just went to file an (alternate) issue and then realized it was super easy to just at least propose a strawman.

This was a drive-by GitHub web UI patch and I did not test it at all. It's trivial so I'm assuming CI will take care of it.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -172,7 +172,7 @@
     <div class="user-stats flash-notice">
       <p>
         Comments posted in the last 6 months: <%= @showing_user.comments.recent.count %><br>
-        &nbsp;&nbsp;...posted on stories you didn't author: <%= @showing_user.comments.recent.on_stories_not_authored_by(@showing_user).count %><br>
+        &nbsp;&nbsp;...posted on stories you didn't mark as self-authored: <%= @showing_user.comments.recent.on_stories_not_authored_by(@showing_user).count %><br>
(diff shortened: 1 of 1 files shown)


pushcx:
We've never described submissions as having authors, just submitters, so I'm at a bit of a loss where that misunderstanding came from. This suggestion unfortunately makes it the text less accurate. We don't track authorship of blogs (or Domains, or Origins), only whether the submitter checked the box.

strugee:
> We've never described submissions as having authors, just submitters, so I'm at a bit of a loss where that misunderstanding came from.

Now that I've figured it out, it makes sense. The problem was essentially that I read "authored" as a generic English word and not a Lobsters term of art with a precise definition in code.

Can you clarify what exactly is less accurate? The proposed text references the checkbox much more directly than the current text (though still not directly, because I wanted to keep it brief). To be honest I'm having a hard time figuring out where you're coming from unless you're reading the diff backwards.

pushcx:
I read the diff backwards somehow. Huh. Sorry for the confusion!

PR #1499, opened 2025-03-24
Use a minus sign (U+2212) instead of hyphen to collapse

pb-:
With a regular hyphen (U+002D) the `[-]` marker uses slightly less space than `[+]` in most standard fonts, resulting in unsatisfying aesthetics especially when toggling the state.

Instead, use a minus character that has the same with as the plus character.

Before
2025-03-24-201355_100x115_scrot

After
2025-03-24-201347_98x117_scrot

(I'm not sure which variant is preferred - encoding the character as UTF-8 and declaring the use of UTF-8 as in this PR or the slightly more conservative solution of escaping the string. Happy to switch it.)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /* common color scheme */
 
 :root {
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks, this is a nice little improvement to UI. And I'm glad to have an explicit charset declared on the CSS, so that's two improvements that I'm very happy to merge.

PR #1504, opened 2025-03-27
Fix display of story _singledetail

### LINKED ISSUE #1502: Bug: Missing comments count on any disccusion when no logger in
This happens when you are not logged in.

In an old discussion from 2 weeks ago that I have open everything is fine (I have the caches button, I see the number of comments and, if there are any, I see the off-topic votes (if any, in the example one there are none)). But now, when I open the same discussion I don't have any of this. 

Old one:
`{{picture}} via {{user}} {{timestamp}} | cache button | {{comments}} | {votes}}`
Image

Now:
`via {{picture}} via {{user}} {{timestamp}}`
Image
In addition to removing everything after the timestamp, it also changes the order of the user's picture to after the `via/authored`.

chamlis:
The first commit fixes displaying a duplicated comment count for unmerged stories (one a sum of all the merged stories, one the comment count on the story itself). We still render only a total comment count for merged stories.

The second commit fixes some unclosed control flow that was skipping more of the template for logged-out users than intended. I've checked this over a few times and I think it's correct now, but it's probably worth another skim over.

Closes #1502

### DIFF
--- a/app/views/stories/_singledetail.html.erb
+++ b/app/views/stories/_singledetail.html.erb
@@ -96,6 +96,7 @@ classes = [
             <span> | </span><%= link_to "unhide", story_unhide_path(ms.short_id), :class => "hider" %>
           <% else %>
             <span> | </span><%= link_to "hide", story_hide_path(ms.short_id), :class => "hider" %>
+          <% end %>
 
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for figuring this out, you're right on both fixes here.

PR #1507, opened 2025-03-29
Use the correct classes on the title of user profiles

chamlis:
These classes don't impact the styling, so are possibly candidates for removal altogether.


pushcx:
Yep, that's a real obvious bug, thank you!

Issue #1284, opened 2024-06-04
Default font size on small screen mobile phones is quite small


knl:
The default font size seems quite small on mobile phones with small screens, like iPhone SE 2nd generation, especially for the older population. Would it be possible to make it bigger or to follow the system settings?

note how in this screenshot the url text is quite bigger than the comment body text.

![IMG_3888](https://github.com/lobsters/lobsters/assets/361496/3e1ba3fc-ce75-43e0-9c5f-d0d8e2eb06e4)

pushcx:
Similar to #642

MatheusRich:
IMO, even in desktop mode, it is small. I can zoom in on my laptop at least.

kconner:
A larger default size for phones would be great.

There is an immediate remedy available to Mobile Safari users, which is to bump the browser text size site-wide. This is available from the page controls button on the left side of the location bar. I personally forget it exists, but it is a no-code solution.

An option we could consider is to apply the user's system text size setting, which is at least possible on iPhone. [Craig Hockenberry explains how to use it in CSS](https://furbo.org/2024/07/04/dynamic-type-on-the-web/). Android also has such a setting, so I wonder if browsers can offer that to pages or if they do it automatically. This would be really nice, but only after bringing text size in line with common body text.

Disclaimer: I set my system text size up to 120%, so lots of sites' text looks small to me. However, text on Lobsters looks smaller than most.

pushcx:
This [article on accessible fonts](https://accessibe.com/blog/knowledgebase/ada-compliant-fonts#character-size) recommends "A safe bet for a WCAG-conforming font size is 16px (or [1em](https://fonts.google.com/knowledge/glossary/em) or [1rem](https://blog.hubspot.com/website/css-rem))". (thanks to gj1118 on [today's office hours stream](https://push.cx/stream/2025-03-24))

fleebeee:
> The most important text on the site is the comment text, so we should pick a good size for that and pick every other size as a variation on it (headlines, bylines, header/footer, etc).

I believe the most accessible and user-friendly option is to let the user dictate their preferred font size. As long as `font-size` isn't overwritten in the root element, rem units respect the browser setting for default font size. Comment text should probably have a size of 1rem while other text elements could have some other multiple of rem.

This [article on CSS units](https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/) does a good job of explaining the issue, and it also presents some solutions to help adapt to rem units.

[...]

PR #1508, opened 2025-03-29
Begin porting pages to CSS Grid

chamlis:
I've started with the user profile page, but I plan to gradually cover all of the pages in this branch. Alongside using CSS grid, I've aimed to stop depending on `line-height` and `<br>` for page layout. The avatar is still floated at the moment, though given the grid cannot wrap around it with the current setup, that could be changed without affecting the (new) design.

I'm no designer, but I tried to make things space out more logically. For instance, in the mobile single-column view, there is more space between a heading and the previous item than between a heading and its item. I'm very much open to feedback both on the visual design, and the CSS implementation.

As well as this not covering every page yet, I'm still not 100% happy with the new profile page (hence the draft PR, and CSS class names with `todo` in them!).

This relates to, but does not close, issue #1279

## Desktop

<details>
<summary>Before</summary>

desktop_before
</details>

<details>
<summary>After</summary>

desktop_after
</details>

## Mobile

<details>
<summary>Before</summary>

mobile_before
</details>

<details>
<summary>After</summary>

mobile_after
</details>

## Views to convert

<details>
<summary>Long list...</summary>

- [x] `app/views/about/404.html.erb`
- [ ] `app/views/categories/edit.html.erb`
- [ ] `app/views/categories/_form.html.erb`
- [ ] `app/views/categories/new.html.erb`
- [ ] `app/views/domains/edit.html.erb`
- [ ] `app/views/filters/index.html.erb`
- [ ] `app/views/hat_requests/index.html.erb`
- [ ] `app/views/hat_requests/new.html.erb`
- [ ] `app/views/hats/doff.html.erb`
- [ ] `app/views/hats/index.html.erb`
- [x] `app/views/home/_category.html.erb`
- [x] `app/views/home/_multi_tag.html.erb`
- [x] `app/views/home/_single_tag.html.erb`
- [ ] `app/views/invitations/build.html.erb`
- [ ] `app/views/keybase_proofs/new.html.erb`
- [ ] `app/views/login/forgot_password.html.erb`
- [ ] `app/views/login/index.html.erb`
- [ ] `app/views/login/set_new_password.html.erb`
- [ ] `app/views/login/twofa.html.erb`
- [ ] `app/views/messages/_form.html.erb`
- [ ] `app/views/messages/index.html.erb`
- [ ] `app/views/messages/show.html.erb`
- [ ] `app/views/moderations/index.html.erb`
- [x] `app/views/mod/index.html.erb`
- [ ] `app/views/mod_notes/index.html.erb`
- [ ] `app/views/mod/reparents/new.html.erb`
- [ ] `app/views/mod/stories/edit.html.erb`
- [ ] `app/views/origins/edit.html.erb`
- [ ] `app/views/search/index.html.erb`
- [x] `app/views/settings/index.html.erb`
- [ ] `app/views/settings/mastodon_authentication.html.erb`
- [x] `app/views/settings/twofa_enroll.html.erb`
- [ ] `app/views/settings/twofa.html.erb`
- [ ] `app/views/settings/twofa_verify.html.erb`
- [x] `app/views/signup/index.html.erb`
- [x] `app/views/signup/invite.html.erb`
- [ ] `app/views/signup/invited.html.erb`
- [x] `app/views/stats/index.html.erb`
- [ ] `app/views/stories/edit.html.erb`
- [ ] `app/views/stories/_form.html.erb`
- [ ] `app/views/stories/_listdetail.html.erb`
- [x] `app/views/stories/_missing.html.erb`
- [ ] `app/views/stories/new.html.erb`
- [ ] `app/views/stories/show.html.erb`
- [ ] `app/views/stories/_singledetail.html.erb`
- [ ] `app/views/suggestions/new.html.erb`
- [ ] `app/views/tags/edit.html.erb`
- [ ] `app/views/tags/_form.html.erb`
- [x] `app/views/tags/index.html.erb`
- [ ] `app/views/tags/new.html.erb`
- [x] `app/views/users/list.html.erb`
- [x] `app/views/users/not_found.html.erb`
- [ ] `app/views/users/standing.html.erb`
- [x] `app/views/users/tree.html.erb`
- [x] `app/views/users/_invitationform.html.erb`
</details>

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -393,6 +393,16 @@ header .inbox_unread {
 	grid-column: main / end;
 }
 
+#inside:has(ol.stories) {
+	display: grid;
+	grid-column: begin / end;
+	grid-template-columns: subgrid;
+}
+
+#inside:has(ol.stories) > * {
+	grid-column: main / end;
+}
+
 /* footer */
 
 footer {
@@ -429,6 +439,12 @@ div#gravatar img {
 
 /* stories */
 
+#inside ol.stories {
+	grid-column: begin / end;
+	display: grid;
+	grid-template-columns: subgrid;
+}
+
 ol.stories,
 ol.comments,
 ol.category_tags {
@@ -458,17 +474,17 @@ ol.comments1 > li.comments_subtree > ol.comments {
 }
 
 div.voters {
-	float: left;
 	margin-top: 0px;
-	text-align: center;
-	width: 40px;
+	justify-self: center;
+	display: flex;
+	flex-direction: column;
+	align-items: center;
 }
 
 div.voters div.score {
 	color: var(--color-fg-contrast-4-5);
 	font-size: 9pt;
-	margin-top: 1px;
-	margin-bottom: -3px;
+	text-align: center;
 }
 
 li.story div.voters div.score {
@@ -479,14 +495,16 @@ li.story div.voters div.score {
 div.voters .upvoter {
 	border-color: transparent transparent var(--color-fg-shape) transparent;
 	border-style: solid;
-	border-width: 6px;
+	border-width: 0 6px 11px;
 	text-decoration: none;
 	width: 0px;
 	height: 0;
-	margin-bottom: 0px;
-	margin-left: 14px;
 	padding: 0;
 	display: block;
+}
+
+.story div.voters .upvoter {
+	margin-top: 25%;
 }
 
 div.voters .upvoter:hover,
@@ -494,17 +512,21 @@ div.voters .upvoter:hover,
 	border-bottom-color: var(--color-fg-accent);
 }
 
-div.voters .upvoter {
-	border-bottom-width: 11px;
-}
-
 li.story {
-	clear: both;
+	grid-column: begin / end;
+	display: grid;
+	grid-template-columns: subgrid;
 }
 ol.stories li.story div.story_liner {
 	padding-top: 0.25em;
 	padding-bottom: 0.25em;
 	word-break: break-word;
+	grid-column: begin / end;
+	display: grid;
+	grid-template-columns: subgrid;
+}
+ol.stories li.story div.story_liner .details {
+	grid-column: main / end;
 }
 .comment {
 	clear: both;
@@ -519,9 +541,6 @@ ol.stories li.story div.story_liner {
 
 ol.stories li:first-child div.story_liner {
 	padding-top: 0.5em;
-}
-li div.details {
-	padding-top: 0.1em;
 }
 
 /*
@@ -767,22 +786,15 @@ li.story.deleted a {
 	color: var(--color-fg-contrast-5) !important;
 }
 
-li div.details {
-	margin-left: 32px;
-}
 div.story_content {
 	margin-bottom: 1em;
 }
 ol.stories.list div.story_content {
 	color: var(--color-fg-contrast-6);
 	max-height: 2.6em;
-	margin: 0.25em 40px 0.25em 0;
+	max-width: 90ch;
 	overflow: hidden;
 	text-overflow: clip;
-}
-
-div.morelink {
-	float: left;
 }
 
 div.morelink,
@@ -819,7 +831,7 @@ div.page_link_buttons span {
 
 div.story_text {
 	margin-bottom: 1.5em;
-	max-width: 700px;
+	max-width: 90ch;
 	word-wrap: break-word;
 }
 div.story_text p {
@@ -1466,16 +1478,13 @@ input[type="submit"].link_post.pushover_button {
 	}
 
 	/* hang voters into left margin to center under logo */
-	ol.stories,
 	#inside > ol.comments {
 		/* .voters width + .voters padding-right */
 		margin-left: -32px;
 	}
-	ol.stories div.voters,
 	#inside > ol.comments div.voters {
 		width: 30px;
 	}
-	ol.stories .upvoter,
 	#inside > ol.comments .upvoter {
 		margin-left: 9px;
 	}
@@ -1605,26 +1614,12 @@ input[type="submit"].link_post.pushover_button {
 		border-left-color: transparent;
 	}
 
-	div.voters {
-		margin-left: 0.25em;
-		margin-top: 0px;
-		width: 30px;
-	}
-	div.voters a.upvoter {
-		margin-top: -5px;
-	}
-
 	div.comment_form_container {
 		margin-left: 0;
 	}
 
 	div.comment_form_container form {
 		margin-left: 35px;
-	}
-
-	div.voters a.upvoter {
-		margin-left: 6px;
-		border-width: 9px;
 	}
 
 	/* explicitly reset color when not upvoted, since previously-upvoted arrow
@@ -1640,23 +1635,19 @@ input[type="submit"].link_post.pushover_button {
 		margin-top: 0;
 	}
 
-	ol.stories.list li.story {
-		display: table;
-	}
 	ol.stories.list li.story div.story_liner {
 		background-color: var(--color-mobile-story-liner-bg);
-		display: table-cell;
-		padding-top: 0.5em;
-		padding-bottom: 0.75em;
-		width: 100%;
+	}
+	ol.stories.list li.story div.story_liner:has(~ .mobile_comments) {
+		grid-column: begin / right;
 	}
 	ol.stories.list li.story .mobile_comments {
-		display: table-cell !important;
+		display: initial !important;
 		font-size: 11pt;
 		min-width: 40px;
 		text-align: center;
 		text-decoration: none;
-		vertical-align: middle;
+		align-self: center;
 	}
 	ol.stories.list li.story .mobile_comments span:before {
 		border-style: solid;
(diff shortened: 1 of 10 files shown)


chamlis:
I've given the user settings a shot, as one of the more complex pages. The most glaring issue is that the "about" textbox is incorrectly sized. As far as I can tell they're automatically sized by JavaScript, and I guess that doesn't play nice with grids.

I've also experimented with using [subgrids](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Subgrid) for this page. They're [not quite so well supported](https://caniuse.com/css-subgrid) as standard grids, so if that's an issue I'll either make sure things fall back reasonably or scrap them altogether. Essentially, they let us have a global grid for the entire page whose direct children don't necessarily follow that grid, but can pass its definition down to their children. This means we can put all the separate forms in one grid, so that we're guaranteed they line up all the way down the page. I've implemented this in [a kind of strange way](https://github.com/lobsters/lobsters/pull/1508/files#diff-5bfcdc54bcda018d14515b827c80d9f9b8ee25704a4271b560a2f87439665f17R1241-R1250), there may well be a more native way of accomplishing this.

pushcx:
This is a great start, thanks. We talked through this a bit in [today's stream](https://push.cx/stream/2025-04-03) for full details, but the big things are:

1. Thanks for taking on a big project that's the source of a lot of janky UI!
2. Maybe one PR per area? This is such a big project I don't want you to get overwhelmed; I find the feeling of incremental progress really satisfying.
3. There's also a need for a full-page grid to align the main page content, and it has one clever bit: stories and top-level comments hang a little into the left margin to align their voters and folders underneath the logo. This is brittle and predates css variables, so `div#inside` has a big comment about it.
4. The story form prompted the desire for CSS grid. It hasn't been possible to use `float` to keep the 'fetch title' button in the right place and the tags field the same width as the others across the test matrix of different browsers, desktop vs mobile, and js vs no-js.

chamlis:
With these most recent commits, I think I'm going to call this first grid PR feature complete.

The layout changes are:
- bringing grid to every page, laying out the header even if the rest of the page isn't affected
- using grid for lists of stories, lists of comments, and threads to line things up in the gutter and between parent and child comments
- using grid to lay out user profiles and settings, lining more things up and hopefully behaving more consistently

I need to clean it up by removing TODOs, fixing indentation where appropriate, and rebasing to give a more cohesive commit log, then I'll mark this ready for review.

chamlis:
While I suspect there'll be feedback and changes to make before this is ready to merge, it's now somewhere I'm mostly happy with.

I hope that isn't me somehow breaking loading the database schema.

pushcx:
We reviewed this PR on [today's office hours stream](https://push.cx/stream/2025-04-10). It's still a WIP, but what great progress it is! This cleans up a lot of brittle CSS, especially the most-complicated bits in the logo and voters. I'm really looking forward to merging this PR.

chamlis:
Thanks for the review, and the positive feedback!

I've brought in as many of the suggestions as I can recall:
- spacing around CSS compound selectors
- making greater use of the cascade to stop unnecessary redundant repetition
- reduce use of margins now that things are in the grid
- generally try to spot and get rid of styles that are irrelevant now

It'd be useful to have an automated way of detecting unused CSS. Chromium can measure coverage on one page, but I'd love something that could accumulate the coverage across multiple pages and screen sizes.

@pushcx also suggested placing the comment folder and upvote buttons in the same column to save on horizontal space - which is especially at a premium on mobile. I've pushed an experimental commit to that effect. I've opted to put the folder above the upvote button. Although, as @pushcx pointed out, we can view folding a comment as a slight negative response to it which implies the upvote should go on top, I feel we want the folder on top so that it doesn't jump around when folding and unfolding. Skeuomorphically, having the folder at the top fits my mental model of what's happening when I fold a comment, too.

<details>
<summary>Preview of a thread with upvoting and folding in the same column</summary>

fold_upvote
</details>

pushcx:
Keeping the folder in a consistent place does make a lot of sense, especially for misclicks.

I have an idea for an experiment about it. If the 'close' folder were below the score, it might work to animate it into the position the 'open' folder sits in (and vice-versa). Duration ~100ms so it's easy to see but doesn't feel slow. I'm not sure it's even possible, and even if possible it might still not feel as good.

chamlis:
For convenience, here's a recording of how animating the folding could look. The approach I've landed on for hiding the upvoter wouldn't work for hiding the tree line, as it depends on being able to set a reasonable max-height. There's probably a way of doing this that'd work for both, but I haven't found it yet.

[folding_animation.webm](https://github.com/user-attachments/assets/fa2f1243-55ac-4cd4-a0cc-0d934db8b65c)

chamlis:
I think I'm once again happy with where this branch is. The folder and upvoter now share a column with an animation, and everything seems to be working. I think to be able to do the animation properly, including the dotted line, will require [`interpolate-size`](https://caniuse.com/mdn-css_properties_interpolate-size) to be more widely supported.

The only issue I know about is that the dotted line does not extend past all of a comment's children. I was able to get this working by hoisting the gutter up a layer in the grid, but then that DOM restructuring broke just about all of the comment-related JavaScript, and I didn't want to add a third language that this PR makes sweeping changes in.

chamlis:
Rebased atop `master` to fix merge conflicts in the CSS.

pushcx:
There's a [new separator for threads on merged stories](https://github.com/lobsters/lobsters/blob/f99696992035c9c5c7564a447f694e44ea65ea54/app/views/stories/show.html.erb#L61-L107) that echoes `_listdetail` and `_singledetail`, [screenshot here](https://github.com/lobsters/lobsters/issues/1456#issuecomment-2803164684).

Sorry to add to the complexity of your project, but in yesterday's stream I got to a point where the design seemed like a really clear improvement for the UI.

chamlis:
Thanks for the preparatory commit, I've rebased again on top of that.

pushcx:
Great news! If you can make tomorrow's office hours stream, I'll do what you suggested by running it with prod data locally to make sure it looks reasonable with real text.

app/assets/stylesheets/application.css
```diff
@@ -430,37 +456,51 @@ ol.category_tags {
...
-}
 ol.comments1 > li.comments_subtree > ol.comments {
 	margin-left: 0;
 	padding-left: 0;
 }
 
+#inside .comment ~ ol.comments {
+	grid-column: 2;
+	display: grid;
+	grid-template-columns: minmax(1.5rem, max-content) 1fr;
```
pushcx, line 480:
```suggestion
	grid-template-columns: max-content 1fr;
```

pushcx:
We reviewed this on stream and found some small tweaks to make:

 * voters are a little too wide (diff suggested)
 * mobile: singledetail has too much padding on the right, probably from the story list layout where there's the text balloon with the comment count
 * mobile: top-level comment box is too narrow; right side is probably previous issue, left side is `div.comment_form_container form` has a `padding-left`
 * mobile: the inline (inserted) commentboxes are also narrow
 * mobile: /comments has the same excess padding on the right
 * comment folder lines at top-level are missing
 * comment folder lines at lower level are truncated (screenshot below)



Most importantly, this is a heroic lift with a ton of improvements to our layout and its maintainability.

pushcx:
We also talked a bit about #1284 and being willing to leave some vertical rhythm stuff a little odd, to be addressed in that issue.

chamlis:
Thanks for the useful feedback, as ever. I think I've addressed all of the bugs we found yesterday with this latest batch of commits. The meta-diff is smaller than it appears: I've switched to naming the main grid lines on the page which makes much of the layout easier to read.

I was able to fix the tree lines strictly in CSS, so no JavaScript woes to worry about that. I think at some point over this redesign the sibling lines became vestigial and their role fulfilled by the parent lines, so I've removed the former. If there's a glaring line-shaped hole somewhere I've missed, I'll add them back!

The only two minor things that annoy me is that the upvoter is not quite aligned with the comment text, and that the top-level's parent/sibling line's column is slightly wider than the others, as it also has the logo. I had a go at fixing both of these with some judicious `calc`, but the numbers I thought should've worked didn't.

Final thing: the comments themselves still have a `max-width` on them, purely to limit the line length for the aim of readability. We could drop that and really surface the main grid lines, but to my eye that didn't look as good.

<details>
<summary>Desktop Comments</summary>

desktop_fixes
</details>

<details>
<summary>Mobile Comments</summary>

mobile_fixes
</details>

app/assets/stylesheets/application.css
```diff
@@ -344,18 +342,20 @@ header#nav .navholder {
 }
 header#nav {
 	box-sizing: border-box;
-	display: flex;
-	padding: 0.5rem 0.5rem 0 0.5rem;
+	grid-column: 1 / -1;
```
pushcx, line 345:
```suggestion
	grid-column: begin / end;
```

app/assets/stylesheets/application.css
```diff
@@ -627,66 +676,42 @@ li .comment_folder.comment_folder_inline.force_inline {
...
 }
 
-li .comment_parent_tree_line.score_shown {
-	top: 44px;
-}
-li .comment_parent_tree_line.can_flag {
-	top: 56px;
-}
 /* hide for comments that don't have children */
 li:not(:has(> .comments > li)) > .comment > .comment_parent_tree_line {
```
pushcx, line 702: This selector has broken. This is a small feature of tree lines and my comment wasn't clear enough (so that should be rewritten, too!). If a comment is the last at its level of reply (or in other words, has no siblings after it), it shouldn't have a line, even if it has replies. The lines are for making siblings clear when the parent might have scrolled off screen, it's sort of assumed that the indent is enough for direct replies.

pushcx (changes requested):
I found one small feature that got lost because my comment was unclear.

chamlis:
Two good catches, (hopefully) two good patches.

pushcx:
Thank you for this big project and its thousand fiddly details! I'm deploying now.

PR #1510, opened 2025-04-03
fix(issue-1505): story form url field retention

im-sakshijain:
[AwesomeScreenshot-3_4_2025,11_19_22pm.webm](https://github.com/user-attachments/assets/87c31f11-e77b-4b6a-9216-64c85fc61bf1)


pushcx:
If the "security" part of the build fails, that's a bug we reported in brakeman that'll be in their next release.

PR #1514, opened 2025-04-08
Add fake story and comment upvotes in fake_data

chamlis:
I've found having a variety of vote counts useful while working on layout, but they weren't generated as part of `fake_data`.

This is by far the slowest part of the fake data generation, I suspect because of the repeated sampling of users. Perhaps bulk inserting votes or only processing a subset of stories/comments would be viable.

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -174,11 +174,14 @@ class FakeDataGenerator
     end
     puts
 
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for improving this script, I'm glad to have it generate more robust data. I'm happy to see it run faster in the future, but that's not worth holding up this PR for.

Issue #1513, opened 2025-04-05
Story comments page missing hyperlink on "N comments" text


ConnorKrammer:
When you click into a story's comments page, the text that either reads "no comments" or "`N` comments" no longer appears to be a hyperlink to that page. If you check the page source, there's an unmatched closing link tag (`</a>`), but no corresponding open tag. I thought this was a pretty recent change, so I poked through the recent commit history and believe I've narrowed it down to 426bce96ffb441f5ef38c376e8bac1bf32d30b41.

The class to look for in the source is `comments_label`, in `_singledetail.html.erb`. Here's the opening tag that I think got lost during the revision:
```html
<a role="heading" aria-level="2" href="<%= story.comments_path %>">
```

I was logged out when I noticed this, and don't have an account to check whether it's present when logged in. Glancing at the code, I _suspect_ it's present either way.

Hope this is helpful!

UditThakkar:
@pushcx 
Let me know if needs to be fixed, I can pick this up and fix it.

pushcx:
Yes, this is a bug and I'd appreciate a PR for it. Thanks for the report and for the volunteering!

UditThakkar:
Let me create a PR for fixing this.

UditThakkar:
Hey @pushcx 
I checked the code and found that we were redirecting to story.comments_path.
After investigating further, I found that the story.comments_path is the same as the story show route [link](https://github.com/lobsters/lobsters/blob/master/app/models/story.rb#L526-L528).

@ConnorKrammer Just to make sure I am referring to the same hyperlink from the issue please refer to the screenshot below. 
If I am going in the right direction clicking on the comments hyperlink will take you to the same page.
![Image](https://github.com/user-attachments/assets/c3ad8a0c-4a3b-4137-bcea-13afc1bf734f).

ConnorKrammer:
@UditThakkar Yep, that's the one. And yes, that link points to its own page; clicking it doesn't move you anywhere. It's for copying, opening in a new tab, etc.

UditThakkar:
@ConnorKrammer Thanks for confirming, I will create a PR for this in some time.

pushcx:
To link @ConnorKrammer's [comment from the PR](https://github.com/lobsters/lobsters/pull/1515#issuecomment-2796055827), there's more to fix here.

UditThakkar:
Sorry for the delay, let me add this fix in a separate PR tomorrow.
Thanks

[...]

PR #1515, opened 2025-04-09
fix: add missing a tag in story comments page

UditThakkar:
This PR is to fix the issue reported here: https://github.com/lobsters/lobsters/issues/1513#issuecomment-2788274255


pushcx:
Thanks for fixing my error, I appreciate it!

PR #1518, opened 2025-04-11
Install Prism 1.4.0 ourselves

chamlis:
This fixes errors from Brakeman, following [this suggestion](https://github.com/presidentbeef/brakeman/issues/1927#issuecomment-2794239272) from @presidentbeef.


pushcx:
Oh great, I assumed it was something compiled into Ruby rather than a gem in the stdlib. Thanks for figuring this out, it's nice to see a green build again!

PR #1520, opened 2025-04-12
Close <ul>s appropriately on /about

chamlis:
Before, the "trivia" section was considered a list item of "technical features" and indented accordingly.


pushcx:
Thanks for spotting this and fixing it, always nice to fix these things up.

PR #1527, opened 2025-04-18
move stats page graphs to cache to prevent 502 #1526

### LINKED ISSUE #1526: /stats is not loading, move queries to jobs
`/stats` is timing out, the couple queries to load the chart data are taking long enough that the request times out and users get a 502.

We've configured ActiveJob, so the obvious fix is to move these slow queries to individual daily jobs that fill the cache, then have the page read the latest cache (and include a "please report this bug" line if the cache is more than 24 hours + 5 minutes old).

It's also possible that this failure is contributing to our new oomkills. It's loading a lot of data and formatting it into svg, which is an xml-based image format.

Dogacel:
## What

Fixes #1526 by moving stats graph calculation from controller to a nightly job.

## Changes

1. Move graph generator functions from `stats_controller` to `stats` model.
2. Prevent controller from invoking heavy queries that try to generate graphs.
3. Create `StatsGraphsJob` and schedule it to run daily at 4:15 AM.

## Testing

Locally run the server, first there was `No Data` warnings in the stats page. Later I have run the job manually by runnings `rails console` and `StatsGraphsJob.perform_now`.


lib/tasks/daily_maintenance.rake
```diff
@@ -4,4 +4,5 @@ desc "Daily maintenance tasks"
 task daily_maintenance: :environment do
   ReadRibbon.expire_old_ribbons!
   CommentStat.daily_fill!
+  StatsGraphsJob.perform_now
```
pushcx, line 7: We don't want to `perform_now`, this should be enqueuing the job.

app/models/stats.rb
```diff
@@ -0,0 +1,105 @@
...
+    timescale_divisions: TIMESCALE_DIVISIONS,
+    add_popups: true,
+    popup_format: "%Y-%m",
+    area_fill: false,
+    min_y_value: 0,
+    number_format: "%d",
+    show_lines: false
+  }
+
+  def self.monthly_graph(cache_key, opts)
```
pushcx, line 90: I don't like this `read_only` option that passed unused through all the methods. I'd rather that there were two methods, one to fill the caches and one to read it (and default to `No data` as you do here). The way the Rails cache API uses blocks mingles reading and writing, but if we're separating them it's ok to have parallel structure in our code. Then the job would be clearer because it would say something like `fill_monthly_graph_cache` instead of `get_cache` and implicitly doing work.

Then the daily job always calls the method to fill the cache and the controller always reads from the cache. They're coupled by `cache_key`.

pushcx (changes requested):
Thanks for taking on this fix! I look forward to merging it soon.

I think the Rails cache API made this a little confused, so if you refactor a little to get away from that the code will be a lot more readable.

lib/tasks/daily_maintenance.rake
Dogacel: I used `perform_later`, let me know if it looks right.

app/models/stats.rb
Dogacel: I remember reading something about how every function with a boolean parameter can be splitted into two other functions. In this case, it certainly makes more sense.

app/models/stats.rb
```diff
@@ -0,0 +1,120 @@
...
+
+  def self.get_users_graph = get_cached_graph(USERS_GRAPH_CACHE_KEY)
+
+  def self.get_active_users_graph = get_cached_graph(ACTIVE_USERS_GRAPH_CACHE_KEY)
+
+  def self.get_stories_graph = get_cached_graph(ACTIVE_USERS_GRAPH_CACHE_KEY)
+
+  def self.get_comments_graph = get_cached_graph(COMMENTS_GRAPH_CACHE_KEY)
+
+  def self.get_votes_graph = get_cached_graph(VOTES_GRAPH_CACHE_KEY)
```
pushcx, line 20: Ruby style for this kind of key is just to make these keys symbols, they're immutable strings. The coupling of a shared key is the same whether it's a constant (`USERS_GRAPH_CACHE_KEY`) or symbol (`:users_graph`). (My favorite introduction to Ruby for experienced programmers is [The Well-Grounded Rubyist](https://www.manning.com/books/the-well-grounded-rubyist-third-edition), it covers lots of Ruby style.)

The `fill_` methods have different queries so let's leave them alone, but we can delete these constants and the individual `get_` methods.

app/jobs/stats_graphs_job.rb
```diff
@@ -0,0 +1,11 @@
+class StatsGraphsJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    Stats.fill_users_graph_cache
```
pushcx, line 5: I like that you do these in series so that we are not taxing the database with 6 big aggregation queries at the same time.

app/models/stats.rb
```diff
@@ -0,0 +1,120 @@
...
+      scale_y_divisions: 10_000,
+      read_only: read_only
+    }) {
+      Vote.pluck(:updated_at).map { |updated_at| updated_at.strftime("%Y-%m") }.tally.sort.flatten
+    }
+  end
+
+  def self.get_cached_graph(cache_key)
+    cached_data = Rails.cache.read(cache_key)
+    cached_data || "<i>No data</i>"
```
pushcx, line 78:
```suggestion
    Rails.cache.read(cache_key) || "<i>No data</i>"
```

Don't need the intermediate variable.

pushcx (changes requested):
Solid improvement! I think there's a couple small changes we can make to take better advantage of Ruby and this will be ready to merge.

Dogacel:
> The brakeman warnings are ignorable, and we talked about copying over the old notes. Let me know if you'd like to do that, if you'd like help, or if you'd like me to take care of it.

No problem, it seems like I need to generate notes for warnings. I will refactor some stuff & fix warnings after work hours.

lib/tasks/daily_maintenance.rake
pushcx: Great!

Dogacel:
> The brakeman warnings are ignorable, and we talked about copying over the old notes. Let me know if you'd like to do that, if you'd like help, or if you'd like me to take care of it.

OK, Just completed it. Also, run the daily task combined with `bin/jobs`, it is picked up automatically and filled the cache 👍 Everything lgtm based on manual testing.

pushcx:
Great, I'm deploying it now!

PR #1528, opened 2025-04-18
Allow nav to spill into the right padding on mobile

chamlis:
I forgot that the nav's `grid-column: 1 / -1` was load-bearing, in that the last column on mobile is some right padding. I've made it more clear what's going on there.


pushcx:
I hear naming things is hard. :smile:

Issue #1532, opened 2025-04-20
Lobste.rs popular comment pages super slow on iOS


marcintustin:
This affects me as a logged in user. 

On iOS on safari, pages with more than a few comments (popular comment pages, the various pages listing comment trees) are very, very slow to render, rerender on scroll, and take ~2s per character entered in a comment box.

Notably, the comments page does not have this problem. https://lobste.rs/comments

This seems to have happened since the recent UI refresh, which is pretty great other than this issue.

marcintustin:
On further investigation this seems to especially affect pages with deeply nested trees of comments, even if there are relatively few comments (14), and in that case will noticeably affected the sections of the page with deeply nested comments worse than the rest of the page.

chamlis:
Thanks for reporting and investigating this. #1534 made things better on one of my test pages, but doesn't seem to fix things in general. My best theory is that WebKit doesn't perform well with nested (sub)grids, which is essentially what comment threads are in the new layout. If that's true, I think we'll need a different approach to laying out comment threads. I've done some initial experiments today with having one global grid but that still performs poorly, perhaps due to still having subgrids.

To pursue the global grid without subgrids, I think we'd need CSS selectors for each level of the comment tree to set the appropriate grid column. An alternative is to abandon using grid for anything other than the top-level page layout, and construct the comment trees out of nested flexboxes, which could perform better.

chamlis:
Here's a contrived example that renders instantly for me on Firefox but slowly on iOS Safari:

<details>
<summary>Long HTML...</summary>

```html
<!doctype html>
<html>
  <head>
    <title>CSS Grid Performance Test</title>
    <style>
.grid {
display: grid;
grid-template-columns: max-content max-content;
}
.subgrid {
display: grid;
grid-template-columns: subgrid;
}
.box {
width: 5rem;
height: 5rem;
background-color: green;
margin: 0.5rem;
}
</style>
  </head>
  <body>
    <div class="grid">
<div class="box"></div>
 <div class="subgrid">
<div class="box"></div>
<div class="grid">
<div class="box"></div>
 <div class="subgrid">
<div class="box"></div>
<div class="grid">
<div class="box"></div>
 <div class="subgrid">
<div class="box"></div>
<div class="grid">
<div class="box"></div>
 <div class="subgrid">
<div class="box"></div>
<div class="grid">
<div class="box"></div>
 <div class="subgrid">
<div class="box"></div>
<div class="grid">
<div class="box"></div>
[...]

PR #1534, opened 2025-04-20
Attempt to fix performance issues on (iOS) safari

chamlis:
I don't understand why this seems to improve things on my device. The logic was to reduce the number of grids that are larger than the viewport, but there's still an outermost `<ol>`.

Targetting #1532, but nowhere near enough confident to say it closes it.


pushcx:
Thangs to you and bfgeek1 for finding this workaround. I'm going to deploy this so at least pages are loading properly, and we can pursue a clearer fix.

PR #1537, opened 2025-04-21
Fix tree lines with added <div>

chamlis:
Sorry about that! Thanks @JulianWgs for spotting this during office hours.


pushcx:
Oh great, I didn't think something like that would work because I assumed the `<div>` would interrupt the nesting. Thanks for the fast fix!

PR #1538, opened 2025-04-21
Add the <div> workaround everywhere comments are rendered

chamlis:
In several places, we fake a thread by manually creating an empty list of children. Add the `<div>` workaround in those places so the styling can apply consistently.

This is yet another commit to revert when we solve this properly. On the bright side, that should be all the more satisfying.


pushcx:
Off the top of my head, either the difference is styling (we generally want the 'disown' buttons to look like links) or I reinvented `button_to` because I didn't remember it existed. Thanks for fixing it.

PR #1539, opened 2025-04-22
Fix styling of disown button

chamlis:
I'd need to dig more into it, but I'm not sure what the differences between `link_post` and `button_to` are, and whether we can switch to only using one of them.


pushcx:
Off the top of my head, either the difference is styling (we generally want the 'disown' buttons to look like links) or I reinvented `button_to` because I didn't remember it existed. Thanks for fixing it.

PR #1540, opened 2025-04-22
Fix styling of "Banned user" on users/show

dschaerer:
This fixes the regression in  c4a68b9d4c9547e38350a9027262e5e341b7f7c7 (see show.html.erb:18) that rendered style attribute outside of tag.


pushcx:
Thanks for spotting and fixing this bug, I'm deploying it now.

dschaerer:
Glad I could help, thank you!!!

PR #1546, opened 2025-04-24
Remove unused nested selector rules

chamlis:
Originally noticed this because it uses the CSS nested selector: a more recent addition. The colour is the same as applied without :hover, and as far as I can see there is no "disable" class, so I don't believe we need this block.


pushcx:
Thanks for spotting and tidying this.

The failing build is my error, so I'm comfortable rebasing to merge.

Issue #1543, opened 2025-04-23
UI: alignment for post / comment entries seems to be bugged and changed to center on safari on ios 15


AhmadFudl:
![image](https://github.com/user-attachments/assets/2d0a24bf-aeb4-4b62-93f4-e51f503f440a)

![image](https://github.com/user-attachments/assets/e36935d8-9d8d-4380-aedc-dc65f41a4306)

pushcx:
This is downstream of #1532 where we pushed a workaround (#1534) for a webkit performance problem. I don't want to revert the workaround because the webkit bug meant pages didn't load at all. Bad layout is better than no content, at least.

It's a little redundant, but I'm going to keep this issue open because seeing the sitewide nav header laid out wrong is a novel surprise to me. I know it also uses `subgrid` now, and I think maybe there's a serious bug in webkit's implementation of it.

chamlis:
Sadly (for us), iOS 15 [doesn't support subgrid](https://caniuse.com/css-subgrid). Clearly more people use non-recent graphical browsers than I'd realised in doing the CSS grid, and I'd absolutely like to still support them. This is more evidence in favour of moving away from subgrid altogether, which is what I'm currently working on.

AhmadFudl:
I love your spirit, and hate that my old phone is causing you guys more work.

pushcx:
Thanks for this bug report. I just pushed [b2965a6](https://github.com/lobsters/lobsters/commit/b2965a6da73fbdc6785bf56cdb2adc7c7aba8a2d) that should fix this and many other bugs.

PR #1555, opened 2025-05-03
Reduce usage of (sub)grid

chamlis:
The [recent layout change](https://github.com/lobsters/lobsters/pull/1508) leant heavily on CSS grid and subgrid. Sadly, subgrid in particular was not as well-supported as I'd hoped, which has lead to a [litany](https://github.com/lobsters/lobsters/issues/1532) of [issues](https://github.com/lobsters/lobsters/issues/1543).

This branch moves away from such heavy use of grid. By accepting that we can't use subgrid, and that automatic sizing of the global page columns is infeasible, it seems we don't actually need grid at all for the story and comment pages. We can fix a column width `--logo-col-width`, which is calculated based on the logo size, and use that fixed size on flexbox elements to lay out the pages.

This necessitated changing the DOM layout, as we need to hoist the gutter containing the folder and upvote columns up one level so that it can span the height of the entire subtree. Along the way I've tried to make the pages I was touching use more semantic HTML. For example, comments and stories use `time` to make their creation date more machine-readable, and stories are now `article`s.

<details>
<summary>Story list</summary>

flexbox_stories
</details>

<details>
<summary>Story view</summary>

flexbox_comments
</details>

Things I know need to happen before this branch can be merged:
- [ ] Make JavaScript upvoting and flagging work with the new DOM layout
- [ ] Make `labelled_subgrid` function without subgrid (fall back to set column widths)
- [ ] Split into more meaningful separate commits

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -483,8 +483,7 @@ ol.comments1 > li.comments_subtree > ol.comments {
 	padding-left: 0;
 }
 
-
-#inside .comment ~ div > ol.comments {
+#inside .comment ~ ol.comments {
 	grid-column: main / end;
 	display: grid;
 	grid-template-columns: [begin] max-content [main] 1fr [end];
@@ -664,7 +663,7 @@ li .comment_folder:before {
 }
 #inside li .comment_folder_button:checked ~ .comment div.comment_text,
 #inside li .comment_folder_button:checked ~ .comment .comment_gutter > *:not(.comment_folder),
-#inside li .comment_folder_button:checked ~ div ol.comments {
+#inside li .comment_folder_button:checked ~ ol.comments {
 	display: none;
 }
 
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -31,6 +31,12 @@ But I make an unprincipled exception for :has because it solves so many big prob
 	--color-shadow-80: rgb(var(--base-shadow) / var(--opacity-shadow-80));
 	--color-shadow-25: rgb(var(--base-shadow) / var(--opacity-shadow-25));
 	--color-shadow-10: rgb(var(--base-shadow) / var(--opacity-shadow-10));
+
+	--logo-size: 16px;
+	--logo-padding: 1px;
+	--logo-margin: 0.5rem;
+	--logo-col-width: calc(var(--logo-margin) + var(--logo-padding) + var(--logo-size) + var(--logo-padding) + var(--logo-margin));
+	--page-padding: 1rem;
 }
 
 /* generics */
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/application.css
```diff
@@ -460,102 +447,71 @@ ol.category_tags {
...
+	flex-grow: 1;
 }
 
 .upvoter:before {
 	content: "\25b2";
 	color: var(--color-fg-shape);
 	font-size: 12pt;
 }
-.upvoted .upvoter:before,
+.comments_gutter:has(~ .comments_body > .comment.upvoted) .upvoter:before,
```
pushcx, line 479: I saw that this was a draft but had to peek.

We can copy the upvoted class to a closer element if it helps keep selectors simple. I'd rather not lean on `:has` if we can avoid it, given your research on what's [widely available](https://github.com/lobsters/lobsters/blob/71b8e1458a9234fa2644c6e65b83ab8d4f8998ce/app/assets/stylesheets/application.css#L4).

app/helpers/application_helper.rb
```diff
@@ -164,11 +164,10 @@ def tag_link(tag)
 
   def how_long_ago_label(time)
     ago = how_long_ago(time)
-    content_tag(:span, ago, title: time.strftime("%F %T %z"))
+    content_tag(:time, ago, title: time.strftime("%F %T%z"), datetime: time.strftime("%F %T%z"))
```
pushcx, line 167: Great catch!

app/assets/stylesheets/application.css
```diff
@@ -1555,14 +1481,6 @@ input[type="submit"].link_post.pushover_button {
 		-webkit-text-size-adjust: none;
 	}
 
-	body {
-		grid-template-columns: [begin] max-content [main] 1fr [right] max-content [end] 0.5rem [padding];
-	}
-
-	#inside {
```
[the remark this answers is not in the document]
pushcx: Also reported as #1564.

pushcx:
Chamlis and I talked through this on office hours today, leading to reverting the CSS grid in [b2965a6](https://github.com/lobsters/lobsters/commit/b2965a6da73fbdc6785bf56cdb2adc7c7aba8a2d) because we thought it would be easier to roll back than forward with fixes.

I still very much like the start of this PR to define a width on the logo and reuse that as the width of the comment gutter, which has a defined semantic element.

Chamlis: I'm going to leave this PR open for your convenience comparing your in-flight work to the big change, though of course it may make more sense to cherry-pick that to a new branch (or something) than deal with merge conflicts.

chamlis:
Thanks for all the reviews and your patience. I'm closing this PR in favour of the smaller #1567 and #1568.

PR #1567, opened 2025-05-13
Use CSS Grid for user profiles and settings

chamlis:
Bringing back some of the changes from #1508 that didn't involve subgrid.

This PR includes the bug fix from a54b145c18735872528e09df24247da6d7bc63fa.


pushcx:
Thanks for coming back to this to start moving things over. I especially appreciate that you caught the bugfix I accidentally reverted.

Could you add a comment to #1279 with your punchlist or rough plan? I'd love if you converted more of the site forms, and it'd be good to publicize your future plans.

PR #1568, opened 2025-05-13
Use the <time> element for display how_long_ago_labels

chamlis:
Use `<time>` instead of `<span>` for these labels, to ensure the times are machine-readable. This doesn't affect the text that appears when a user hovers, though that has a different format than the machine-readable portion.


pushcx:
Nice semantic improvement, thanks!

PR #1569, opened 2025-05-13
Expand BAD_NETS to cover more special addresses

### LINKED ISSUE #1559: Sponge should block 100.64.0.0/10
I was talking on today's office hours stream and while talking about how we parse domains @grayhatter reminded that public domain names like `localhost.gr.ht` can resolve to local IP address like `127.0.0.1`. We filter those in `extras/sponge.rb`, the `BAD_NETS` constant.

When reviewing it, I noticed that I didn't see the `100...` addresses that [Tailscale uses](https://tailscale.com/kb/1015/100.x-addresses). The list should include that range.

Also their doc links to [RFC 5735](https://datatracker.ietf.org/doc/html/rfc5735) and for this issue, you must review it to see if there are any other ranges we need to prohibit linking to.

chamlis:
Primarily based off of RFC 6890, which obsoleted RFC 5735. Commented to give rationale for deviations from that RFC.

Closes #1559


pushcx:
Great to have a more complete list, and thanks for your comprehensive comments.

PR #1570, opened 2025-05-15
add humanity section to about page

GrayHatter:


pushcx:
To contextualize this, we had another [big thread on vibe coding](https://lobste.rs/s/gkpmli/if_ai_is_so_good_at_coding_where_are_open) that rolled into the [chat room](https://lobste.rs/chat) about what the policy on submitting slop comments is. I said that I've been removing them and asked @GrayHatter to PR a change to [/about](https://lobste.rs/about) to improve the visibility of this.

This is a great first draft, thanks for taking this on. It's not easy to explain norms and write good policy.

So one concern I have with what's written is that it's not something I can practically enforce. It says "use AI to help you generate any part of any comments". While directly copy-and-pasting slop becomes clear pretty quickly (for now \*ominous music plays\*), using a tool like Grammarly to fix style issues or have Claude suggest an outline and then writing a comment towards those points would be pretty undetectable. I'm sure there's a proper political science term for this, but a rule where violations are easy and undetectable means the rule is nearly unenforceable in practice, which deligitimizes the law and makes and enforcement look like capricious selective prosecution.

It also might not be good policy. A non-native speaker touching up their grammar is someone reaching for a tool to help them communicate well. It's novel now, but [familiar algorithms](https://en.wikipedia.org/wiki/Levenshtein_distance) are ubiquitous and unobjectionable. My hunch is that as LLMs get better at running on-device for privacy and speed, they're going to be regarded similarly. I don't have a lot of confidence that this accurately reflects the community's sentiment and we should probably have a [meta thread](https://lobste.rs/t/meta) to discuss this guideline before merging.

This is a solid first draft, let's polish this and some of the other normal first draft issues (discussed on [today's office hours stream](https://push.cx/stream/2025-05-15)) and then bring it to the community.

PR #1574, opened 2025-05-17
Fix user test to reflect increased banned_reason max size

chamlis:
Fixup for f174623be9d627509d79ed18d80e2ec5cfaa7e39


pushcx:
Thanks for catching this, I missed it while I was in a hurry.

PR #1575, opened 2025-05-17
Stop inserting the custom lobster emoji list marker manually

chamlis:
This makes things line up a bit nicer, though sadly for me the aesthetic center of the lobster emoji still doesn't match that of the disc markers.

Before:
image

After:
image


pushcx:
Great simplification, thanks!

Issue #1255, opened 2024-02-22
Add data to /tags


pushcx:
The [/tags](https://lobste.rs/tags) page should have more info on it like how many stories use each tag, how many comments there are on those stories, how many users are [filtering](https://losbte.rs/filters). Depending on the presentation, maybe differentiate between recent (say, 6 months?) and all-time. We have a nice svg charting library in use on the [/stats](https://lobste.rs/stats) page, so there's options there.

nogweii:
One idea I had during the latest stream was applying a clustering algorithm to see tags relationships. For example, it's possible that "rant" and "systemd" are used together very often.

This probably would be more useful on an individual tag's page rather than trying to fit it into the overview page.

-----

Another idea is to see the average number of comments per tag, not just the total number of comments on stories with a tag. @pushcx has already written the applicable SQL (prettied up):

```sql
WITH tag_story_counts AS
  (SELECT tags.tag,
          comments.story_id,
          count(*) AS c
   FROM tags
   JOIN taggings ON tags.id = taggings.tag_id
   JOIN comments ON comments.story_id = taggin gs.story_id
   GROUP BY tags.tag,
            comments.story_id)
SELECT tag,
       avg(c)
FROM tag_story_counts
GROUP BY tag
ORDER BY 2 DESC
LIMIT 10;
```

rahul1990gupta:
@pushcx 

In my mind, there are three ways, tags visualization page can be designed. 
-  **TileMap view**: Each tile  correspond to a tag with size of the tag proportional to story count. Upon hovering on the tip, pointer will show other stats such as total_comments, number of filtering_users etc. e.g. https://finviz.com/map.ashx
- ** Bump Chart **: It looks like a leadersboard through time. Each row correspond to a tag ranked by story count that year. See the chart below at the bottom of the page. https://ossinsight.io/collections/web-framework/
- ** Data Table **: Regular table with sorting buttons in the header. Filters can be added at the top.  e..g. https://www.tradingview.com/screener/

Data Table will probably be easier to implement. Bump chart and TileMap look cool, but may require more effort and writing a custom js library. :man_shrugging:

PR #1577, opened 2025-05-20
Lay out the pages for tags and categories using CSS Grid

chamlis:
This is work towards #1279 

<details>
<summary>Uninteresting screenshots</summary>

Screenshot 2025-05-20 at 14-02-55 Tags Lobsters
Screenshot 2025-05-20 at 14-03-09 Tags Lobsters
Screenshot 2025-05-20 at 14-03-29 Edit Category Lobsters
Screenshot 2025-05-20 at 14-03-41 Edit Category Lobsters
Screenshot 2025-05-20 at 14-03-49 Edit Tag Lobsters
Screenshot 2025-05-20 at 14-04-01 Edit Tag Lobsters
Screenshot 2025-05-20 at 14-04-09 Create Tag Lobsters
Screenshot 2025-05-20 at 14-04-17 Create Tag Lobsters
Screenshot 2025-05-20 at 14-04-24 Create Category Lobsters
Screenshot 2025-05-20 at 14-04-32 Create Category Lobsters

</details>

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -6,6 +6,7 @@ class CategoriesController < ApplicationController
   def new
     @category = Category.new
     @title = "Create Category"
+    @title_h1 = true
   end
 
   def create
@@ -22,6 +23,7 @@ class CategoriesController < ApplicationController
   def edit
     @category = Category.where(category: params[:category_name]).first!
     @title = "Edit Category"
+    @title_h1 = true
   end
 
   def update
--- a/app/views/categories/_form.html.erb
+++ b/app/views/categories/_form.html.erb
@@ -1,11 +1,9 @@
 <%# locals: (category:) -%>
 <%= form_with model: category, url: category.persisted? ? update_category_path : categories_path, method: :post do |f| %>
-  <div class="boxline">
-    <%= f.label :category, 'Name:', class: 'required' %>
+  <div class="labelled_grid">
+    <%= f.label :category, 'Name', class: 'required' %>
     <%= f.text_field :category  %>
   </div>
 
-  <div class="boxline">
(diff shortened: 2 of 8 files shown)


pushcx:
Please move the the checkboxes inline with the titles on the mobile view.

And let me cross-reference #1255 - this PR doesn't take that on, but it's in the neighborhood.

pushcx:
Oh, and if you are in the mood to move checkboxes inline and don't mind slipping an extra small fix into this PR, I noticed on the checkbox on the message form for whether to modlog the use of a hat should also be inline.

pushcx:
I can see the complex selectors, yeah. I value keeping things separate, too. If one of these selector drafts [397](https://github.com/w3c/csswg-drafts/issues/397), [1067](https://github.com/w3c/csswg-drafts/issues/1067), or [11300](https://github.com/w3c/csswg-drafts/issues/11300) landed (and several years passed...) this would be very simple to express. Until then, I see two options.

1. Compromise a little and add a `for_checkbox` class on the labels for checkboxes.

2. Nest `<label><input type="checkbox"></label>` rather than use `for` to target it. I can't immediately picture where the help/hints land, and it's a bit of churn in the code, but the three selectors would collapse to `.labelled_grid > label:has(> input[type="checkbox"])` which is succinct and clear.

chamlis:
Thanks for both of the suggestions for getting rid of the complex selectors. I experimented with the first one, but I think we would then have to use a subgrid within the label itself if we wanted to line up the labels for a series of checkboxes. I've therefore opted for the second suggestion. I applied the new class predominantly through editor macros to the existing grid forms, so I apologise if I missed any!

pushcx:
Thanks for another view cleanup, nice to improve these form layouts so they display reliably.

Issue #1548, opened 2025-04-28
Migrate cron scritps to jobs


pushcx:
Now that I have [SolidQueue running recurring jobs](https://github.com/lobsters/lobsters/issues/1299#issuecomment-2831309995) I'd like to start moving over our cron jobs and remove that as a moving part. 🖖

Anyone's welcome to contribute and it's fine if you want to PR one single script or (for the more complicated ones) only part of a script. (Please be explicit about your planned scope in your PR.)

So the to-do list is our two cron jobs. [Daily](https://github.com/lobsters/lobsters-ansible/blob/master/roles/lobsters/templates/sbin/lobsters-daily.j2) has these scripts, which should be ported to [ActiveJob](https://guides.rubyonrails.org/active_job_basics.html) jobs - see [app/jobs](https://github.com/lobsters/lobsters/tree/master/app/jobs) and [recurring.yml](https://github.com/lobsters/lobsters/blob/master/config/recurring.yml):

 - [ ] Refresh the sitemap, note it's probably broken #1496 
 - [ ] Write a job to delete old rack attack entries. The cron job is the old file-based setup, ignore it. Delete from SolidCache any entry whose key starts `production:rack::attack` older than 2 days. (Maybe it has a nice `expire` method?)

[https://github.com/lobsters/lobsters-ansible/blob/master/roles/lobsters/templates/sbin/lobsters-cron.j2](Every 5 minutes):

 - [ ] [fill_flagged_cache](https://github.com/lobsters/lobsters/blob/master/script/fill_flagged_cache.rb): call FlaggedCommenters.new("1m").commenters; this only needs to run every 10m (easy!)
 - [ ] [traffic_range](https://github.com/lobsters/lobsters/blob/master/script/traffic_range) (easy!)
 - [ ] [mastodon_sync_list](https://github.com/lobsters/lobsters/blob/master/script/mastodon_sync_list.rb) run hourly
 - [ ] [send_new_webmentions](https://github.com/lobsters/lobsters/blob/master/script/send_new_webmentions) rather than a recurring job, have story submission enqueue this with a 5 minute delay (which the AJ doc says SQ supports, but I don't know how)
 - [ ] [mastodon_post](https://github.com/lobsters/lobsters/blob/master/script/mastodon_post.rb) also on 5 min delay from submission
 - [ ] [mail_new_activity](https://github.com/lobsters/lobsters/blob/master/script/mail_new_activity.rb) this script does several independent jobs:
[...]

PR #1586, opened 2025-05-24
Port `fill_flagged_cache` to an ActiveJob

vivshaw:
Howdy! This PR resolves one small chunk of issue #1548. I start by porting only the `fill_flagged_cache` job, because my Rails is a tad rusty and I wanted to make sure I'm doin' it right before I tackle a larger chunk.

See the `lobsters-ansible` sister PR for cleanup of the cron job: lobsters/lobsters-ansible#89

Wasn't sure what order you'd prefer to ship the ActiveJob vs. the cron removal. I figure if you shipped the ActiveJob first, `fill_flagged_cache` would run twice for a while and I wasn't sure if it was idempotent. So I structured this as though the cron job would be taken offline first in the other PR, then this would ship second. Let me know if this is incorrect and I can restructure accordingly.

### DIFF
--- a/app/jobs/fill_flagged_cache_job.rb
+++ b/app/jobs/fill_flagged_cache_job.rb
@@ -1,0 +1,8 @@
+class FillFlaggedCacheJob < ApplicationJob
+  queue_as :default
+
+  # fills the cache - used by app/views/users/show.html.erb for mods
+  def perform(*args)
+    FlaggedCommenters.new("1m").commenters
+  end
+end
--- a/script/fill_flagged_cache.rb
+++ b/script/fill_flagged_cache.rb
@@ -1,7 +1,0 @@
-#!/usr/bin/env ruby
-
-require File.expand_path("../../config/environment", __FILE__)
-
-# fills the cache - used by app/views/users/show.html.erb for mods
-
-FlaggedCommenters.new("1m").commenters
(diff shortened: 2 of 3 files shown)


app/jobs/fill_flagged_cache_job.rb
```diff
@@ -0,0 +1,8 @@
+class FillFlaggedCacheJob < ApplicationJob
+  queue_as :default
+
+  # fills the cache - used by app/views/users/show.html.erb for mods
```
vivshaw, line 4: Not sure if there's a comment convention for ActiveJobs, nor whether y'all wanna keep this comment from the old script

script/fill_flagged_cache.rb
```diff
@@ -1,7 +0,0 @@
-#!/usr/bin/env ruby
```
vivshaw, line 1: As mentioned in PR description, happy to bring this back if there's a different sequence y'all wanna follow for deploying this stuff. Just let me know!

app/jobs/fill_flagged_cache_job.rb
pushcx: I don't know of a formatting convention, but I appreciate you keeping it for context. These informal little cross-references help me a lot.

pushcx:
This looks great, I'd rather run the job twice for a few minutes during the deploy.

Thanks for starting this conversion work. This one's real simple so I have no concerns, hope to see you take on more. :)

Issue #1496, opened 2025-03-20
Sitemap generation is broken


davispuh:
I tried
```
$ rails sitemap:refresh
rails aborted!
NoMethodError: undefined method 'create!' for an instance of SitemapGenerator::LinkSet (NoMethodError)

        (@link_set ||= reset!).send(*args, &block)
                              ^^^^^
Did you mean?  create
config/sitemap.rb:9:in 'SitemapGenerator::Interpreter.run'
Tasks: TOP => sitemap:refresh => sitemap:create
```

In [config/sitemap.rb:9](https://github.com/lobsters/lobsters/blob/master/config/sitemap.rb#L9) there is `SitemapGenerator::Sitemap.create!` but it probably should be `create`.

Also when looking at https://lobste.rs/sitemap.xml.gz it doesn't look to be updated.
```
$ curl -s https://lobste.rs/sitemap.xml.gz | gunzip - | xmllint --format -
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd">
  <sitemap>
    <loc>https://lobste.rs/sitemap1.xml.gz</loc>
    <lastmod>2023-09-16T00:08:31+00:00</lastmod>
  </sitemap>
  <sitemap>
    <loc>https://lobste.rs/sitemap2.xml.gz</loc>
    <lastmod>2023-09-16T00:08:31+00:00</lastmod>
  </sitemap>
</sitemapindex>

$ curl -s https://lobste.rs/sitemap1.xml.gz | gunzip - | grep -E '2024-|2025-'
$ curl -s https://lobste.rs/sitemap2.xml.gz | gunzip - | grep -E '2024-|2025-'
```

krydos:
hey @pushcx 
Do you run this `rails sitemap:refresh` in your deployment script? It probably should generate enormous sitemap as it goes [through all the Story records](https://github.com/lobsters/lobsters/blob/master/config/sitemap.rb#L18). 
It seems like this loop is actually grouping stories too which probably produces less enormous sitemap but I still wonder how much time does it take to run `rails sitemap:refresh`

pushcx:
[...]

PR #1598, opened 2025-05-29
Fix sitemap generation

krydos:
Fix this issue over here - https://github.com/lobsters/lobsters/issues/1496

It's not my first PR here so my apologies for stealing `good first issue` from other people. I just wanted to generate a sitemap locally and noticed this issue which I then found already created. 

The issue was introduced here - https://github.com/lobsters/lobsters/commit/2332d2802254af5d77ff4808a69b42523fdae4c3 
I'm unable to find reasoning behind this change and I can't find proofs that `create!` has ever existed in `sitemap_generator` gem. Probably that `!` was just introduced by accident. 

Hope it's gonna fix the cronjob you have in prod.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -6,7 +6,7 @@ check_hourly = 4.days.ago
 check_daily = 2.weeks.ago
 top_score = Story.all.maximum("score")
 
-SitemapGenerator::Sitemap.create! do
+SitemapGenerator::Sitemap.create do
   %w[/about /chat].each do |path|
     add path, changefreq: "monthly", lastmod: nil
   end


dreig:
Looks like a false positive of [Rails/SaveBang](https://docs.rubocop.org/rubocop-rails/cops_rails.html#examples-railssavebang)
But the Cop has since been disabled in the `standard-rails` plugin: https://github.com/standardrb/standard-rails/pull/37/files

pushcx:
@dreig Thanks, I started looking at this and was very puzzled why I would've changed this when updating gems... and then saw that you ran down where this error was introduced. Thank you.

And of course thank you @krydos for fixing this. You don't at all have to apologize for taking a 'good first issue', they're available for anyone. I would love to see you PR as many as I can talk you into helping with. :smile:

PR #1600, opened 2025-05-29
Fix note on hidden stories

sknebel:
The note shown on hidden stories explaining that it is hidden and what that means had broken grammar, I reworded it slightly to fix that.


pushcx:
Definitely clearer, thanks for the improvement.

PR #1605, opened 2025-06-01
Port over the sitemap refresh cron job to an ActiveJob

vivshaw:
This PR does another chunk of #1548: the sitemap job. A sister PR, lobsters/lobsters-ansible#90, cleans up the cron.

### DIFF
--- a/app/jobs/refresh_sitemap_job.rb
+++ b/app/jobs/refresh_sitemap_job.rb
@@ -1,0 +1,7 @@
+class RefreshSitemapJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    SitemapGenerator::Interpreter.run(config_file: ENV["CONFIG_FILE"], verbose: false)
+  end
+end
--- a/config/recurring.yml
+++ b/config/recurring.yml
@@ -19,6 +19,10 @@ production:
     class: FillFlaggedCacheJob
     queue: background
     schedule: every 10 minutes
+  refresh_sitemap:
+    class: RefreshSitemapJob
+    queue: background
+    schedule: every day at 12:00am
 
 development:
   expire_old_ribbons:
@@ -41,3 +45,7 @@ development:
     class: FillFlaggedCacheJob
     queue: background
     schedule: every 10 minutes
+  refresh_sitemap:
+    class: RefreshSitemapJob
+    queue: background
+    schedule: every day at 12:00am


app/jobs/refresh_sitemap_job.rb
```diff
@@ -0,0 +1,7 @@
+class RefreshSitemapJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    SitemapGenerator::Interpreter.run(config_file: ENV['CONFIG_FILE'], verbose: false)
```
vivshaw, line 5: This syntax looks a lil' different than the familiar `rake -s sitemap:refresh`, so let me explain.

I actually started with something to directly call the Rake task, like so:

```Ruby
Rails.application.load_tasks
Rake::Task['sitemap:refresh'].invoke
```

This "worked", but behaved a little oddly. On calls after the first it didn't seem to regenerate the sitemap. A `Rake::Task['sitemap:refresh'].reenable` didn't seem change that, but _did_ seem to call the sitemap ping an increasing number of time per invocation. It was also somewhat annoying to silence the job to mimic `rake -s`. This all set my spidey senses tingling, so I changed tactics.

Second approach: I peeked at [the sitemap_generator Rake tasks](https://github.com/kjvarga/sitemap_generator/blob/272e790bf76d23964b651f502260f062c9a72833/lib/sitemap_generator/tasks.rb#L43-L54) and manually did what they do. I noticed that #1496 stopped doing the sitemap ping, so I didn't both to implement that here. So unless I'm missing something, this _should_ be analogous to `rake-s sitemap:refresh:no_ping`.

app/jobs/refresh_sitemap_job.rb
```diff
@@ -0,0 +1,7 @@
+class RefreshSitemapJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    SitemapGenerator::Interpreter.run(config_file: ENV["CONFIG_FILE"], verbose: false)
```
vivshaw, line 5: This syntax looks a lil' different than the familiar `rake -s sitemap:refresh`, so let me explain.

I actually started with something to directly call the Rake task, like so:

```Ruby
Rails.application.load_tasks
Rake::Task['sitemap:refresh'].invoke
```

This "worked", but behaved a little oddly. On calls after the first it didn't seem to regenerate the sitemap. A `Rake::Task['sitemap:refresh'].reenable` didn't seem change that, but _did_ seem to call the sitemap ping an increasing number of time per invocation. It was also somewhat annoying to silence the job to mimic `rake -s`. This all set my spidey senses tingling, so I changed tactics.

Second approach: I peeked at [the sitemap_generator Rake tasks](https://github.com/kjvarga/sitemap_generator/blob/272e790bf76d23964b651f502260f062c9a72833/lib/sitemap_generator/tasks.rb#L43-L54) and manually did what they do. I noticed that #1496 stopped doing the sitemap ping, so I also skipped that here. So unless I'm missing something, this _should_ be analogous to `rake -s sitemap:refresh:no_ping`.

vivshaw: (We could alternatively just shell out with `%x( rake -s sitemap:refresh )`, I suppose, but IDK how y'all feel about arbitrary shelling out in your ActiveJobs)

pushcx: Thanks for figuring out how the gem had the task wired up. This does just what we want, and is preferable to shelling out to start a new rails process.

pushcx:
Thanks for this, nice to cross one more cron job off the list!

PR #1619, opened 2025-06-13
Move remaining timed jobs (TrafficHelper and Mastodon sync) to Solid Queue

vivshaw:
This PR ports the two remaining timed jobs from #1548 over to Solid Queue.


app/jobs/mastodon_sync_list_job.rb
```diff
@@ -0,0 +1,50 @@
+class MastodonSyncListJob < ApplicationJob
```
vivshaw, line 1: This ports over the logic from `mastodon_sync_list.rb` verbatim.

This was somewhat challenging to test locally, given that it cares about prod Mastodon keys &c. Might wanna kick the tires a tad.

script/lobsters-cron
```diff
@@ -19,8 +19,6 @@ fi
 
 bundle exec ruby script/mail_new_activity.rb &
 bundle exec ruby script/mastodon_post.rb &
-bundle exec ruby script/mastodon_sync_users.rb &
```
vivshaw, line 22: Was this job already broken? I don't see any file at `script/mastodon_sync_users.rb`, but I do see `script/mastodon_sync_list.rb`

app/jobs/mastodon_sync_list_job.rb
pushcx: A recent project of mine has been to wire up Telebugs for better production error reporting. One benefit of your work to move all these into jobs is that now errors will flow into the tracker instead of into noisy cron emails that instantly get buried in my personal inbox.

Which I say to explain how I appreciate your contributions, and also to justify yoloing this into prod without testing because I agree that testing a prod integration is a pain.

script/lobsters-cron
pushcx: Yes, and apparently it always has been, when I look at the [lobsters-ansible commit introducing it](https://github.com/lobsters/lobsters-ansible/commit/7689a7fcd51fc5ceeb3ce431099346864c06a8fb).

This was a pretty basic port of the old script to maintain a list of users on Twitter. I knew Mastodon lists weren't public so I didn't check up on whether it was running in prod.

Thanks for catching this error.

pushcx:
This looks great, thank you. I really appreciate that you're doublechecking my work to catch misconfigurations and errors like this.

PR #1624, opened 2025-06-15
replace after create notification hooks with a notify job

thomasdziedzic:
I decided to do a refactor of the notifications into an ActiveJob to get a little more familiar with the code I'm going to touch and I remember on stream you mentioned you weren't very fond of the lifecycle hooks. As a bonus I'm also learning jj with this pr.


pushcx (changes requested):
You're taking on a hairy bit of functionality, it's great to have it split out into clear reusable chunks, and the tests are a weight off my shoulders.

I think `NotifyJob` probably wants to be two different jobs, `NotifyMessageJob` and `NotifyCommentJob`. The `case` in when splits into two separate tracks and I didn't see that they share use of other methods from there. Did I miss something, or could you split them up?

thomasdziedzic:
I was initially thinking about keeping a unified notification job to handle all types of notifications in the app. I did split up the job though since the comments and message notifications are still separate in the logic.

app/jobs/notify_comment_job.rb
```diff
@@ -0,0 +1,81 @@
+class NotifyCommentJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
```
pushcx, line 4: Could you tell me about this idiom? I didn't notice it with the `case`, but I see it's not in our other jobs. The two call sites call the jobs with a single comment or a single message but you've made it more general. And rather than take an array, you've used the args, so someone passing an array would have to call `.perform_later(comment1, comment2)` or `.perform_later(*comments)`. Is there an integration with other parts of ActiveJob I'm missing context on?

thomasdziedzic: That's just part of the default job template. You can see that idiom being used in the [rails guides](https://guides.rubyonrails.org/active_job_basics.html#create-and-enqueue-jobs) and in a newly generated job:

```
lobsters$ rails g job test
lobsters$ grep perform app/jobs/test_job.rb 
  def perform(*args)
```

Of course the rails guides and the rails api docs are inconsistent using the splat operator for the args because the rails api docs don't use them: https://api.rubyonrails.org/classes/ActiveJob/Base.html

If you want I can remove the splat from both of the jobs if you want to conform to the other jobs in the repo.

thomasdziedzic: I dove into the code on how rails works with that argument: https://github.com/rails/rails/blob/main/activejob/lib/active_job/execution.rb#L68

It uses the splat operator which works with both a single argument and also a perform defined with splats, so I don't think there is a right or wrong way of splats vs no splats, it just comes down to preference at that point.

pushcx: Thanks for explaining. That is a really strange idiom but I have learned it's usually better not to try to "correct" Rails on these things because there'll be some obscure piece of vital plumbing that assumes it and starts throwing bizarre exceptions 18 months later.

As for the inconsistency... that's a direct consequence of the last big temper tantrum, so it's also better not to think too hard about.

pushcx:
Thanks for this, it's great to start removing ActiveRecord hooks with side effects. At some point I'll even get to make a Rubocop rule against them. :)

pushcx:
xref #1548 because that's organizer for the big ongoing push towards jobs, though this isn't a cron script.

PR #1626, opened 2025-06-17
Move Send Webmentions to ActiveJob

vivshaw:
This PR does the next chunk of #1548, and the first which is an async enqueued job vs. a timed cron job. There were some larger changes in this one. I'll notate them in self-review.


app/controllers/stories_controller.rb
```diff
@@ -35,7 +35,11 @@ def create
           raise ActiveRecord::Rollback
         end
       end
-      return if @story.persisted? # can't return out of transaction block
+
+      if @story.persisted?
+        SendWebmentionJob.set(wait: 5.minutes).perform_later(@story)
```
vivshaw, line 40: I'm legit not sure if enqueueing from the controller vs enqueueing from an `after_save` callback is more idiomatic. This was my first guess for KISS reasons.

Steelmanning the point _against_ my choice, and in favor of a callback:
 - this isn't very "skinny controllers, fat models" of me
 - if you ever add other paths to create stories (API, separate mod interface, etc), webmentions would not fire, and you might want them to

Thoughts?

vivshaw: (Separate Q- should we skip this job in a dev environment?)

app/jobs/send_webmention_job.rb
```diff
@@ -0,0 +1,86 @@
+class SendWebmentionJob < ApplicationJob
+  queue_as :default
+
+  def endpoint_from_body(html)
```
vivshaw, line 4: helper methods should all be verbatim copies from the old script

app/jobs/send_webmention_job.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Don't check SSL certificate here for backward compatibility, security risk
+    # is minimal.
+    sp.ssl_verify = false
+    sp.fetch(endpoint.to_s, :post, {
+      "source" => URI.encode_www_form_component(source),
+      "target" => URI.encode_www_form_component(target)
+    }, nil, {}, 3)
+  end
+
+  def perform(story)
```
vivshaw, line 61: first change: this now runs on a single story, so there's no longer any need to use `LAST_STORY_KEY`, iterate over multiple stories, etc.

app/jobs/send_webmention_job.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Could have been deleted between creation and now
+    return unless story.persisted?
```
vivshaw: second change: since this is tied to a single story, but async, we need to think about what happens if a story is created and then immediately deleted before the 5 mins are up. (unless I'm mistaken and SolidQueue somehow Does That For You ™️)

app/jobs/send_webmention_job.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Need a URL to send the webmention to
+    return if story.url.blank?
+
+    sp = Sponge.new
+    sp.timeout = 10
+    begin
+      response = sp.fetch(ERB::Util.url_encode(story.url), :get, nil, nil, {
```
vivshaw: third change: this used to use `WEBrick::HTTPUtils.escape`, but WEBrick was removed from stdlib some time ago and isn't in the gemfile either. so I figured, why not use something else you already have?

app/controllers/stories_controller.rb
pushcx: I would like to remove all ActiveRecord hooks with side effects. I'm OK with using them to enforce correctness like initializing an attribute, but I don't like when they touch other records or fire jobs.

You've got a great questions about skipping the job in dev. We definitely don't want to be firing them. I guess we'd like something like how ActionMailer in dev/test puts emails in a known place for inspection and test assertions, but I'm still getting familiar with Active Job and don't know the idiom here.

app/jobs/send_webmention_job.rb
pushcx: The other benefit of `LAST_STORY_KEY` is that it is advanced regardless of whether sending a webmention fails, so we don't forever retry when we have a bug or other servers are down. If the job tries once and stops, great, that's a bit of implicit logic we don't have to maintain.

app/jobs/send_webmention_job.rb
pushcx: Solid Queue doesn't detect that, no. If a story `.is_gone?` when the job fires, it should immediately `return`.

app/jobs/send_webmention_job.rb
pushcx: Great call.

pushcx:
I had a minute and wanted to toss in a couple notes.

BTW, your work moving Mastodon to a job is paying off; because those are properly wired into our exception tracking (where cron errors were yeeted into the outer darkness), I caught and [fixed](https://github.com/lobsters/lobsters/commit/32595ff6b122354e9595504a1aa8c00d1b517c12) a lurking bug. Thanks again!

app/jobs/send_webmention_job.rb
vivshaw: > If the job tries once and stops, great, that's a bit of implicit logic we don't have to maintain.

great, then we should be good to go! unless I'm wildly off the mark, SolidQueue doesn't do retries by default (need to slap a `retry_on` in the ActiveJob if you want that). so this should indeed try it once and stop.

app/controllers/stories_controller.rb
vivshaw: as a first effort, added an early return over in the job, with the other guards at the top of `perform`. so in dev, it _would_ enqueue the job (allowing you to see it happen, inspect params, poke at it in Mission Control), but it would then noop.

> I guess we'd like something like how ActionMailer in dev/test puts emails in a known place for inspection and test assertions, but I'm still getting familiar with Active Job and don't know the idiom here.

I am also still getting familiar with ActiveJob. I'm not sure if there's a per-job way to do something like that. I think if one wanted that behavior globally, one could maybe use [`ActiveJob::QueueAdapters::TestAdapter`](https://api.rubyonrails.org/v8.0.2/classes/ActiveJob/QueueAdapters/TestAdapter.html). but I get the sense that's typically reserved for actual tests?

Gemfile
```diff
@@ -57,6 +57,7 @@ gem "rexml" # no release for https://github.com/lumean/svg-graph2/pull/48/files
 gem "rack-attack" # rate-limiting
 gem "lograge" # for JSON logging
 gem "silencer" # to disable default logging in prod
+gem "webrick", "~> 1.9" # for specifically URL escaping in jobs
```
vivshaw: WEBrick was used by the old `send_new_webmentions` script, but was removed from stdlib several versions ago. I tried replacing with stuff that is already available, like `ERB::Util.url_encode`, `CGI::escape`, or ` URI.encode_www_form_component` but those don't actually do the same thing. They are for encoding URL _params_ and _path components_, so they will escape _all_ the characters in the URL, and give ya something like this:

```
Lobsters(dev)> ERB::Util.url_encode(Story.find(114).url)
=> "https%3A%2F%2Fbeatonma.org%2Fwebmentions_tester%2F"
```

when what we'd prefer is a valid URL like this:

```
Lobsters(dev)> WEBrick::HTTPUtils.escape(Story.find(114).url)
=> "https://beatonma.org/webmentions_tester/"
```

digging a bit on how to get the desired behavior, a few sources [[1](https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape), [2](https://bibwild.wordpress.com/2023/02/14/escaping-encoding-uri-components-in-ruby-3-2/)] suggest "don't do it, handle paths and query strings separately". which is understandable, if not precisely the most helpful answer.

Options I see:
 - bring back WEBrick jsut for this. not ideal, but this gets it working with the same behavior as it once had
 - use `Addressable` like those sources mentioned. (it's already i the lockfile, though not the gemfile)
 - just `URI.parse` and don't worry about escaping? maybe for this use case there's nothing to worry about for some reason?
 - secret fourth thing

vivshaw: aaaaaand "secret fourth thing" it is! I dug into the `URI` repo issues and found a recommendation to use URI::RFC2396_PARSER.escape. and that works exactly like you'd hope:

```
Lobsters(dev)> URI::RFC2396_PARSER.escape(Story.find(114).url)
=> "https://beatonma.org/webmentions_tester/"
```

so, updated to use that and once again remove WEBrick.

app/jobs/send_webmention_job.rb
```diff
@@ -0,0 +1,88 @@
...
+    return if story.is_gone?
+    # Need a URL to send the webmention to
+    return if story.url.blank?
+    # Don't try to send webmentions in dev
+    return if Rails.env.development?
+
+    sp = Sponge.new
+    sp.timeout = 10
+    begin
+      response = sp.fetch(URI::RFC2396_PARSER.escape(story.url), :get, nil, nil, {
```
vivshaw, line 72: third change: as discussed in the other WEBRick-related comments, moves to an escaping util that's already available in stdlib, replacing WEBrick

app/jobs/send_webmention_job.rb
vivshaw: well, turned out it was actually a kinda busted call, it broke the job. that led me to [try reintroducing WEBrick](https://github.com/lobsters/lobsters/pull/1626#discussion_r2153626342), then [realize I didn't need to](https://github.com/lobsters/lobsters/pull/1626/files#r2153736677). and one more yak was shaved smooth as silk that day.

vivshaw:
OK, after a few false starts I _think_ we're now in a good spot. This works to the fullest extent I can test without a staging env, i.e., it can run the job without blowing up, targeting a real site with webmentions, and said job fires the POST that I would expect to the place I would expect. (I can't be 100% because actually confirming that the webmention sticks would require a public URL to the page that's hosting the mention)

app/jobs/send_webmention_job.rb
vivshaw: great, updated to `is_gone?`

pushcx:
I think the stalebot is catching my mistake, that I've let this slip through the cracks. @vivshaw is this ready to review for merge? From skimming the conversation it sure looks complete, and I'm sorry if I've accidentally left it to go stale.

vivshaw:
@pushcx well hey, I _also_ let it slip through the cracks. was outta town for a while, was Planning Season at work for a while, and completely forgot I had never followed up with this. such is life!

I just merged in `master` and tested it locally once again. far as I'm aware, this is indeed ready for review!

app/controllers/stories_controller.rb
pushcx: Looks like the answer on job enqueing... another callback hook, [`enqueue_after_transaction_commit`](https://github.com/rails/solid_queue?tab=readme-ov-file#jobs-and-transactional-integrity) to make sure the objects referred to will exist when the job runs.

I still don't know the answer to skipping in dev/test. I think I saw something in a talk (probably a Margheim talk on ChaoticJob) about having to manually invoke the job runner in test. No better idea for dev.

app/jobs/send_webmention_job.rb
pushcx: Great!

app/jobs/send_webmention_job.rb
pushcx: I don't love that the codebase is gaining _another_ way of parsing/escaping URLs, but I do like that it's in the stdlib rather than something we handrolled...

app/controllers/stories_controller.rb
pushcx: Found the answer for test: https://guides.rubyonrails.org/testing.html#testing-jobs

For dev, please add a `return if Rails.env.development?` It may not be the blessed way, but at least it's obvious what it's doing if we ever figure out what the blessed way is.

pushcx:
Just did another review, there's only two small changes before this is ready to go. Again, I'm sorry I let this hang out so long, thanks for continuing to work on it.

1. Add the `enqueue_after_transaction_commit = true` linked in the 'testing' thread. Which... I think should go up in `ApplicationJob`. It seems like it protects us from a footgun and we should be leaving a comment in each job that wants to opt out of it.
2. Add the `return if dev` check from the same thread.

pushcx:
I didn't want this to wait around, so I made those last couple tweaks in the merge commit.

Thanks for implementing this!

Issue #1623, opened 2025-06-14
Escape tags like <script> in comments


FreeFull:
Requires #1234 (update CommonMarker) first.

Currently, raw HTML tags like `<script>` get replaced by `<!-- raw HTML omitted -->` in the page source, which can get confusing for comments actually trying to discuss the `<script>` tag. It would probably be preferable to turn `<script>` into `&lt;script&gt;`. Example comment where the `<script>` has been stripped away: https://lobste.rs/s/qz1njr/html_spec_change_escaping_attributes#c_yyhrj7

Possible edge cases:
* I do not know if there are tags like `<b>Text</b>` that CommonMarker might intentionally let through
* Existing uses of `&lt;` and `&gt;` shouldn't get double-escaped into `&amp;lt;` or `&amp;gt;`

kivikakk:
fwiw, the behaviour you probably want is setting the `escape` render option in CommonMarker. This will escape all tags in output (including `<b>` etc.) instead of omitting them as is currently done.

All inline HTML tags are either escaped (per `escape`), omitted (if `unsafe_` is unset), filtered by the `tagfilter` extension (if enabled), or output as-is.

The stronger solution is to post-process the output HTML and use an actual sanitisation library, but using `escape` will avoid some confusion for now at least. :)

`&lt;` etc. will be passed through as-is without being double-escaped, since they're recognised and processed.

leregdos:
Seems like #1627  already closed this issue?

FreeFull:
> Seems like [#1627](https://github.com/lobsters/lobsters/pull/1627) already closed this issue?

That PR wasn't merged, seems github-actions closed it out of inactivity

pushcx:
#1627 is now merged, though GitHub didn't recognize the merge. This behavior is now fixed.

PR #1627, opened 2025-06-17
Update commonmarker to 2.3.0 and enable escaping HTML

### LINKED ISSUE #1234: Update CommonMarker
We’re on a years-old version of CommonMarker. It’s changed its API, [dropping](https://github.com/gjtorikian/commonmarker/issues/104) a method we used in `extras/markdowner.rb` to add attributes to links, rewrite some HTML, and make the `@username` feature work. (Heck, they even renamed their constant from `CommonMarker` to `Commonmarker`.)

That issue links to a gem they suggest, but perhaps we should just expand our use of Nokogiri? Or maybe there’s a very different way to tie into the renderer at the time it’s emitting HTML, which I’d be more confident in than re-parsing and fixing things up externally. So this one requires some technical investigation!

chamlis:
This PR updates the site to use the latest version of [Commonmarker](https://github.com/gjtorikian/commonmarker), and switches to escaping raw HTML in the Markdown, rather than replacing it with a HTML comment. Remarkably little code needed to change to pass the existing test suite.

I've marked this as a draft as this is a large and very user-facing change, that also has the possibility for security issues. ~~One thing that would make me feel more comfortable about this is if we could re-enable the CSP header to block inline and external scripts. Reading the comments there I understand we can't yet fully block inline styles, but I think we should take the security benefit for all the other categories if we can.~~ We added a CSP in Report-Only mode in #1631, it only remains to check that policy doesn't block anything we need, and begin enforcing it. 

Closes #1234 and #1623


pushcx:
Thanks for taking this on. I've been really dragging my feet because I couldn't guess how much work it would be. (And I can't immediately find it to link again, but I saw something that looked like our version was getting old enough to run into compatibility issues with new versions of Ruby, so this feels very timely.)

Sure sounds like a healthy dose of caution to me. Do you want to take the CSP header? It's been long enough for that project to fall out of my head, but at a quick `grep` we've removed all the inline `<script>` tags.

I'll pull the branch to run `Comments.find_each { |c| puts c.short_id if c.markeddown_description != CommonMarker.to_html(c.comment) }` to start seeing the scale of changed output.

chamlis:
> Do you want to take the CSP header? It's been long enough for that project to fall out of my head, but at a quick `grep` we've removed all the inline `<script>` tags.

By all means, I'll look into that at some point this week.

> I'll pull the branch to run `Comments.find_each { |c| puts c.short_id if c.markeddown_description != CommonMarker.to_html(c.comment) }` to start seeing the scale of changed output.

That's a great idea, thanks for doing that. Crossing my fingers there's not too much to review.

pushcx:
Two differences, probably related, that hit enough comments that I'd like to address them I don't have to write something that can recognize and ignore known differences:

Some kind of extra quoting:

```
  Comment Load (0.8ms)  SELECT `comments`.* FROM `comments` ORDER BY `comments`.`id` DESC LIMIT 1
=> "<p>“no one” is a bit of a stretch, but it certainly can feel like it at times.</p>\n"
Lobsters(dev)> Commonmarker.to_html(Comment.last.comment)
  Comment Load (0.9ms)  SELECT `comments`.* FROM `comments` ORDER BY `comments`.`id` DESC LIMIT 1
=> "<p>&quot;no one&quot; is a bit of a stretch, but it certainly can feel like it at times.</p>\n"
```

And then this "fun" one:

```
Lobsters(dev)> c.markeddown_comment
=> "<p>very engaging post, made me feel an odd sense of nostalgia for something that I wasn’t even around for the first time.</p>\n"
Lobsters(dev)> Commonmarker.to_html(c.comment)
=> "<p>very engaging post, made me feel an odd sense of nostalgia for something that I wasn't even around for the first time.</p>\n"
Lobsters(dev)> c.markeddown_comment == Commonmarker.to_html(c.comment)
=> false
```

I wish GitHub had spoiler tags for this one. Argh.

```
Lobsters(dev)> c.markeddown_comment.codepoints.join(' ')
=> "60 112 62 118 101 114 121 32 101 110 103 97 103 105 110 103 32 112 111 115 116 44 32 109 97 100 101 32 109 101 32 102 101 101 108 32 97 110 32 111 100 100 32 115 101 110 115 101 32 111 102 32 110 111 115 116 97 108 103 105 97 32 102 111 114 32 115 111 109 101 116 104 105 110 103 32 116 104 97 116 32 73 32 119 97 115 110 8217 116 32 101 118 101 110 32 97 114 111 117 110 100 32 102 111 114 32 116 104 101 32 102 105 114 115 116 32 116 105 109 101 46 60 47 112 62 10"
Lobsters(dev)> Commonmarket.to_html(c.comment).codepoints.join(' ')
=> "60 112 62 118 101 114 121 32 101 110 103 97 103 105 110 103 32 112 111 115 116 44 32 109 97 100 101 32 109 101 32 102 101 101 108 32 97 110 32 111 100 100 32 115 101 110 115 101 32 111 102 32 110 111 115 116 97 108 103 105 97 32 102 111 114 32 115 111 109 101 116 104 105 110 103 32 116 104 97 116 32 73 32 119 97 115 110 39 116 32 101 118 101 110 32 97 114 111 117 110 100 32 102 111 114 32 116 104 101 32 102 105 114 115 116 32 116 105 109 101 46 60 47 112 62 10"
```

If you stare at this long enough, you will see a `8217` in the first value and a `39` in the second. Our current version of CommonMarker has been doing 'smart punctuation', transforming an ASCII apostrophe into a curly quote. Probably it should not have been on, but I'd like to turn it back on for this test. Maybe this is related to the `"` escaping that we want to turn off in the first diff?

chamlis:
I think in those examples you're calling `Commonmarker` directly, rather than via our `Markdowner` which enables smart-punctuation. I'm getting the curly quotes locally:
```
Lobsters(dev)> Markdowner.to_html(" I wasn't ... \"no one\"")
=> "<p>I wasn’t … “no one”</p>\n"
```

pushcx:
...yup, that's my mistake. That corrects the apostrophe, and the example I gave of the "no one" comment.

And then I see the inverse behavior where https://lobste.rs/c/gmoyyU transforms from `"<p>Depends &ndash; is security hip today?</p>\n"` to `"<p>Depends - is security hip today?</p>\n"`. I prefer the latter, but it makes comparison hard.

```
Lobsters(dev)> short_ids = Comment.all.find_each.map { |c| c.short_id if c.markeddown_comment != Markdowner.to_html(c.comment) }.compact; nil
nil
Lobsters(dev)> short_ids.count
=> 112427
Lobsters(dev)> Comment.count
  Comment Count (125.0ms)  SELECT COUNT(*) FROM `comments`
=> 586860
```

I think I have a few moles to whack.

chamlis:
What's the Markdown for that comment? From reading the code and testing some things out, it seems both the old and new Markdowners will turn `--` into an en-dash.

pushcx:
```
Lobsters(dev)> puts (Comment/:gmoyyU).comment
  Comment Load (11.1ms)  SELECT `comments`.* FROM `comments` WHERE `comments`.`short_id` = 'gmoyyU' LIMIT 1
Depends - is security hip today?
=> nil
Lobsters(dev)> puts (Comment/:gmoyyU).comment.codepoints.join(' ')
  Comment Load (0.6ms)  SELECT `comments`.* FROM `comments` WHERE `comments`.`short_id` = 'gmoyyU' LIMIT 1
68 101 112 101 110 100 115 32 45 32 105 115 32 115 101 99 117 114 105 116 121 32 104 105 112 32 116 111 100 97 121 63
=> nil
Lobsters(dev)>
```

chamlis:
Took some historical spelunking to figure out this one! That's an old comment, and (as far as I can tell), the last time the `markeddown_comment` field was updated across all comments was in 342d6ff911b2455b9d4605ee98d068b46938c1de (when the column was first added). At that time, the code used [RDiscount](https://github.com/davidfstr/rdiscount) for Markdown conversions, which in the version used then [converted a single `-` to an en-dash](https://github.com/davidfstr/rdiscount/blob/1.6.8/ext/generate.c#L993). The current Commonmarker implementation should produce the same (non en-dashed) output as from this branch, it just never ran on that comment.

We switched to Commonmarker in fa70ab1 - maybe it's worth only comparing comments after this date?

pushcx:
I spent a few more minutes on this on [today's stream](https://push.cx/stream/2025-06-26).

There's a harmless change from 8y ago where we added `rel="nofollow"` to links.

There's a regression where the new version is inserting a `<br>` for a linebreak in a paragraph. The source of [yvcey9](https://lobste.rs/c/yvcey9) exercises this, where the new CommonMarker is inserting a `<br>` after `based.`:

```
> A modular, capability-based operating system

Ooh, you got me at capability-based.
This alone makes me very excited :)
```

There's a regression where code is picking up some kind of styling, maybe syntax highlighting? [5r3cxm](https://lobste.rs/c/5r3cxm):

```
> Why couldn’t it just tell you when it got merged?

I don’t know why GitHub doesn’t, but Git can:

    $ git tag --contains b15024d6a1537c69fc446601559a89dc8b84cf6f
    v1.8.6

That doesn’t address the cherry-picking case though. I’m not aware of any built-in tooling for that. Generally Git avoids relying on metadata for things that can be inferred from the data (wit
h file renames being the poster child of the principle), so I’m not surprised that cherry-picks like this aren’t tracked directly. Theoretically they could be inferred (i.e. it’s “just” a matt
er of someone building the tooling), but I’m not sure that’s doable with a practical amount of computation. (There are other operations Git elects not to try to be fast at (the poster child be
ing `blame`), but many of them still end up not being impractically slow to use.)
```

After thinking through these some more, I need to build tooling to comparing these and mark specific changes (like `&mdash` -> `-` or adding `rel="nofollow"`) as acceptable. Then keep iterating on recognizing changes as acceptable or recognizing them as regressions and address them. Comments are the most important part of the site, I have to take proper care that we don't accidentally make significant changes.

chamlis:
Thanks for those notes. I've force-pushed to disable syntax highlighting and the hard line-breaking. Looking through the rest of the Commonmarker README, I don't think anything else should need disabling (famous last words...).

Is it worth comparing the output of this branch with the output from master (by dumping to a file and diffing, or similar)? It'd sidestep any issues with the evolution of the code, though I guess it depends if dropping the cached column and rendering historic comments with the new code is on the cards, and if we want to avoid changing how they'd display.

pushcx:
Thanks. This one is on me right now, then. I need to pull your branch, compare the cached renders to the new ones, expand the tests to include some comment output, and then write up a punch list of anything left before we merge your PR.

chamlis:
One issue that came up on office hours is that www. is automatically turned into a link with this branch. By my reading of the spec, it shouldn't be, but [this issue](https://github.com/github/cmark-gfm/issues/197) points out an alternative interpretation of the spec. As this comment demonstrates, GitHub considers it a link too, and it's their flavour of markdown. I don't know if it's worth trying to get in touch with any upstreams about this.

pushcx:
The behavior is wrong. The produced link implies an implausible, confusing DNS setup and I can't imagine a spec would want to support that strange situation.

I think the bug report you linked got lost in the weeds of considering whether the spec could be tortured to support it, and the age implies people are bouncing off fixing the bug because of it. I think you should file a new bug, not linking to that one or this thread to avoid the distraction, and then after the bug is fixed, comment that 197 can be closed. It's a little mercenary, but it stalled for 5 years.

All that said, this bug is small enough that I don't consider it a blocker.

So... not that I should've need it, but as one more example of how breaks and downtime are essential, an hour after the stream I realized I could keep the useful diffs but eliminate the newline distractions by changing my comparison to `if markeddown.delete("\n") == new_rendered.delete("\n")` but then diff the two versions without deleting the `\n`.

I picked up from 2017-04-12 when the switch was made to Markdowner. A few more tweaks got it through > 20% of comments and each run jumps over a bigger gap. So I'm still tinkering there.

I did find [one comment](https://lobste.rs/c/dd9vih) that looks like a bug:

    Lobsters(dev)> Markdowner.to_html("on a related note, [neocities](https//neocities.org) is also highly recommendable ")
    => "<p>on a related note, <a href=\"https//neocities.org\">neocities</a> is also highly recommendable</p>\n"

The link generated is missing the `rel="ugc"` tag. I suspect there's some kind of lookalike unicode character in this; if I retype the input string, the link has the correct attribute.

pushcx:
Second bug, parse of a [relative link](https://lobste.rs/c/qm3img) also misses the `rel="ugc"`:

    Lobsters(dev)> Markdowner.to_html "that [oilshell guy](/u/andyc)?"
    => "<p>that <a href=\"/u/andyc\">oilshell guy</a>?</p>\n"

chamlis:
I agree the behaviour isn't really meaningful - another example of me getting caught up in spec minutiae and losing track of the forest, I fear. I'll open another issue on that repo a bit later, but it will probably take a while for any fixes to reach us - first cmark-gfm would need to be fixed, then those changes reflected in comrak, and finally a new release of commonmarker for us to use. Thankfully, I think that's about the only URL that could be affected by this, so it might be possible to fix that up in our wrapper.

**Edit**: Production currently has the `www` issue as well, so hopefully it's not very prevalent.

As to the `ugc` issues, I think we should have had the same behaviour before? The intent seems to be that internal Lobsters links don't get the attribute (which first started happening [here](https://github.com/lobsters/lobsters/commit/6894ac51168e6f5688969c5e1da11e19287535b4)). In the first case, there's a missing `:` after `https` which results in a relative URL, as is clearer in the latter example. I'm certainly not opposed to dropping that condition and adding the attribute to every link, but I'd like to make sure I'm not knocking down this particular fence prematurely.

pushcx:
The Neocities link is external and definitely a bug. I also turned up the `my websites` link in [this comment](https://lobste.rs/c/pqimc7) that doesn't get `rel="ugc"` added by the new renderer. I'd like to have a regression test of one or both of these because I am pretty surprised by this bug.

chamlis:
I think we might be seeing different behaviour? If I go to [this comment](https://lobste.rs/c/dd9vih) or [this one](https://lobste.rs/c/pqimc7), the links in them are to 404s on Lobsters, because they're parsed as relative URLs. In the first case, the `https` scheme is missing the `:`, and in the second case the `https` scheme is missing one of the two `/`s.

pushcx:
Bug: [this comment](https://lobste.rs/c/oqccb3) has a strikethrough but the new render parses it into a code block and truncates it.

source:

```
~~~I think the difference is over who holds the copyright. IIUC, you can _assign_ copyright permanently, but you can't irrevocably _license_ a work. So when you write code for a GNU project, the FSF is the copyright holder. But when you write code for Linux, you retain the copyright over your own contributions (and can therefore rescind licenses to them). In the former case, the FSF _can_ rescind the license since they are the copyright holders; they just probably _won't_, whereas big amalgamations of various people's copyrights (e.g. Linux) are somewhat more risky in this regard.~~~

Edit: Actually it seems like you also can't irrevocably transfer copyright.
```

new render:

```
<pre lang="I"><code>
Edit: Actually it seems like you also cant irrevocably transfer copyright.
```

pushcx:
Aha, I see what you mean with those two comments. I was so focused on the missing attribute that I didn't notice they had typos that transformed them into relative links. Not a bug, then.

chamlis:
> Bug: [this comment](https://lobste.rs/c/oqccb3) has a strikethrough but the new render parses it into a code block and truncates it.

The current Commonmarker implementation also misses out the first paragraph, so there was a change in how it parsed it between 2018 and the current production version. I spent about half an hour trying to bisect it, but wasn't able to get the older versions of Commonmarker running on newer Ruby versions. At a quick glance, though, it looks like that comment was relying on [this bug](https://github.com/github/cmark-gfm/issues/118) to render correctly. That bug has been resolved in the Commonmarker now in production, and the comment is parsed as a fenced code block with a long info string.

I guess that's the abstract question here - if extant comments are depending on bugs in the Markdown renderer from the time they were written, how far do we want to go to preserve that rendering?

chamlis:
Still going - it's far more laborious to review than it was to develop.

pushcx:
I merged this in 236bce18 but GitHub doesn't recognize it.

@chamlis Sorry this took so long. I finally got through all the comments. I have two comments to manually fix, then there are two minor cleanup issues that I'll file now.

pushcx:
So I have it somewhere, I manually refreshed the comments with:

```
Comment.all.find_each { |c| c.update_columns markeddown_comment: Markdowner.to_html(c.comment) }
```

At some point we should probably replace the `markeddown_` columns with fragment caching.

chamlis:
Thanks for merging this, and for all of your patience in reviewing it!

PR #1629, opened 2025-06-18
Reinstate `/top/rss` and allow RSS for any `/top/…/rss` path

### LINKED ISSUE #1602: /top/rss stopped working
The /top/rss endpoint stopped working as of yesterday.

caius:
## What?

- [x] Added routing specs to ensure we end up in the right action with correct params
- [x] Wire up any `/top/…/rss` path to controller endpoint
- [x] Add controller specs for HTML/RSS feeds
- [x] Fix redirect bug in `home#top` when accessing with `format: "rss"`

## Why?

Fixes #1602, and adds additional functionality for the top feeds because its easier to do that than only allow the `/top/:length/rss` feeds.

Technically `/top/rss` isn't a valid feed anymore, but it redirects to `/top/1w/rss` which is now a valid feed and any well behaved feed reader should handle that. The upside is you can subscribe to a feed of any length you want now, eg `/top/1m/rss` or `/top/3d/rss`.

Closes #1606 I think, one of the tests is incorrect and the other is better tested in the routing/controller specs here.

### DIFF
--- a/spec/controllers/home_controller_spec.rb
+++ b/spec/controllers/home_controller_spec.rb
@@ -125,4 +125,94 @@ describe HomeController do
       expect(@controller.view_assigns["stories"]).not_to include(other_story)
     end
   end
+
+  describe "#top" do
+    let(:old_story) { create(:story, created_at: 10.days.ago, updated_at: 10.days.ago) }
+    let(:recent_stories) do
+      # Create descending by score so they are sorted in this order from controller
+      num = StoriesPaginator::STORIES_PER_PAGE + 5
+      Array.new(num) { |n| create :story, score: (num - n) }
+    end
+
+    before do
+      story
+      old_story
+    end
+
+    describe "/top redirect" do
+      it "redirects to default length path" do
+        get :top
+
+        expect(response).to redirect_to("/top/1w")
+      end
+    end
+
+    describe "/top/rss redirect" do
+      it "redirects default length path maintaining format" do
+        get :top, format: "rss"
+
+        expect(response).to redirect_to("/top/1w/rss")
+      end
+    end
+
+    describe "/top/:length" do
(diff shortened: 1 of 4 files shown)


spec/controllers/home_controller_spec.rb
```diff
@@ -125,4 +125,94 @@
       expect(@controller.view_assigns["stories"]).not_to include(other_story)
     end
   end
+
+  describe "#top" do
+    let(:old_story) { create(:story, created_at: 10.days.ago, updated_at: 10.days.ago) }
```
pushcx, line 130: Change these `let` to `let!` and you can delete the `before` that references them.

caius: Whoops, didn't realise that was convention here. Cheers for fixing

spec/controllers/home_controller_spec.rb
```diff
@@ -125,4 +125,94 @@
...
+        expect(@controller.view_assigns["rss_link"]).to eq(
+          title: "RSS 2.0 - Top Stories of the Past Week",
+          href: "/top/1w/rss"
+        )
+        expect(@controller.view_assigns["stories"]).to include(recent_stories[StoriesPaginator::STORIES_PER_PAGE + 1])
+        expect(@controller.view_assigns["stories"]).not_to include(old_story)
+      end
+    end
+
+    describe "/top/:length/page/2/rss" do
```
pushcx, line 204: Why did you add rss urls for each page? RSS is for fetching the latest updates?

caius: Was easier to add than not in the controller code, and was additional functionality we didn't support before. We could back it out without breaking anyone's workflow I'm sure.

Issue #800, opened 2020-01-01
email bounces should disable future email notification


alanpost:
Users can elect to receive email notification for comment replies, comment mentions, and private messages.  If the provided email address begins bouncing, we should turn off notification for that user.  Further it would be nice to send them a private message indicated we modified their preferences and explaining why.

Bounces are returned to our no-reply address, and we have an existing script for mailing list mode that could be amended to handle non-deliverable addresses.

hmadison:
In case anyone wants to pick up and work on this, the script in question is [`script/parse_inbound_mail`](https://github.com/lobsters/lobsters/blob/master/script/parse_inbound_mail). Its used [by postfix](https://github.com/lobsters/lobsters-ansible/blob/335992067cd6fa0a604c0dce9e25801639a81c29/roles/postfix/templates/etc/postfix/master.cf#L125-L126) to forward all incoming mail to the application. To the best of my understanding, a patch resolving this issue would do the following: 

- Parse any incoming bounce notifications and attempt to map them back to users (via their email).
- For each bouncing user disable all of their [email related](https://github.com/lobsters/lobsters/blob/master/app/models/user.rb#L56-L75) settings 
- It would also be nice to send them [a message](https://github.com/lobsters/lobsters/blob/1b448d916e68f4e116a0899d2c71768da94560aa/app/models/message.rb) from System letting them know their email has been bouncing.

pushcx:
To update @hmadison's pointer, we're now using ActionMailbox, so this is a little simpler and more rails-y, and certainly easier to test.

Something that's occurred to me since we posted it is that we should record outgoing `Message-Id`s or recognize bounces by the email being `Reply-To: lobsters-<valid key>@` so that an attacker cannot spoof bounces to forcible unsubscribe people.

PR #1630, opened 2025-06-18
postfix: added postscreen - pre-SMTP client screening

wojtekregis:
Enables basic Postfix-native "postscreen" functionality to shield SMTP from... mostly just crappy bots and botnet peers running on consumer hardware. This is not an anti-SPAM functionality but rather a shield from garbage mail sent from vulnerable consumer devices, vulnerable WordPress hosting account and such. A reasonable first line of defense when using Postfix.

### DIFF
--- a/hatchbox/main.cf
+++ b/hatchbox/main.cf
@@ -17,7 +17,6 @@ smtpd_tls_key_file=/home/deploy/.local/share/caddy/certificates/acme-v02.api.let
 smtpd_tls_CAfile=/etc/ssl/certs/ca-certificates.crt
 smtpd_tls_security_level=may
 smtpd_use_tls=yes
-postscreen_tls_security_level=may
 smtpd_tls_session_cache_database = btree:${data_directory}/smtpd_scache
 smtp_tls_loglevel = 1
 smtp_tls_security_level=may


wojtekregis:
I suggest postponing deployment of any resource-hungry SPAM-fighting measures and acting upon results of measures deployed earlier. A gradual increase of SPAM-fighting defensive measures with a pressure on avoiding over-killing it. With no maillog access I'm mostly guessing here.

pushcx:
To back up a little, could you say more about the benefit? Our incoming mail is configured to basically drop everything except that addressed to `lobsters-(valid mailing list key)@`. Postfix drops anything that's not well-formed before we take the computation of booting Rails and checking for that key in the database.

So if we're dropping everything else already, does this get us much?

We do need better handling of bounces in the future - like, right now, the app doesn't recognize if someone has mailing list mode enabled and closes their email address. So maybe this will help when we start looking at those? (xref #800)

wojtekregis:
@pushcx You mentioned having incoming SPAM problem. This PR introduces Postfix-native functionality which does not deal with SPAM on content basis, but deals with an earliest stage of client-server connection, before resources are spent on the client-server SMTP chat. 

Postscreen cuts low-effort spambots. It is not dealing with later SMTP session - envelope, From/To addresses nor other headers, message body etc. Just the connection stuff.

Example:

```
> nc -v sysartist.com 25
Connection to sysartist.com 25 port [tcp/smtp] succeeded!
220-sysartist.com ESMTP Postfix
220 sysartist.com ESMTP Postfix
```

The first "220-whatever" is an improper server response. Low-effort-coded bots go YOLO and start throwing SMTP commands at this point. Normal RFC-based clients wait for the "220 whatever" line (notice the lack of dash) and only then start talking SMTP to the server. There's a delay of few seconds between the first fake line and second SMTP-valid line but it's much lower delay than 5-15 or so minutes you have with "graylisting". 

Following connections are whitelisted automatically. Notice the "220-" line no longer present in servers' response.

```
> nc -v sysartist.com 25
Connection to sysartist.com 25 port [tcp/smtp] succeeded!
220 sysartist.com ESMTP Postfix
```

It's reasonable to introduce native functionality - postscreen - before deploying heavy, resources-hungry solutions like SpamAssassin or rspamd.

This is not in any way related to "bounces". Bouncing means that the server doesn't accept messages from a client during the SMTP connection. Client server informs their "local" sending party that the message could not have been sent.

When recipient's MTA accepts a message and is unable to deliver it, it stores it and returns it to the sender after days or weeks - that's not a bounce but just shitty approach. One of SMTP-related mantras is to never accept messages you can't/won't deliver.

pushcx:
OK, thanks. I guess I have a small question and a larger one.

Small: What does the `127.0.0.2*3` syntax do? I know the IP for localhost, but this is a not quite that. In the linked readme the `*2` is a weight, and I'm guessing this set of weights is based on your experience?

Larger: Does this fail open? What happens to our setup if one of these blocklists goes away? Is there a protocol standard here that gets revised occasionally? I ask because we won't immediately notice any kind of change to update our config. Or to put it another way: will this config be resilient in the face my inexpertise and benign neglect?

wojtekregis:
> What does the 127.0.0.2*3 syntax do? 

DNSBL / DNSWL are (reasonably so) using DNS as key-value store. IP addresses are used for compatibility. This stuff is less prone to be broken by some implementations than TXT records. Responses and classification may differ between providers but the usual translation of DNSBL codes to these IP addresses is:

``` 
127.0.0.2  = Confirmed spam source
127.0.0.3  = Dial-up/dynamic IP  
127.0.0.4  = Open proxy
127.0.0.5  = Open SMTP relay
127.0.0.9  = Policy block (PBL)
127.0.0.10 = Compromised machine
127.0.0.11 = CSS (hacked website)
```

Different "IP" responses from any single list can be configured with different weights multiple times locally.

> I'm guessing this set of weights is based on your experience?

Yes but it's an accordingly tuned-down slice of my opinionated setup. More on the safe side, not to screw up lobsters' incoming email flow. The idea - for now - is to slightly(!) limit SPAM with zero impact on legitimate emails.

> What happens to our setup if one of these blocklists goes away?

Overall score for given incoming message is not impacted. Inspect logs, remove not responding hosts from config, possibly tune scoring (rarely needed unless the majority of these DNSBL/DNSWLs are unreachable), and you should be good.

-- 

@pushcx You asked me about fighting incoming SPAM on IRC and this is what I believe is the bare minimum I see missing for the time being. I have not seen these incoming messages' headers and "how legit" they are (spammers tend to get SPF, DKIM, DMARC, FCrDNS and whatnot configured better than legitimate senders - this solution rarely works for this kind of spam) but my best guess is that Postfix-native [postscreen](https://www.postfix.org/POSTSCREEN_README.html) is the first measure to implement, before jumping into SpamAssassin or other heavier solutions.

I have little doubt one of these will be required to finally get your inbox clean, but the basic resource-light Postfix mechnisms are the first step before you throw remaining messages at `spamd`-like heavy-duty stuff.

Happy to answer questions here or on IRC. I'll join tomorrow's Office Hours but - sorry - I'm not willing to register an account on Twitch which would allow me to comment there. I'm dropping all my Amazon accounts and AWS is the single one left to off-board...

pushcx:
Great, I understand what this is doing and am really glad to have a clear reference available for when I look back at this in a year or two. Thanks!

wojtekregis:
@pushcx Please note that the list of DNSWL/DNSBL hosts and weights may call for adjustments as time flows.

pushcx:
My last comment was unclear: this is merged and deployed now. I was trying to joke about how I'll surely forget everything in a year or two when circumstances* require, but wojtekregis's explanation in these comments will make everything clear.

 \* This means bugs caused by my own inexperience.

PR #1631, opened 2025-06-19
Add a CSP header in report-only mode

chamlis:
This is a tighter policy than the one commented out before. Due to the JavaScript importmap counting as an inline script, we need to use nonces to allow that without also allowing all inline scripts.

This switches to a random nonce rather than deriving it from the session. The benefit of the session ID approach is to enable conditional caching, but the site does not make use of this. Using a random nonce reduces the chance of security issues (for example, logging in currently doesn't reset the session as far as I can see) and means we can generate nonces for guest requests who hit Rails.

For guest users who hit the cache, we need to set a CSP in Caddy as well. As we can't use nonces here, I've opted to disallow scripts altogether. This approach fails an edge case on the signup page, where a guest user is served JavaScript for the markdown formatting dropdown. In this situation, that script and functionality won't work. I've marked this PR a draft for that reason, so we can discuss how we want to handle that.

For now this is all using report-only mode so that any oversights don't break the site for users.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -7,21 +7,25 @@
 # https://guides.rubyonrails.org/security.html#content-security-policy-header
 
 Rails.application.configure do
-  #  18 inline styles to clean up before enabling this
-  #  config.content_security_policy do |policy|
-  #    policy.connect_src :self, :https
-  #    policy.default_src :self, :https
-  #    policy.font_src :self, :https, :data
-  #    policy.img_src :self, :https, :data
-  #    policy.object_src :none
-  #    policy.script_src :self, :https
-  #    policy.style_src :self, :https
-  #    policy.report_uri "/csp-violation-report"
-  #  end
+  config.content_security_policy do |policy|
+    policy.default_src :none
 
-  # Generate session nonces for permitted importmap, inline scripts, and inline styles.
-  config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
-  config.content_security_policy_nonce_directives = %w[script-src style-src]
+    # Needed for AJAX calls, mini-profiler
+    policy.connect_src :self
+
+    # Data URL used for Pushover logo in settings
+    policy.img_src :self, :data
+    policy.script_src :self
+
+    # 18 inline styles to clean up before enabling this
+    policy.style_src :self, :unsafe_inline
+
+    policy.report_uri "/csp-violation-report"
+  end
+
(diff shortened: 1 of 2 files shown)


pushcx:
For the markdown helper, can we do `<details>/<summary>` or the checkbox hack?

In the CSP controller, we currently record violations to the mail rails log where they vanish into the infinite haystack of hits. It'd be better to use our new Telebugs setup here. We can [log a message](https://docs.sentry.io/platforms/ruby/guides/rails/), I'll add it to our wrapper.

chamlis:
I've added a Telebugs message on CSP violations. I haven't tested that locally, but have added specs for the input validation. I've removed the check that something is logged - I can't see a way to assert that Telebugs/Sentry did receive a message, but I may have missed something.

The input validation doesn't feel like idiomatic Rails to me. I tried using route constraints for the `Content-Type` check, but that made the specs fail at finding a route rather than giving a useful error code. I'm keen to know how I could write this better.

pushcx:
It's our new wrapper around Telebugs' new feature used for logging exceptions that hopefully never happen, I'm OK with YOLOing this one into prod to see what happens rather than trying to set up some kind of test harness.

Rails routing is not expressive enough to handle the Content-Type header. The only idiomatic improvement here is that you could write `return head :bad_request unless request.content_type == "application/csp-report"`. The json check a couple lines down could also use the nil propagation operator but that doesn't have syntax for hash/array access so you'd have to write `unless json&.public_send(:[], 'csp-report')&.is_a?(Hash)`, which is a little distractingly unclear.

chamlis:
Thanks for the reminder about using `unless` as a modifier, that's a Rubyism that doesn't come naturally to me at the minute. I've combined all of the validation, and then standard/rubocop were able to merge two of hthe checks with nil-propagation in a way that doesn't look too bad to me.

pushcx:
Great. I should probably admit that `&.` is officially called the "safe navigation operator" but I call it the "nil propagation operator" because I am not the biggest fan of nullable types.

I guess the last thing to do here is see if we can remove the script from the markdown toggle. Did you want to try that or should I?

chamlis:
> Great. I should probably admit that `&.` is officially called the "safe navigation operator" but I call it the "nil propagation operator" because I am not the biggest fan of nullable types.

Good to know, thanks.

> I guess the last thing to do here is see if we can remove the script from the markdown toggle. Did you want to try that or should I?

#1636 :wink:

pushcx:
I hadn't seen that, thanks. Just left a quick review there, and I'll merge these two at the same time.

PR #1632, opened 2025-06-20
Statistics collection script

thomasdziedzic:
Performance has come up a couple of times as a topic of interest in the #539 and now as part of #1076 . I realized I don't really have a good way to test and see performance issues locally. The fake_data task isn't enough since it produces 0 rows in the mentioned view.

I wrote up a script that I hope you could review/run on lobsters and paste the stats output somewhere so that I could come up with a more realistic data set for my testing.

run with `rails data_stats` and the output file will be under /tmp/stats.json

This would allow me to work on a script that would generate more realistic data so that I can see the performance problems locally for debugging and testing.

### DIFF
--- a/lib/tasks/data_stats.rake
+++ b/lib/tasks/data_stats.rake
@@ -18,15 +18,57 @@ task data_stats: :environment do
 
         case column.sql_type_metadata.type
         when :binary
-          descendant.connection.exec_query("select avg(length(`#{column.name}`)) avg_length, stddev_pop(length(`#{column.name}`)) stddev_pop_length, count(`#{column.name}`) column_count from `#{descendant.table_name}`").to_a.first
+          descendant.connection.exec_query(
+            <<~SQL
+              select
+                avg(length(`#{column.name}`)) avg_length,
+                stddev_pop(length(`#{column.name}`)) stddev_pop_length,
+                count(`#{column.name}`) column_count
+              from `#{descendant.table_name}`
+            SQL
+          ).to_a.first
         when :boolean
-          descendant.connection.exec_query("select count(case `#{column.name}` when 0 then 1 end) count_false, count(case `#{column.name}` when 1 then 1 end) count_true, count(`#{column.name}`) column_count from `#{descendant.table_name}`").to_a.first
+          descendant.connection.exec_query(
+            <<~SQL
+              select
+                count(case `#{column.name}` when 0 then 1 end) count_false,
+                count(case `#{column.name}` when 1 then 1 end) count_true,
+                count(`#{column.name}`) column_count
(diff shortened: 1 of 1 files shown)


pushcx:
Sure, glad to run it!

I'm pretty sure it would be ok to gate particular users to see the mini-profiler tab in prod. It leaks PKs of tables and I try not to publish those to avoid making low user IDs or round number comment IDs into a status symbol. But with the `env` toggled off I don't think there's a security concern. It shows queries but not results, so a user could see their own session token and maybe password hash, but not dumps of other user objects that would reveal tokens. @chamlis any thoughts? Maybe it's info you'd be interested in as well.

pushcx:
[stats.json](https://github.com/user-attachments/files/20839091/stats.json)

Might be worth linking [/stats](https://lobste.rs/stats) for context on how activity has grown and shrunk over time. And that averages are usually misleading in social software; almost all user activity is log-normal distributed: number of votes cast, received, stories submitted, flags received, etc.

thomasdziedzic:
> [stats.json](https://github.com/user-attachments/files/20839091/stats.json)

Thanks!

> Might be worth linking [/stats](https://lobste.rs/stats) for context on how activity has grown and shrunk over time.

That stats page is useful to see the high level trends, again, thanks!

> And that averages are usually misleading in social software; almost all user activity is log-normal distributed: number of votes cast, received, stories submitted, flags received, etc.

Good to know, I did find https://www.johndcook.com/blog/2022/02/24/find-log-normal-parameters/ which I think means I can derive the necessary variables if need be given the existing data that has been collected though I would need to verify this. The log normal distribution idea sounds like a great way to simulate social activity over time for a more realistic dataset. I might give it a shot when generating the data if it turns out to be relatively easy.

thomasdziedzic:
I hit a bit of a snag creating the script, it is extremely slow in creating rails records.

I initially tried creating records the usual rails way via `create!` and friends, but I was getting single digit record insertions per second, and when I wanted to create 10s of thousands or 100s of thousands for some classes, this was going to take many hours to complete running.

I then attempted to use `insert_all!` but this bypasses lifecycle hooks, an important part of getting the site working bug free. And the site doesn't really render properly without those extra hooks running. I would then at this point be chasing things to fix in the site. The only good news though that I have is that this is much faster, measured in seconds to insert 100s of thousands of records.

In the past I relied on the bulk `insert_all!` strategy when I wanted to create many records in the database at once, though those apps didn't rely so heavily on the lifecycle hooks to have a working app.

Do you have any ideas on how to create a performant script and keep the site working? If not, I might just have to split out the script into many scripts so that I can make progress in small increments, and eat the cost of waiting for activerecord to do it's thing on a per record basis.

PR #1633, opened 2025-06-23
Address HTML markup issues reported by Herb

marcoroth:
This commit addresses the low-hanging fruit discovered by Herb when analyzing the codebase. This analysis also revealed a lot of cases where the Herb parser falls short. So once I've fixed those upstream I'm planning on coming back to fix the remaining ones.

Let me know if you need anything updated.

### DIFF
--- a/app/views/categories/_multi_category_tip.html.erb
+++ b/app/views/categories/_multi_category_tip.html.erb
@@ -1,4 +1,4 @@
 <%# locals: () -%>
 <p class="hint">
   Tip: read stories across multiple categories with <code>/categories/foo,bar</code>
-</p >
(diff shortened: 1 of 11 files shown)


pushcx:
Thanks for catching these. Please do `#mention` this PR (or `@pushcx`) when you improve Herb, I'm always thrilled to see how the codebase can be reused in these interesting ways.

Thanks for the fixes. It's interesting to see how many of them are `</p>` tags; I wonder if it's a consequence of our style quirk of not indenting their contents.

ETA: Also, I do hope you write a formatter that is as opinionated and option-free as standardrb.

PR #1634, opened 2025-06-26
Refactor StoryRepository: newest_by_user method

nav-mike:
removed `newest_by_user` method

### DIFF
--- a/spec/models/story_spec.rb
+++ b/spec/models/story_spec.rb
@@ -597,6 +597,53 @@ describe Story do
         expect(Story.saved(user)).to eq([first_story])
       end
     end
+
+    describe "newest_by_user" do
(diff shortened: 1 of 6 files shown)


spec/models/story_spec.rb
```diff
@@ -597,6 +597,53 @@ def fake_response(content, type, code = "200")
         expect(Story.saved(user)).to eq([first_story])
       end
     end
+
+    describe "newest_by_user" do
+      let(:user) { create(:user) }
+      let(:submitter) { create(:user) }
+
+      def story_titles_from(submitter:, basic: user)
```
pushcx, line 605: Clever little helper here, nice way of making tests readable!

PR #1635, opened 2025-06-28
add support for notifications 

thomasdziedzic:
Work for https://github.com/lobsters/lobsters/issues/1076

lobsters_wip

Currently notifications are isolated from the other parts of the site like unread messages/unread comments because I'm not yet sure what the migration / integration will look like so I'm going to leave it open for now. This is why you might see "1" unread at the top, but the screenshot has 5 unread notifications.

For messages, the bold text under the byline is the subject, to differentiate it from the body.

Clicking on unread works as expected for both messages and comments.

Added pagination to all notifications, but not the unread, since we clear them out after clicking on the tab.

I might have found a bug while taking some of the pagination code from the replies. The replies controller limits to [REPLIES_PER_PAGE](https://github.com/lobsters/lobsters/blob/master/app/controllers/replies_controller.rb#L17) but the replies view only shows the pagination if there are [more than REPLIES_PER_PAGE replies](https://github.com/lobsters/lobsters/blob/master/app/views/replies/show.html.erb#L22). I tested this out locally and I couldn't get the pagination to display when set to a small number like 2.

Looking for lots of feedback, especially around how you would like to rollout this feature.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -13,7 +13,7 @@ class NotificationsController < ApplicationController
       .offset((@page - 1) * notifications_per_page)
       .limit(notifications_per_page)
       .order(created_at: :desc)
-      .includes(notifiable: {story: [:tags, :user], user: [], author: []})
+      .includes(user: [:hidings, :votes], notifiable: {story: [:tags, :user], user: [:comments], author: [], parent_comment: []})
 
     @has_more = @user.notifications.count > (@page * notifications_per_page)
 
@@ -28,7 +28,7 @@ class NotificationsController < ApplicationController
       .notifications
       .where(read_at: nil)
       .order(created_at: :desc)
-      .includes(notifiable: {story: [:tags, :user], user: [], author: []})
+      .includes(user: [:hidings, :votes], notifiable: {story: [:tags, :user], user: [:comments], author: [], parent_comment: []})
 
     respond_to do |format|
       format.html { render :all }
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -13,7 +13,7 @@ class NotificationsController < ApplicationController
       .offset((@page - 1) * notifications_per_page)
       .limit(notifications_per_page)
       .order(created_at: :desc)
-      .includes(notifiable: [:user, :author])
+      .includes(notifiable: {story: [:tags, :user], user: [], author: []})
 
     @has_more = @user.notifications.count > (@page * notifications_per_page)
 
@@ -28,7 +28,7 @@ class NotificationsController < ApplicationController
       .notifications
       .where(read_at: nil)
       .order(created_at: :desc)
-      .includes(notifiable: [:user, :author])
+      .includes(notifiable: {story: [:tags, :user], user: [], author: []})
 
     respond_to do |format|
       format.html { render :all }
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,0 +1,53 @@
+# typed: false
+
+class NotificationsController < ApplicationController
+  before_action :require_logged_in_user
+  before_action :set_page, only: [:all]
+  after_action :update_read_at
+
+  def all
+    notifications_per_page = 25
+
+    @notifications = @user
+      .notifications
+      .offset((@page - 1) * notifications_per_page)
+      .limit(notifications_per_page)
+      .order(created_at: :desc)
+      .includes(notifiable: [:user, :author])
+
+    @has_more = @user.notifications.count > (@page * notifications_per_page)
+
+    respond_to do |format|
+      format.html
+      format.json { render json: @notifications }
+    end
+  end
+
+  def unread
+    @notifications = @user
+      .notifications
+      .where(read_at: nil)
+      .order(created_at: :desc)
+      .includes(notifiable: [:user, :author])
+
+    respond_to do |format|
+      format.html { render :all }
+      format.json { render json: @notifications }
+    end
+  end
+
+  private
+
+  def update_read_at
+    @notifications.update_all(read_at: Time.current)
+  end
+
+  def set_page
+    @page = params[:page].to_i
+    if @page == 0
+      @page = 1
+    elsif @page < 0 || @page > (2**32)
+      raise ActionController::RoutingError.new("page out of bounds")
+    end
+  end
+end
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -4,31 +4,37 @@ class Notification < ApplicationRecord
 
   include Token
 
-  def good?
+  GoodFaithResult = Struct.new(:good_faith?, :bad_properties)
+
+  def check_good_faith
     case notifiable
     when Message
-      good_message?
+      check_good_faith_message
     when Comment
-      good_comment?
+      check_good_faith_comment
     end
   end
 
-  def good_message?
-    true
+  def check_good_faith_message
+    GoodFaithResult.new(true, [])
   end
 
-  def good_comment?
+  def check_good_faith_comment
     comment = notifiable
     story = comment.story
     parent_comment = comment.parent_comment
 
-    good_story = story.score > story.flags
-    good_comment = comment.score > comment.flags && !comment.is_gone?
-    good_parent_comment = parent_comment.nil? ? true : parent_comment.score > parent_comment.flags && !parent_comment.is_gone?
-    user_has_flagged_replier = Vote.joins(:comment).where(user: user, story: story, vote: -1, comment: {user: comment.user}).any?
-    user_has_hidden_story = HiddenStory.where(user: user, story: story).any?
-    filtered_tags = story.tags & user.tag_filter_tags
+    bad_properties_to_check = []
 
-    good_story && good_comment && good_parent_comment && !user_has_flagged_replier && !user_has_hidden_story && filtered_tags.empty?
+    bad_properties_to_check << [:bad_story, story.score <= story.flags]
+    bad_properties_to_check << [:bad_comment, comment.score <= comment.flags || comment.is_gone?]
+    bad_properties_to_check << [:bad_parent_comment, parent_comment.nil? ? false : parent_comment.score <= parent_comment.flags || parent_comment.is_gone?]
+    bad_properties_to_check << [:user_has_flagged_replier, Vote.joins(:comment).where(user: user, story: story, vote: -1, comment: {user: comment.user}).any?]
+    bad_properties_to_check << [:user_has_hidden_story, HiddenStory.where(user: user, story: story).any?]
+    bad_properties_to_check << [:user_has_filtered_tags_on_story, !(story.tags & user.tag_filter_tags).empty?]
+
+    bad_properties = bad_properties_to_check.filter_map { |reason, flag| flag && reason }
+
+    GoodFaithResult.new(bad_properties.empty?, bad_properties)
   end
 end
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,0 +1,33 @@
+class Notification < ApplicationRecord
+  belongs_to :user
+  belongs_to :notifiable, polymorphic: true
+
+  include Token
+
+  def good?
+    case notifiable
+    when Message
+      good_message?
+    when Comment
+      good_comment?
+    end
+  end
+
+  def good_message?
+    true
+  end
+
+  def good_comment?
+    comment = notifiable
+    story = comment.story
+    parent_comment = comment.parent_comment
+
+    good_story = story.score > story.flags
+    good_comment = comment.score > comment.flags && !comment.is_gone?
+    good_parent_comment = parent_comment.nil? ? true : parent_comment.score > parent_comment.flags && !parent_comment.is_gone?
+    user_has_flagged_replier = Vote.joins(:comment).where(user: user, story: story, vote: -1, comment: {user: comment.user}).any?
+    user_has_hidden_story = HiddenStory.where(user: user, story: story).any?
+
+    good_story && good_comment && good_parent_comment && !user_has_flagged_replier && !user_has_hidden_story
+  end
+end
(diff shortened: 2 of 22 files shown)


app/views/comments/_comment.html.erb
```diff
@@ -28,7 +29,9 @@
   ">
 
   <div class="voters">
-    <label for="comment_folder_<%= comment.short_id %>" class="comment_folder"></label>
+    <% if !defined?(show_folder_control) || show_folder_control %>
```
```suggestion
    <% if show_folder_control %>
```

It shouldn't be possible for it to be undefined because of the default in the "signature".

app/views/comments/_threads.html.erb
```diff
@@ -72,7 +73,9 @@
   ">
 
   <div class="voters">
-    <label for="comment_folder_<%= comment.short_id %>" class="comment_folder"></label>
+    <% if !defined?(show_folder_control) || show_folder_control %>
```
pushcx:
```suggestion
    <% if show_folder_control %>
```

app/views/notifications/all.html.erb
```diff
@@ -0,0 +1,39 @@
...
+          <%= render "message", message: message, is_unread: notification.read_at.nil? %>
+        </li>
+      <% end %>
+    <% end %>
+  </ol>
+<% else %>
+  <p class="help">No notifications to show.</p>
+<% end %>
+
+<% if action_name == "all" && @has_more %>
```
pushcx:
```suggestion
<% if @has_more %>
```

I try not to have templates peeking back up the stack at what's rendering them, and in this case it can just decide based on `@has_more`.

config/routes.rb
```diff
@@ -139,6 +139,10 @@
     post "mod_note"
   end
 
+  get "/notifications" => "notifications#all"
+  get "/notifications/page/:page" => "notifications#all"
```
pushcx:
```suggestion
  get "/notifications(/page/:page)" => "notifications#all"
```

db/migrate/20250628172500_create_notifications.rb
```diff
@@ -0,0 +1,14 @@
+class CreateNotifications < ActiveRecord::Migration[8.0]
+  def change
+    create_table :notifications, id: {type: :bigint, unsigned: true} do |t|
+      # We shouldn't need an index on the user_id alone because the unique index below provides user_id in the leftmost column
+      t.references :user, null: false, foreign_key: true, type: :bigint, unsigned: true, index: false
+      t.references :notifiable, polymorphic: true, null: false, type: :bigint, unsigned: true
```
pushcx, line 6: It's future-proofing, but please add a `t.string :token, null: false` and in the model `include Token`.

app/views/comments/_comment.html.erb
thomasdziedzic: I needed to include a check for `!defined?(show_folder_control)` here because the heinous_inline_partial(comments/_comment.html.erb) in comments/_threads.html.erb already includes strict locals at the top of that view, and so when the heinouse_inline_partial inlines comments/_comment.html.erb along with it's locals, it gets ignored, causing an error to appear for that variable when visiting http://127.0.0.1:3000/threads

app/views/comments/_threads.html.erb
thomasdziedzic: Left a comment related to heinous_inline_partial above

app/views/notifications/all.html.erb
thomasdziedzic: Decided to test out the pagination a bit more, and fixed some additional bugs. removed the reference to action_name

config/routes.rb
thomasdziedzic: changed.

db/migrate/20250628172500_create_notifications.rb
thomasdziedzic: added

app/views/comments/_comment.html.erb
thomasdziedzic: On second thought, I added show_folder_control: true to the _threads partial view to fix that error and also to get rid of the !defined?, let me know if that works for you.

Ideally we could somehow merge the _comment partial view's strict locals into the parent (_threads) but that's a 1 to many relationship and I think it would get messy.

Not sure what the proper fix is for this in heinous views.

pushcx: Ah, I hadn't realized that possible context. Improving the consistency is a great approach.

I don't think there's any fix for the heinous views, they're living down to their name.

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,53 @@
+# typed: false
+
+class NotificationsController < ApplicationController
+  before_action :require_logged_in_user
+  before_action :set_page, only: [:all]
+  after_action :update_read_at, only: [:unread]
```
pushcx: Why only on `unread`? If it appears on `all`, the user can be assumed to have read it.

pushcx:
This looks great, and I love how clear and straightforward the code is.

Thinking through what's needed for this to take over:

 - [ ] remove the `ReplyingComments` model and spec
 - [ ] remove `User#unread_replies_count`, `clear_unread_replies!`, `inbox_count`, `unread_message_count`, and `update_unread_message_count`
 - [ ] a migration to remove the `replying_comments` view and all `keystores` entries matching `user:.*:unread_messages` or `user:.*:unread_replies`
 - [ ] rm `db/views`
 - [ ] remove the `scenic` gem and bundle (I'd rather the previous migration execute raw sql `drop view` than split this up into two PRs)
 - [ ] update the `inbox` link - actually, would this feature be nicer under `/inbox` than `/notifications`?
 - [x] `notifications/all.html.erb` needs a call to `possible_flag_warning` at the top
 - [x] add anti-flamewar predicates, see below
 - [ ] redirect routes under `/replies`
 - [x] add `ModNote` [button](https://github.com/lobsters/lobsters/blob/3aed0db34be05122d2de8245bd6bf1da5e1f827b/app/views/messages/show.html.erb#L29-L34) to messages so mods can record interactions

A lot of the [`replying_comments` view](https://github.com/lobsters/lobsters/blob/master/db/views/replying_comments_v10.sql) exists to avoid pouring gas on a flamewar. Some of that logic ports over really cleanly, like: don't create a notification if the story or parent comment have more flags than upvotes, or if the user has hidden the story. Some really can't effectively, like don't create a notification if the parent comment has been deleted or removed by a moderator, because the parent comment must exist at the time the reply is created, but I guess we can have comment deletion/moderation remove notifications. I'm very open to ideas for improving to these, they really do seem to have helped.

If you'd rather soft-ship this alongside replies without links into it and do all this replacement work in another PR, that's fine by me.

app/controllers/notifications_controller.rb
thomasdziedzic: I did it to match the [behavior in the replies_controller](https://github.com/lobsters/lobsters/blob/master/app/controllers/replies_controller.rb#L7). I changed it to include both all & unread actions

pushcx: I sure can't recall why I did that, sorry for the confusion.

thomasdziedzic:
> This looks great, and I love how clear and straightforward the code is.

Thanks!

> Thinking through what's needed for this to take over:
> 
> * [ ]  remove the `ReplyingComments` model and spec
> * [ ]  remove `User#unread_replies_count`, `clear_unread_replies!`, `inbox_count`, `unread_message_count`, and `update_unread_message_count`
> * [ ]  a migration to remove the `replying_comments` view and all `keystores` entries matching `user:.*:unread_messages` or `user:.*:unread_replies`
> * [ ]  rm `db/views`
> * [ ]  remove the `scenic` gem and bundle (I'd rather the previous migration execute raw sql `drop view` than split this up into two PRs)
> * [ ]  update the `inbox` link - actually, would this feature be nicer under `/inbox` than `/notifications`?

I'll throw the above in a follow up PR mainly cause I don't want to implement and replace replies all in one go.

> * [ ]  `notifications/all.html.erb` needs a call to `possible_flag_warning` at the top

Added this

> * [ ]  add anti-flamewar predicates, see below

Will add anti-flamewar logic into this PR.

> * [ ]  redirect routes under `/replies`

Were you thinking something like redirecting the old /replies route to /notifications?

> * [ ]  add `ModNote` [button](https://github.com/lobsters/lobsters/blob/3aed0db34be05122d2de8245bd6bf1da5e1f827b/app/views/messages/show.html.erb#L29-L34) to messages so mods can record interactions

This is already on the message which has a link on the "time_ago" for messages. If you still want this, what kind of workflow were you thinking should this support?

> A lot of the [`replying_comments` view](https://github.com/lobsters/lobsters/blob/master/db/views/replying_comments_v10.sql) exists to avoid pouring gas on a flamewar. Some of that logic ports over really cleanly, like: don't create a notification if the story or parent comment have more flags than upvotes, or if the user has hidden the story. Some really can't effectively, like don't create a notification if the parent comment has been deleted or removed by a moderator, because the parent comment must exist at the time the reply is created, but I guess we can have comment deletion/moderation remove notifications. I'm very open to ideas for improving to these, they really do seem to have helped.

I think the above sounds reasonable. I would like to centralize this logic in some sort of helper class so that we can make sure that the logic isn't spread out throughout the rails app.

> If you'd rather soft-ship this alongside replies without links into it and do all this replacement work in another PR, that's fine by me.

I would prefer soft-shipping alongside replies. I'll start working on adding anti flamewar logic to this PR. One thing I also thought about is what should we do about backfilling old notifications? Since we're planning on replacing replies, if we don't backfill, the user wont have access to their old replies.

thomasdziedzic:
Added an initial implementation of flamewar prevention by showing only good notifications as defined in the replying comment where clauses. It's not ideal because the property doesn't live in a database, but it's the simplest method I could come up with, everything else I thought of would have taken up a lot more time and would have been much more complicated. Let me know what you think of this approach, and if you like it, I'll write up some tests.

pushcx:
> make sure that the logic isn't spread out throughout the rails app

Flamewar prevention is maybe the most pervasive concept in the codebase, I think about it for almost every bit of copy, UI, and feature. :) It is the most common reason why bugs/features aren't tagged 'good first issue'. So I take your meaning here, and I like the way your implementation way of names predicates to minimize comments, but don't sweat it too hard.

If you don't want to remove /replies in the same PR, which is prudent, I think we don't need any redirects yet. We can soft launch this page, play with it a couple days, then redirect /replies to it.

And, yeah... I guess we want to write a backfill script? I say script instead of migration because we'll want to run it now to play with the feature in prod, then again when redirecting so that people have notifications with correct 'unread' marks. I thought a few minutes about whether we should delete them after a couple years, but the median number of notifications for comments have to be 1 (65% of comments are replies + n people mentioned, which is overwhelmingly 0). That's not a concerning growth rate and I can realistically see people wanting to use it to find years-old comments. You know, maybe the best way to handle this would be to make it a background job per comment (590k) and message (16k), so the migration just enqueues them all? It would be a great low-stakes stress test for our SolidQueue setup to shove 600k jobs in there and watch it drain.

pushcx:
OK, so: I think this PR is maybe ready to merge? We hit the must-haves, it's in a usable state so we can go click on it in prod and see how it works, and we can come back to write the backfill script when we migrate the rest.

---

We visited this on [today's office hours stream](https://push.cx/stream/2025-06-30) and @thomasdziedzic wants to add some tests and probably the backfill script before merging.

thomasdziedzic:
Added some tests for the flamewar prevention logic, and created a backfill script.

You can run the backfill script with `rails backfill_notifications` which takes around 18 minutes for me.

thomasdziedzic:
~Note that the backfill script doesn't set read_at which means all the backfilled notifications will show up as unread for the user. Let me know if you want me to set read_at in the backfill script.~

I set read_at on the script since I think that was the more sensible default.

lib/tasks/backfill_notifications.rake
```diff
@@ -0,0 +1,42 @@
...
+    end
+    finished_at = Time.current
+    puts "Created missing comment notifications, the time it took was #{finished_at - started_at} seconds"
+
+    started_at = Time.current
+    Message.includes(:recipient).find_each do |message|
+      Notification.create(
+        user: message.recipient,
+        notifiable: message,
+        read_at: read_at
```
pushcx:
```suggestion
        read_at: message.has_been_read ? read_at : nil
```

app/models/notification.rb
```diff
@@ -0,0 +1,33 @@
...
+    story = comment.story
+    parent_comment = comment.parent_comment
+
+    good_story = story.score > story.flags
+    good_comment = comment.score > comment.flags && !comment.is_gone?
+    good_parent_comment = parent_comment.nil? ? true : parent_comment.score > parent_comment.flags && !parent_comment.is_gone?
+    user_has_flagged_replier = Vote.joins(:comment).where(user: user, story: story, vote: -1, comment: {user: comment.user}).any?
+    user_has_hidden_story = HiddenStory.where(user: user, story: story).any?
+
+    good_story && good_comment && good_parent_comment && !user_has_flagged_replier && !user_has_hidden_story
```
pushcx: Let's support tag filters here too, please. Off the top of my head, I guess add a `Story#tag_filtered_by_user?` or something.

thomasdziedzic: added support

thomasdziedzic:
We discussed a couple more changes we would like on the stream today:

1. The filter on notifications in the view could potentially display less than 25 notifications, would placeholders for filtered out notifications work for you here @pushcx ?
2. Fix 2 * n+1 queries in the Notification#good? method, specifically for votes and hidden stories.

I think that was all.

thomasdziedzic:
> We discussed a couple more changes we would like on the stream today:
> 
> 1. The filter on notifications in the view could potentially display less than 25 notifications, would placeholders for filtered out notifications work for you here @pushcx ?
> 2. Fix 2 * n+1 queries in the Notification#good? method, specifically for votes and hidden stories.
> 
> I think that was all.

Displayed a placeholder for notifications that are filtered that looks like:

display hidden notifications

Also fixed the n+1 queries, though I'm really not happy with the tools rails gives me to fix them. I essentially have to query for more than I need if I want to do it through activerecord. Happy to learn about better techniques if they exist.

pushcx:
That's pretty much how you have to handle it, yep.

One of the benefits of sqlite is that in 1 + ns the n round trips don't have the per-query overhead that out-of-process, usually over the network, databases do. So not only do we not have to do complicated preloading, we'll avoid selecting more than we need, especially if we start using the [russian doll caching](https://guides.rubyonrails.org/caching_with_rails.html#russian-doll-caching).

I have some minor style tweaks that I'll make in the merge.

PR #1636, opened 2025-06-29
Stop using JavaScript for Markdown help

chamlis:
Switch to using a `<details>` element for this help, so it doesn't require JavaScript. I elected for this over the checkbox hack as it's more semantically appropriate.

The look of the dropdown has changed (though I matched the font size so it remains less distracting). Personally, I like the new look as it more clearly communicates what's going to happen, and leans on the native browser rendering, in line with the rest of the site. I'm not opposed to trying to more closely match the existing design, however.

image

image

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -534,10 +534,6 @@ onPageLoad(() => {
 
   // Global
 
-  on('click', '.markdown_help_label', (event) => {
-    qS(parentSelector(event.target, '.markdown_help_toggler'), '.markdown_help').classList.toggle('display-block');
-  });
-
   on('click', '#modal_backdrop', () => {
     Lobster.removeFlagModal()
   });
--- a/app/views/signup/invited.html.erb
+++ b/app/views/signup/invited.html.erb
@@ -47,13 +47,7 @@
     <div>
       <%= f.text_area :about%>
 
-      <div class="markdown_help_toggler">
-        <div class="markdown_help_label">
-          Markdown formatting available
-        </div>
-
-        <%= render :partial => "global/markdownhelp" %>
-      </div>
+      <%= render :partial => "global/markdownhelp" %>
     </div>
   </div>
 
--- a/app/views/stories/new.html.erb
+++ b/app/views/stories/new.html.erb
@@ -6,20 +6,14 @@
       <p></p>
 
       <div class="box">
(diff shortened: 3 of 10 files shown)


app/views/stories/new.html.erb
```diff
@@ -6,20 +6,14 @@
       <p></p>
 
       <div class="box">
-        <div class="boxline actions markdown_help_toggler">
-          <div class="markdown_help_label">
-            Markdown formatting available
-          </div>
+        <div class="boxline actions">
+          <%= render :partial => "global/markdownhelp",
+            :locals => { allow_images: @story.can_have_images? } %>
```
pushcx, line 11: Funny to see an improvement to the highest-touch UI in the site mention what's probably the least-used feature in the entire codebase.

app/views/signup/invited.html.erb
```diff
@@ -9,12 +9,6 @@
   <%= errors_for(@new_user) %>
 
   <div class="labelled_grid">
-    <% if not Rails.application.open_signups? %>
```
pushcx: Please don't delete this. We haven't had it enabled for ages, but new sister sites always turn it on at first.

app/javascript/application.js
```diff
@@ -534,10 +534,6 @@ onPageLoad(() => {
 
   // Global
 
-  on('click', '.markdown_help_label', (event) => {
```
pushcx, line 537: Yay!

pushcx (changes requested):
Having the triangle is a very nice UI hint, good call. Thanks for picking this up.

One small bit of code to restore, but this is ready to go.

app/views/signup/invited.html.erb
chamlis: Sorry about that, this was one of my local hacks to let me load the invited page without a valid invite code and shouldn't have made its way into the branch. I've force-pushed to restore it.

pushcx: Ah, thanks for explaining, I was puzzled and guessed you were tidying apparently-dead code.

There's probably a lot of these sorts of minor hassles that I'm oblivious to because I almost always develop against a copy of the prod db and it's very easy for me to grab `Invitation.where(used_at: nil).last.token` or whatever in the rails console. If you see similar issues, please do post feature requests for things that would improve your developer experience.

PR #1637, opened 2025-06-30
Mockturtle irc bot port to ruby

Church-:
Basic port of the current mockturtle nodeJS codebase to a ruby port under the main lobsters repo.

Implements the same dot commands mockturtle currently does as well as the rss watching notifications we already have for new site stories and repo commits.

Still needs:
* DB persistence
* Telebugs
* Config/arg parsing


mockturtle/app/scheduled.rb
```diff
@@ -0,0 +1,76 @@
...
+require "sqlite3"
+
+class RSSWatcher
+  def initialize(bot, db)
+    sleep 5
+    @bot = bot
+    @db  = db
+  end
+
+  def watch_github
```
Church-, line 15: Need to fix persistence so we can actually start scraping stories.

Should probably fail open and set lastSeen as a class variable. So even if we don't have working persistence we can still scrape rss.

mockturtle/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Running
+
+```sh
+$ ruby app/mockturtle.rb 'irc://username:password@example.com:6697/#channel1,##channel2`
+```
+
+## License
+
+```
+Copyright 2025 Noah Hendrickson
```
pushcx: I'd prefer to use the license that's already used for the site (3 clause BSD, if my quick glance is correct) because code will be shared between the bot and site. Do you have a strong preference that we do it differently?

mockturtle/app/scheduled.rb
pushcx: Yeah. We should use ActiveRecord here. For all I grumble about it, the consistency of matching the site database access would be worth it.

mockturtle/app/commands.rb
```diff
@@ -0,0 +1,122 @@
+require "ircinch"
+require "json"
+require "nokogiri"
+require "open-uri"
+require "time"
+
+class Seen
+  class SeenStruct < Struct.new(:who, :where, :time)
```
pushcx, line 9: Since these aren't modified after creation, recent versions of Ruby (which we're running) have a [`Data`](https://docs.ruby-lang.org/en/3.2/Data.html) class that's immutable.

I'd also prefer to just be `Seen` without the hungarian notation. Ruby is very timey-wimey about types.

Gemfile
```diff
@@ -58,6 +58,15 @@ gem "rack-attack" # rate-limiting
 gem "lograge" # for JSON logging
 gem "silencer" # to disable default logging in prod
 
+# mockturtle
```
pushcx: Because this is going to run in a separate process, let's make a `mockturtle/Gemfile` for it. I don't want the Rails processes to take up any more RAM than they already do.

Gemfile
```diff
@@ -58,6 +58,15 @@ gem "rack-attack" # rate-limiting
...
+gem "ircinch"
+gem "json"
+gem "nokogiri"
+gem "open-uri"
```
pushcx: Please remove this and use `extras/sponge.rb` to fetch anything. It prevents people from saying in chat "https://127.0.0.1:22" etc and using the bot to portscan/exploit us or others.

Gemfile
```diff
@@ -58,6 +58,15 @@ gem "rack-attack" # rate-limiting
...
+gem "rss"
+gem "sqlite3"
+gem "time"
```
pushcx: Not needed, it's in the standard library and loaded automatically.

pushcx (commented):
Off to a good start here, and I've seen the activity in the test channel. I've added some Ruby pointers and I look forward to seeing this in prod. I'll talk to you offline about that idea of setting up some pairing time.

pushcx:
Open integration questions for myself:

 * Should the bot require rails so it has access to that entire env, or is it enough to integrate by reading the db? I think I'd prefer to have mockturtle treat the prod db as read-only.
 * How should a background job on the rails side command the bot? I'd like to have a rails bg job fire the bot posting, but I don't see how to send data over.

mockturtle/README.md
Church-: No strong preference, will copy.

caius: If we're using the same license, you can probably just delete the license from this README and the whole repo is covered by the top level license.

Gemfile
caius: We don't need it in the Gemfile, we do need to require it (which we might be getting for free with the Gemfile entry?)

Methods like `Time.parse()` aren't loaded unless you explicitly require the time stdlib.

caius:
> * Should the bot require rails so it has access to that entire env, or is it enough to integrate by reading the db? I think I'd prefer to have mockturtle treat the prod db as read-only.

We should be able to configure ActiveRecord readonly connections, the safest way would be to have a readonly Postgres user the bot connects as.

> * How should a background job on the rails side command the bot? I'd like to have a rails bg job fire the bot posting, but I don't see how to send data over.

The bot could listen on a local port or socket, and have the app make a request in the background to the bot over localhost to trigger it?

mockturtle/Gemfile
```diff
@@ -0,0 +1,17 @@
...
+gem "json"
+
+# parsing
+gem "nokogiri"
+
+# mockturtle
+gem "ircinch"
+gem "rss"
+
+# performance monitoring
```
pushcx, line 16:
```suggestion
# bug reporting
```

pushcx:
Building on @caius's comment, I think we could cheaply get a robust way for the rails app to boss the irc bot around by using Solid Queue. The rails app could write jobs to the queue for a worker process in the mockturtle process to run. This saves us writing our own code for things like serialization or supervising the worker. Solid Queue includes a [puma plugin](https://github.com/rails/solid_queue/blob/main/lib/puma/plugin/solid_queue.rb) for that process to supervise a solid queue worker, and I think this is very close to what we'd want. So maybe we get something very mature very cheaply!

Issue #1494, opened 2025-03-18
Show feedback after the server processes a request to upvote a story or comment


roryokane:
# Current behavior

When logged in, when I click the arrow next to a story or comment to upvote, the arrow changes color right away. This is an example of optimistic UI.

| State | Light theme | Dark theme |
|--------|--------|--------|
| Before upvoting | <img width="135" alt="comment with gray upvote arrow in light theme" src="https://github.com/user-attachments/assets/b8092840-b1c9-4aad-81dc-00297dd84f2b" /> | <img width="135" alt="comment with gray upvote arrow in dark theme" src="https://github.com/user-attachments/assets/5b66a9c5-2628-45ed-9c46-09ce02f677d4" /> |
| After clicking the upvote arrow | <img width="135" alt="comment with solid red upvote arrow in light theme" src="https://github.com/user-attachments/assets/71e1b836-c6ab-4409-8d3a-285da1a1fee2" /> | <img width="135" alt="comment with solid red upvote arrow in dark theme" src="https://github.com/user-attachments/assets/9a6f120d-080e-4c21-9237-f2444c6bb251" /> |

# Problem

There have been times when I clicked this arrow, then quickly navigated to another page, and I later discovered that my upvote had not been registered. I needed to click upvote again, then wait on the page long enough to be confident that my upvote had been registered.

Due to this issue, I have developed a habit. After I upvote something, I often wait a few seconds, then refresh the page to check whether my upvote was received. This takes about 4 seconds. This also has the side effect of marking any new comments on a story as read, which I might not want, because I might have wanted to stop reading the comment thread for now.

It would be more convenient if the site gave feedback when my upvote had been received and saved without my having to refresh the page.

# Idea for solution: three colors for the upvote button

| State | Light theme | Dark theme |
|--------|--------|--------|
| Before upvoting | <img width="135" alt="comment with gray upvote arrow in light theme" src="https://github.com/user-attachments/assets/b8092840-b1c9-4aad-81dc-00297dd84f2b" /> | <img width="135" alt="comment with gray upvote arrow in dark theme" src="https://github.com/user-attachments/assets/5b66a9c5-2628-45ed-9c46-09ce02f677d4" /> |
[...]

PR #1643, opened 2025-07-03
Improve accessibility for upvotes, fix #1641

### LINKED ISSUE #1641: Upvote button color does not change enough when clicked
Here is how upvote buttons look like:

Image

Perhaps you might think it is easy for me to understand which posts I upvoted. But it looks something like that to me (and some other color-blind people):

Image

I am struggling to tell whether I clicked the upvote or not. It is especially bad when I have the comments page open; there are no other "upvote" buttons to compare. Did I upvote this submission? Did I not? I have no idea, just by looking at this:

Image

I have to click the upvote button a few times to ensure that my click brought the number up. Perhaps, it is possible to have an easier-to-understand UI?

veqqq:
Now, △ becomes ▲ when upvoting. Previously,▲was always used.

N.b. for #1494 ◭ `\25ed` could be used, until the server confirms the upvote (pushcx recommended this on twitch.)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -455,16 +455,13 @@ div.voters {
 }
 
 .upvoter:before {
-	content: "\25b2";
+	content: "\25b3";
 	color: var(--color-fg-shape);
 	font-size: 12pt;
 }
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for taking this on. Seems like a clear and reasonable fix, expressive with or without color.

PR #1644, opened 2025-07-03
Unify upvote arrow sizes

veqqq:
`\25b2` and `\25b3` were not in fact the same size, as 3 had a border. Now we use the same character and outline both! This improves on 352dbb0 .


pushcx:
Mentioning #1641 

We worked through this on [today's office hours stream](https://push.cx/stream/2025-07-03). Thanks for working through this to make the upvote arrow sizes consistent, this looks very nice now.

PR #1645, opened 2025-07-03
Fix upvote hover state on iOS

marcoroth:
I usually don't like fixing CSS bugs with `!important`, but it seems to do the trick here:

| Before | After |
| --- | --- |
| IMG_7648 | IMG_7649 |

Feel free to close this if you can think of another way of fixing this!

/cc @veqqq


pushcx:
I can't see why it's needed, there's not a lot of classes here, but trying to debug iOS Safari takes me back to pre-Firebug days, and not in a good way. Thanks for finding a working fix.

PR #1646, opened 2025-07-03
Improve Upvote Arrow Visibility

veqqq:
To continue our bikeshedding, I thickened and lightened the border (color now shared with the reply/edit line) embiggened the arrows(' font).

I tested it on Chrome's blindness tools from a few distances. I tested many arrows and mathematical symbols, but the original is best. N.b. 2px was too bold and attention-grabbing (especially on the front page. This color change greatly increases distinctiveness compared to the background and the smaller border requirement makes the filled out arrow more distinct from the empty one.)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -455,14 +455,14 @@ div.voters {
 }
 
 .upvoter:before {
-	content: "\25b2";
+  content: "\25b2";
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks. I had to deal with a merge conflict from #1641. While I was there I set it to `1px`, I really don't want it to be too dark.

PR #1647, opened 2025-07-04
Fix preview comment indent

### LINKED ISSUE #1639: Top-level reply preview is indented
This has probably been the case since the site started and I never noticed, but while writing a [top-level reply]() to a story I noticed the preview has an odd indentation to it. I think it would look a lot nicer if the left edge of the preview comment's text block was lined up with the left edge of the textarea.

Image

So probably there is also a similar improvement available for replies, but I'm procrastinating enough as it is.

supermarin:
Fixes #1639

This PR does 2 things and I'm happy to split it into 2 PRs if you decide to merge this.
  1. When moving the preview comment back left, the preview line was still displayed and completely off. You probably don't want it there in the first place.
  2. Removes the dotted line for comments that aren't children.

I checked [wayback machine](https://web.archive.org/web/20250105171457/https://lobste.rs/s/9anym5/what_was_best_research_paper_you_read_2024) and based on the codebase, it seems like the intention was to draw the dotted line only on comments that have replies.

The current implementation on the website seems a bit off (there's dotted lines everywhere, even over upvote arrows). Side note - think the comment form should be moved to the left as well.

After:

Nested replies

Before:

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -908,10 +908,6 @@ div.comment_text code {
 	font-size: 9pt;
 }
 
-.comment .preview {
-	padding-left: 17px;
-}
-
 div#story_preview {
 	margin-top: 2em;
 	margin-left: 3.5em;
@@ -952,9 +948,6 @@ div#story_box textarea {
 	width: 600px;
 }
 
-div.comment_form_container {
-	margin-left: -11px;
-}
 
 div.comment_form_container form {
 	margin-left: 40px;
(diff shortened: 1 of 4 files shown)


pushcx:
Thanks, this fixes the bug in the top-level and inline comment previews.

I'll have to write up the tree lines as a separate issue. It was a regression we accepted to fix some other layout issues, but we haven't gotten back to it to fix it. Sorry that you had to spend some time puzzling it out! Your change here is an improvement.

PR #1648, opened 2025-07-04
Fix top stories page linking to RSS

Tyrubias:
The change in #1629 caused the previous and next page links on the Top Stories pages to link to the RSS pages instead.

I think it's because the new top RSS route matches before the regular top HTML route matches. I fixed this by swapping the order of the top page routes and adding a more specific `/top/rss` route for the case where that's requested. I also added a test to make sure the URL generated for the previous and next page links is correct.

I rarely write Ruby and have never contributed to a Rails codebase so please let me know if my approach is idiomatic.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,9 +30,8 @@ Rails.application.routes.draw do
   get "/upvoted", to: redirect("/upvoted/stories")
   get "/upvoted/page/:page", to: redirect("/upvoted/stories/page/%{page}")
 
-  get "/top/rss" => "home#top", :format => "rss"
-  get "/top(/:length(/page/:page))" => "home#top", :as => "top"
   get "/top(/:length)/rss" => "home#top", :format => "rss"
+  get "/top(/:length(/page/:page))" => "home#top", :as => "top"
 
   get "/threads" => "comments#user_threads"
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,8 +30,9 @@ Rails.application.routes.draw do
   get "/upvoted", to: redirect("/upvoted/stories")
   get "/upvoted/page/:page", to: redirect("/upvoted/stories/page/%{page}")
 
-  get "/top(/:length(/page/:page))/rss" => "home#top", :format => "rss"
+  get "/top/rss" => "home#top", :format => "rss"
   get "/top(/:length(/page/:page))" => "home#top", :as => "top"
+  get "/top(/:length(/page/:page))/rss" => "home#top", :format => "rss"
 
   get "/threads" => "comments#user_threads"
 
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -30,8 +30,9 @@
   get "/upvoted", to: redirect("/upvoted/stories")
   get "/upvoted/page/:page", to: redirect("/upvoted/stories/page/%{page}")
 
-  get "/top(/:length(/page/:page))/rss" => "home#top", :format => "rss"
+  get "/top/rss" => "home#top", :format => "rss"
   get "/top(/:length(/page/:page))" => "home#top", :as => "top"
+  get "/top(/:length(/page/:page))/rss" => "home#top", :format => "rss"
```
pushcx:
```suggestion
  get "/top(/:length)/rss" => "home#top", :format => "rss"
```

I don't want to add an rss url for each page.

pushcx: Hang on, I'm confused: https://github.com/lobsters/lobsters/pull/1629/files#r2185462790

I'm going to hang onto this a sec to hear back from @caius.

pushcx:
Thanks for puzzling this out.

My edit broke the build; I didn't notice caius added URLs in 1629 and I'd like to understand it.

The surprise that broke this pagination is that the bottom of `app/views/home/index.html.erb` is shared across the several views, so it generates its route with the controller and action. Rails was taking the first route in the file that matched those constraints, which happened to be the rss route. An alternate fix would be to name the format so the RSS route wouldn't match.

caius:
We could also just move the RSS route second, I'm surprised Rails router picks the one with specific format on it, having the RSS one second would stop it being picked.

Tyrubias:
@caius I tried moving the RSS route second and it broke the tests that #1629 added for RSS routes.

supermarin:
I believe 5d4cf137f08ffafeeac942ba86deb16b0251892c breaks the intended functionality of ddfc7872, which is adding /rss on paginated results like /top/1w/page/2/rss (although I have no idea why anyone would want that).
Think all you need is reverse the two routes and add a /top/rss one on top: #1657 

EDIT: sorry, I didn't read all the comments and commits here, feel free to close my PR. It does solve for all the cases listed above though.
/top
/top/rss
/top/1w
/top/1w/rss
/top/1w/page/2
/top/1w/page/2/rss

Tyrubias:
@supermarin I think @pushcx wanted to get rid of the paginated RSS links. I think this PR both does this and fixes #1650

pushcx:
Thanks for working through all the behavior here. In addition to the direct fixes from @Tyrubias, I appreciate the collaboration from @caius and @supermarin. You made sense out of a puzzling combination of bugs that made this very easy to merge!

PR #1655, opened 2025-07-09
disable transactions in fake_data task

pcai:
As discussed, we can turn off AR transactions which will speed up the insertion for fake_data task. Tailing the development logs with this change will confirm explicit transactions are no longer used:
image

BEFORE (on my machine):

real    3m45.308s
user    2m0.959s
sys     0m26.652s

AFTER:

real    2m32.567s
user    1m37.593s
sys     0m21.596s

A modest speedup but probably still worth it?


pushcx (approved):
Thanks for this improvement. It's certainly safe enough for setting up a dev env, and I'm very happy to save time for future contributors.

PR #1658, opened 2025-07-13
updates docker-compose yaml and docs to allow debugging

jacklynhma:
Goal: Allow debugging while you are running lobster in docker and update the docs 
Issue: `docker compose up` runs both servers at the same time, but we only want to interact with the app service.


docker-compose.yaml
```diff
@@ -1,6 +1,8 @@
 
 services:
   app:
+    stdin_open: true
```
jacklynhma, line 4: `stdin_open` allows input 
`tty` allows interactive shell access

docs/setup_with_docker.md
```diff
@@ -55,3 +55,16 @@ Solution:
 
 * Run `docker compose run app db:drop`
 * Redo steps starting at "Switch back to the tab running the mariadb image and restart the server by:"
+
+# Debugging in Docker
+
+* Open a new terminal
+* Run your app server:
+  * IF your app server is running, Run `docker attach <CONTAINER_NAME>`
```
jacklynhma, line 63: docker compose up runs both servers at the same time, but we only want to interact with the app service. 

`docker attach lobsters-app` allows us to do this by specifying which service we want to interact with

pushcx:
Great, thank you for explaining this, both in the docs and in these comments. I'm really glad that it'll be easier for new contributors and sister sites.

PR #1659, opened 2025-07-14
draft: guideline about generative ai output

pushcx:
This is a rough draft for community decisionmaking and editing.

It needs a specific title, right now it has a couple ideas.

I try to phrase the guidelines 'positively', in terms of things we'd like to see. This one sort of necessarily has to be a negative.

I wrote [some policy thoughts](https://github.com/lobsters/lobsters/pull/1570#issuecomment-2884067535) over on another draft PR.

### DIFF
--- a/app/views/about/about.html.erb
+++ b/app/views/about/about.html.erb
@@ -67,6 +67,19 @@
     As a rule of thumb, self-promo should be less than a quarter of one's stories and comments.
   </p>
 
+  <p>
+
+  <strong>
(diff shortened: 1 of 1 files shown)


pushcx:
This went stale because I was traveling. There's been a [related meta thread](https://lobste.rs/s/cpadta/proposal_mark_punish_ai_imagery) and I'd like to use it to inform this wording, maybe following with another meta thread before merging the guideline.

PR #1664, opened 2025-07-15
Improve format rendering and vote logic

### LINKED ISSUE #1662: Upvote/unvote always returns "ok" when logged in
It appears when calling upvote/unvote that it always returns "ok" even if the user is already upvoted or unvoted a story/comment. 

I tested this by doing a POST request to `http://localhost:3000/stories/:story_id/upvote.json` with the cookie set to the `lobster_trap` cookie manually taken from the website. 

When not logged in "not logged in" is returned.

I don't know if this warrants another issue but when using the json format the format is not in json as `plain:` is used for the output.

twodayslate:
Closes #1662

While testing/coding I noticed that a lot of the HTTP status codes were 400 and I thought there were more appropriate errors codes for some things. 

I also noticed that a lot of endpoints always `return render plain:` instead of JSON where appropriate. 

In order to know if a vote already happened I had to add a return Boolean value to `vote_thusly_on_story_or_comment_for_user_because`.

I updated the test cases where appropriate and added some new ones

Let me know what you think!

### DIFF
--- a/app/controllers/concerns/authenticatable.rb
+++ b/app/controllers/concerns/authenticatable.rb
@@ -61,11 +61,14 @@ module Authenticatable
     end
   end
 
-  def require_logged_in_user_or_400
+  def require_logged_in_user_or_401
     if @user
       true
     else
-      render plain: "not logged in", status: 400
+      respond_to do |format|
+        format.html { render plain: "not logged in", status: 401 }
+        format.json { render json: {error: "not logged in"}, status: 401 }
+      end
       false
     end
   end
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -5,7 +5,7 @@ class CommentsController < ApplicationController
 
   caches_page :index, :threads, if: CACHE_PAGE
 
-  before_action :require_logged_in_user_or_400,
+  before_action :require_logged_in_user_or_401,
     only: [:create, :reply, :upvote, :flag, :unvote, :update]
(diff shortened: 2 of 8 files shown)


app/controllers/concerns/authenticatable.rb
```diff
@@ -61,11 +61,14 @@ def require_logged_in_admin
     end
   end
 
-  def require_logged_in_user_or_400
+  def require_logged_in_user_or_401
```
pushcx, line 64: While the status's name "Unauthorized" is more-specific, the RFC defines it _even more_ specifically and [requires that it set another header](https://www.rfc-editor.org/rfc/rfc7235#section-3.1) to lead into the HTTP basic auth or digest auth flow, neither of which Lobsters supports.

So 401 is not appropriate here. I could [see a case for 403](https://www.rfc-editor.org/rfc/rfc7231#section-6.5.3), I don't think there's anything wrong there. Our login page is long enough that browsers shouldn't preempt it with their builtin pages.

pushcx (changes requested):
I left one small note before I saw what all these changes are, so I left longer feedback over in the linked issue #1662. I appreciate that you want to contribute, but I don't understand how this would improve the site and there's maintenance costs here.

app/controllers/concerns/authenticatable.rb
twodayslate: Ah, thanks. I didn't know about that header field. 

If you would like me to resubmit this PR with just the 403 change, please let me know and I will do that.

PR #1666, opened 2025-07-15
Upgrade ffi to 1.17.2

twodayslate:
I was getting the following error when running `bundle install`:
```
Error: invalid CFI advance_loc expression .cfi_def_cfa x1, 40;
```

I believe https://github.com/libffi/libffi/issues/852 is the root cause. I am on a M1 Macbook Air.


pushcx:
Thanks for diagnosing this, I'm happy to bump dependencies for this kind of issue.

PR #1667, opened 2025-07-15
fix(story): check for ascii only titles

yuanji-dev:
When submitting a story with Japanese title in local dev env, met the validation error.

The background is I'm planning to run a non-English speaking site using Lobsters code base, when exploring the features the code base provides, I was stopped at submitting stories.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -244,7 +244,7 @@ class Story < ApplicationRecord
     if title.match(GRAPHICS_RE)
       errors.add(:title, " may not contain emoji, dingbats, or other graphics")
     end
-    if !title.empty? && title == title.upcase
+    if !title.empty? && title.ascii_only? && title == title.upcase
       errors.add(:title, " doesn't need to scream, ASCII has supported lowercase since June 17, 1963.")
     end
 


pushcx:
Thanks for this fix. I'm sure there are other English-based assumptions in the codebase, and I'm happy to make the code easier for sister sites in other languages if it doesn't mean a significant maintenance burden.

yuanji-dev:
Thanks for the quick review and merge! @pushcx 

I've successfully made the codebase run in a docker container(using the [Dockerfile](https://github.com/rails/rails/blob/v8.0.2/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt) generated by Rails) with few modifications. Most of them are related to some hardcoded paths like `/home/deploy/lobsters/shared/log/action.log`.

Besides those path issues, I could send another PR to fix some common issues which will be helpful other ones to deploy their own sites.

pushcx:
I hardcoded many of our production paths to reduce work when migrating our deployment to Hatchbox. It would be an improvement to use `Rails.root` instead. Now that our recurring scripts have been folded into jobs (#1548) I think every reference outside of the `hatchbox/` directory could use `Rails.root` instead, but I did not check closely.

[Hatchbox](https://hatchbox.io) clones the git repo to a unique directory, then symlinks that to `/home/deploy/lobsters/current`. In that directory, it also symlinks `log`, `storage`, and `tmp` to `/home/deploy/lobsters/shared/...` so they are preserved across deploys.

yuanji-dev:
Thanks for the info about hatchbox, will take a look into it and maybe send another PR to fix the hardcoded path issue.

PR #1669, opened 2025-07-16
fix: make it eaiser to deploy in container

yuanji-dev:
Some minor but common stuffs may help people deploy the codebase in production.

background: https://github.com/lobsters/lobsters/pull/1667#issuecomment-3077897938

### DIFF
--- a/config/initializers/00_zeitwerk.rb
+++ b/config/initializers/00_zeitwerk.rb
@@ -3,7 +3,7 @@
 # This file is a relic of how Lobsters used to manage API keys and some other config variables.
 # It only exists to give developers and sister sites an error message as they setup.
 
-if Rails.application.credentials.secret_key_base.blank?
+if ENV["SECRET_KEY_BASE_DUMMY"].blank? && Rails.application.credentials.secret_key_base.blank?
   config = <<~CONFIG
     ** SETUP REQUIRED
 
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -59,3 +59,6 @@ end
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
+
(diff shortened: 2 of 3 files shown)


config/puma.rb
```diff
@@ -59,3 +59,6 @@
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
+
+# Run the Solid Queue supervisor inside of Puma for single-server deployments
+plugin :solid_queue if ENV["SOLID_QUEUE_IN_PUMA"]
```
yuanji-dev, line 64: `SOLID_QUEUE_IN_PUMA` is used in solid_queue's docs. https://github.com/rails/solid_queue?tab=readme-ov-file#puma-plugin

config/initializers/00_zeitwerk.rb
```diff
@@ -3,7 +3,7 @@
 # This file is a relic of how Lobsters used to manage API keys and some other config variables.
 # It only exists to give developers and sister sites an error message as they setup.
 
-if Rails.application.credentials.secret_key_base.blank?
+if ENV["SECRET_KEY_BASE_DUMMY"].blank? && Rails.application.credentials.secret_key_base.blank?
```
yuanji-dev, line 6: `SECRET_KEY_BASE_DUMMY` is used in Rails's official Dockerfile template.

https://github.com/rails/rails/blob/97dac99b5e80cc0aab3764e86fbcd12f4c316112/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L97

config/puma.rb
pushcx: Thanks for finding a way to do this that will not break production.

pushcx (approved):
Thanks, this is clear and useful. I appreciate that you linked documentation and examples, that made it very easy for me to review.

PR #1672, opened 2025-07-18
Remove unknown cookies and end empty session 1594

jacklynhma:
**Issue**: https://github.com/lobsters/lobsters/issues/1594 

**To do:** 

- [x] Add an ApplicationController before_action to delete any cookies that aren't the two we use. 
- [x] Reference the names using TAG_FILTER_COOKIE and Rails.application.config.session_options[:key].
- [x] If the session is empty because the user logged out or never logged in, delete it. ApplicationController has clear_lobster_trap but it's not working. 
- [x] Rename to clear_session_cookie for clarity and generality.
- [x] Write tests
- [x] Verify all tests pass locally
- [x] Update docker notes with command on how to run tests locally via docker.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,10 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :prepare_exception_notifier
   before_action :mini_profiler
   before_action :set_traffic_style
-
   before_action :remove_non_lobster_cookies
-  # 2023-10-07 one user in one of their browser envs is getting a CSRF failure, I'm reverting
-  # because I'll be AFK a while.
   after_action :clear_session_cookie
 
   # match this nginx config for bypassing the file cache
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,9 +10,10 @@ class ApplicationController < ActionController::Base
   before_action :mini_profiler
   before_action :set_traffic_style
 
+  before_action :remove_non_lobster_cookies
   # 2023-10-07 one user in one of their browser envs is getting a CSRF failure, I'm reverting
   # because I'll be AFK a while.
-  # after_action :clear_lobster_trap
+  after_action :clear_session_cookie
 
   # match this nginx config for bypassing the file cache
   TAG_FILTER_COOKIE = :tag_filters
@@ -63,9 +64,18 @@ class ApplicationController < ActionController::Base
     true
   end
 
+  def remove_non_lobster_cookies
+    # clear all cookies except the tag filters cookie and the session cookie
+    cookies.each do |key, _value|
+      next if key == TAG_FILTER_COOKIE # don't clear tag filters cookie
+      next if key == Rails.application.config.session_options[:key] # don't clear session cookie
+      cookies.delete(key)
+    end 
+  end 
+
   # clear Rails session cookie if not logged in so nginx uses the page cache
   # https://ryanfb.xyz/etc/2021/08/29/going_cookie-free_with_rails.html
-  def clear_lobster_trap
+  def clear_session_cookie
     key = Rails.application.config.session_options[:key] # "lobster_trap"
     cookies.delete(key) if @user.blank?
     # this probably should test session.empty? && controller...
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :prepare_exception_notifier
   before_action :mini_profiler
   before_action :set_traffic_style
-  before_action :remove_non_lobster_cookies
+  before_action :remove_unknown_cookies
   after_action :clear_session_cookie
 
   # match this nginx config for bypassing the file cache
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -64,9 +65,17 @@ def check_for_read_only_mode
     true
   end
 
+  def remove_non_lobster_cookies
```
jacklynhma: 🤔 Not sure about this name. Open to suggestions.

jacklynhma:
~Taking a look at build failures~ All fixed! 🥳  PR is ready for review

app/controllers/application_controller.rb
pushcx:
```suggestion
  def remove_unknown_cookies
```

How's this?

spec/features/homepage_spec.rb
```diff
@@ -19,6 +22,11 @@
 
     scenario "homepage" do
       visit "/"
+
+      expect(page.driver.request.cookies.length).to be(2)
+      expect(page.driver.request.cookies["lobster_trap"]).to_not be_nil
+      expect(page.driver.request.cookies["__profilin"]).to_not be_nil
```
pushcx: I don't understand this test, shouldn't this second cookie be removed? Or should it be testing a `response` object to see the instruction to delete cookies?

app/controllers/application_controller.rb
jacklynhma: PERFECT! Thank you :)

spec/features/homepage_spec.rb
jacklynhma: @pushcx You make a very good point.  I looked more into this and discovered the `__profilin` cookie is set by `Rack::MiniProfiler`  here https://github.com/lobsters/lobsters/blob/master/app/controllers/application_controller.rb#L86 

I looked into the Rack::miniprofiler docs and saw that you can't technically delete the `__profilin` cookie. If you try to delete it then it just resets. The only way to remove it is to disable it, but I don't know if that is what we want to do here, since it is lobster that is setting this cookie. Please correct me if I am wrong. 
https://github.com/MiniProfiler/rack-mini-profiler?tab=readme-ov-file#patching-nethttp

pushcx: Huh, I never noticed it. One of the useful features of this functionality is that it turns cookies from a sort of super-global variable that could be used by almost any part of the front or backend into a tracked feature. Your code just forced an unknown cookie to reveal itself.

The MiniProfiler is visible to devs in local dev and site mods. So please update `remove_unknown_cookies` to delete it `unless Rails.env.development? || @user&.is_moderator?`.

pushcx:
As long as I'm writing: you had asked some good questions about getting a dev env setup because the README wasn't clear. On the [office hours stream today](https://push.cx/stream/2025-07-21) I finished writing up production deployment with Hatchbox and used it as an opportunity to rearrange a few things, so the [readme now](https://github.com/lobsters/lobsters/blob/master/README.md) is organized better to explain why things are the way they are. Thanks again for telling me about working through setup!

spec/features/homepage_spec.rb
jacklynhma: Will do! 
Would you prefer I write a test for this condition? 
My gut tells me I should write a test for this, but while I was writing one, I came across this in the readme https://github.com/lobsters/lobsters/blob/master/README.md?plain=1#L22, and I just want to sure I am following the guidelines before proceeding

jacklynhma:
> As long as I'm writing: you had asked some good questions about getting a dev env setup because the README wasn't clear. On the [office hours stream today](https://push.cx/stream/2025-07-21) I finished writing up production deployment with Hatchbox and used it as an opportunity to rearrange a few things, so the [readme now](https://github.com/lobsters/lobsters/blob/master/README.md) is organized better to explain why things are the way they are. Thanks again for telling me about working through setup!

The README looks BEAUTIFUL!!! 🥳

spec/features/homepage_spec.rb
pushcx: That guideline is kinda my excuse not writing tests for a bunch of functionality. I can claim it's a principled engineering tradeoff but it is also just being real limited in dev time and hoping I'm making the most of it.

In PR review I'll be explicit when I need a test before I'll merge. I think it would be nice here but doesn't require it.

spec/features/homepage_spec.rb
```diff
@@ -19,9 +22,20 @@
...
       expect(page).to have_content(story.title)
     end
 
+    scenario "homepage with unknown cookies" do
+      page.driver.request.cookies[:unknown_cookie] = "value"
+      visit "/" # should clear unknown cookies
+
+      expect(page.driver.request.cookies.length).to be(2)
+      expect(page.driver.request.cookies["lobster_trap"]).to_not be_nil
+      expect(page.driver.request.cookies["__profilin"]).to_not be_nil # Cookie set by Rack::MiniProfiler. This can not be cleared.
```
pushcx, line 35: Whew, this is one of those comments that catch my eye because I can tell it's like 3 hours of debugging condensed down to 10 words. This must have been frustrating to run down, but your explanation made it immediately apparent why this is the case, thanks.

pushcx (approved):
This looks great! Thanks for improving how we handle cookies, we'll hopefully get to serve a lot more pages out of the cache and scale better.

PR #1673, opened 2025-07-19
fix: replace hardcoded paths outside hatchbox dir

yuanji-dev:
This PR is to replace all hardcoded paths outside of hatchbox & script dirs.

Although there is one more case in `config/puma.rb`,  it can be bypassed by using `PIDFILE` env variable.(or maybe we can just remove the path there for production env, let target deployment set the PIDFILE explicitly) 

https://github.com/lobsters/lobsters/blob/d6194605f17fd2354f524077f9875ec2ce1fc26c/config/puma.rb#L30

Background: https://github.com/lobsters/lobsters/pull/1667#issuecomment-3080262996

### DIFF
--- a/app/jobs/restic_job.rb
+++ b/app/jobs/restic_job.rb
@@ -3,6 +3,11 @@ class ResticJob < ApplicationJob
 
   def perform(*args)
     shared = "/home/deploy/lobsters/shared"
+    # Check if the shared directory exists
+    unless File.directory?(shared)
+      Rails.logger.warn "ResticJob: Shared path '#{shared}' does not exist. Skipping backup."
+      return
+    end
     system("source #{shared}/etc/restic-env ; restic backup --no-scan #{shared}/etc #{shared}/log")
   end
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -68,7 +68,7 @@ Rails.application.configure do
   # https://hatchbox.relationkit.io/articles/61-accessing-your-server-logs-in-hatchbox
   config.logger = ActiveSupport::BroadcastLogger.new(
     ActiveSupport::Logger.new($stdout),
-    ActiveSupport::Logger.new("/home/deploy/lobsters/shared/log/rails.log")
+    ActiveSupport::Logger.new(Rails.root.join("log/rails.log"))
   )
   config.logger.formatter = config.log_formatter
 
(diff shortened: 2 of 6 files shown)


config/storage.yml
```diff
@@ -1,7 +1,3 @@
 local:
   service: Disk
   root: <%= Rails.root.join("storage") %>
-
-production:
-  service: Disk
-  root: /home/deploy/lobsters/shared/shared/storage
```
yuanji-dev, line 7: If `Rails.root.join("storage")` symlinks to `/home/deploy/lobsters/shared/shared/storage`, there is no need to define another Disk service for production :thinking:

config/environments/production.rb
```diff
@@ -122,7 +122,7 @@
   # Skip DNS rebinding protection for the default health check endpoint.
   # config.host_authorization = { exclude: ->(request) { request.path == "/up" } }
 
-  config.active_storage.service = :production
+  config.active_storage.service = :local
```
yuanji-dev, line 125: by reading the Rails's `storage.yml` example, the `local` can be seen as local disk, not local env, so i guess it also makes sense that production env uses local storage.

https://github.com/rails/rails/blob/97dac99b5e80cc0aab3764e86fbcd12f4c316112/railties/lib/rails/generators/rails/app/templates/config/storage.yml.tt#L5-L15

config/storage.yml
pushcx: Ah, yeah, I found the documentation at https://api.rubyonrails.org/v8.0.0/classes/ActiveStorage/Service.html and I see that these two sections are equivalent.

pushcx:
Thanks for running these down. I'm going to deploy these now so I can make sure it works in prod. :)

yuanji-dev:
@pushcx Hope everything works fine in the Hatchbox based deployment in prod.

Actually i don't have an account on Lobsters so many of the features I can't actually view and check, do you mind sending me a invitation? Although i don't know if it will succeed or not, I do plan to run a site based on the code base for a local community, anyhow I feel good to contribute on the codebase itself.

app/jobs/restic_job.rb
```diff
@@ -3,6 +3,11 @@ class ResticJob < ApplicationJob
 
   def perform(*args)
     shared = "/home/deploy/lobsters/shared"
+    # Check if the shared directory exists
+    unless File.directory?(shared)
+      Rails.logger.warn "ResticJob: Shared path '#{shared}' does not exist. Skipping backup."
+      return
+    end
```
yuanji-dev, line 10: BTW, when adding this check logic, i found the `system` call below won't raise exception even the path or restic doesn't exist on server. So I hope as Lobsters' sysadmin @pushcx  you have other ways to monitor the backup is actually working. Or maybe add `exception: true` to `system` to let it raise errors when exception happens, at least failed jobs can be found at `/jobs` 

One more thing, I'm curious how many finished jobs Lobsters currently keep given the config below, I ran an instance on my machine for 8 days, it has around 3.1k jobs in history now.

https://github.com/lobsters/lobsters/blob/c2b7f0503d1421d0389777f83fce161d0a4b609c/config/environments/production.rb#L89

PR #1679, opened 2025-07-27
fix: hidden by info should come after hide link

yuanji-dev:
When testing the hide/unhide feature locally, i found the "hidden by" info is displayed after "save/unsave" section, ideally it should come after the "hide/unhide" section?

Since i don't have an account on lobste.rs, i'm not sure if current behaviour is expected or not. if it's on purpose please feel free to close the PR.

---

Before:

image

After:

image


pushcx:
Yes, this is an improvement to put the two pieces of info next to each other. I made many changes to this template recently, I must have accidentally put them in the wrong order. Thanks!

PR #1682, opened 2025-07-31
fix(mod_note): prevent creating duplicate mod note when preview

yuanji-dev:
Found a bug(creating duplicate mod notes) when a new user submit a story from a unseen domain.

### DIFF
--- a/app/models/mod_note.rb
+++ b/app/models/mod_note.rb
@@ -209,7 +209,7 @@ class ModNote < ApplicationRecord
   # and save attempts, leading to duplicate notes
   def self.create_without_dupe!(attrs)
     latest = attrs[:user].mod_notes.last
-    return latest if latest && attrs[:moderator] == latest.moderator && attrs[:note] == latest.note
+    return latest if latest && attrs[:moderator] == latest.moderator && attrs[:note].strip == latest.note
     ModNote.create!(attrs)
   end
 end
--- a/app/models/mod_note.rb
+++ b/app/models/mod_note.rb
@@ -31,7 +31,7 @@ class ModNote < ApplicationRecord
   end
 
   def note=(n)
-    self[:note] = n.to_s.strip
+    self[:note] = n
     self.markeddown_note = generated_markeddown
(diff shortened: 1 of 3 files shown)


app/views/stories/new.html.erb
```diff
@@ -21,7 +21,7 @@
 </div>
 
 <div id="story_preview">
-  <% if @story.previewing && @story.valid? %>
+  <% if @story.previewing && @story.errors.empty? %>
```
yuanji-dev, line 24: Firstly, i saw `@story.valid?` was called twice(one in controller, another in view) and both call `tattle_on_story_domain`.

app/models/mod_note.rb
```diff
@@ -209,7 +209,7 @@ def self.tattle_on_traffic_attribution!(story)
   # and save attempts, leading to duplicate notes
   def self.create_without_dupe!(attrs)
     latest = attrs[:user].mod_notes.last
-    return latest if latest && attrs[:moderator] == latest.moderator && attrs[:note] == latest.note
+    return latest if latest && attrs[:moderator] == latest.moderator && attrs[:note].strip == latest.note
```
yuanji-dev: However if `create_without_dupe!` works fine it should prevent creating duplicate mod notes, then i started digging deeper, i found there is:

https://github.com/lobsters/lobsters/blob/407a212f68fb1e047766725a9a39fe9cf97bdfb8/app/models/mod_note.rb#L33-L36

which makes `attrs[:note] == latest.note`  never be true. :disappointed:

pushcx: Thank you for running down this bug, it has irritated me for years.

I think I'd rather you remove the `.strip` from `note=` than try to replicate it everywhere. I can't guess why I added it in the first place; the whitespace will either be stripped by the markdown processor or ignored by the browser.

pushcx (changes requested):
This has bothered me for years! I have suggested a slightly more robust fix, and I look forward to merging this very soon.

PR #1683, opened 2025-07-31
Link message read status with notifications.

thomasdziedzic:
related to https://github.com/lobsters/lobsters/issues/1076#issuecomment-3114031461

I think this is still a work in progress, but wanted to get feedback for it.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -3,7 +3,7 @@
 class NotificationsController < ApplicationController
   before_action :require_logged_in_user
   before_action :set_page, only: [:all]
-  after_action :update_read_at, only: [:unread]
+  after_action :update_read_at
 
   def all
     notifications_per_page = 25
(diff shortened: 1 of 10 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -13,7 +13,7 @@ def all
       .offset((@page - 1) * notifications_per_page)
       .limit(notifications_per_page)
       .order(created_at: :desc)
-      .includes(user: [:hidings, :votes], notifiable: {story: [:tags, :user], user: [:comments], author: [], parent_comment: []})
+      .preload(user: [:hidings, :votes], notifiable: {story: [:tags, :user], user: [:comments], author: [], parent_comment: []})
```
thomasdziedzic, line 16: only `preload` is supported for polymorphic types so don't leave it up to chance for includes to pick eager.
https://stackoverflow.com/a/46838595

spec/controllers/notifications_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
+# typed: false
+
+require "rails_helper"
+
+describe NotificationsController do
+  let(:author) { create(:user) }
+  let(:recipient) { create(:user) }
+
+  describe "GET all" do
+    it "marks the notification and associate message as read" do
```
pushcx:
```suggestion
    it "marks the notification and associated message as read" do
```

pushcx:
Looks like a good set of fixes!

We talked through this in detail on [today's stream](https://push.cx/stream/2025-07-31).

PR #1684, opened 2025-07-31
hide bad faith notifications

thomasdziedzic:
in response to https://github.com/lobsters/lobsters/issues/1076#issuecomment-3140470057


pushcx:
Thanks for making this change. I know it's harder to delete a neat feature!

PR #1689, opened 2025-08-02
markdown help: Document the syntax to refer to users

PuercoPop:
Technically it is markdown, but its part of markup that lobste.rs supports.

### DIFF
--- a/app/views/global/_markdownhelp.html.erb
+++ b/app/views/global/_markdownhelp.html.erb
@@ -33,6 +33,10 @@
   text</pre></td>
     <td>prefix text with at least<tt>&nbsp;&nbsp;&nbsp;&nbsp;4 spaces</tt></td>
   </tr>
+  <tr>
+    <td>~</td>
+    <td>Refer to users by prefixing their username with a tilde</td>
+  </tr>
   <% if allow_images %>
     <tr>
       <td>(inline image)</td>


pushcx (commented):
Thanks for adding this. Sorry that it came at the same time I was bobbling the 'master' -> 'main' branch rename, so those changes got mixed in here.

I appreciate you documenting this feature, though it doesn't actually work this way right now. `@mention` works but `@mention` doesn't. And it doesn't yet because it's waiting behind #1234 which I'm slowly working through in #1627. So this is going to hang out for a bit before I can merge it!

app/views/global/_markdownhelp.html.erb
```diff
@@ -33,6 +33,10 @@
   text</pre></td>
     <td>prefix text with at least<tt>&nbsp;&nbsp;&nbsp;&nbsp;4 spaces</tt></td>
   </tr>
+  <tr>
+    <td>~</td>
+    <td>Refer to users by prefixing their username with a tidle</td>
```
kconner: s/tidle/tilde

PuercoPop: @kconner Thanks for catching this, fix'd

pushcx:
It's ok, stalebot. This one is blocked.

pushcx:
Sorry this took a minute to get back to, but the site now supports the doc.

PR #1697, opened 2025-08-08
fix(unvote): clear arrow highlight after unvoting

### LINKED ISSUE #1680: Undoing vote doesn't clear arrow highlight
Situation: I've voted for a story/comment, so the upvote arrow is highlighted red. I click on the arrow again (trying to undo my vote).

Expected behavior: The vote count goes down by one and the arrow turns back to grey.

Actual behavior: The vote count goes down by one, but the arrow stays highlighted red (until I refresh the page)

I've verified this happens on Android, using both Vivaldi and Chrome.

yuanji-dev:
Fix #1680 

That bug i think was introduced by #1645, previously there is a logic for mobile device to explicitly clear the highlight color, however that logic was ignored by the `!important`.

This PR reverts the change made in #1645 and provides another solution to fix the original issue described in #1645

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -469,7 +469,7 @@ div.voters {
 	-webkit-text-stroke: 1px var(--color-fg-contrast-4-5);
 }
 .upvoted .upvoter:before, .upvoter:hover:before {
-	color: var(--color-fg-accent) !important;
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for looking into this. The `brakeman` failure is mine; an edit updated one of the brakeman signatures and I didn't catch the build failure.

PR #1698, opened 2025-08-08
Convert settings and invitations views to CSS grid

chamlis:
Not much interesting here - the usual conversion of forms, with a couple of views that used `class="box"` unnecessarily, which I've removed in the hopes of one day getting rid of all floats across the site.

Tiny changes that are bundled in:
- fixing a small typo on the invitation request page (in a separate commit)
- changing an `na` class to a `hint`, which seem to apply equivalent CSS. `na` is used in a few more places, if I change another page I see it on, I'll make the same change.


pushcx (approved):

pushcx:
Thanks, nice to get another set of pages converted over!

PR #1699, opened 2025-08-08
Fix fake_data not flagging comments as a user that replied to them

chamlis:
This is a blind fix, but I think comments.last was referring to the reply, as it comes after the second `Comment.create!`.


pushcx:
Ah, that's a pretty silly mistake, thank you. That's what happens when I write code that juggles so much state on-stream.

PR #1700, opened 2025-08-11
[1581] Extract to_html_label from hat model to hats helper

### LINKED ISSUE #1581: Refactor Hat.to_html_label to helper
This model method generates HTML but isn't called from anywhere outside the view, it should be refactored into a helper.

sjeanc911:
resolves #1581

### DIFF
--- a/app/helpers/hats_helper.rb
+++ b/app/helpers/hats_helper.rb
@@ -1,0 +1,32 @@
+# typed: false
+
+module HatsHelper
+  def hat_html_label(hat)
+    hl = hat.link.present? && hat.link.match(/^https?:\/\//)
+
+    h = "<span class=\"hat " \
+      "hat_#{hat.hat.gsub(/[^A-Za-z0-9]/, "_").downcase}\" " \
+      "title=\"Granted #{hat.created_at.strftime("%Y-%m-%d")}"
+
+    if !hl && hat.link.present?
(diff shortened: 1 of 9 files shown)


sjeanc911:
This is ready for review. Thank you!

app/helpers/hats_helper.rb
```diff
@@ -0,0 +1,32 @@
+# typed: false
+
+module HatsHelper
+  def hat_html_label(hat)
+    hl = hat.link.present? && hat.link.match(/^https?:\/\//)
+
+    h = "<span class=\"hat " \
```
pushcx: I should've been clearer in my bug report. One of the things I'd like out of this refactoring is to get rid of all this unsafe string concatenation. If you look at the other helpers, you'll see use of `link_to` and `content_tag` to generate HTML without needing to use `.html_safe`. Can you continue the refactor to that safer approach?

sjeanc911: Got it! I'll make the changes.

sjeanc911: Done making the changes.

pushcx:
This looks great, thank you! It'll be much easier to edit this safely. You've cleaned up the last place an ActiveRecord model produced HTML for the frontend.

I do hope you'll choose another issue to contribute to, feel free to comment or email if you'd like help picking one out. :)

PR #1701, opened 2025-08-13
Mention on /chat that IRC keeps no history

chamlis:
Semi-frequently, somebody will join #lobsters, say hi, get no response in 5 minutes, and leave. Explain that IRC won't show any history, and give an idea of the activity level of the channel. The messages per day statistic comes from my Weechat logs.

I've put this before the link to the web chat, under the theory that as soon as somebody sees how to join, they'll stop reading the page.

### DIFF
--- a/app/views/about/chat.html.erb
+++ b/app/views/about/chat.html.erb
@@ -22,15 +22,20 @@
   </p>
 
   <p>
-  If you're new to IRC, you'll have the easiest time joining with the official <a href="https://web.libera.chat/#lobsters">Libera web chat</a>.
+  If you're new to IRC, bear in mind that the channel may appear empty; you won't see any messages that were sent before you joined or whilst you were away.
+  The channel is most active during US work hours and there are usually 100&ndash;1000 messages a day.
+  If you leave the channel open you'll probably catch a discussion.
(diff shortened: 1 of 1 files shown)


chamlis:
And of course as soon as I open this, I see that the same page touches on some adjacent topics further down.

> The channel isn't busy enough that there's always someone talking, so don't be surprised if you try to start a conversation and no one responds. Leave it open and you'll see activity, though we're most active during US work hours. 

I think it'd still be good to explain why the channel looks empty when you first join, but maybe it's worth combining these topics.

pushcx:
Yeah, I think it's one of those limitations that's [so familiar I forgot it needs explaining](https://en.wikipedia.org/wiki/Curse_of_knowledge). IRC is now [so weird](https://push.cx/discord-vs-irc-notes) that it makes sense to introduce a couple of key points like that.

Could you bring the existing note about active hours up into one paragraph starting "If you're new to IRC..." with the webchat link, your new text about missing scrollback + activity level, and the existing point about roughly US hours?

And then the `irc://` link can be its own paragraph.

chamlis:
I've combined the two sections as you suggested, and restructured them slightly in an attempt to sate my inner Strunk & White.

PR #1703, opened 2025-08-14
Refactor story repository: categories

### LINKED ISSUE #1523: refactor away StoryController to scopes
When `StoryRepository` was written it was a good way to collect related functionality; Rails scopes weren't as powerful and in common use. There were originally some other details in this class (constants, maybe a helper method or two) that have been refactored away over the last decade.

At this point, StoryRepository doesn't have a purpose. Its methods should be extracted to individual scopes on Story (like `.recent` was a while ago).

The StoryRepository methods roughly parallel HomeController pages. As part of this PR, please confirm there is a smoke test for each of those. You don't have to get more detailed about testing these scopes.

nav-mike:
close #1523

Move `categories` method from the repository to `Story` model as a scope.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -93,11 +93,10 @@ class Story < ApplicationRecord
   }
 
   scope :categories, ->(user, categories) {
-    tagged_story_ids = Tagging.select(:story_id).where(tag_id: Tag.where(category: categories).select(:id))
-
     base(user)
       .positive_ranked
-      .where(id: tagged_story_ids)
+      .joins(:tags)
+      .where(tags: {category: categories})
       .order(created_at: :desc)
   }
 


app/models/story.rb
```diff
@@ -92,6 +92,15 @@ class Story < ApplicationRecord
     top.order(score: :desc)
   }
 
+  scope :categories, ->(user, categories) {
+    tagged_story_ids = Tagging.select(:story_id).where(tag_id: Tag.where(category: categories).select(:id))
```
pushcx: I don't think I thought about this code before, but this and `tagged` look really inefficient. They fetch the list of every story id with the given tagging. There's now 15,000 stories tagged 'programming', and more every day.

Do you mind refactoring these two to use `.join` so the database can be smart about this? If you don't want to take it on, I can file a new issue for the performance work, but I figured I should ask as you now know these queries and tests best.

nav-mike: should be better

Issue #1670, opened 2025-07-16
Block disposable emails


pushcx:
* Add a `recurring.yml` task job to fetch the blocklist once a week to `storage/disposable_email_blocklist.conf`.
 * Add a helper to check the list; instead of their [reject_email_blocklist](https://github.com/disposable-email-domains/disposable-email-domains?tab=readme-ov-file#ruby-on-rails) function, check `/^#{email[1+email.index('@')..]}$/.match File.read(...)` to save allocating an array we don't care about. Return `false` if the list isn't present on disk.
 * Add a `User` validation.
 * Add an `InvitationRequest` validation.
 * Add an `Invitation` validation.
 * On `/settings`, add a warning to the email field if the email is on the blocklist. "We don't send email to disposable email domains; change your email address."
 * Configure ActionMailer to discard any attempts to email those domains.

pushcx:
This was completed in the linked PR.

PR #1704, opened 2025-08-15
Block disposable emails

Akarsh-Choudhary:
This PR aims to work towards [this issue](https://github.com/lobsters/lobsters/issues/1670)
It retrieves the block list and stores the list locally.
It adds validation to the relevant models and to the `ApplicationMailer`
There is still work to be done but I thought I would share the progress I have made.

This is my first time contributing to a open-source project so you will have to bare with me
I haven't been able to test these changes from the console because I keep getting errors and I thought I would be better to ask for help. 
```/home/akarsh/OpenSource/lobsters/vendor/bundle/ruby/3.3.0/gems/trilogy-2.9.0/lib/trilogy.rb:18:in `_connect': trilogy_connect - unable to connect to db:3306: TRILOGY_DNS_ERROR (Trilogy::BaseConnectionError)``` 
Occurs whenever I try to access an existing record via the rails console.

I am getting a similar problem when I try to debug why some of the tests are failing:
`$ bundle exec rspec ./spec/features/user_administration_spec.rb:39`
```
ActiveRecord::DatabaseConnectionError:
  There is an issue connecting with your hostname: db.

  Please check your database configuration and ensure there is a valid connection to your database.
```
I have setup my local dev environment according to the [setup with docker guide](https://github.com/Akarsh-Choudhary/lobsters/blob/main/docs/setup_with_docker.md).

Finally, a few questions:

- On /settings, add a warning to the email field if the email is on the blocklist. "We don't send email to disposable email domains; change your email address."

On this point I'm not too sure where to add the warning. Server side validation on the model is already done and it seems that form errors are displayed so I'm not sure what else to add.

I read the guidelines and it seems that testing is very limited, so I'm not sure if I should write any tests?
If you want me to I definitely can.

I'm thinking of talking about this work in a presentation I have coming up, so all feedback is appreciated.
Thanks.

### DIFF
--- a/app/helpers/email_blocklist_helper.rb
+++ b/app/helpers/email_blocklist_helper.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module EmailBlocklistHelper
+  def email_on_blocklist?(email)
+    /^#{email[1 + email.index("@")..]}$/.match? File.read(FetchEmailBlocklistJob::STORAGE_PATH)
+  end
+end
--- a/app/helpers/email_blocklist_helper.rb
+++ b/app/helpers/email_blocklist_helper.rb
@@ -1,7 +1,0 @@
-# frozen_string_literal: true
-
-module EmailBlocklistHelper
-  def email_on_blocklist?(email)
-    /^#{email[1 + email.index("@")..]}$/.match? File.read(FetchEmailBlocklistJob::STORAGE_PATH)
-  end
-end
--- a/bin/setup
+++ b/bin/setup
@@ -25,11 +25,6 @@ FileUtils.chdir APP_ROOT do
     FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
   end
 
-  unless File.exists?('storage/disposable_email_blocklist.conf')
-    FileUtils.mkdir_p('storage')
-    FileUtils.touch('storage/disposable_email_blocklist.conf')
-  end
-
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
 
--- a/bin/setup
+++ b/bin/setup
@@ -25,6 +25,11 @@ FileUtils.chdir APP_ROOT do
     FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
   end
 
+  unless File.exists?('storage/disposable_email_blocklist.conf')
+    FileUtils.mkdir_p('storage')
+    FileUtils.touch('storage/disposable_email_blocklist.conf')
+  end
+
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
 
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1247,7 +1247,8 @@ label.required {
 
 .labelled_grid textarea,
 .labelled_grid input[type="text"],
-.labelled_grid input[type="password"] {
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by pushcx** [comment]
I think this error means you have the app correctly running from `Dockerfile.dev` but not mariadb wired up via `docker-compose.yml`. Maybe @jacklynhma has seen an error like this and would have a quick idea?

**Comment by jacklynhma** [comment]
@Akarsh-Choudhary If you are running docker and trying to run tests, then you have to use this command https://github.com/Akarsh-Choudhary/lobsters/blob/main/docs/setup_with_docker.md#running-tests-in-docker (ie: `docker compose run app rspec` ) 

If that doesn't work, can you post a copy of `database.yml` and what the output is if you run `docker ps`.  (https://github.com/Akarsh-Choudhary/lobsters/blob/main/docs/setup_with_docker.md#debugging-in-docker) 

Try to also run through the steps `Debugging in Docker`.

**Comment by Akarsh-Choudhary** [comment]
@jacklynhma That solved the problem thanks so much for your help!

**Comment by Akarsh-Choudhary** [comment]
The tests seem to be failing on the github actions because the `storage/disposable_email_blocklist.conf` file does not exist even though in `setup.rb` it should create the file if it doesn't exist. Not sure what to do about this.

File: bin/setup [hunk 1]
```diff
@@ -25,6 +25,11 @@ FileUtils.chdir APP_ROOT do
     FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
   end
 
+  unless File.exists?('storage/disposable_email_blocklist.conf')
```
pushcx: Please just delete this, let's tweak the reader to be forgiving if the dir/file doesn't exist. We can't count on all developers to run this script, especially if they already have a dev env set up.

app/helpers/email_blocklist_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module EmailBlocklistHelper
+  def email_on_blocklist?(email)
+    /^#{email[1 + email.index("@")..]}$/.match? File.read(FetchEmailBlocklistJob::STORAGE_PATH)
```
pushcx:
```suggestion
    /^#{email[1 + email.index("@")..]}$/.match? File.read(FetchEmailBlocklistJob::STORAGE_PATH)
  rescue FileNotFound
    return false
```

Explained below!

app/helpers/email_blocklist_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module EmailBlocklistHelper
```
pushcx: Rename to `EmailBlocklistValidation` and `extend ActiveSupport::Concern` (look at `app/models/concerns/token.rb` which is quite similar), then move the model `validate` call and validation method into this module.

If we need the method elsewhere, I'm OK with a call saying `EmailBlocklistValidation.email_on_blacklist?` referring directly to the object from the global namespace, I don't think the OO style is improved by mixing it into more objects. But having a mixin would let us remove the duplication in these models!

pushcx (changes requested):
Great start, you've got the heart of the feature implemented! There's one bit of OO style I'd like you to change.

And then we discussed the `/settings` feature on [today's office hours stream](https://push.cx/stream/2025-08-25). The idea is that because we're now considering existing data to be invalid, we want to show users the moment they load the form rather than only when they submit (they might not even be editing that field, so it could be an unpleasant surprise).

Issue #872, opened 2020-06-25
Disable on-site reply notifications totally or per-comment


pushcx:
In the upper-right of every page, there's a link to the Replies page with a number that turns red to indicate new replies. Users have reported finding this distracting, and especially frustrating when they want to move on from a particular conversation.

Let's add three boolean options that to `/settings` to turn off on-site comment reply, comment mention, and PMs, defaulting to `true`.

And then on individual comments, let's add a nullable boolean indicating if the user wants replies for this particular comment, with the default `null` indicating the site should use the user-level option. This seems like the least surprising and annoying way to handle the corner cases that arise when users turn on and off the global setting and individual comments. On each comment, show a `disable replies` link when it's `null`/`false`, and `enable replies` if it's `true`. No UI to set it back to `null` because it doesn't seem worth it given that few comments 

Also, email/pushover notifications should not send if this comment-level boolean is set to `false`. I don't want to try to squeeze the word "on-site" into the UI and it seems really likely that someone who doesn't want to see a number on the site also doesn't want their phone to get a notification.

Issue #1140, opened 2022-11-09
Inbox count is inconsistent


davidchisnall:
For the last few months, the red indicator next to Inbox has been inconsistent.  Often:

 - When loading `lobste.rs` for the first time, it is not there.
 - When clicking on a story, it will show a value.
 - When clicking on the inbox, it will then show a different number of unread messages (sometimes zero).

In addition, read messages periodically pop back up in the unread list.

pushcx:
This is an old bug that was squashed but not closed, then recreated:

In the new Notification model the `good_faith`/`should_display` boolean is in Ruby rather than a database scope.

(There is probably another dupe of this hanging around, I thought I'd seen it diagnosed before.)

PR #1711, opened 2025-08-31
Add inbox_mentions setting

untitaker:
Since people keep mentioning cadey in the friendlysock thread, this
seems like a good time to add a setting.

This is a really basic subset of https://github.com/lobsters/lobsters/issues/872

I can add the other booleans as well, but thought i'd start with the most minimal needed change.

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -275,7 +275,7 @@ class SettingsController < ApplicationController
   def user_params
     params.require(:user).permit(
       :username, :email, :password, :password_confirmation, :homepage, :about,
-      :email_replies, :email_messages, :email_mentions,
+      :email_replies, :email_messages, :email_mentions, :inbox_mentions,
       :pushover_replies, :pushover_messages, :pushover_mentions,
       :mailing_list_mode, :show_email, :show_avatars, :show_story_previews,
       :show_submitted_story_threads, :prefers_color_scheme, :prefers_contrast
--- a/app/jobs/notify_comment_job.rb
+++ b/app/jobs/notify_comment_job.rb
@@ -15,7 +15,9 @@ class NotifyCommentJob < ApplicationJob
   def deliver_mention_notifications(comment, notified)
     to_notify = comment.plaintext_comment.scan(/\B[@~]([\w\-]+)/).flatten.uniq - notified - [comment.user.username]
     User.active.where(username: to_notify).find_each do |u|
-      u.notifications.create(notifiable: comment)
+      if u.inbox_mentions?
+        u.notifications.create(notifiable: comment)
+      end
 
       if u.email_mentions?
         begin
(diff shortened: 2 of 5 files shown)


pushcx:
Thanks for taking this on, a partial solution seems worth implementing.

I think you did this about the same time I wrote 917f0a1c3c215e0ac36fef86a14b8c0429a70af1 so I see there's a conflict, and it's not using the pattern I tried to document there. We want to always create a Notification and then that class has a method to filter them out based on user settings (and events in the thread). The means that as the user changes their settings, those Notifications immediately disappear or appear as desired.

The name 'good faith' and 'bad faith' in the Notification class are a little misleading and reflects that they can also be hidden by events in the comment thread like flags. If you have ideas to better reflect the full functionality and would make it obvious how they're to be used, that would be a help here.

untitaker:
I think I addressed your review, but I wonder:

> The means that as the user changes their settings, those Notifications immediately disappear or appear as desired.

this doesn't seem fully desirable to me, if i turn on inbox_mentions after many years of having it off, my inbox is completely backlogged?

untitaker:
ehh this works but there is still a bug with the red notification count next to `Inbox`. it's getting increased regardless of whether the user chose to hide mention notifications entirely. i think that's pretty "unhealthy" when i'm trying to escape the noise but i see that number going up

pushcx:
That bug is #1140.

untitaker:
So looking at your comments in #1720 i feel like I should just mark the notification instantly as read (right after creation) if not `should_display?`, and that'd fix a good chunk of issues related to #1140?

app/models/notification.rb
```diff
@@ -9,20 +9,41 @@ class Notification < ApplicationRecord
...
+    true
+  end
+
+  private
+
+  def user_wants_notification?
+    comment = notifiable
+
+    # Check if this is a mention notification
+    if comment.plaintext_comment.match?(/\B[@~]#{Regexp.escape(user.username)}\b/)
```
pushcx: I just wrote this exact same regexp in `extras/markdowner.rb`, let me lift it up to a constant you can reuse.

pushcx: 20d85ebf

pushcx:
Yeah, that's a clean way to handle it. (Sorry such a simple answer took so long, I was traveling.)

I added a constant for you to reuse in 20d85ebf. With those two small changes, this is ready to go.

Thanks for improving the name of those functions.

app/models/notification.rb
pushcx:
```suggestion
    if comment.plaintext_comment.match?(Markdowner::USERNAME_MENTION)
```

app/models/notification.rb
```diff
@@ -9,20 +9,41 @@ class Notification < ApplicationRecord
```
pushcx, line 9:
```suggestion

  before_validation on: :create do
    self.read_at = Time.current if !should_display?
  end

```

pushcx:
Thanks for coding this, @untitaker. I made the last couple small tweaks to get this merged before merge conflicts can really start accumulating or stalebot comes around to hassle.

Issue #1702, opened 2025-08-13
Page /domains/X may misreport results


ollollol:
I found an instance when the _/domains/X_ page ([for_domain view](https://github.com/lobsters/lobsters/blob/HEAD/app/views/home/_for_domain.html.erb)) claims non-zero results, but shows nothing.

Case in point: <https://lobste.rs/domains/quad9.net> reads "1 story from 1 submitter", but in fact doesn't show any.

The issue reproduces when both logged in and out.

pushcx:
Thanks for reporting this. We've been meaning to redo pagination (#394) but this is actually a bug in how deleted stories are displayed.

Akarsh-Choudhary:
It seems that when a user searches for a domain that has never been submitted on the site before they're shown the 404 page. I'm not sure if this is intended but I feel like it would make sense to display some kind of message saying "No stories with this domain found" while still showing the deleted stories for moderators/admins. 
If we want to go down this route it would probably be useful to have 2 cache counters on the domain model, "stories_count", "active_stories_count". 
What do you think?

Edit:
I think the reason a 404 is shown is for RSS and JSON responses? Would the correct response for when a domain record exists, but it has no associated stories to return a 404 then?
Also it seems like to get the number of "active" story counts initially for domain some kind of script would need to be run on production console.

Issue #1084, opened 2022-04-22
Negative story count


ron-wolf:
I believe [user @apg](https://lobste.rs/u/apg) has a negative story count. Not sure how this could come about as I am unfamiliar with the source code (unless I'm stumbling on some sort of elaborate prank 😄). They are a user of eight years, though, so I guess their data may have been through a migration or two.

!["Long screenshot" of the relevant user profile, including the negative story count](https://user-images.githubusercontent.com/7342888/164701306-f86f6e1c-b543-4c5b-aef7-c4e87a74c5f3.PNG)

Note: I tried the same when logged out. I did not try from a different IP but I doubt that matters in this case.

NattyNarwhal:
I believe this happens if the user deleted their account and comes back, presumably with disownment before.

norpol:
Just noticed for another user who's profile is completely deactivated with disowning and no restoration.

PR #1713, opened 2025-09-01
Fix domain story counter cache

Akarsh-Choudhary:
This PR aims to work towards this [issue](https://github.com/lobsters/lobsters/issues/1702)

This bug arises through the use of `counter_cache` which is dependent on records being created and destroyed.
When a story is "deleted" it is not actually deleted rather it is marked as deleted via `is_deleted`.
This is in place so users can "undelete" their stories. 

In order to store the actual number of not deleted stories, I have a migration to introduce a new column on domain `active_stories_count` although I'm not 100% on the naming of this. 
We would need to pre-populate this field before merging any changes.

```
Domain.all.each do |domain|
  domain.active_stories_count = stories.where(is_deleted: false).count
end
```
 Although there is probably a more efficient way to accomplish this ^ (I need to learn more about N+1 queries)
 
 A new setter method is created on `story` so that when we flip from deleted to not deleted state, the domain's `active_stories_count` is updated accordingly. 
 Finally these changes are displayed for normal users, moderators will still be able to see the total number of stories associated with a domain (deleted and not deleted)

Let me know if this is not the correct approach to take and I am also happy to write tests :)

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -246,6 +246,8 @@ class HomeController < ApplicationController
   def for_domain
     @domain = Domain.find_by!(domain: params[:id])
 
+    return render "about/404", status: :not_found if @domain.active_stories_count.zero? && !@user&.is_moderator?
+
     @stories, @show_more = get_from_cache(domain: @domain.domain) do
       paginate @domain.stories.base(@user).order(id: :desc)
     end
--- a/app/models/domain.rb
+++ b/app/models/domain.rb
@@ -15,6 +15,7 @@ class Domain < ApplicationRecord
   validates :selector, length: {maximum: 255}
   validates :replacement, length: {maximum: 255}
   validates :stories_count, numericality: {only_integer: true, greater_than_or_equal_to: 0}, presence: true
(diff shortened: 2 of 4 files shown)


pushcx:
Thanks for taking a look at this bug, I think you've correctly diagnosed it. There's a related bug #1084 for users that probably has the same or a very similar cause, so I think you have diagnosed two bugs. :)

I look at all the code you had to write and I think that points to the codebase not using the correct approach from the start. There are not so many rows that counting them is expensive (120k users; 600k stories). We shouldn't use a counter_cache for these at all, we should just make sure that there are indexes:

 * comments: index on user_id
 * stories: index on user_id
 * stories: index on domain_id
 * stories: index on origin_id

Does that plan make sense? Would you like to open a PR with a migration to remove these counter caches, update the controllers to load the correct counts, and confirm the indexes exist in `schema.rb`?

Akarsh-Choudhary:
That makes sense, thank you.
I will continue to work on this when I can, but I have got back to university and I am very busy with class etc. 

Hopefully I can find some time for this :)

Issue #1718, opened 2025-09-09
Mention database credentials in setup instructions


chamlis:
The other day somebody came by the IRC chat running into an issue whilst setting up the site for development. It looks like the setup instructions imply that the database password should be set through `rails credentials:edit`, whereas they actually live in `config/database.yml`.

We could simply mention this extra file in the setup instructions, but I noticed it doesn't actually exist until it's [created by `bin/setup`](https://github.com/lobsters/lobsters/blob/main/bin/setup#L23). If we did this, the setup would go something like:
1. Run `bin/setup`, get an error when it tries to set up the database;
2. Find that the README explains this error, and add the credentials to `config/database.yml`;
3. Run `bin/setup` again.

Is there anything stopping us shipping `config/database.yml` directly in the repository, so we can ask people to add their credentials there before running `bin/setup` at all?

pushcx:
This was completed in the linked PR.

PR #1722, opened 2025-09-17
Improve setup, instructions

fpsvogel:
I'm getting ready to pick up an issue, but first here are improvements around friction that I ran into during setup.

This PR improves the setup experience:

- Avoids a database setup error from `bin/setup` by including the default database username and password in the sample `database.yml`.
  - If the sample `database.yml` should not be changed, I made [a separate prior commit](https://github.com/lobsters/lobsters/pull/1722/commits/d6038eda164f184d9661f23bdb53baee0d487ebc) where a point is simply added to the README about manually editing `database.yml` to add a username and password, after the setup error occurs. (This step [is already in the Docker setup instructions](https://github.com/fpsvogel/lobsters/blob/main/docs/setup_with_docker.md#:~:text=Add%20the%20below%20code%20to%20yoru%20development%20primary%20and%20test%20primary%3A).)
  - Note: there is [an open Issue](https://github.com/lobsters/lobsters/issues/1718) about this same database setup error.
- In the README, points out the `test` user password so that readers don't have to guess or go hunting for it in `seeds.rb`.
- Corrects a few misspellings that I noticed in the docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ Do not submit code written by LLM-powered coding tools because of the [uncertain
 While this project's license allows for modification and use to run your own website,
 this source code repository is specifically for the code running the website at [lobste.rs](https://lobste.rs/).
 
-We're very deliberate about new features and behavior changes because they have difficult-to-forsee social effects or maintenance costs.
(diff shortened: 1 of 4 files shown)


pushcx (approved):
This is all great! You're one of the first people to work through these instructions, so I really appreciate you fixing so many rough edges. (So many typos!)

Issue #1268, opened 2024-04-04
impossible to search for c++


fanf2:
I wanted to find my past comments that mentioned C++, but it seemed that the ++ was ignored. I tried quoting with " and escaping with \ neither of which worked.

I looked at the source code and it seems `strip_operators` in `lobsters/app/models/search.rb` removes almost all punctuation.

Could it please be relaxed to retain `+` (for C++) and `#` (for C# and F#)?

pushcx:
It would be great to allow `+`. It’s a [search operator](https://mariadb.com/kb/ru/full-text-index-overview/#in-boolean-mode) in MariaDB, so this needs to robustly escape. This will need a bunch of tests to confirm.

`#` doesn’t have this same concern.

Note that there’s multiple code paths worth checking here, for `:tag`, `:term`, and `:quoted`.

Supporting either of these probably [requires] (https://dev.mysql.com/doc/refman/8.4/en/fulltext-fine-tuning.html) creating a [user-defined collation](https://dev.mysql.com/doc/refman/8.4/en/full-text-adding-collation.html) to specify that these are “word characters” rather than punctuation to be ignored. There may be other viable approaches I’m unaware of, but I’ll note that I’m unwilling to accept a patch to the MariaDB source and build a custom version as I won’t be able to maintain that across database updates.

pushcx:
Spent some time today hacking away at this one. It definitely needs a user-defined collation to accept `+` and `-`; it's not enough to allow the characters through `strip_operators`.

MatheusRich:
I have yet to check the search code, but I want to highlight that rails have some potential useful sanitization methods:
 https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_like

pushcx:
I tried to use those but I think they didn't one that works in the context of an  `IN BOOLEAN MODE` search so I conservatively wrote `strip_operators` to discard based on an allowlist rather than try to escape.

PhilArmstrong:
Looking at the mariadb docs, user defined collations are deprecated & probably going to be removed in a future release.

However it seems that adding an extra charset which marks + & # as letter characters instead of punctuation & building FULLTEXT indices with that charset will still be supported & doesn’t require rebuilding mariadb, just adding a new charset file to the relevant directory in the mariadb server install: https://dev.mysql.com/doc/refman/8.4/en/fulltext-fine-tuning.html#fulltext-modify-character-set

Issue #1185, opened 2023-07-17
Search "tag" keyword is case sensitive


dhnaranjo:
When using Lobsters on a device that auto capitalizes the first character of a sentence, ex: iOS, the "tag" keyword doesn't work. This could be resolved by making the keyword consumer case insensitive or by adding 'autocapitalize="none"' to the input.

Example: https://lobste.rs/search?q=Tag%3Aweb+HTML+semantics&what=stories&order=newest vs https://lobste.rs/search?q=tag%3Aweb+HTML+semantics&what=stories&order=newest

PR #1726, opened 2025-09-30
Auto-detect domains in search

### LINKED ISSUE #1691: Search for domain names
Feature request: it would be useful if you could search for a domain name on the https://lobste.rs/search page. For example, https://lobste.rs/search?q=example.com&what=stories would return all stories from the example.com domain. Many thanks.

Ecco:
Fixes #1691 

Here are the choices I made when parsing domains:
 - For the sake of simplicity, I did not enforce the "no two consecutive dashes" rules
 - To limit false positives, I used the official [IANA list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) of TLDs (I added a command-line snippet to update it if needed)
 - Domain names are only matched if lowercase. The code is cleaner and I think it's likely to prevent false positives (I don't think I've ever typed a domain name using uppercase letters).
 - Parsing is more strict than the previous regex, and will reject more invalid domains. For instance, `com` (just the tld) is now ignored, as well as `.com` (tld and period but no domain) and `foo..com` (two consecutive periods).


app/models/search_parser.rb
```diff
@@ -22,8 +24,12 @@ class SearchParser < Parslet::Parser
 
   # User::VALID_USERNAME
   rule(:commenter) { str("commenter:") >> match("[@~]").repeat(0, 1) >> match("[A-Za-z0-9_\\-]").repeat(1, 24).as(:commenter) >> space? }
-  # reproduce the <domain> named capture in URL_RE
-  rule(:domain) { str("domain:") >> match("[A-Za-z0-9_\\-\\.]").repeat(1).as(:domain) >> space? }
+  rule(:domain) {
+    (
+      (match("[a-z0-9]") >> match("[a-z0-9\\-]").repeat(1, 62) >> str(".")).repeat(1) >>
+      Parslet::Atoms::Alternative.new(*IANA_TLDS.sort_by{-it.length}.map{ str(it) }) # Consider lengthier TLDs first because Parslet is greedy (and would match "co" before "com")
```
pushcx: Great comment for a clever bit of code.

app/models/search_parser.rb
```diff
@@ -22,8 +24,12 @@ class SearchParser < Parslet::Parser
 
   # User::VALID_USERNAME
   rule(:commenter) { str("commenter:") >> match("[@~]").repeat(0, 1) >> match("[A-Za-z0-9_\\-]").repeat(1, 24).as(:commenter) >> space? }
-  # reproduce the <domain> named capture in URL_RE
-  rule(:domain) { str("domain:") >> match("[A-Za-z0-9_\\-\\.]").repeat(1).as(:domain) >> space? }
+  rule(:domain) {
+    (
+      (match("[a-z0-9]") >> match("[a-z0-9\\-]").repeat(1, 62) >> str(".")).repeat(1) >>
```
pushcx: Please prefix an optional `str("domain:")` so we keep supporting the old syntax.

app/models/search_parser.rb
```diff
@@ -4,6 +4,8 @@
 
 # https://mariadb.com/kb/en/full-text-index-stopwords/#innodb-stopwords
 MYISAM_STOPWORDS = %w[a about an are as at be by com de en for from how i in is it la of on or that the this to was what when where who will with und the www].sort_by { it.length }.reverse.freeze
+# curl -s https://data.iana.org/TLD/tlds-alpha-by-domain.txt | grep -v "^#" | tr '\n' ' ' | tr '[:upper:]' '[:lower:]'
```
pushcx: We've had a few constants like this over the years and never manged to keep them up-to-date before we started writing jobs to do it.

Take a look at `config/recurring.yml` and use `fetch_email_blocklist` as a template for a monthly `FetchTLDsJob`. Then instead of the strategy in `app/models/concerns/email_blocklist_validation.rb` of falling through to `false` when the file doesn't exist, return a fallback list of %w[com org net ping co]` with a comment that it's for the SearchParser specs.

spec/models/search_parser_spec.rb
```diff
@@ -95,10 +95,16 @@ def to_primitive input
   end
 
   describe "domain rule" do
```
pushcx, line 97: Great set of examples here.

pushcx (changes requested):
Thanks for taking on a fairly complicated area of the codebase. I agree with all the choices you made in parsing. I really like your strategy of using TLDs and appreciate that you found an official IANA list.

There's two changes. The minor one is to keep supporting the explicit `domain:` operator. The larger one is to refresh the list with a job.

One optional feature idea: when presenting results for a search of a url, the results page could link to the search results but replacing the`url` with the `domain`. So a search for `foo url:example.com/bar` would offer a link to the search results for `foo example.com`, with text like "You may find more results if you search [the example.com domain]". If you don't want to take this on, no worries, I'll file it as a featurereq after merging this.

pushcx:
To satisfy the failing linter, you can probably run `standardrb --fix` for an automatic fix.

For the failing tests:

 * `spec/models/search_spec.rb:158` looks like the note about the `domain:` operator, and also assert that it works without the explicit `domain:`
 * `spec/models/search_spec.rb:269` doesn't need to repeat that extra assertion, so no changes once `domain:` works

Ecco:
Thanks for the review! I think I addressed all the (very valid) points you mentioned.

config/recurring.yml
```diff
@@ -26,6 +26,9 @@ production:
   fetch_email_blocklist:
     command: FetchEmailBlocklistJob
     schedule: every 7 days
+  fetch_iana_tlds:
+    command: FetchIanaTldsJob
```
pushcx:
```suggestion
    class: FetchIanaTldsJob
```

config/recurring.yml
```diff
@@ -43,3 +46,6 @@ development:
...
```
pushcx:
```suggestion
    class: FetchIanaTldsJob
```

pushcx:
Great! The way you split out the functionality in the job made for clear, readable methods. I fixed a small bug (that I had coincidentally just [fixed in prod](https://github.com/lobsters/lobsters/commit/f0c2a0115548a989d829400283da31b636d70241) a minute before!) and this is ready to go.

Thanks for taking this on, it's a nice little improvement to our search engine.

You know, if you enjoyed working on the search engine, there's a couple more bugs you could take on :smile: : #1185, #1268, #1373, #1378.

Issue #1735, opened 2025-10-06
restyle <label> tags


pushcx:
We currently use `<label>` tags on form fields and on displays of info, like the top of the user profile.

[According to research](https://www.nngroup.com/articles/required-fields/), bold (or fading) is not an effective way to indicate a field is required. We should restyle these to match common practice: no bold, remove the trailing `:`, and suffix only the required fields with a ` *`, like `Username *`. (There is a clever bit of css (from @chamlis) that only shows the colons on the desktop layout, so we'll need to remove that.)

A number of form labels had `class="required"` so that they would be bold and be visually consistent with other fields. This caused an error when #1724 changed to use the `required` attribute - those fields weren't actually required!

We should also remove the class `bold` that was introduced in #1724. Labels should either all be bold or none be bold; I don't have a strong opinion here. Removing bold feels maybe better because it removes a distinction that is not a difference. (Maybe I should crack my Bringhurst's Elements of Typographic Style.)

Should labels be bold when _viewing_ data? It seems a little more usually there, to visually separate them from the data, but probably the layout is enough to be clear.

mateusdeap:
Might I take a stab at this? I might have some ideas.

fpsvogel:
@mateusdeap I have a draft PR open: https://github.com/lobsters/lobsters/pull/1737. I started working on it yesterday and saw your message just now.

I was going to add one more thing there before marking it ready (see PR description), but I'm open to making improvements if you have ideas. Or if a follow-up PR would make more sense, I expect that would be fine too.

mateusdeap:
Hey, no problem. I'll check it out. Maybe I can help

fpsvogel:
@mateusdeap I added that last change to my PR, and put screenshots in the PR description: https://github.com/lobsters/lobsters/pull/1737. LMK if you have any suggestions.

pushcx:
@mateusdeap We have plenty of [good first issues](https://github.com/lobsters/lobsters/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) and [design issues](https://github.com/lobsters/lobsters/issues?q=is%3Aissue%20state%3Aopen%20label%3Adesign) if you're looking for something to do, or let me know what your areas of interest are and I can reach into my backlog. There are ALWAYS more opportunities for stuff to do here. 😄

PR #1737, opened 2025-10-07
Restyle <label> tags

fpsvogel:
## Summary

Resolves https://github.com/lobsters/lobsters/issues/1735

- EDIT: Restores bold and colon (:) on the Profile page. Without some sort of extra styling, the labels on that page are harder to visually parse, especially on mobile.
- Adds an asterisk (*) after labels of required elements
- Removes bold styling from all labels
- Removes colon (:) from all labels
- Removes the `required` attribute from radio button inputs
  - Rationale: these radio buttons are always pre-selected and cannot be unselected, so there's never a scenario where the user needs to be reminded to make a selection in them. Keeping the `required` attribute in these would mean that an asterisk would be shown after the label, which I'm suggesting is unnecessary:

image

## Screenshots

Here is a sampling of affected views, before and after the changes in this PR.

### Desktop

|  | search | submit | settings | settings (cont.) | profile | hat requests | edit domain |
|---|---|---|---|---|---|---|---|
| before | image | image | image | image | image | image | image |
| after | image | image | image | image | image | image | image |

### Mobile

|  | search | submit | settings | profile | hat requests | edit domain |
|---|---|---|---|---|---|---|
| before | image | image | image | image | image | image |
| after | image | image | image | image | image | image |

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1184,6 +1184,10 @@ label:has(+ *[type="hidden"] + *:required)::after {
 	font-size: 150%;
 	vertical-align: middle;
 }
+/* Labels on the (read-only) Profile page get extra contrast */
+.profile label::after {
+	content: ":";
+}
 .box img {
 	vertical-align: middle;
 }
@@ -1644,6 +1648,10 @@ input[type="submit"].link_post.pushover_button {
 	}
 	div#inside {
 		margin: 0.5rem;
+	}
+
+	label:not(:has(> input[type="radio"])) {
+		font-weight: bold;
 	}
 
 	.comment {
--- a/app/views/search/index.html.erb
+++ b/app/views/search/index.html.erb
@@ -321,7 +321,7 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
   <% end %>
 <% else %>
   <div class="help">
-    Search hints:
+    <label>Search hints:</label>
     <ul>
       <li>Search terms are required. Use quote marks if you want <span class="searchq">"words together"</span></li>
       <li>Terms are exact, so <span class="searchq">code</span> doesn't match <span class="searchq">coder</span> or <span class="searchq">coding</span>. You can approximate stemming with <span class="searchq">cod*</span></li>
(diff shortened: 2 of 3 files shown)


pushcx:
On [today's office hours stream](https://push.cx/stream/2025-10-09) @fpsvogel caught that the user profile on mobile no longer had enough distinction to be immediately clear.

I looked into it, this is the only place on the site where we use `<label>` tags in a similar layout to forms. On forms the difference between plain text labels and interactive elements is clear enough that the form structure is apparent, but it's not clear on profiles.

So we do need to figure out some distinction to add to this one view so it's clear what's happening. That could be bold labels, a different layout, a significant redesign of the page, or something else. The simplest thing do to would be to update `app/views/users/show.html.erb` so it's `<section class="profile">` and add a rule `.profile label { font-weight: bold; }`, and I'm happy to merge just that while we consider bigger changes.

Otherwise, I think this simplification is a clear improvement overall, nice to be in line with the most common UI pattern.

fpsvogel:
@pushcx In the added commit I made a change similar to what you suggested, but going a bit beyond it:

- labels are bold *on mobile, on all pages*
- labels are also followed by a colon *on the Profile page only* (mobile and desktop)

See the updated the screenshots in the PR description.

The reason I ended up adding bold to labels is that even outside the Profile page, without extra styling can be hard to visually parse, as in these examples where bold is omitted:

| search | settings | hat requests |
|---|---|---|
| Screen Shot 2025-10-09 at 11 20 01 AM | image | image |

(As an alternative to bold on mobile, I played around with showing a colon via a CSS rule, similar to [the rules](https://github.com/lobsters/lobsters/pull/1737/files#diff-5bfcdc54bcda018d14515b827c80d9f9b8ee25704a4271b560a2f87439665f17L1563) that are removed in this PR, only more general. But [some styling](https://github.com/lobsters/lobsters/pull/1737/files#diff-5bfcdc54bcda018d14515b827c80d9f9b8ee25704a4271b560a2f87439665f17R1184-R1185) in this PR that makes the asterisk larger and more noticeable would interact with that in a bad way, making the colon weirdly large. So I gave up on that and used bold instead.)

pushcx:
This looks really good, thank you for putting so much care into it.

PR #1738, opened 2025-10-07
Disable autocapitalization on the search field

### LINKED ISSUE #1185: Search "tag" keyword is case sensitive

Ecco:
Fixes #1185

This felt like the right way to fix this bug:
1. It's super easy
2. I don't think a search field should be autocapitalized anyway.

For the record, the search field on google.com has `autocapitalize="off"` .

### DIFF
--- a/app/views/search/index.html.erb
+++ b/app/views/search/index.html.erb
@@ -2,7 +2,7 @@
 <div class="box">
   <%= form_with url: '/search', method: :get do |f| %>
     <div class="boxline">
-        <%= f.text_field "q", { :value => @search.q, :size => 40 }.
+        <%= f.text_field "q", { :value => @search.q, :size => 40, :autocapitalize => "none" }.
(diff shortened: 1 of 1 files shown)


pushcx:
How nice, this isn't just a bugfix but a new CSS feature I hadn't heard of.

It's not yet in the "widely available" baseline standard we aim for, but it does seem universal in the mobile browsers where this issue is overwhelmingly likely to occur. If the browser doesn't support it, it'll be ignored rather than breaking any layout.

So I'm going to merge this, and then also add this attribute I just learned about to some more forms on the site, like login.

Issue #1723, opened 2025-09-19
Story liner has different background color on mobile


evangirardin:
On mobile, the `story_liner` class has a different background-color (`#fbfbfb`) than the rest of the page (`rgb(254, 254, 254)`). The effect is that story liners look surrounded by a non-full-page-width rectangle with a slightly darker background than the surroundings (e.g. the mobile comments link and some slices of space near the top and bottom of the page).

<img width="436" height="406" alt="Image" src="https://github.com/user-attachments/assets/5011fee0-4c97-4131-8dd5-449470a45e6a" />

It might be worth just removing the mobile-only background-color from the `story_liner` class since it's inconsistent with the background color on the rest of the site anyway.

pushcx:
Thanks, that's one of those design quirks that's been hanging around forever. I'm a big fan of keeping the design and typography quiet, that variation should try indicate a meaningful difference.

PR #1741, opened 2025-10-09
Remove story_liner background color for visual consistency

fpsvogel:
## Summary

Resolves https://github.com/lobsters/lobsters/issues/1723

Makes background colors more consistent by removing a slightly different background color that appears on part of some pages in the light color theme only.

## Alternative solution

Another approach would be to keep the slightly different background color, and extend it to the other color themes. Taking this approach, the slightly different background color in the "before" screenshot below would also appear in the dark and high-contrast themes.

In my opinion, a consistent background color (this PR's current approach) looks better, but happy to consider other points of view.

## Screenshots

The "before" image shows a slightly darker background color on the inner part (not including the comment bubbles and page selector).

| before | after |
|---|---|
| image | image |

For the other color themes, this background color is the same as the main background color, i.e. `--color-mobile-story-liner-bg` is the same as `--base-bg`. In the output below, only the last pair (`254 254 254` and `#fbfbfb`) are actually different colors.

```sh
lobsters git:(main) $ git grep -E "(--color-mobile-story-liner-bg|--base-bg):"
app/assets/stylesheets/dark-high.css:        --base-bg: 0 0 0;
app/assets/stylesheets/dark-high.css:        --color-mobile-story-liner-bg: #000;
app/assets/stylesheets/dark-normal.css:        --base-bg: 12 12 12;
app/assets/stylesheets/dark-normal.css:        --color-mobile-story-liner-bg: #0c0c0c;
app/assets/stylesheets/light-high.css:        --base-bg: 255 255 255;
app/assets/stylesheets/light-high.css:        --color-mobile-story-liner-bg: #fff;
app/assets/stylesheets/light-normal.css:        --base-bg: 254 254 254;
app/assets/stylesheets/light-normal.css:        --color-mobile-story-liner-bg: #fbfbfb;
```

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1704,7 +1704,6 @@ input[type="submit"].link_post.pushover_button {
 		display: table;
 	}
 	ol.stories.list li.story div.story_liner {
-		background-color: var(--color-mobile-story-liner-bg);
 		display: table-cell;
 		padding-top: 0.5em;
 		padding-bottom: 0.75em;
--- a/app/assets/stylesheets/dark-high.css
+++ b/app/assets/stylesheets/dark-high.css
@@ -73,7 +73,6 @@
 	--color-lobsters-hat-sysop-brim-stroke: #6f4444;
 
 	/* mobile */
-	--color-mobile-story-liner-bg: #000;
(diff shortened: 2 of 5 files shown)


pushcx:
Thanks for your screenshots, that made this change very clear.

Issue #1582, opened 2025-05-22
Split comment deleting for users and mods


pushcx:
Similar to story editing, the action for moderators to delete comments needs to be separated from the user action to delete comments. That means a separate controller in `app/controllers/mod` and a separate 'mod delete' link in `app/views/comment/_comment.html.erb`. 

Right now a mod looking at their own old comments can never see the 'disown' button because they can always delete comments. And then the 'delete' action throws away the information of whether the user deleting a comment is doing so as a moderator or a user, which `Comment.delete_for_user` tries to recreate, with the result that moderators can't delete their own comments like regular users, they always end up down the 'mod delete' path of creating modlog entries.

Basically #1431 for comments, same reasoning.

pushcx:
@taimourz addressed the feature request I wrote... which prompted me to realize I want even more features. (I always want more features.)

Specifically, there's a feature for authors to undelete comments that also contains the logic for mods undeleting comments. It also needs to be disentangled to the new `Mod::CommentsController`.

pushcx:
#1742 has a bug. I removed some trolling this morning and none of the comment deletions were modlogged. I'm working on a fix and I'll create the missing modlog entries, but I wanted to post here to keep the info together.

EDIT: I finished investigating this and no modlog entries were missing, so I did not backfill any. The functionality for mod deletion was broken so my attempts to delete the comments had silently failed. When a user deactivates or is banned, their comments with negative scores are deleted. I committed the fix for comment deletion in 6ca740c79ff0b0f4763f0c7bb0c0ebd867e5141b.

PR #1742, opened 2025-10-09
split comment deletion for users and moderators

taimourz:
addresses issue [1582](https://github.com/lobsters/lobsters/issues/1582)

This PR changes the deletion workflow as follows:

- For mod users:
1. When deleting their own comments, they have 2 options now:
  a. delete: Normal delete with no logging 
  b. mod delete: delete with logging

2.  When deleting other user comments 
  a. Uneffected
   b. Still need to provide reason for logging (only one delete button available)

- For Normal users:
    -  Unaffected

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -90,6 +90,11 @@
           <% if comment.is_deletable_by_user?(@user) %>
             |
             <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
+          <% elsif comment.is_disownable_by_user?(@user) %>
+            |
+            <%= link_post("disown", disown_comment_path(comment.short_id),
+            params: { type: 'comment', show_story: show_story, show_tree_lines: show_tree_lines })
+            %>
           <% end %>
           <% if @user&.is_moderator? %>
             |
--- a/app/views/comments/_threads.html.erb
+++ b/app/views/comments/_threads.html.erb
@@ -131,11 +131,16 @@
           |
           <a class="comment_undeletor" href="#c_<%= comment.short_id %>">undelete</a>
         <% elsif !comment.is_gone? %>
-          <% if @user && @user.id == comment.user_id && comment.is_deletable_by_user?(@user) %>
+          <% if comment.is_deletable_by_user?(@user) %>
             |
             <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
+          <% elsif comment.is_disownable_by_user?(@user) %>
+            |
+            <%= link_post("disown", disown_comment_path(comment.short_id),
+            params: { type: 'comment', show_story: show_story, show_tree_lines: show_tree_lines })
+            %>
           <% end %>
-          <% if @user && @user.is_moderator? %>
+          <% if @user&.is_moderator? %>
             |
             <a class="comment_moderator" href="#c_<%= comment.short_id %>">mod delete</a>
           <% end %>


app/controllers/mod/comments_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    comment ||= Comment.find(params[:id]) if @user&.is_admin?
+
+    if @user && comment
+      comment.current_vote = Vote.where(user_id: @user.id,
+        story_id: comment.story_id, comment_id: comment.id).first
+      comment.vote_summary = Vote.comment_vote_summaries([comment.id])[comment.id]
+      comment.current_reply = @user.ids_replied_to([comment.id])
+    end
+
+    comment
```
pushcx:
```suggestion
    Comment.find_by(short_id: params[:id])
```

pushcx: We don't need this display functionality for an action that's only hit via js.

app/views/comments/_comment.html.erb
```diff
@@ -25,6 +25,7 @@
       "></div>
 <div id="c_<%= comment.short_id %>"
   data-shortid="<%= comment.short_id if comment.persisted? %>"
+  data-id="<%= comment.id %>"
```
```suggestion
```

Don't want to leak PKs, they become a weird status symbol.

app/views/comments/_threads.html.erb
```diff
@@ -69,6 +69,7 @@
       "></div>
 <div id="c_<%= comment.short_id %>"
   data-shortid="<%= comment.short_id if comment.persisted? %>"
+  data-id="<%= comment.id %>"
```
pushcx:
```suggestion
```

app/javascript/application.js
```diff
@@ -818,8 +818,8 @@ onPageLoad(() => {
     const formData = new FormData();
     formData.append('reason', reason);
     const comment = parentSelector(event.target, '.comment');
-    const commentId = comment.getAttribute('data-shortid');
-    fetchWithCSRF('/comments/' + commentId + '/delete', { method: 'post', body: formData })
+    const commentId = comment.getAttribute('data-id');
```
```suggestion
    const commentId = comment.getAttribute('data-shortid');
```

pushcx (changes requested):
Thanks for taking this on! Not an exciting bit of functionality, but useful for making the site more reliable.

I have a couple small changes to simplify the code, and I want to keep using shortids rather than leak PKs.

app/views/comments/_threads.html.erb
mtaimourz: deleted comment.id

app/javascript/application.js
mtaimourz: used shortid

app/views/comments/_comment.html.erb
mtaimourz: sure deleted id and used shortid

app/controllers/mod/comments_controller.rb
mtaimourz: hmm makes sense :))

app/views/comments/_threads.html.erb
```diff
@@ -132,11 +132,15 @@
...
+          <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>          
+          |
+          <a class="comment_moderator" href="#c_<%= comment.short_id %>">mod delete</a>
+          <% elsif @user && @user.is_moderator? %>
             <a class="comment_moderator" href="#c_<%= comment.short_id %>">delete</a>
-          <% else %>
-            <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
-          <% end %>
+         <% else %> 
+            <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
```
pushcx:
```suggestion
         <% else %>
            <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
```

app/views/comments/_threads.html.erb
```diff
@@ -132,11 +132,15 @@
           <a class="comment_undeletor" href="#c_<%= comment.short_id %>">undelete</a>
         <% elsif !comment.is_gone? && comment.is_deletable_by_user?(@user) %>
           |
-          <% if @user && @user.is_moderator? && @user.id != comment.user_id %>
+        <% if @user && @user.is_moderator? && @user.id == comment.user_id %>
```
pushcx: There should be two independent links here.

1. User wrote the comment and it's within the delete window, show the delete link.

2. User is a moderator, show the 'mod delete' link.

mtaimourz: refactored code to make both conditions independent

app/models/comment.rb
```diff
@@ -283,10 +283,19 @@ def current_upvoted?
 
   def delete_for_user(user, reason = nil)
     Comment.record_timestamps = false
+    self.is_deleted = true
 
+    save!(validate: false)
+    Comment.record_timestamps = true
```
```suggestion
```

app/models/comment.rb
```diff
@@ -283,10 +283,19 @@ def current_upvoted?
 
   def delete_for_user(user, reason = nil)
     Comment.record_timestamps = false
```
```suggestion
```

These `record_timestamps` lines were outdated. You couldn't know it, but Comment used to use `updated_at` to record the last time a comment was edited. We shifted to use a new `last_edited_at` column because Rails uses `updated_at` for clearing the cache anytime a record was changed for any reason... which we certainly want to do when a comment is deleted.

app/views/comments/_comment.html.erb
```diff
@@ -86,13 +86,15 @@
         <% if comment.is_gone? && comment.is_undeletable_by_user?(@user) %>
           |
           <a class="comment_undeletor" href="#c_<%= comment.short_id %>">undelete</a>
-        <% elsif !comment.is_gone? && comment.is_deletable_by_user?(@user) %>
-          |
-          <% if @user && @user.is_moderator? && @user.id != comment.user_id %>
-            <a class="comment_moderator" href="#c_<%= comment.short_id %>">delete</a>
-          <% else %>
+        <% elsif !comment.is_gone? %>
+          <% if @user && @user.id == comment.user_id && comment.is_deletable_by_user?(@user) %>
```
```suggestion
        <% elsif !comment.is_gone? %>
          <% if comment.is_deletable_by_user?(@user) %>
```

Removed some redundant predicates that are checked in `is_deleteble_by_user`.

app/views/comments/_comment.html.erb
```diff
@@ -86,13 +86,15 @@
...
-          |
-          <% if @user && @user.is_moderator? && @user.id != comment.user_id %>
-            <a class="comment_moderator" href="#c_<%= comment.short_id %>">delete</a>
-          <% else %>
+        <% elsif !comment.is_gone? %>
+          <% if comment.is_deletable_by_user?(@user) %>
+            |
             <a class="comment_deletor" href="#c_<%= comment.short_id %>">delete</a>
           <% end %>
+          <% if @user && @user.is_moderator? %>
```
```suggestion
          <% if @user&.is_moderator? %>
```

pushcx (commented):
Thanks for tidying the id stuff. This looks basically correct, I made a couple small tweaks you couldn't have guessed.

I realized this very late, but I only just now realized that in the issue I should also have asked for mod UNdeleting to be separated out as well. You know this code now and it's a very similar change, would you like to add it into this PR? Otherwise I'm happy to merge and I'll update the issue to include the missing feature request for someone to pick up later.

mtaimourz:
Thanks for the review and really appreciate you taking the time doing these tweaks on this one. 
Regarding the undeleting functionality: yes, I’d definitely be happy to pick that up later in the week after you’ve updated the issue. A clean, fresh start for that part sounds ideal.

app/views/comments/_comment.html.erb
```diff
@@ -86,13 +86,15 @@
         <% if comment.is_gone? && comment.is_undeletable_by_user?(@user) %>
           |
           <a class="comment_undeletor" href="#c_<%= comment.short_id %>">undelete</a>
-        <% elsif !comment.is_gone? && comment.is_deletable_by_user?(@user) %>
-          |
-          <% if @user && @user.is_moderator? && @user.id != comment.user_id %>
-            <a class="comment_moderator" href="#c_<%= comment.short_id %>">delete</a>
-          <% else %>
+        <% elsif !comment.is_gone? %>
+          <% if comment.is_deletable_by_user?(@user) %>
```
pushcx, line 90: The build is failing because this is split out into a nested conditional. The goal is that the user can 'delete' young comments (14 days) but 'disown' after that, to avoid turning old discussions into swiss cheese with lots of missing comments. The two buttons are mutually exclusive, the user should only see the one they're allowed to do.

I think the easiest fix would probably be to recombine the predicates to restore the `is_deletable ... elsif disown` and move the mod delete button down below.

pushcx:
The brakeman config is a false alarm about a new version available. It's handled on `main` and #1786 improved our CI config so this shouldn't be a hassle for contributors in the future.

Thanks for working through this feature, I'll update the issue now.

PR #1743, opened 2025-10-09
Fix pagination links on /inbox/all

nortti0:
The links still pointed to the old /notifications endpoint, which no longer exists.


pushcx:
Thanks for catching and fixing these links.

PR #1745, opened 2025-10-10
Update uri to 1.0.4 for security patches

caius:
Fixes a CVE in the uri library

    Name: uri
    Version: 1.0.3
    CVE: CVE-2025-61594
    Criticality: Unknown
    URL: https://www.ruby-lang.org/en/news/2025/10/07/uri-cve-2025-61594
    Title: CVE-2025-61594 - URI Credential Leakage Bypass over CVE-2025-27221
    Solution: update to '~> 0.12.5', '~> 0.13.3', '>= 1.0.4'

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -425,7 +425,7 @@ GEM
     unicode-display_width (3.2.0)
       unicode-emoji (~> 4.1)
     unicode-emoji (4.1.0)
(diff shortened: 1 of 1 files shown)


pushcx:
One of those where we're very unlikely to be affected, but it's worth being ahead of. Thanks for spotting it.

PR #1746, opened 2025-10-10
Upgrade to Ruby 3.4.7

caius:
## What?

- [x] Pin ruby version in bundler so we can't accidentally run on the wrong version
- [x] Upgrade to Ruby 3.4.7

## Why?

Ruby upgrade contains patches that we want to use, and pinning the version means we can't accidentally boot the app using the wrong ruby. We lock our dependencies to ensure it runs the same everywhere, let's do the same with the runtime. It reads from `.ruby-version` like all other tooling, so the only "maintenance burden" is remembering to run `bundle install` after updating the ruby version.


Gemfile
```diff
@@ -1,5 +1,7 @@
 source "https://gem.coop"
 
+ruby file: ".ruby-version"
```
pushcx, line 3: Ah nice, I didn't know this was possible.

pushcx:
Thanks for noticing the new version available and opening this, we do try to track those pretty closely.

One of the few very individualist quirks of the codebase is that I don't bump the Ruby version until I can install it on my desktop from [the Arch package of ruby-build](https://archlinux.org/packages/extra/any/ruby-build/). I just flagged the Arch package as out-of-date, linking to its new release. (I'm open to other options of a similar reliability and convenience!) So this PR is going to hang out a couple days until that's updated to include 3.4.7.

caius:
> One of the few very individualist quirks of the codebase is that I don't bump the Ruby version until I can install it on my desktop from [the Arch package of ruby-build](https://archlinux.org/packages/extra/any/ruby-build/). I just flagged the Arch package as out-of-date, linking to its new release. (I'm open to other options of a similar reliability and convenience!) So this PR is going to hang out a couple days until that's updated to include 3.4.7.

Ha, I'd missed this was specifically the _Arch package_ previously. I saw https://ruby.social/@postmodern/115335993973299710 and figured it was there, so PR was ready to go 😁

pushcx:
It landed, merging.

PR #1747, opened 2025-10-10
Upgrade Bundler, add more platforms to lockfile & checksum dependencies for security

caius:
I failed to think of a decent title for this PR, soz. 😛

## What?

- [x] Upgrade bundler to 2.7.2
- [x] Add ruby platform to `Gemfile.lock` to allow native gem precompilation on non-pinned platforms
- [x] Add macOS platform to `Gemfile.lock` and pin native gems for it
- [x] Add checksums to `Gemfile.lock`

## Why?

Upgrade to the latest version of bundler, which is also a compatibility release ahead of Bundler 4 coming out at some future point. If we want to check our compatibility, we can add `export BUNDLE_SIMULATE_VERSION="4"` to our shells and bundler will behave as it will under 4. I've been running it locally with this set and not noticed any issues yet so we should be good.

Running any bundler commands locally keeps updating the `Gemfile.lock` with `arm64-darwin-24` platform and adding the native gems in, which is annoying to elide from any commits I'm making. There's no real reason to avoid having the platform in there, although we can add it as `arm64-darwin` so it doesn't need updating with macOS releases every year anymore.

There is a potential attack vector through our upstream gems provider (currently gem.coop, but it applies to any gem source we could point at) where a `.gem` is served to us initially and we vet it works, doesn't create issues, etc but then the service (again, hypothetically) serves up a _different_ `.gem` file with different/extra code contained within it at the same version number. This shouldn't ever happen, but if a bad actor got into any of the upstreams we pull from it could be a possibility. At that point we're running different code to what we originally depended on. Bundler added a feature to help mitigate/prevent this happening, where it stores a checksum of the initial `.gem` file (including the platform specific builds) and checks against that anytime it fetches the file again in future. This will warn loudly if the upstream file changes anytime after we first pinned it.


pushcx:
Thanks for this. I'd heard of checksums this last week or two in Ruby news but they're way down my to-do list so I appreciate you volunteering this. It looks like it closes off some unlikely but nasty risks, and there's public attention on improving it to handle more risks. So for us this will be an improvement with few or no costs that sets us up for future improvements.

GitHub doesn't handle stacked PRs ([jujutsuhub when](https://lobste.rs/c/id7yok)) so I'll manually note that this is blocked on #1746 and I'll merge it immediately after, assuming I don't generate any merge conflicts by updating other gems in the meantime. (None planned, and I'll try to put anything minor off.)

caius:
> Thanks for this. I'd heard of checksums this last week or two in Ruby news but they're way down my to-do list so I appreciate you volunteering this. It looks like it closes off some unlikely but nasty risks, and there's public attention on improving it to handle more risks. So for us this will be an improvement with few or no costs that sets us up for future improvements.

I'd added it locally weeks ago but forgot to push 😅 Picking up the repo for the ruby update reminded me about it.

> GitHub doesn't handle stacked PRs ([jujutsuhub when](https://lobste.rs/c/id7yok)) so I'll manually note that this is blocked on #1746 and I'll merge it immediately after, assuming I don't generate any merge conflicts by updating other gems in the meantime. (None planned, and I'll try to put anything minor off.)

Annoyingly you can kinda do it but not when you're opening across-forks. I normally do it by having the first PR open on top of master/main, then the second PR open on top of the first. GitHub sorts out changing the base on the second PR as the first PR is merged. Turns out this doesn't work cross-repo for forks though, the second PR would just be open within my forked repo, so you wouldn't see it. Edge cases, wooo. (Also yes, something that supports `jj` ftw.)

pushcx:
I wasn't thinking about this when I merged #1755 so I created a merge conflict about the checksums. Do you mind refreshing that?

caius:
No worries at all, partly why I marked it as draft initially 🙃

Rebased on main and updated all gems. Noticed there's not an arm64-darwin precompile for commonmarker yet, so added the ruby platform as well to allow gem compilations in the case there's not any precompiled version for that platform yet. Arguably we could remove `arm64-darwin` as a platform and let all Mac users compile the four gems that have native extensions, but also I'm a Mac user and like quicker gem installs.

pushcx:
Ran into a [nice post on checksums](https://mensfeld.pl/2025/01/the-silent-guardian-why-bundler-checksums-are-a-game-changer-for-your-applications/) [today](https://lobste.rs/s/kgrikp/bundler_2_6_enhanced_security_with_built).

Issue #1694, opened 2025-08-04
Refactor repeated apply_current_vote


pushcx:
There's a repeated bit of code that hydrates a list of comments with the votes of the current user. It appears as `apply_current_vote` in StoriesController, RepliesController (soon to disappear, see #1649), and NotificationsController, and in CommensController#comments.

There's also some parallel loading using `for_presentation`. Maybe we don't need to touch that, it's reasonably well-encapsulated in the scope, but I do wonder if there's a missing concept here around "a collection of comments to render". That collection is currently an ActiveRecord Results object/an array, but maybe it should be wrapped up together.

pushcx:
Quick note: the list of places that set `current_vote` is probably incomplete, so check `app/controllers` for more.

One thing that's lost in the current design: ideally we wouldn't be looping over comments and votes and things in the controllers. A class could maintain parallel data structures and hydrate the user data on as-accessed. I've been trying to avoid extra loops over comments, it's very much our hot path!

pushcx:
(Also I made this worse today by adding `.current_reply` in ed9171ed798992a85f8499c951af29ef9edba646 and I did have to touch a lot of duplicated code.)

leregdos:
Hi there,

I thought this would be a good issue for me to pick up as my first contribution to the project, although I'm a bit confused because `apply_current_vote` is not present in StoriesController, but only in InboxController. Does it mean that in the controllers you mentioned, there is a repeated logic concerning votes? TIA for your patience.

pushcx:
Thanks for taking a look at this, I'm sorry I didn't notice your comment when I was traveling.

In the initial review I referred to our then-in-progress work to replace RepliesController with InboxController (and associated database modeling). That was done by the time you commented, so that's probably the discrepancy you see between the initial description of the problem and the state of the code.

I took a look at your PR. You remove the duplication, which is a clear improvement. I haven't audited the code to see if any more instances remain, can you say if you have?

I said above:

> One thing that's lost in the current design: ideally we wouldn't be looping over comments and votes and things in the controllers. A class could maintain parallel data structures and hydrate the user data on as-accessed. I've been trying to avoid extra loops over comments, it's very much our hot path!

Is this a refactoring you'd like to take on? I'm open to merging your improvement and doing this later, or you could take it on in #1748.

leregdos:
Hi, thanks for your feedback. 

[...]

PR #1748, opened 2025-10-10
Refactor repeated vote hydration

leregdos:
This PR attempts to solve #1694 by creating a new class following the example of `StoriesPaginator` class that achieved similar functionality. The new class is then integrated into controller methods that are looping over every comment, hydrating them with the current user's vote.

### DIFF
--- a/app/models/comment_vote_hydrator.rb
+++ b/app/models/comment_vote_hydrator.rb
@@ -22,6 +22,8 @@ class CommentVoteHydrator
     @comments.each { |c| yield self[c] }
   end
 
+  private
+
   def [](comment)
     comment.current_vote ||= @votes[comment.id]
     comment.vote_summary ||= @vote_summaries[comment.id]


app/models/comment_vote_hydrator.rb
```diff
@@ -0,0 +1,27 @@
+class CommentVoteHydrator
+  def initialize(comments, user)
+    @comments = comments
+    @comment_ids = comments.map(&:id)
+    @user = user
+    @hydrated = false
```
pushcx: Ah, `@hydrated` is clever, but Rails is ahead of you on this one: when you create a query, it doesn't actually run against the database until you try to access it. You can see how this works in the console by running `Rails.logger = Logger.new STDOUT` and then something like `c = Comment.last; nil` (the trailing `; nil` means pry won't immediately try to access the query to inspect it). Then if you run `c.user_id` you'll see the query happen as you try to access.

So what this means here is that you can delete `@hydrated` and bring those queries from `hydrate` up here to the initializer.

app/models/comment_vote_hydrator.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def [](comment)
+    hydrate unless @hydrated
+    comment.instance_variable_set(:@current_vote, @votes[comment.id])
```
pushcx: Your strategy of hydrating comments on read to avoid looping over the array of comments is what I was looking for, thanks.

Why did you edit these to `instance_variable_set` instead of `comment.current_vote =`? That should work just as well from the hydrator here.

leregdos: right, yeah, I'll use `comment.current_vote=` since that's more consistent with the codebase and looks cleaner. I forgot why I opted to use `instance_variable_set()` though

app/models/comment_vote_hydrator.rb
```diff
@@ -0,0 +1,22 @@
...
+    else
+      comment_ids = comments.map(&:id)
+      @votes = Vote.comment_votes_by_user_for_comment_ids_hash(@user.id, comment_ids)
+      @vote_summaries = Vote.comment_vote_summaries(comment_ids)
+      @current_user_reply_parents = @user.ids_replied_to(comment_ids)
+    end
+  end
+
+  def [](comment)
+    comment.current_vote = @votes&.dig(comment.id)
```
pushcx: These assignments should all use `||=` to avoid repeated work. And none of them should use `&.` on the class's instance variables, you've written the initializer so those are always instantiated to the right type.

app/controllers/comments_controller.rb
```diff
@@ -307,15 +298,7 @@ def index
...
-    comment_ids = @comments.map(&:id)
-    @votes = Vote.comment_votes_by_user_for_comment_ids_hash(@user&.id, comment_ids)
-    summaries = Vote.comment_vote_summaries(comment_ids)
-    current_user_reply_parents = @user&.ids_replied_to(comment_ids) || Hash.new { false }
-    @comments.each do |c|
-      c.current_vote = @votes[c.id]
-      c.vote_summary = summaries[c.id]
-      c.current_reply = current_user_reply_parents.has_key? c.id
-    end
+    @vote_hydrator = CommentVoteHydrator.new(@comments, @user)
```
pushcx: Instead of having the vote_hydrator be a separate instance variables that the templates know about (and have to prepare), let's make `CommentVoteHydrator` a decorator. This line of code would change to `@comments = CommentVoteHydrator.new(@comments, @user)`.

Have the class `include Enumerable` and define an `each`:

```ruby
  def each
    @comments.each { |c| yield self[c] }
  end
```

Then the templates won't have any edits related to the hydrator - as far as they're concerned, comments arrive hydrated and are an array to be looped over.

pushcx (changes requested):
I outlined another way to make the hydrator a little more complex internally so that all the code using it can be simpler.

app/controllers/comments_controller.rb
leregdos: That's really great, thank you!

app/models/comment_vote_hydrator.rb
```diff
@@ -0,0 +1,33 @@
+class CommentVoteHydrator
+  include Enumerable
+
+  delegate :size, :length, :empty?, :any?, to: :@comments
```
pushcx, line 4: Excellent, thanks for being thorough here.

pushcx:
Thanks for working through my many suggestions here. This ended up being a really clean and performant way of hydrating the data, and now we have a reusable tool so that future lists of stories can start out correct instead of their usual practice of launching and earning a bug report about not showing upvotes correctly.

PR #1749, opened 2025-10-11
Add rel="ugc" to all links in Markdown

### LINKED ISSUE #1728: Markdowner missing rel="ugc" on some links
In #1627 (GitHub doesn't recognize it's merged) we updated CommonMarker. It has a behavior change in links, a handful of them (I think all mildly invalid, note the `://` is `:/`) lack `rel="ugc"`.

Here's [one comment](https://lobste.rs/s/finatu/fighting_coronavirus_with_freebsd#c_ago3zf), source:

```
join [World Community Grid](https:/worldcommunitygrid.com), as they
```

result:

```
Markdowner.to_html("join [World Community Grid](https:/worldcommunitygrid.com), as they")
=> "<p>join <a href=\"https:/worldcommunitygrid.com\">World Community Grid</a>, as they</p>\n"
```

It also happens on relative links:

```
Markdowner.to_html("see [the about](/about) page")
=> "<p>see <a href=\"/about\">the about</a> page</p>\n"
```

This is a low priority bug.

fpsvogel:
## Summary

Resolves #1728 

Adjusts `Markdowner::to_html` to add `rel="ugc"` to all links.

Previously, it was added only to links with a non-nil `URI::HTTP#host`, so links were not getting `rel="ugc"` when the URL lacked a host, or the host could not be parsed. This was the case in the two examples in the Issue (relative links, and URLs missing one protocol slash).

## Note: why should _all_ links get `rel="ugc"`?

The UGC attribute [should be present in all links that appear in user-generated content](https://ahrefs.com/seo/glossary/ugc-link-attribute). That's why I changed `Markdowner` to add `rel="ugc"` to _all_ links, rather than trying to account for just the two cases that the Issue mentioned.

### DIFF
--- a/spec/extras/markdowner_spec.rb
+++ b/spec/extras/markdowner_spec.rb
@@ -49,17 +49,30 @@ describe Markdowner do
         "test</a></p>\n")
   end
 
-  it "correctly adds ugc" do
-    expect(Markdowner.to_html("[ex](http://example.com)"))
-      .to eq("<p><a href=\"http://example.com\" rel=\"ugc\">" \
-            "ex</a></p>\n")
+  it "adds ugc" do
+    expect(Markdowner.to_html("[full URL](http://example.com)"))
+      .to eq("<p><a href=\"http://example.com\" rel=\"ugc\">full URL</a></p>\n")
 
-    expect(Markdowner.to_html("[ex](//example.com)"))
-      .to eq("<p><a href=\"//example.com\" rel=\"ugc\">" \
-            "ex</a></p>\n")
+    expect(Markdowner.to_html("[protocol-relative URL](//example.com)"))
+      .to eq("<p><a href=\"//example.com\" rel=\"ugc\">protocol-relative URL</a></p>\n")
 
-    expect(Markdowner.to_html("[ex](/~abc)"))
-      .to eq("<p><a href=\"/~abc\">ex</a></p>\n")
+    # invalid URLs that are still parsed as links (not an exhaustive list)
+    expect(Markdowner.to_html("[missing protocol](example.com)"))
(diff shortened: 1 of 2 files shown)


spec/extras/markdowner_spec.rb
```diff
@@ -49,17 +49,30 @@
...
+    expect(Markdowner.to_html("[relative link](/example)"))
+      .to eq("<p><a href=\"/example\" rel=\"ugc\">relative link</a></p>\n")
+
+    expect(Markdowner.to_html("[relative link to user profile](/~abc)"))
+      .to eq("<p><a href=\"/~abc\" rel=\"ugc\">relative link to user profile</a></p>\n")
+
+    # autolink
+    expect(Markdowner.to_html("www.example.com"))
+      .to eq("<p><a href=\"http://www.example.com\" rel=\"ugc\">www.example.com</a></p>\n")
   end
```
fpsvogel, line 76: I kept the three previous expectations, just removing a source code line break and changing each one's link text for clarity.

I added four new expectations:

Expectations for `rel="ugc"` being added to the two the cases mentioned in the Issue:

- L63-64, wrong number of slashes after the protocol: `[wrong number of slashes after protocol](http:/example.com)`
- L67-68, relative link: `[relative link](/example)`

Similar expectations for two additional cases not mentioned in the Issue:

- L74-75, autolink: `www.example.com`
- L60-61, missing protocol: `[missing protocol](example.com)`

pushcx:
Aha, I assumed the `rel="ugc"` was coming from CommonMarker, I didn't remember that it's our code [even though I touched it](https://github.com/lobsters/lobsters/commit/feea257c648f3636722f33520912cc4b40749f6d).

Thanks for fixing this. Great call to expand the test suite with the examples of invalid data.

PR #1754, opened 2025-10-13
Fix HTML markup errors found by Herb

marcoroth:
Follow up on #1633 

This pull request fixes some of the parse errors Herb discovered using the [`herb analyze`](https://github.com/marcoroth/herb?tab=readme-ov-file#command-line-usage) command.

I also noticed that the fixed content in these files are identical and might benefit from being extracted to a partial:
- `app/views/home/index.html.erb`
- `app/views/search/index.html.erb`
- `app/views/stories/_listdetail.html.erb`

I intentionally didn't address any of the Herb Linter offenses yet, but I'm happy to go over those too if you want! 🙌🏼

### DIFF
--- a/app/views/about/about.html.erb
+++ b/app/views/about/about.html.erb
@@ -260,7 +260,7 @@
   client.  These e-mails are then converted and submitted to the website as
   comments, just as if the comment was posted through a web browser.
   <%- raise "Don't just delete that error message. You didn't set up mailing list mode, right? This really isn't your about page, write one from scratch." if Rails.env.production? and Rails.application.name != 'Lobsters' -%>
-  </p>
+  </p></li>
 
   <li><p>
   <strong>Private messaging</strong> enables users to communicate privately
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -147,10 +147,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
             <span> | </span>
             <a role="heading" aria-level="2" href="<%= Routes.title_path story %>">
             <% if story.comments_count == 0 %>
-              no comments</a>
+              no comments
             <% else %>
-              <%= story.comments_count %> <%= 'comment'.pluralize(story.comments_count) %></a>
+              <%= story.comments_count %> <%= 'comment'.pluralize(story.comments_count) %>
             <% end %>
+            </a>
           </span>
         <% end %>
       <% end %>
--- a/app/views/search/index.html.erb
+++ b/app/views/search/index.html.erb
@@ -276,10 +276,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
             <span> | </span>
             <a role="heading" aria-level="2" href="<%= Routes.title_path story %>">
             <% if story.comments_count == 0 %>
-              no comments</a>
+              no comments
             <% else %>
-              <%= story.comments_count %> <%= 'comment'.pluralize(story.comments_count) %></a>
+              <%= story.comments_count %> <%= 'comment'.pluralize(story.comments_count) %>
             <% end %>
+            </a>
           </span>
         <% end %>
       <% end %>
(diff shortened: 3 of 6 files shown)


pushcx:
Thanks for catching these bugs!

They are indeed a partial, you're seeing [heinous_inline_partial](https://github.com/lobsters/lobsters/blob/4e59086fcb6e1903ef8d620f620eada7fa55d8d6/config/initializers/heinous_inline_partial.rb). My awful hack to avoid the significant overhead of looking up a partial.

As for Herb Linter: I hope to never again take a dep on node/npm/yarn/whatever in this project (...or any project, if I can help it). On top of the pain of having a second language toolchain, js needs attention way too frequently for this hobby project, it felt like I had to touch it a couple times a year. My experience with rubocop ([big setup](https://github.com/lobsters/lobsters/pull/460) + endless decisions in updates) kind of [burned me out](https://github.com/lobsters/lobsters/commit/c3f8625788245129903f31622bb452957223fc44) of linters with options. While I'm impressed by Herb Linter and it fixes a very real pain, the dev time to maintain an install would stress our limited budget.

That's why I'm seriously considering [papercraft](https://github.com/digital-fabric/papercraft) now that its [performance is in the neighborhood of erb](https://github.com/digital-fabric/p2/pull/1). It's a big step out of mainstream Rails and not especially popular, but fixes Erb's [ignorance of structure](https://github.com/ruby/ruby-bench/pull/358#issuecomment-2856472389) (like this and [your previous PR](https://github.com/lobsters/lobsters/pull/1633) fix), security footguns, and painful instance variable coupling so we can test and [demo](https://lobste.rs/cabinet) views, and we get linting with zero setup and maintenance cost.

I think I nerd-sniped myself into collecting up all my opinions about Erb. I'd appreciate any criticism you or others have, especially if I've missed considerations. I guess I'll post this on the socials and hope for kibitzing. :smile:

marcoroth:
Thanks for merging, @pushcx!

> I hope to never again take a dep on node/npm/yarn/whatever

The good thing is that you don't need any NPM or Node dependency in your project, you can just have `npx @herb-tools/linter` as part of your `linters` check and you are good, no need to install anything. Unless that's what you mean with dependencies. 

https://github.com/lobsters/lobsters/blob/c97c5907a101ff41ce1979e0f5699541b27b4215/.github/workflows/check.yml#L66-L68

---

> That's why I'm seriously considering [papercraft](https://github.com/digital-fabric/papercraft) now that its https://github.com/digital-fabric/p2/pull/1. It's a big step out of mainstream Rails and not especially popular, but fixes Erb's https://github.com/ruby/ruby-bench/pull/358#issuecomment-2856472389 (like this and https://github.com/lobsters/lobsters/pull/1633 fix), security footguns, and painful instance variable coupling so we can test and [demo](https://lobste.rs/cabinet) views, and we get linting with zero setup and maintenance cost.

This is exactly why I started working on `Herb::Engine` as a replacement for `Erubi::Engine` (which is what Rails uses by default today for ERB files). It pretty much addresses all these concerns. 

It's a new ERB engine that's fully HTML-aware and doesn't allow you to produce invalid HTML. On top of the improved DX you get in development, it also has the potential to outperform existing Action View today, since we can optimize the compiled templates more strategically. We have a lot more introspection because Herb analyzes the HTML structure and the Ruby inside of the ERB tags. A lot of this also has potential to make it back into upstream Rails, given the transparent nature of these performance and DX improvements. 

We might also be able to start inlining partial render calls (or at least replace them with the Action View compiled method), optimize any Action View helpers back to static strings, and potentially more. Which is very similar in nature to what you were doing with `heinous_inline_partial`, but all while staying backwards compatible and while keeping it Railsy.

If you want to read more about it, I'd suggest to read this [blog post](https://marcoroth.dev/posts/rails-world-2025-recap).

pushcx:
I'm not taking any dependency on javascript.

PR #1755, opened 2025-10-14
Stop autolinking "www." and bare schemes

### LINKED ISSUE #1727: Markdowner: don't hyperlink bare www., http://, or https://
#1627 (which GitHub doesn't recognize as merged) updated CommonMarker. It has a behavior difference we should fix. When a user mentions `www.`, `http://`, or `https://` it hyperlinks them. Test cases:

```
Markdowner.to_html("my www. subdomain")
=> "<p>my <a href=\"http://www\" rel=\"ugc\">www</a>. subdomain</p>\n"
```

similar extra links on these:

```
Markdowner.to_html("the http:// protocol")
Markdowner.to_html("the https:// protocol")
```

Maybe there's a setting in CommonMarker or the underlying library we can disable these with, otherwise we'll need to detect and strip them.

chamlis:
Updates Commonmarker to bring in stricter autolink validation from Comrak, and adds a spec to verify the fix.

Closes #1727

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -105,7 +105,7 @@ GEM
     childprocess (5.1.0)
       logger (~> 1.5)
     chunky_png (1.4.0)
-    commonmarker (2.3.0-x86_64-linux)
+    commonmarker (2.5.0-x86_64-linux)
     concurrent-ruby (1.3.5)
     connection_pool (2.5.4)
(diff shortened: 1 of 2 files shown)


pushcx:
Looks great, thank you. I'll leave a comment over in issue to keep the discussion from spreading out.

PR #1756, opened 2025-10-15
Add Speculation Rules prefetching for visitors

### LINKED ISSUE #1687: Add speculationrules for visitors
I saw an [article](https://news.ycombinator.com/item?id=44747241) about the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API).

Even if it's Chrome-only, we might as well support this for visitors because serving them pages from the full-page cache (see `caches_page` in `app/controllers/*`) is effectively free.

We don't want to enable this for users because when they `GET` a story page it updates their `ReadRibbon` to make comments as read, so a speculative fetch that is not completed would lose state.

fpsvogel:
## Summary

Resolves #1687

### Logged out

Prefetches are visible in the Chrome dev tools Application tab → Speculative loads → Rules.

https://github.com/user-attachments/assets/34c0a4f9-5c77-4eef-b661-3bf0de435baa

In the video above, prefetch failures start to appear after the second successful prefetch because Chrome allows only two prefetches at a time, and the "failures" come from the oldest prefetches being discarded as new ones are successfully loaded. You can see this in Speculative loads → Speculations:

image

### Logged in

No speculation rules are defined, because the script is absent from the page.

speculationrules-logged-in |

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -46,7 +46,7 @@
       href="<%= @comments_rss_link[:href] %>">
   <% end %>
   <meta name="robots" content="noai, noimageai">
-
+<% # only for logged-out users so we don't speculatively advance ReadRibbons %>
   <% if !@user %>
     <script type="speculationrules" nonce="<%= content_security_policy_nonce %>">
       {


app/views/layouts/application.html.erb
```diff
@@ -46,6 +46,21 @@
...
+
+  <% if !@user %>
+    <script type="speculationrules" nonce="<%= content_security_policy_nonce %>">
+      {
+        "prefetch": [
+          {
+            "where": {
+              "href_matches": "/*"
+            },
+            "eagerness": "moderate"
```
fpsvogel, line 58: Eagerness options are listed at https://developer.chrome.com/blog/speculation-rules-improvements

`"moderate"`:

> This performs speculations if you hover over a link for 200 milliseconds (or on the [pointerdown](https://developer.mozilla.org/docs/Web/API/Element/pointerdown_event) event if that is sooner, and on mobile where there is no hover event).

Alternatively, we could omit this property, in which case eagerness will default to `"conservative"`, which is defined as:

> This speculates on pointer or touch down.

But given that many of the prefetches will be served from the cache (as noted in [the issue](https://github.com/lobsters/lobsters/issues/1687)), `"moderate"` seems reasonable.

app/views/layouts/application.html.erb
```diff
@@ -46,6 +46,21 @@
       href="<%= @comments_rss_link[:href] %>">
   <% end %>
   <meta name="robots" content="noai, noimageai">
+
```
pushcx:
```suggestion
<% # only for logged-out users so we don't speculatively advance ReadRibbons %>
```

PR #1758, opened 2025-10-16
Add extrapolated current month to stats

### LINKED ISSUE #1580: Stats: add linear extrapolation for current month
The [charts on /stats](https://lobste.rs/stats) (known bug #1562) look odd because the final number is partial data for the current month. I colored it gray to hint that it was different, but it would be nice to also have a second dot for the current month that is a simple linear extrapolation of where the month will wind up. To say it another way, (current month's stats / day of the month * days in the month).

fpsvogel:
## Summary

Resolves #1580

Adds an extra data point to Stats graphs, representing the extrapolated final value of the current month.

## Questions

- Is it clear what the green point represents, based solely on visual context?
  - The grey point was arguably easy to understand, since a user would see a smaller value for this month and probably make the connection that the month isn't over and grey = incomplete.
  - The green point might be harder to guess because (a) the graph is so compressed by this point (showing 13 years) that it's hard to notice that the green point is on the same place as the grey point on the x axis, and (b) even if a user does notice that alignment, it seems like slightly more of a stretch to understand that green (or any color) = extrapolated value.
- So, should there be legend at the top telling what the grey and green dots mean? Ideally we wouldn't need to do this, but I can't think of a simple visual way to clarify the colors.

## Screenshots

The grey point is, as before, this month's value so far (10). The new green point is this month's extrapolated value (20, since we're about halfway through the month).

|  | before | after |
|---|---|---|
| dark | stats-before-dark | stats-after-dark |
| light | stats-before-light | stats-after-light |

### DIFF
--- a/app/views/stats/index.html.erb
+++ b/app/views/stats/index.html.erb
@@ -36,8 +36,12 @@ Want more info?
 .xAxisLabels { fill: var(--color-fg); }
 .yAxisLabels { fill: var(--color-fg); }
 .key1, .dataPoint1 { fill: var(--color-fg-accent); }
-circle:nth-last-of-type(2) { fill: var(--color-fg-contrast-4-5); }
-circle:last-of-type { fill: var(--color-fg-affirmative); }
+circle:nth-last-of-type(2), circle:last-of-type {
+  fill: none;
+  stroke: var(--color-fg-accent);
+  stroke-width: 1;
+}
+circle:last-of-type { stroke-dasharray: 1; }
 .line1 { fill: var(--color-fg-accent); }
 .guideLines { stroke: var(--color-bg); stroke-dasharray: none; }
 .dataPointPopupMask { fill: var(--color-fg); stroke: var(--color-bg); }


app/models/stats.rb
```diff
@@ -53,15 +53,15 @@ def self.fill_votes_graph_cache
     }
   end
 
-  def self.cache_key(name)
-    "stats_graphs/#{name}"
-  end
```
fpsvogel, line 58: I moved this down and made it private as part of a commit where I separated public and private methods: https://github.com/lobsters/lobsters/pull/1758/commits/9d52e1d8fc59247e61c8fe5c6a2947a997240434

As a rule of thumb I find it helpful to show which methods are intended as a class's public API, but if it seems unnecessary, we can leave out that commit.

pushcx: I don't have a strong feeling about it; it's harmless to be public and harmless to hide.

app/models/stats.rb
```diff
@@ -89,15 +89,28 @@ def self.get_cached_graph(name)
     number_format: "%d",
     show_lines: false
   }
+  private_constant :TIME_SERIES_DEFAULTS
```
fpsvogel: This is my least favorite part of the above-mentioned commit (https://github.com/lobsters/lobsters/commit/9d52e1d8fc59247e61c8fe5c6a2947a997240434), but I included it for consistency since `TIME_SERIES_DEFAULTS` is only used within this class.

app/views/stats/index.html.erb
```diff
@@ -35,9 +35,10 @@ Want more info?
 .graphBackground { fill: var(--color-bg); }
 .xAxisLabels { fill: var(--color-fg); }
 .yAxisLabels { fill: var(--color-fg); }
-.key1, .dataPoint1 { fill: #ac130d; }
-circle:last-of-type { fill: #cccccc; }
-.line1 { fill: #ac130d; }
+.key1, .dataPoint1 { fill: var(--color-fg-accent); }
+circle:nth-last-of-type(2) { fill: var(--color-fg-contrast-4-5); }
+circle:last-of-type { fill: var(--color-fg-affirmative); }
+.line1 { fill: var(--color-fg-accent); }
```
fpsvogel, line 45: Using standard colors improves contrast, since the colors are defined differently according to color scheme. See the screenshots in the PR description.

app/models/stats.rb
pushcx: I get the impression that your day job is in a much, much bigger codebase than Lobsters. We haven't needed to use `private` because we're just not that big and we don't have a lot of contributors. We've defaulted to everything being `public` to promote reuse rather than default to everything being `private` to discourage misuse and mislayering.

I'd prefer not to make these constants and class methods private. I've had to debug the Stats in the prod console pretty often. If these methods get called from an inappropriate place it'll be glaringly odd design, we don't need to make things private to really get the message home.

README.md
fpsvogel, line 1: I touched up the setup instructions because I had in mind to add a step about using `bin/rails dev:cache` to toggle caching in dev mode, but decided against it because that's standard to Rails and not specific to Lobsters.

Still, I kept the touch-ups, which improve mostly punctuation and layout. The changes are easiest to view [with whitespace ignored](https://github.com/lobsters/lobsters/pull/1758/files?w=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

One of the spacing changes is to prevent the `<ul>` from being split in two by a code block that's not being nested in the list:

| before | after |
|---|---|
| Image | Image |

fpsvogel:
@pushcx Thanks for the feedback! That all makes sense to me.

I removed the commit where methods are made private, and I added a new commit that styles the current month as hollow circles, with the extrapolated circle having a dashed outline:

| dark | light |
|---|---|
| Screen Shot 2025-10-16 at 12 23 39 PM | Screen Shot 2025-10-16 at 12 28 48 PM |

If we want the current month to have more contrast / a "greyed out" appearance, here's what it would look like with `--color-fg-contrast-10` on [index.html.erb:41](https://github.com/lobsters/lobsters/pull/1758/files#diff-58ace09d8e2535d5b7a94dbdb4df1be0eafa84c833fb928dad808b2210b700dbR41):

| dark | light |
|---|---|
| image | image |

pushcx:
Great to have your fixes to the readme, and thanks for polishing this so it's really clear what it means. I'm going to deploy this now and pull up the prod console to call that public method to refresh the graphs. :smile:

pushcx:
This looks great, and I'm adding a line of explanation at the top like you suggested.

The current month data for the active users graph is implausible, it's ~2x higher than trend. I don't see what's wrong with the query at a glance, but it is the only one that combines multiple queries.

fpsvogel:
@pushcx I wonder if the incorrect extrapolation in the active users graph comes from that value growing non-linearly within a month.

The values in the other graphs increase linearly within a month: when the month is half over, we can expect about half of new users to have joined that will join in the month, half of the stories submitted, half of the comments posted, and half of the votes cast.

But with active users, it could be that when the month is half over, most of the users who will have submitted/posted/commented, have already done so at least once. Does this sound plausible?

If that's true, we have to either use some other formula for the extrapolation of the active users graph, or just omit the extrapolation for that one.

pushcx:
That's gotta be it, yeah. The active users are overwhelmingly the same day-to-day.

Issue #1725, opened 2025-09-24
Display a warning about self-promotion when the user checks the "I am the author" checkbox


linkdd:
When submitting a link, the user is able to check the "I am the author" checkbox.

When this checkbox is selected, displaying a warning about the abuse of self promotion could help lighten the workload on moderators and give more visibility to this rule. New users, or occasional users might not be aware of it.

HamptonMakes:
Is there some exact recommended warning text?

pushcx:
How about:

> If you're new here, please review the <a href="https://lobste.rs/about#self-promo">self-promo guideline</a>.

If this first attempt doesn't help, it'll be easy enough to revise with more stern wording before looking to try code.

Issue #1186, opened 2023-07-19
New users using the "authored by" radio button should get a warning


ngoldbaum:
If a new user's blog has been seen before there's nothing adding friction to doing self-promotion as a new user. See e.g. this domain:

https://lobste.rs/domains/allaboutcoding.ghinda.com

EDIT: (since submitting the issue the second story got deleted, https://lobste.rs/s/qyxaya/where_promote_article_about_ruby)

If a new user clicks the "authored by" button, they should probably get the same warning a new user submitting an unseen domain gets. Of course they could still resubmit without the "authored by" but at least well-intentioned users will get a warning.

pushcx:
To make sure I follow - you're thinking of this validation?

> [example.com] is an unseen domain from a new user. We restrict this to discourage self-promotion and give you time to learn about topicality. Skirting this with a URL shortener or tweet or something will probably earn a ban.

I'm asking because it's a Story validation, so the new user can't submit their link at all. If this warning you're thinking of is added it'd be a bit of a policy change, right now users are allowed to submit links to their own stuff if someone has submitted the domain before (...or it's github, youtube, or another popular site for self-publishing). We do [explicitly encourage authors](https://github.com/lobsters/lobsters/blob/a2831a81ca1e23fb374789b6dd950af76a8429e6/app/views/about/about.html.erb#L117) who've had their work submitted to ask to join the site. Right now people just get treated pretty skeptically if the line under the headline says "by" instead of "via" and the username is green.

I can see how that would be an appropriate rule change. It feels like we've seen a significant increase in people using the site exclusively for self-promo this year.

A softer approach would be to add text to the submission page (similar to the one that detects and suggests users remove site names), it could explain that they may want to wait until they've first built up some trust with the community by posting comments and submitting other people's interesting stuff. That would continue to permit authors who get invited to immediately post. Though I'm not sure if that's worth keeping; typically they find their way to the site while an article of theirs is already on the home page; spacing out more from their site might also be an improvement.

ngoldbaum:
Yeah, I guess an “are you *sure*?” check that links to a page describing community norms before letting them submit would also work.

All of this is predicated on someone clicking the “authored by” button, which is optional, so a policy change probably isn’t warranted some additional friction would at least tell well-intentioned but ignorant submitters that they are doing something that might raise eyebrows if they’re not also contributing to the site in other ways.

PR #1760, opened 2025-10-16
Add note to /submit about 25% self-promo rule

beyarkay:
[`roryokane`][1] suggested I open a PR, after I accidentally self-promo'd too much ([thread](https://lobste.rs/s/roygg3/why_your_boss_isn_t_worried_about_ai_can_t#c_svgdyx)). I'm sorry about that, I feel bad for being a dick and spamming my stuff without being part of the community. I didn't realise the 25% guideline was a thing, because it's on the /about page which I had skimmed but not read. I *did* read the /submit page, but that page doesn't say anything about the 25% guideline.

Hopefully this PR will help future users not make the same mistake.

This PR will partially resolve https://github.com/lobsters/lobsters/issues/1725 (although maybe not perfectly) and is somewhat related to https://github.com/lobsters/lobsters/issues/1186.

[1]: https://lobste.rs/~roryokane

### DIFF
--- a/app/views/stories/_form.html.erb
+++ b/app/views/stories/_form.html.erb
@@ -127,6 +127,14 @@ See the guidelines below for more." %>
         (It can't recognize and remove blog/author names for you, though.)
         </p></li>
 
+        <li><p>
+        Self-promotion: It's great to have authors participate in the
+        community, but not to exploit it as a write-only tool for product
+        announcements or driving traffic to their work. As a rule of thumb,
+        self-promo should be less than a quarter of one's stories and
+        comments.
+        </p></li>
+
       </ul>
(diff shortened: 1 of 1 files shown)


pushcx:
(You can ignore the brakeman failure; it's warning an update is available. We've improved the CI on `main` to not distract from PRs.)

This isn't what #1725 asked for. It's not in the right part of the page, it's not the desired message, and it doesn't appear when the user checks the box so that it draws attention. The folks who immediately start in on self-promo regularly violate the other guidelines here (especially not correctly formatting their titles) that make it clear they don't read this text at all on their route to submitting a link, so I don't think there's much value to be had here.

PR #1762, opened 2025-10-17
Add https source to image policies

### LINKED ISSUE #1714: Images are broken due to Content-Security-Policy header
I noticed recently images in story's description are broken, perhaps due to https://github.com/lobsters/lobsters/blob/f7091164909add7242e7c47a472dcc678eb1bdef/config/initializers/content_security_policy.rb#L17 the CSP header here.

Several weeks ago when i was investigating the mailing list feature i found this story https://lobste.rs/s/jg3eet/lobsters_by_mail , at that time i remembered i could see images on that page, but now they are broken.

Will it be a good idea to add `:https` to ` policy.img_src` or current behavior is expected?

luigieli:
This PR should close #1714.

The CSP error when loading images was happening because of the policies on images, which were `self` and `data.` That way, images could only be fetched from the same origin as the original document (e.g., lobste.rs/) or a [data URI](https://en.wikipedia.org/wiki/Data_URI_scheme).

This PR has the intent to add the `https` source as well. That way, images from `https` sources like [Imgur](https://imgur.com/) will be rendered without error.

Before:

image

After:

image

While testing, I had some issues with `page_caching`, where this bug doesn't occur on cached pages. After some investigation, I found out that this was because the Puma web server used in the development environment didn't handle CSP headers for cached pages, which led me to wonder that [Caddyfile](https://github.com/lobsters/lobsters/blob/a16def55062b9e3457b10e1fb243b2cffeb3d9ed/hatchbox/Caddyfile.pre#L65) should also include the `https` headers, but I'm not sure. Is it worth testing it locally, or will it not cover all the differences between development and production environments?

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -14,7 +14,7 @@ Rails.application.configure do
     policy.connect_src :self
 
     # Data URL used for Pushover logo in settings
-    policy.img_src :self, :data
+    policy.img_src :self, :data, :https
     policy.script_src :self
 
     # 18 inline styles to clean up before enabling this


luigieli:
Closing this PR since it doesn't cover the whole issue, and that way it won't close #1714. See the [discussion](https://github.com/lobsters/lobsters/issues/1714#issuecomment-3416449239) for more details.

PR #1763, opened 2025-10-18
Omit extrapolated value from active users graph

fpsvogel:
## Summary

Follow-up to #1758

Omits the extrapolated value from the Active Users graph, because our linear extrapolation logic does not apply to that graph's data.

See [this comment](https://github.com/lobsters/lobsters/pull/1758#issuecomment-3413522508) in that previous PR.

## Screenshots

| before | after |
|---|---|
| image | image |

### DIFF
--- a/app/models/stats.rb
+++ b/app/models/stats.rb
@@ -16,7 +16,8 @@ class Stats
   def self.fill_active_users_graph_cache
     cache_monthly_graph(:active_users, {
       graph_title: "Active users by month",
-      scale_y_divisions: 500
+      scale_y_divisions: 500,
+      extrapolate: false
     }) {
       stories = Story.pluck(:created_at, :user_id).map { |created_at, user_id| [created_at.strftime("%Y-%m"), user_id] }
       votes = Vote.pluck(:updated_at, :user_id).map { |updated_at, user_id| [updated_at.strftime("%Y-%m"), user_id] }
@@ -66,6 +67,7 @@ class Stats
     height: 300,
     graph_title: "Graph",
     show_graph_title: false,
+    extrapolate: true,
     no_css: false,
     key: false,
     scale_x_integers: true,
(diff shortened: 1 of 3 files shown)


lib/time_series.rb
```diff
@@ -20,4 +20,16 @@ def get_x_labels
...
     get_y_values.collect { |v| number_with_delimiter(v.to_i) }
   end
+
+  # Override to add a custom CSS class to the SVG element, for graphs
+  # that should omit an extrapolated value for the current month.
+  def burn_svg_only(add_no_extrapolation_css_class: false)
+    svg = super()
+
+    if add_no_extrapolation_css_class
+      svg.sub!("<svg", "<svg class='no-extrapolation'")
```
fpsvogel: I resorted to manipulating the HTML output string because AFAICT [the svggraph gem](https://github.com/lumean/svg-graph2) doesn't provide a way to set a custom class on the output.

fpsvogel: Now that I’ve slept on it, I think it’s worth taking another look into the gem to see if there’s anything we can monkeypatch so as to add the CSS class in a way that’s less brittle to potential future changes to the gem (e.g. if the gem starts adding a class, causing this to break). Or at least to assess the likelihood of a breakage like that. I’ll mark this PR as ready again after I’ve done that.

app/models/stats.rb
```diff
@@ -93,23 +95,11 @@ def self.get_cached_graph(name)
...
-    current_month = data[-2]
-    current_month_extrapolated_value = (
-      data.last / (
-        Time.now.utc.day.to_f /
-        Time.days_in_month(Time.now.utc.month, Time.now.utc.year)
-      )
-    ).round
-
-    [*data, current_month, current_month_extrapolated_value]
-  end
```
fpsvogel, line 114: Moved to TimeSeries ([below](https://github.com/lobsters/lobsters/pull/1763/files#diff-677c0202713f7209e4786949303de458c54c4d7f138d58fefd895116b15d8bb0R51-R61)).

lib/time_series.rb
fpsvogel: I refactored to [add in the CSS class while the graph is being built](https://github.com/lobsters/lobsters/pull/1763/files#diff-677c0202713f7209e4786949303de458c54c4d7f138d58fefd895116b15d8bb0R40-R51), rather than into the HTML output string.

pushcx:
Thanks for this fix, I like the way you injected the CSS to make this cleaner. I'll fix up the help text at the top that'll be inaccurate after I merge this.

If we really wanted to gold-plate this, `extrapolate` could be a block that either does the existing linear multiplication or something more clever for the 'active users' graph, but I'm happy to ship "not distractingly inaccurate". :smile: :chart_with_upwards_trend:

Issue #1525, opened 2025-04-16
Replace meta tags with Open Graph link previews


pushcx:
https://getoutofmyhead.dev/link-preview-meta-tags/

We currently use the Twitter tags approach described here (`app/controllers/stories_controller.rb`, ctrl-f `twitter`). It looks like Open Graph is the clear winner now.

One improvement: the image for our stories is always the Lobsters logo (or missing). After discussion in #547 I'm comfortable introducing libvips. Let's extract the image for the story we're linking to and overlay the Lobsters logo in the lower-left corner. We'll have more variety in our feeds and hopefully a clear visual distinction to people who see both a direct link to a post and our link to our story of that post.

If that doesn't work in practice, we could make put a border around the original and have our logo on a different visual layer like:

```
   +--------------+
   |              |
   |   Original   |
 +---+            |
 ||  |            |
 |L_ |------------+
 +---+
```

Related: #1248

PR #1764, opened 2025-10-20
Replace Twitter tags with Open Graph

luigieli:
Part of #1525

This PR intends to replace the Twitter tags with the Open Graph ones. Since there is no 1 to 1 conversion between Twitter and Open Graph tags, the closest ones were chosen to be used based on [Open Graph](https://ogp.me/) and [X (formely Twitter)](https://developer.x.com/en/docs/x-for-websites/cards/overview/markup) docs.

I used [threadcreator.com](https://threadcreator.com/tools/twitter-card-validator) to check the tags and a Twitter preview.

|Before| After|
|---|---|
|2025-10-20-102834_hyprshot|2025-10-20-102732_hyprshot|

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -167,13 +167,17 @@ class StoriesController < ApplicationController
     respond_to do |format|
       format.html {
         @meta_tags = {
-          "og:type"       => "website",
-          "og:site_name"   => "Lobsters",
-          "og:title"       => @story.title,
+          "og:type" => "article",
+          "og:site_name" => "Lobsters",
+          "og:title" => @story.title,
           "og:description" => @story.comments_count.to_s + " " +
             "comment".pluralize(@story.comments_count),
-          "og:image"       => Rails.application.root_url + "touch-icon-144.png",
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for picking this up, and you made it real convenient to review by linking the docs. I think it makes sense to replace our `type` with `article`, each story is usually about one article. Then we can add back the `author` field, which I think is valuable for giving credit to our submitters.

luigieli:
Thanks for the reply.
> I think it makes sense to replace our `type` with `article`, each story is usually about one article. Then we can add back the `author` field, which I think is valuable for giving credit to our submitters.

Sure, I'll change it right away.

pushcx:
Thanks! I can see a limitation in how our meta tags are implemented; the `author` tag can appear multiple times to give multiple links, so I'm going to merge down your contribution and then refactor a bit to better support this.

Thanks for taking this on, your PR has taken care of a fairly old issue!

luigieli:
> Thanks for taking this on, your PR has taken care of a fairly old issue!

Thanks for improving the logic on `author` tags and merging my first PR, I'm really happy it was useful.

PR #1765, opened 2025-10-26
Fix .ruby-version is needed inside container for running with Docker

fedemp:
Add `.ruby-version` to the files copied into the container when running the project with Docker. `bundler` requires it since it's referenced in the `Gemfile`.


pushcx:
Ah, the reference was just added in #1746, thanks for catching the oversight. It is especially nice when bug reports arrive in the form of fixes, thank you.

caius:
Whoops, didn't notice we had a Dockerfile when I put the change in the Gemfile. Thanks!

Issue #1720, opened 2025-09-14
Remove Message.has_been_read


pushcx:
We added a Notification class to handle read tracking on comments/messages. The old code on Message for read tracking should be removed, in favor of using the Notification model.

`InboxController` has a method for marking messages read that needs work. First, it shouldn't override existing values for `read_at` - we want the earliest timestamp a user read a Message or Comment, not the latest.

And then related to #1711, we should mark Notifications that are filtered out as read so that if they are revealed the user is not overwhelmed by old unread Notifications.

pushcx:
Relates to #1140

differentreality:
Is it safe to assume that removing `Message.has_been_read` does not require updating any other values beforehand? 

Meaning we can remove the attribute, and not miss any actual information from our business logic.

differentreality:
> And then related to https://github.com/lobsters/lobsters/pull/1711, we should mark Notifications that are filtered out as read so that if they are revealed the user is not overwhelmed by old unread Notifications.

This seems to be already addressed here https://github.com/lobsters/lobsters/blob/main/app/models/notification.rb#L11, right ? Is there anything more you'd have in mind regarding this?

pushcx:
Yep, you're correct that we can just drop the column, it's entirely redundant now.

Good catch on the completed work! That sentence predates the line of code. You're right that it's done.

PR #1767, opened 2025-10-27
Remove Message.has_been_read #1720

differentreality:
This addresses #1720 (partly; for the first part about removing the attribute `has_been_read`)

* Remove db field `has_been_read` from messages table 
* Remove the attribute wherever it was used with the codebase
* Remove a test, testing creation of notification for already read message, since we no longer have this information available 

This PR assumes that we no longer need `has_been_read` at all, ie we don't need to migrate this information in another db table. This information will no longer be available at all.

### DIFF
--- a/app/jobs/notify_message_job.rb
+++ b/app/jobs/notify_message_job.rb
@@ -8,7 +8,7 @@ class NotifyMessageJob < ApplicationJob
   end
 
   def deliver_message_notifications(message)
-    read_at = message.notification&.read_at ? read_at : nil
+    read_at = message.notification&.read_at
     message.recipient.notifications.create(notifiable: message, read_at: read_at)
 
     if message.recipient.email_messages?


app/controllers/inbox_controller.rb
```diff
@@ -51,7 +51,6 @@ def unread
   def update_read_at
     @notifications.touch_all(:read_at)
     message_ids = @notifications.of_messages.pluck(:notifiable_id)
```
pushcx:
```suggestion
```

This only existed for the next line.

app/jobs/notify_message_job.rb
```diff
@@ -8,7 +8,7 @@ def perform(*messages)
   end
 
   def deliver_message_notifications(message)
-    read_at = message.has_been_read ? Time.current : nil
+    read_at = message.notification&.read_at ? read_at : nil
```
pushcx:
```suggestion
    read_at = message.notification&.read_at
```

Your edit would've set `read_at` to nil, a quirk of ruby is that introducing a variable on both sides of an assignment is not an error: `a = a` just sets it to nil.

pushcx:
Thanks for cleaning this up. I made a couple tiny tweaks because this was so close to ready to go that I wanted to get it live.

PR #1768, opened 2025-10-27
1720 inbox update unread notifications

differentreality:
Part of #1720

* Update `read_at` only for notifications which were previously unread

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # local bundler config
 /.bundle
+bundle/
 
 # default databases
 db/**/*.sqlite3


spec/controllers/inbox_controller_spec.rb
```diff
@@ -10,12 +10,17 @@
     it "marks the notification and associated message as read" do
```
pushcx, line 10: When #1767 is merged, this test can be rewritten (possibly into 2 tests) to confirm that the notification is read whether or not its visible.

pushcx:
This is blocked on #1767, as a test it touches will need to be updated based on that.

pushcx:
Thanks for this, it just needs to be updated following your previous PR and #1720 will be ready to close.

differentreality:
Rebased. 
I am following up with test cases here https://github.com/lobsters/lobsters/pull/1778

pushcx:
I didn't notice that #1778 was stacked on top of this one, and I merged it down. So I think this is done and can close?

PR #1769, opened 2025-10-28
Fixes #1752 "caches" dropdown clips to the right

### LINKED ISSUE #1752: Caches dropdown menu displayed off screen on narrow viewports
The "caches" dropdown menu may be partly displayed off the screen on certain narrow viewports.

(The screenshot is taken from the Firefox Developer Tools, but I caught the issue initially in real life on my old iPhone.)

Related: #1566

Image

fedemp:
Migrate the "caches" dropdown to use the Popover API.

Calculate its width with CSS so if the dropdown may overflow outside the viewport, the width will shrink and the text will wrap.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -343,7 +343,7 @@ header#nav {
 }
 header#nav nav.links {
 	display: flex;
-	width: 100%
+	width: 100%;
 }
 header#nav a, header#subnav a, header#subnav input[type="submit"] {
 	color: var(--color-fg-contrast-6);
@@ -671,10 +671,20 @@ li .byline img.avatar {
 li.story .byline {
 	margin-top: 1px;
 }
+
+.byline button,
 .byline a {
 	color: var(--color-fg-contrast-4-5);
 	text-decoration: none;
+	border: 0;
+	padding: 0;
+	line-height: 1;
+	outline: revert;
 }
+.byline button:hover {
+	background-color: var(--color-bg);
+}
+
 .comment.replied .flagger,
 .comment.flagged .comment_replier {
 	color: var(--color-fg);
@@ -863,6 +873,7 @@ div.comment_text code {
 
 .dropdown_parent {
 	position: relative;
+	display: inline-block;
 }
 
 .comments_subtree .dropdown_parent > #flag_dropdown {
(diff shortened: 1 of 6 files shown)


fedemp:
[Screencast From 2025-10-28 17-01-49.webm](https://github.com/user-attachments/assets/91496321-69ad-43ae-b324-557c5ff8eb3b)

The "flag" dropdown does not work like this yet cause I haven't moved it to Popover.

fedemp:
Back to draft. I've just noticed that story pages are not working. :man_facepalming:

fedemp:
There are no more instances of the old checkbox. It's ready now.

fedemp:
~~Back to draft. It doesn't work on merged stories.~~

Fixed.

pushcx:
This looks really nice. In your video there's some ugly linebreaks as you resize. Could that be fixed with `white-space: preserve nowrap;`? I'm not an expert on [white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/white-space) but I think that's likely to work.

I'm waiting on your thoughts there, but I have no other questions and consider this ready to merge.

fedemp:
Unfortunately, it all comes down to browser support. On Firefox, the positioning is kind of a hack; popovers are supposed to open in the middle of the screen, but here I'm positioning it relative to the "caches" button. In this case, we are really limited, and we must allow the word break to prevent the content from overflowing outside the viewport again.

On the other hand, thanks to your comment, I found that, since Chrome (or Blink) natively supports linking the popover to an anchor (the button), we can set fallback positions if the popover would overflow. I've updated the PR. Here's a video showing how it works on Chromium.

[Screencast From 2025-11-05 00-06-49.webm](https://github.com/user-attachments/assets/59e8170c-dd32-44f2-b94a-c34ddfeadc3a)

app/assets/stylesheets/application.css
```diff
@@ -878,6 +889,37 @@ div.comment_text code {
...
+	width: max(100px, max-content);
+	inset: auto;
+	padding: 0;
+	margin: 5px 0 0 0;
+	position-area: bottom span-right;
+	overflow: visible; /* Allow focus ring to be visible */
+	position-try-fallbacks: --custom-left, --custom-bottom
+}
+
+@supports (position-area: left) {
```
pushcx, line 916: So far I've learned about a great browser feature from every one of your PRs, thank you. :)

pushcx:
Thanks for all this detailed design work on core UI, it's a relief to get this into a more attractive, semantic, and long-term reliable state.

pushcx:
This caused #1787 - I'm reverting to fix prod because I don't immediately see the bug.

PR #1770, opened 2025-10-29
feat JS: refactor parentSelectorOrNull to use closest()

fedemp:
No need to traverse all parent elements when you can use [`closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -29,19 +29,8 @@ export const onPageLoad = (callback) => {
   document.addEventListener('DOMContentLoaded', callback);
 };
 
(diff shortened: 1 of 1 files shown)


pushcx:
Oh, great! If there's a well-supported browser implementation, let's use it and entirely delete the `parentSelectorOrNull` function, calling `closest` in its place.

fedemp:
Ok. I have completely removed the old function and replaced every call with `closest()`.

pushcx:
I love getting to delete code like this, thank you.

PR #1773, opened 2025-10-29
Add https source to image policies on stories allowed to have images

luigieli:
Part of #1714 

This PR has the intent of adding the `https` CSP header on `img-src`, only on stories who are allowed to have images, that means only posts made by a user who is currently a moderator. Also added tests to check if the header is set correctly.

|Post by moderator|Post by normal user|
|---|---|
|```content-security-policy: default-src 'none'; connect-src 'self'; img-src 'self' data: https:; script-src 'self' 'nonce-jCAyrwjCPBB9gpHHCyBEBA=='; style-src 'self' 'unsafe-inline'```|```content-security-policy: default-src 'none'; connect-src 'self'; img-src 'self' data: ; script-src 'self' 'nonce-lC33t7pkf9HKauSsd+/COw=='; style-src 'self' 'unsafe-inline'```|

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/spec/requests/stories_spec.rb
+++ b/spec/requests/stories_spec.rb
@@ -272,6 +272,34 @@ describe "stories", type: :request do
         end
       end
     end
+
+    context "can have images" do
+      it "puts https CSP header" do
+        moderator = create(:user, :moderator)
+
+        story = create(:story, user: moderator, title: "test", url: "http://example.com/")
+        expect(story.can_have_images?).to eq(true)
+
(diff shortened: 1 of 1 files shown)


pushcx:
Ah, that's a very clear way to express what we're going for, and it's great that you added a couple tests. Thanks!

luigieli:
Thanks for reviewing and merging it. I'm glad my PR was useful.

PR #1774, opened 2025-10-29
fix #639 use <details> for story previews

### LINKED ISSUE #639: story preview overflow hidden is very annoying
The use of overflow:hidden on list elements hides text in an unfriendly way. Browser search will still find it, but gives very little visual indication of what or where.

The repro works (at the present time) from https://lobste.rs/t/databases using desktop chrome.

Search for "postgres" and observe several hits. Note that the *second* hit is the first one that is visible, for the "PostgreSQL vs. fsync" story. The story before that, Database as Filesystem, includes the word postgres somewhere in its summary but that is chopped off and hidden by CSS. Story link: https://lobste.rs/s/h2qhan/database_as_filesystem

I'm not sure what the fix is, but this was pretty infuriating when I didn't know what was happening. In that case, I was searching for postgres on the front page, which hit on two stories, but only one visible match. I stared and stared at the screen trying to find the highlighted text when it was invisible the whole time.

The simple fix of text-overflow:ellipsis doesn't seem to work. I think because the max-height is set on the bounding element? I only poked around for a little in the inspector.

fedemp:
Use the `<details>` element when showing story previews, so browsers will show matches inside when using the native search in page feature.

### DIFF
--- a/app/views/stories/_listdetail.html.erb
+++ b/app/views/stories/_listdetail.html.erb
@@ -53,10 +53,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
     <% end %>
 
     <% if @user&.show_story_previews? %>
-      <% if (sc = story.description_or_story_text(500)).present? %>
-        <div class="story_content">
+      <% if (sc = story.description_or_story_text()).present? %>
+        <details class="story_content">
+          <summary><span><%= story.description_or_story_text(250) %></span></summary>
           <%= sc %>
-        </div>
+        </details>
       <% end %>
     <% end %>
 
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -740,13 +740,23 @@ li div.details {
 div.story_content {
 	margin-bottom: 1em;
 }
-ol.stories.list div.story_content {
+ol.stories.list details.story_content {
 	color: var(--color-fg-contrast-6);
-	max-height: 2.6em;
 	margin: 0.25em 40px 0.25em 0;
-	overflow: hidden;
-	text-overflow: clip;
 }
+
+ol.stories.list details.story_content:open {
+	color: var(--color-fg-contrast-10);
+}
+
+ol.stories.list details.story_content:open summary {
+	float: left;
+	line-height: 1;
+	margin-top: 0.2rem;
+}
+
+ol.stories.list details.story_content:open summary span {
+	display: none;
 
 div.morelink {
 	float: left;
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -68,10 +68,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
     <% end %>
 
     <% if @user&.show_story_previews? %>
-      <% if (sc = story.description_or_story_text(500)).present? %>
-        <div class="story_content">
+      <% if (sc = story.description_or_story_text()).present? %>
+        <details class="story_content">
+          <summary><span><%= story.description_or_story_text(250) %></span></summary>
           <%= sc %>
-        </div>
+        </details>
       <% end %>
     <% end %>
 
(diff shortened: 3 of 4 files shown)


fedemp:
I'm open to suggestions. As mentioned in the original issue, the browser may find text that is hidden due to `overflow: hidden`. `<details>` automatically opens when that happens.

The downside is that the first characters of the description are duplicated inside the `<summary>`. I lessen that by changing the color and font size of the summary when details is open. 

image

image

db/queue_schema.rb
```diff
@@ -10,132 +10,5 @@
...
-    t.index ["key"], name: "index_solid_queue_semaphores_on_key", unique: true
-  end
-
-  add_foreign_key "solid_queue_blocked_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
-  add_foreign_key "solid_queue_claimed_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
-  add_foreign_key "solid_queue_failed_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
-  add_foreign_key "solid_queue_ready_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
-  add_foreign_key "solid_queue_recurring_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
-  add_foreign_key "solid_queue_scheduled_executions", "solid_queue_jobs", column: "job_id", on_delete: :cascade
+ActiveRecord::Schema[8.0].define(version: 0) do
```
pushcx: Looks like a hassle in your dev env; the queue didn't get set up correctly and so Rails wiped out the schema. Please revert changes to this file.

db/schema.rb
```diff
@@ -493,38 +493,6 @@
     t.index ["user_id", "story_id"], name: "user_id_story_id"
   end
```
pushcx, line 495: Revert these changes as well.

db/cache_schema.rb
```diff
@@ -10,15 +10,5 @@
...
-    t.binary "key", limit: 1024, null: false
-    t.binary "value", limit: 536870912, null: false
-    t.datetime "created_at", null: false
-    t.integer "key_hash", limit: 8, null: false
-    t.integer "byte_size", limit: 4, null: false
-    t.index ["byte_size"], name: "index_solid_cache_entries_on_byte_size"
-    t.index ["key_hash", "byte_size"], name: "index_solid_cache_entries_on_key_hash_and_byte_size"
-    t.index ["key_hash"], name: "index_solid_cache_entries_on_key_hash", unique: true
-  end
+ActiveRecord::Schema[8.0].define(version: 0) do
```
pushcx: Revert.

app/views/stories/_listdetail.html.erb
```diff
@@ -53,10 +53,11 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
     <% end %>
 
     <% if @user&.show_story_previews? %>
-      <% if (sc = story.description_or_story_text(500)).present? %>
-        <div class="story_content">
+      <% if (sc = story.description_or_story_text()).present? %>
```
pushcx, line 56: This change also needs to be made in `app/views/stories/_singledetail.html.erb`

fedemp:
I have added a rule so the `summary` is hidden when `details` is open, without sacrificing the ability to toggle it. I'm just concerned how screen readers will react to the `summary` suddenly disappearing; we will have to wait for feedback. :shrug: 

image

app/views/stories/_listdetail.html.erb
fedemp: There are no story texts in `singledetails`. And to be honest, it actually calls my attention that every story *in the dev environment* has a story text but when I visit the single details page, there's nothing there.

pushcx: Ah, sorry for the confusion, I was thinking about the display of descriptions in `app/views/stories/show.html.erb`, but I also forgot that those are always visible, not collapsible. So there's nothing to do here!

pushcx:
If it causes problems, we should update `truncate` to try to split at the whitespace nearest to but less than 500 characters, and return both the summary and the rest of the description. Then the browser should run them together cleanly. I'm OK merging it and waiting for feedback before we guess at something more complicated to implement, though.

Thanks for making this improvement to our markup.

PR #1775, opened 2025-10-30
feat Add aria-current="page" in navigation

fedemp:
`aria-current` can be used when you have a group of related elements and there's already a visual indication for the currently selected item. The `page` value represent the current page.


pushcx:
[MDN's documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current)

Looks like a great little improvement, thank you.

PR #1776, opened 2025-10-30
Fix error user not found while running fake_data

differentreality:
`docker compose run app rails fake_data` would cause 2 errors regarding missing `user` and `disabler`

```
NoMethodError: undefined method 'id' for nil (NoMethodError)

      m.moderator_user_id = user.id
                                ^^^
/lobsters/app/models/user.rb:435:in 'block in User#grant_moderatorship_by_user!'
/lobsters/app/models/user.rb:430:in 'User#grant_moderatorship_by_user!'
/lobsters/lib/tasks/fake_data.rake:71:in 'block in FakeDataGenerator#generate'
```

```
NoMethodError: undefined method 'id' for nil (NoMethodError)

      self.disabled_invite_by_user_id = disabler.id
                                                ^^^
/lobsters/app/models/user.rb:238:in 'block in User#disable_invite_by_user_for_reason!'
/lobsters/app/models/user.rb:236:in 'User#disable_invite_by_user_for_reason!'
/lobsters/lib/tasks/fake_data.rake:74:in 'block in FakeDataGenerator#generate'
```

The methods are relevant only if the user is set, so we don't need to run anything in case we don't have a user.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -233,6 +233,8 @@ class User < ApplicationRecord
   end
 
   def disable_invite_by_user_for_reason!(disabler, reason)
+    return unless disabler
+
     User.transaction do
       self.disabled_invite_at = Time.current
       self.disabled_invite_by_user_id = disabler.id
@@ -427,6 +429,8 @@ class User < ApplicationRecord
   end
 
   def grant_moderatorship_by_user!(user)
+    return unless user
+
     User.transaction do
       self.is_moderator = true
       save!


pushcx:
I don't want to risk production code like this. If the disabler isn't passed in prod, this function would fail silently.

If `fake_data` is failing, the bug is that the task is trying to pass a `nil`. We want to fix the bug at the source, not accomodate it elsewhere in the stack.

PR #1777, opened 2025-10-30
Move comment score from model to helper

### LINKED ISSUE #1517: Score helper to limit width
We print comment scores in a column with the upvoter arrow. This works fine now for comment scores that are typically 1 or 2 digits, and rarely 3. While reviewing #1508  we realized that when comment scores are 4 digits or more the column will get wider. And then either child comments get inconsistent widths for this column (ugly) or they have overbroad widths for their low scores (also ugly).

Move `Comment#score_for_user` to a helper (it'll have to live in `ApplicationHelper`, the comment view is reused many places) because that's a better home for this kind of "how to display UI" concern.

Then add a branch so that positive scores get rounded when > 999, so they're always 3 characters wide at most. So 1000 -> 1k, 10000 -> 10k, 100000 -> ???, 1000000 -> 1m. Add the exact score as a title attribute so people can see it on hover.

There no tests for `score_for_user` so this is a great time to add some.

differentreality:
Closes #1517 

1. Move comment score to ApplicationHelper (instead of Comment model)
2. Add tests for the new method

Additional commits for changing how we display the score: 

1. Format number with `number_to_human`  to show 1K, 10M, etc.
2. Use translations for abbreviated unit 

Still an issue with more than 3 characters:
* Negative numbers less than -99 (eg. -99K)
* 100K and 100M 

image

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -194,20 +194,16 @@ module ApplicationHelper
   # Returns the score value, if the user can see the comment's score
   # Returns "~" otherwise
   def comment_score_for_user(comment, user)
-    return {} unless comment
+    return nil unless comment
 
-    score = if comment.show_score_to_user?(user)
+    if comment.show_score_to_user?(user)
       comment.score
     else
       "~"
     end
+  end
 
-    score_formatted = if score.is_a?(Integer)
-      number_to_human(score, format: "%n%u")
-    else
-      score
-    end
-
-    {score_value: score, score_formatted: score_formatted}
+  def upvoter_score(score)
+    number_to_human(score, format: "%n%u") if score.is_a? Integer
   end
 end
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -15,13 +15,13 @@
       (comment.score > Comment::COLLAPSE_SCORE && !comment.current_flagged?) ? "" : "checked" %>>
 
       <% can_flag = @user && @user.can_flag?(comment) %>
-      <% score_display = comment_score_for_user(comment, @user) %>
+      <% score = comment_score_for_user(comment, @user) %>
       <% draw_tree_line = comment.reply_count > 0 %>
 
       <div class="
         <%= draw_tree_line ? "comment_parent_tree_line" : "" %>
         <%= can_flag ? "can_flag" : "" %>
-        <%= score_display[:score_value] != "&nbsp;" ? "score_shown" : "" %>
+        <%= score != "~" ? "score_shown" : "" %>
       "></div>
 <div id="c_<%= comment.short_id %>"
   data-shortid="<%= comment.short_id if comment.persisted? %>"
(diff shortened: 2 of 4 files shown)


differentreality:
Plz do let me know if you think of any additional test case which makes sense.

app/helpers/application_helper.rb
```diff
@@ -184,4 +184,29 @@ def how_long_ago_label(time)
   def how_long_ago_link(url, time)
     content_tag(:a, how_long_ago_label(time), href: url)
   end
+
+  ##
+  # This replaces comment.score_for_user(user)
+  # Returns the score value, if the user can see the comment's score
+  # Returns "~", if the user can flag the comment
```
pushcx: I didn't see it until your comment laid it out, but this design decision no longer makes sense. The ~ was there because the downvote/flag arrow [used to be underneath the score](https://github.com/lobsters/lobsters/commit/30057c7416e0507abeff97f2e99d29a98453d5c2), but I moved it away years ago to get better signal from flags.

Please update this function (and comment) to show the score if the user should see it and `~` otherwise; no `&nbsp;`.

app/helpers/application_helper.rb
```diff
@@ -184,4 +184,29 @@ def how_long_ago_label(time)
...
+      "&nbsp;".html_safe
+    end
+
+    score_formatted = if score.is_a?(Integer)
+      number_to_human(score, format: "%n%u")
+    else
+      score
+    end
+
+    {score_value: score, score_formatted: score_formatted}
```
pushcx: I really don't like returning an ad-hoc data structure from a function. I see that you want to have the full score on the `title` attribute and display the shortened score, but I would rather you created a second helper:

```ruby
  def upvoter_score(s)
    number_to_human(s, format: "%n%u") if s.is_a? Integer
  end
```

app/helpers/application_helper.rb
```diff
@@ -184,4 +184,29 @@ def how_long_ago_label(time)
   def how_long_ago_link(url, time)
     content_tag(:a, how_long_ago_label(time), href: url)
   end
+
+  ##
+  # This replaces comment.score_for_user(user)
+  # Returns the score value, if the user can see the comment's score
+  # Returns "~", if the user can flag the comment
+  def comment_score_for_user(comment, user)
+    return {} unless user
```
pushcx: Delete this guard clause, logged out users can still see scores.

app/helpers/application_helper.rb
differentreality: OK that now makes more sense :slightly_smiling_face:

differentreality:
> Thanks for taking this on this long-term fix. It'll be a while before we start seeing scores above a thousand (and quite a bit longer before we hit a quadrillion...) but it's nice to be ready.
> 
> Maybe for 100K and 100M we could present those as 0.1M and 0.1B, but I can see how that'd require custom code rather than getting to lean on the framework like this. We have long enough to hit 100K that I'm very happy to get this merged and worry about that in a few decades.

Few decades it is then! 

I've addressed all points. Let me know if anything else is needed.

pushcx:
Thanks for contributing this fix. :)

PR #1780, opened 2025-10-30
fix #1090 Upvote arrows are invisible in high contrast mode

### LINKED ISSUE #1090: a11y: forced colors support
Windows and Firefox support "forced colors", the ability to override a website color palette with user-defined colors. Windows calls this "High-Contrast Mode"; Firefox exposes it in the "Manage Colors" modal of `about:preferences`.

Testing on Firefox 100 on Linux: forced colors causes vote arrows to render as white squares. I imagine the problem also exists on Windows High Contrast Mode. Here's a screenshot for reference:

Comment thread in lobsters; the upvote arrow is a white rectangle over a dark background.

Possible solutions, in descending order of desirability from my perspective:

- Render the upvote as an SVG. A simple static SVG-Tiny should do the trick.
- Render the upvote as a plain old PNG
- Render the upvote using icon fonts

fedemp:
Add a CSS file with shared rules for light and dark mode, when the OS forces high contrast mode. Should not apply for users that chose higher contrast in their profile.

This has only been tested on Firefox and Chromium, on Linux.


fedemp:
Firefox only allows blue for links, and a primary color for text.

Light mode in Firefox:

image

Dark mode in Firefox:

image

Dark mode in Chromium:

image

Light mode in Chromium:

image

fedemp:
I have added a fix for the logo that is invisible too in high contrast mode. It's in a separate commit in case you want me to revert it.

I had to inline the svg in the html so the svg could inherit the link color in high contrast mode (in firefox, at least).

image

image

app/assets/stylesheets/application.css
```diff
@@ -325,8 +325,7 @@ header {
 	line-height: 18px;
 }
 
-#logo {
-	background: transparent url(logo-transparent.svg) no-repeat center;
+#nav #logo {
```
pushcx: Why change this selector?

app/views/layouts/application.html.erb
```diff
@@ -62,7 +62,12 @@
 </head>
 <body data-username='<%= @user.try(:username) %>' data-now-unix='<%= Time.current.to_i  %>'>
   <header id="nav">
-    <a id="logo" style="<%= @traffic_style %>" href="/" title="<%= Rails.application.name %> (Current traffic: <%= @traffic_intensity.to_i %>%)"></a>
+    <a id="logo" aria-label="<%= Rails.application.name %>" style="<%= @traffic_style %>" href="/" title="<%= Rails.application.name %> (Current traffic: <%= @traffic_intensity.to_i %>%)">
+      <svg aria-hidden viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```
pushcx: Why inline the logo?

app/assets/stylesheets/system-system.css
```diff
@@ -1,6 +1,7 @@
 /* system color scheme (system contrast) */
 
 @import url("light-normal.css"); /* default */
+@import url("application-high.css") (prefers-contrast: more);
```
pushcx: Why only do this when the browser requests high contrast, instead of also when the user's sets that they prefer high contrast? This seems like a bug.

pushcx (changes requested):
> Firefox only allows blue for links

Could you share a link about this? I hadn't heard of this restriction.

I appreciate that you saw some issues and sent in a fix, but I'm not a CSS expert so I have added some pretty basic questions about how this works and why these changes are needed. :)

app/views/layouts/application.html.erb
fedemp: The logo is white and only visible because there's a background around. Firefox removes (almost) everything related to colors, including background colors; so now we have a white image on a white background. Notice how in dark mode, we can see the white logo on the dark background.

The good thing is that if we inline the SVG, we can modify it with CSS and make it inherit the text color. Since Firefox sets the link color to blue, the SVG can inherit the same blue. This is not possible if we add the image as background through CSS (as it's currently done) or if we add the image with `:before` (I've just tried).

fedemp:
> > Firefox only allows blue for links
> 
> Could you share a link about this? I hadn't heard of this restriction.
> 
> I appreciate that you saw some issues and sent in a fix, but I'm not a CSS expert so I have added some pretty basic questions about how this works and why these changes are needed. :)

The article *[Change the fonts and colors websites use](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use#w_contrast-control)* shows the contrast settings and how web pages will only use four colors.

app/assets/stylesheets/system-system.css
fedemp: I see your point. I was under the impression that the high contrast CSS files were only available for registered users.

But given your suggestion, then I would need to duplicate the rules from `application-high` in both `light-high` and `dark-high`. It that alright?

app/assets/stylesheets/application.css
fedemp: Because of the `#fff` I added below. There's a `#header a` that has precedence over this when setting the font color.

app/views/layouts/application.html.erb
pushcx: Could you check that this works with the novelty logos? They're in `app/helpers/traffic_helper.rb`, function `novelty_logo`. The easiest way to check is to edit the date conditionals to just `if true`.

I'm also concerned by adding 900 bytes to every pageload. I'd rather see this moved into a `.svg` in the asserts directory than inline.

app/assets/stylesheets/system-system.css
pushcx: Not really, those files only have color variables. If we start putting selectors and rules in there, contributors have increased hassle finding what they want to edit.

fedemp: Adding a `@media (prefers-contrast: more)` rule in the `application.css` for high contrast exceptions would be better?

app/views/layouts/application.html.erb
fedemp: > I'm also concerned by adding 900 bytes to every pageload. I'd rather see this moved into a .svg in the asserts directory than inline.

But then we are back to the original problem; the SVG can't be modified if not inlined and will render white over white on a light system theme.

BUT I have another offer. What if we ditch the image in high contrast mode and just use CSS generated content to add an uppercase L as plain text?

Edit: For context, I mean "High contrast by OS", not from the site preferences.

Default/Light/Dark on Firefox:

Screenshot From 2025-11-11 15-04-54 Screenshot From 2025-11-11 15-04-47 Screenshot From 2025-11-11 15-05-04

----

Default/Light/Dark on Chrome:

Screenshot From 2025-11-11 15-04-01 Screenshot From 2025-11-11 15-04-10 Screenshot From 2025-11-11 15-08-50

app/assets/stylesheets/system-system.css
pushcx: Yes, but I'm still left the question of why you're multiplying the complexity by having differences between os-requested high contrast mode and the user setting for high contrast mode.

fedemp: I forgot to move the bookmark; that's why I didn't get what you were talking about.

I deleted the `application-high.css` and will only show the text logo for users that set High Contrast at OS level, because I don't want to show the fallback one for those registered users that chose the High Contrast theme.

pushcx:
This looks like a good compromise approach to get _something_ decent working in the reduced colors mode, thanks. Please match the indentation and I'll get this merged down.

fedemp:
> This looks like a good compromise approach to get _something_ decent working in the reduced colors mode, thanks. Please match the indentation and I'll get this merged down.

Indentation should be fixed now.

pushcx:
Great! Thanks for spotting this issue and working to such a nice fix for it.

PR #1783, opened 2025-11-04
Fix #1685: Keep session alive to verify CSRF token

### LINKED ISSUE #1685: Filters can no longer be set while offline
Hello !

Today, after a session reset, I could no longer save filters via cookies, when logged out.

The POST request, with a seemingly proper body, gets 302'd to the same location, except without any Set-Cookie, so it's effectively a no-op.

1. Log out/use a private window

2. Go to filters, check any box, save

3. Observe POST (and unchecked boxes)

Image

Image

The redirected GET response (just in case):

Image

I tried under a different browser profile (Firefox by the way), still happens. No system updates so that can be ruled out.

Furthermore, the site is now broken under w3m (no line) when it used to work fine, so I couldn't test it there either. Is it somehow related/worthy of a separate issue ?

toretore:
This fixes #1685 (typo in commit messsage) by keeping the session cookie around until the `tag_filters` cookie can be set. It works by *not* deleting the cookie at the end of the `FiltersController#index` request, which will make the session available to the next request, possibly `FiltersControllerupdate`, where the CSRF token (stored in the session cookie) can be used to verify the form submission.

The `clear_session_cookie` action has been rewritten to take into account a session that's been loaded (and not `empty?`), but only contains default keys. The session gets loaded by `verify_authenticity_token`.

Caveat: If the user clicks away from the filters page, that request will also have a session cookie, and not be cached.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -89,7 +89,7 @@ class ApplicationController < ActionController::Base
   def clear_session_cookie?
     # If the session has been loaded, it will contain some default keys and not
     # be `empty?`
-    !@user && session.keys.all?{|k| SESSION_DEFAULT_KEYS.include?(k.to_s) }
+    !@user && session.keys.all? { |k| SESSION_DEFAULT_KEYS.include?(k.to_s) }
   end
 
   def find_user_from_rss_token


pushcx:
Thanks for taking a crack at this, this looks like a clean and maintainable solution. I'm ready to merge this when the linter's quiet.

The security/brakeman failure is a safety measure. I'll update it on main and you can pull/rebase to clear that part of the build.

toretore:
Fixed linter errors. Sorry about the git mess, I was just throwing commands at it until I got it to work, I'm not used to this workflow..

pushcx:
No worries, I value fixing bugs a lot more than making vcs history pretty!

Thanks for diagnosing and fixing this bug.

Issue #1588, opened 2025-05-24
Reposted URL not Flagged as Previously Submitted


veqqq:
https://lobste.rs/s/worahh/petnames_humane_approach_secure

Though these URL's are identical:

`https://files.spritely.institute/papers/petnames.html` 
`https://files.spritely.institute/papers/petnames.html`

The older story: https://lobste.rs/s/worahh/petnames_humane_approach_secure is not cached in the url repeat filter. Trying to submit the URL a 3rd time triggers the filter with the newer submission: https://lobste.rs/s/worahh/petnames_humane_approach_secure

The older story does show the newer story under `Stories with similar links:` but the newer story does not show the older!

chamlis:
Agreed that this is another expression of #1529. Of those three stories, the oldest has a different URL entirely (that now redirects to the same URL as the other two). The middle story was before the normalisation change, so its normalised URL ends in `.html`. The newest story is after the normalisation change, so its normalised URL doesn't end in `.html`.

Calculating similar stories [doesn't use the cached normalised value for the starting story](https://github.com/lobsters/lobsters/blob/main/app/models/story.rb#L426) but does for the candidate destination stories. As this case demonstrates this is not always symmetric.

I hope to be able to fix these two issues with the suggested normalisation migration tomorrow.

PR #1784, opened 2025-11-04
Add functions to update cached normalized URLs

### LINKED ISSUE #1529: Duplicate URLs are getting submitted
We have a bunch of logic to prevent URLs from being resubmitted within 30 days, with more complex rules after that. It's not working:

https://lobste.rs/s/bblc5z and https://lobste.rs/s/blhj1m have identical links to https://diziet.dreamwidth.org/11840.html (so this is not just a bug in `Utils.normalize`). The submitter didn't get a warning they were resubmitting and wasn't required to add a comment to start a new discussion.

https://lobste.rs/s/90djfu This link was submitted twice a couple days apart, differing only by `.html`.

This bug surprises me! I thought we had thorough tests of url normalization, submission of duplicate links, and permitted resubmission of links.

chamlis:
If the normalization algorithm changes, as it did to start stripping a ".html" suffix, new stories may miss old cached matches that were normalized differently. Whenever it does change, we want to update the cached results in Story and Link.

@pushcx: In #1529 you mentioned wanting a migration for this. I copied the public method approach for updating markeddown text, but I'm happy to rework this into a migration if that's preferable.

Closes #1529 and #1588

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -76,7 +76,7 @@ class Link < ApplicationRecord
   def self.refresh_normalized_urls!
     # Note that we assume normalization doesn't affect whether a URL links to a comment or story
     # on site, as calculated in url=
-    Link.where.not(url: "").find_each { |l| l.update_columns normalized_url: Utils.normalize(l.url) }
+    Link.find_each { |l| l.update_columns normalized_url: Utils.normalize(l.url) }
   end
 
   private


app/models/link.rb
```diff
@@ -70,6 +70,15 @@ def self.recreate_from_story! s
     end
   end
 
+  # Recalculates the cached normalized URL for each link, should be run whenever
+  # Utils.normalize changes so that new story submissions can locate links in old
+  # discussions correctly.
+  def self.refresh_normalized_urls!
+    # Note that we assume normalization doesn't affect whether a URL links to a comment or story
+    # on site, as calculated in url=
+    Link.where.not(url: "").find_each { |l| l.update_columns normalized_url: Utils.normalize(l.url) }
```
pushcx:
```suggestion
    Link.find_each { |l| l.update_columns normalized_url: Utils.normalize(l.url) }
```

Empty URLs are prohibited by validation and I ran `select count(*) from links where url = "";` to confirm there are none.

pushcx:
Thanks for writing this. I feel like it happens infrequently enough that I don't care much if it's a formal migration or a thing I manually do in the console.

It's a shame the [Rails normalization feature](https://api.rubyonrails.org/classes/ActiveModel/Attributes/Normalization.html) misses our use case of storing both the original and normalized. Though I'd like to replace the `markeddown_` columns with fragment caching sometime, seeing this repeated pattern in the code has me suspecting there's a missing generalization around refreshing calculated values as well; we also do it in score, karma, counter caches, etc.

chamlis:
> It's a shame the [Rails normalization feature](https://api.rubyonrails.org/classes/ActiveModel/Attributes/Normalization.html) misses our use case of storing both the original and normalized. Though I'd like to replace the `markeddown_` columns with fragment caching sometime, seeing this repeated pattern in the code has me suspecting there's a missing generalization around refreshing calculated values as well; we also do it in score, karma, counter caches, etc.

I'll have a think about this - it does feel like this should be solvable. This case, and that of markdown,feels distinct from the others you list in that it's the algorithm is prone to changing as well as the input data, which requires updating the cache for all of the items. I don't know how that could work in a cleaner way than manually triggering it as appropriate.

The others feel like they should fit the platonic ideal of having generated columns for them, and in principle the DB could know when they need updating and defer recalculating them until things are idle or they're needed. I remember hearing there was a project to stop using views altogether and treat the DB as a simpler datastore, so I guess this kind of thing didn't work out well in practice?

If SQLite ends up being a good fit for the site, the latency of recomputing some of these might be small enough that we could drop the cache altogether, but that's just more speculation on my part.

> One small change, I'll commit it.

Good catch, thanks! Too blind of a copy-paste on my part.

PR #1786, opened 2025-11-05
Add dependabot.yml that checks only for brakeman

### LINKED ISSUE #1785: Configure dependabot to bump brakeman
We conservatively configured the brakeman security scanner to fail the build if there's a newer version available. Inconveniently, this often means that random contributors see what looks like a security warning, like in #1783. 

Dependabot is freely available and already configured to give us security warnings. It would be nice to [configure it](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates) to open a PR when new versions of brakeman are available. Maintainers would immediately hear of the new version to more easily update it, and contributors may see the update to better understand their failing builds.

fpsvogel:
## Summary

Resolves #1785 

Adds a Dependabot config that checks daily for new Brakeman versions, and doesn't check any other gems. So when a new version of Brakeman is released, Dependabot will open a version bump PR on the same day.

Also, tangentially related, this PR removes a line from the Gemfile related to a Brakeman issue that has been resolved.

## Questions

- Would it make sense to also remove the `--ensure-latest` option in the ["Run brakeman"](https://github.com/lobsters/lobsters/blob/dd14f3ac16ba97ed88d5c6f234a56c49935284d4/.github/workflows/check.yml#L82-L84) CI step? In other words, can we remove that blocker to PRs now that Brakeman will *generally* be up to date?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Run brakeman
         run: |
-          bundle exec brakeman --ensure-latest --ensure-ignore-notes
+          bundle exec brakeman --ensure-ignore-notes
 
   leak-check:
     runs-on: ubuntu-latest


Gemfile
```diff
@@ -68,7 +68,6 @@ group :test, :development do
   gem "foreman"
   gem "letter_opener"
   gem "listen"
-  gem "prism" # rm after https://github.com/presidentbeef/brakeman/issues/1909 closes
```
fpsvogel, line 71: That issue is closed. Also, as I mentioned in the commit message for this, https://github.com/lobsters/lobsters/commit/236bce1827e88413af8c5cc2375973dd5f9fb4bc made Prism 1.4 a direct dependency, so it no longer needs to be listed in the Gemfile. (Gemfile.lock still specifies Prism 1.4.0.)

pushcx: Oh great! I forgot about this issue, thanks for checking in on it and closing it out.

pushcx:
Thanks for adding this config.

I think you have a good idea to remove the `--latest` from our brakeman CI. Doing so removes a discouraging annoyance from contributors and this dependabot config means maintainers (me) will be notified of available versions _faster_, so it's not an increased risk to the site. Do you mind rolling it into this PR?

fpsvogel:
@pushcx Sure, I added [`e3ccbfd` (#1786)](https://github.com/lobsters/lobsters/pull/1786/commits/e3ccbfd0adad73f7420b4407f1cf572f792d3fdf) which removes that option in the CI step.

pushcx:
Great, thanks for improving this PR. I'm sure this will mean less friction for future contributors.

PR #1788, opened 2025-11-06
fix #639 use <details> for story previews

### LINKED ISSUE #639: story preview overflow hidden is very annoying

fedemp:
Use the `<details>` element when showing story previews, so browsers will show matches inside when using the native search in page feature.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -740,12 +740,23 @@ li div.details {
 div.story_content {
 	margin-bottom: 1em;
 }
-ol.stories.list div.story_content {
+ol.stories.list details.story_content {
 	color: var(--color-fg-contrast-6);
-	max-height: 2.6em;
 	margin: 0.25em 40px 0.25em 0;
-	overflow: hidden;
-	text-overflow: clip;
+}
+
+ol.stories.list details.story_content:open {
+	color: var(--color-fg-contrast-10);
+}
+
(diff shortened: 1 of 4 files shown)


fedemp:
This PR is a duplicate of #1774 that fixes a typo/bug in the css file.

pushcx:
Ah, sorry for your inconvenience, you didn't need to reopen these. I pushed the `main` bookmark back after my first revert attempt failed. I've restored it, so this is all set.

PR #1789, opened 2025-11-06
Fixes #1752 "caches" dropdown clips to the right

### LINKED ISSUE #1752: Caches dropdown menu displayed off screen on narrow viewports

fedemp:
Migrate the "caches" dropdown to use the Popover API.

Calculate its width with CSS so if the dropdown may overflow outside of the viewport, the width will shrink and the text will wrap.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -343,7 +343,7 @@ header#nav {
 }
 header#nav nav.links {
 	display: flex;
-	width: 100%
+	width: 100%;
 }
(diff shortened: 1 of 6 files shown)


fedemp:
This is a duplicate of #1769 that was reverted. The fix that justified the revert is included in #1788

PR #1790, opened 2025-11-06
feat JS: refactor parentSelectorOrNull to use closest()

fedemp:
No need to traverse all parent elements when you can use [`closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)


fedemp:
Duplicate of #1170 that was reverted.

pushcx:
Ah, sorry for your inconvenience, you didn't need to reopen these. I pushed the `main` bookmark back after my first revert attempt failed. I've restored it, so this is all set.

PR #1792, opened 2025-11-07
Use fieldset and legend for radio groups

fedemp:
The best way to give a title to a group of radio inputs is using `<fieldset>` and `<legend>`.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1147,6 +1147,7 @@ table.data.zebra tr:nth-child(odd) td,
 table.data tr.nobottom td {
 	border-bottom: 0px;
 	padding-bottom: 0px;
+
 }
 table.data tr.nobottom td.wrap {
 	overflow-wrap: anywhere;
@@ -1500,7 +1501,6 @@ input[type="submit"].link_post.pushover_button {
 }
 
 /* search */
-
 .searchresults summary .heading {
 	font-weight: bold;
 	font-size: 11pt; /* h2 */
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application.css
```diff
@@ -1530,6 +1531,16 @@ input[type="submit"].link_post.pushover_button {
...
 
+@media only screen and (max-width: 450px) {
+	#q {
+		width: 100%;
+
+		& + input {
+			margin-top: 0.25rem;
+		}
+	}
+}
```
fedemp: This is a little nitpick. When the screen is **too** small, the "search" button nearly overlapped the input.

Image

app/assets/stylesheets/application.css
```diff
@@ -1176,6 +1176,7 @@ table.data.zebra tr:nth-child(odd) td,
 table.data tr.nobottom td {
 	border-bottom: 0px;
 	padding-bottom: 0px;
+
```
pushcx, line 1150:
```suggestion
```

pushcx (commented):
Thanks for spotting the bug and opening a PR.

We already have a breakpoint at 480, what's happening in the 30px that we need to introduce a second one?

app/assets/stylesheets/application.css
```diff
@@ -1547,6 +1558,55 @@ input[type="submit"].link_post.pushover_button {
...
+			margin-bottom: 0.5rem;
+		}
+
+		.radio-group {
+			gap: 1rem;
+
+			label {
+				display: inline;
+				float: none;
+				width: auto !important;
```
pushcx: `!important` is not a shortcut I want to take here, we don't have that much css.

pushcx:
Could you say some more about what makes this the best way to group radio buttons? Do screen readers treat it better or something? I noticed you added a missing aria role, which is nice.

fedemp:
> Could you say some more about what makes this the best way to group radio buttons? Do screen readers treat it better or something? I noticed you added a missing aria role, which is nice.

I'm just starting to focus more on accessibility, so screen readers are a mystery to me yet. Even so, for this specific case, I tried using Orca on the search page. For both tests, I started with the text input focused, turned on Orca and then reloaded the page.

For the current iteration in production, on page load, Orca said "Leaving form, entry" (yes, I'm confused on what that even means). After I pressed `Tab` to focus the radio buttons, it said "Comments, selected, radio button"

In my proposal, on page load, it says "Form search, search query entry". When I tab into the radio buttons, it says "Search panel, Comments, selected, radio button".

app/assets/stylesheets/application.css
fedemp: I've deleted it along with the `float`. I was probably trying to override some other rule and forgot to remove. It works as expected on both screen sizes.

app/assets/stylesheets/application.css
```diff
@@ -1500,6 +1500,15 @@ input[type="submit"].link_post.pushover_button {
 }
 
 /* search */
+@media only screen and (max-width: 450px) {
```
pushcx: Please match the tab indentation used in the rest of the file.

I'm still curious about why this needs a new breakpoint that's so close to the existing one.

fedemp: Fixed the indentation.

About the breakpoint, it's because if the screen is too narrow, the Search button is placed too close to the input. In the screenshot, the viewport is 387px wide and the text input is focused. The current breakpoint is triggered before the button is moved down so adding a `margin-top` there would misalign the elements.

**Before**

image

**After**

image

pushcx:
Just to check in, I think there's only two things to do here:

1. Match the existing indentation, which should be easier with #1802 .
2. Explain/remove the new mobile breakpoint that's close to the existing one.

fedemp:
> Just to check in, I think there's only two things to do here:
> 
>     1. Match the existing indentation, which should be easier with [Add .editorconfig #1802](https://github.com/lobsters/lobsters/pull/1802) .
> 
>     2. Explain/remove the new mobile breakpoint that's close to the existing one.

Done and done.

pushcx:
Thanks for working through this with me!

PR #1795, opened 2025-11-07
Fix #1791: cache button issue in firefox

### LINKED ISSUE #1791: The "caches" button is bold when using Firefox, due to Firefox' default styles
The default CSS rules of Firefox are such that `button` elements are bold. For example, this:

```html
<!DOCTYPE html>
<html>
  <body>
    <button>Button text</button>
  </body>
</html>
```

Renders like so:

Image

The reason is that in Firefox the default `font-weight` value for `button` elements is 700 i.e. bold.

The result is that the Lobsters homepage looks like this:

Image

Would it be possible to use a font weight that's the same as the other links/buttons next to the "caches" button? Having "caches" in bold is a tad distracting and makes the button look more important than it probably is. Thanks! 😃

DoctorFateKS:
This PR should corrects issue [#1791](https://github.com/lobsters/lobsters/issues/1791?utm_source=chatgpt.com) , where the caches button appears bold in Firefox due to the browser’s default button styling.

- I do not have a Firefox environment to test this
- Verification in Firefox is recommended before merging.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -678,10 +678,9 @@ li.story .byline {
 	text-decoration: none;
 	border: 0;
 	padding: 0;
+	font-weight: normal;
 	line-height: 1;
 	outline: revert;
-}
-.byline button:hover {
 	background-color: var(--color-bg);
 }
 
@@ -885,10 +884,6 @@ div.comment_text code {
 .dropdown_parent {
 	position: relative;
 	display: inline-block;
-}
-
-.archive-button {
-	font-weight: normal;
 }
 
 .comments_subtree .dropdown_parent > #flag_dropdown {


app/assets/stylesheets/application.css
```diff
@@ -887,6 +887,10 @@ div.comment_text code {
 	display: inline-block;
 }
 
+.archive-button {
```
yorickpeterse:
```suggestion
.archive_button {
```

The class name is `archive_button`, with a `_` instead of `-`. Using `.archive_button { font-weight: normal }` in the Firefox inspector does the trick for me.

Looking at the inspector it seems most of the styling comes from the `.byline button, .byline a` selector, so it may be worth adjusting that one instead.

While we're at it: it seems there's also a background added by the `button, input[type="button"], input[type="reset"], input[type="submit"], a.button` selector, but this isn't cleared for this particular button, resulting in this (note the faint grey background):

image

DoctorFateKS: Woah... It seems I don't wake up well this morning to made an error like this...
Sorry for this syntax error and thanks for the correction @yorickpeterse  !

You're right for the selector, I will add the `font-weight: normal` property in the `.byline button`

For the background, it's effectively a very light background color (I struggled to see it in the inspector with Google chrome). However, is there an issue to correct with this ? I mean, do you have the background-color on your cache button with firefox ?

yorickpeterse: > For the background, it's effectively a very light background color (I struggled to see it in the inspector with Google chrome). However, is there an issue to correct with this ? I mean, do you have the background-color on your cache button with firefox ?

The issue is that the background is meant to be used for things that _look_ like buttons (at least based on it being introduced by the `button` selector), but the "caches" button here is meant to look the same as the other links that are adjacent to it. 

In other words, it technically _is_ a button but doesn't look like it, hence the background color is out of place.

DoctorFateKS:
The CSS property for the cache button was moved into the correct class selector, and the background styling was corrected. The commit also removes the hover rule that was no longer needed in my opinion.

pushcx:
I put a comment over in your report in #1791, but: we had to revert that code for other bugs. I'll retrieve this code if we use a `button` this way again.

PR #1796, opened 2025-11-07
Build Devcontainer setup

ChaelCodes:
# Description

Add [devcontainer](https://containers.dev) setup for Lobste.rs. Devcontainers open an instance of VS Code with existing theme, extensions, and auth inside a docker container. The experience is like working on a local machine, but it has all the benefits of a disposable, containerized dev environment. It is a layer on top of Docker to make it easier to work with in VS Code (or other theoretical editors that implement the specification). It also natively supports codespaces - cloud-editing.

Traditionally, when using docker, you have to prepend all commands with `docker compose`, there are permissions issues when a file is created in one container and accessed locally, and running commands like `rspec` happens in a new container, leaving behind container cruft if you forget to mark the container for removal. Also, basically no extensions work because your editor doesn't have the correct ruby version installed. (Ruby is installed inside docker.)

Devcontainers feels more like a local setup, because you're running VS Code _inside_ the docker container.

Benefits:
- Easy, consistent setup
- Developer can throw away the container and all dependencies when they're done contributing
- One-click GitHub codespaces support

_And yes, it has an open standard, but it does come from Microsoft._

# Architecture

[.devcontainer/devcontainer.json](https://github.com/lobsters/lobsters/pull/1796/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686)
This devcontainer.json file is responsible for how VS Code is configured and can run a few limited lifecycle commands on running containers.

[Dockerfile.dev](https://github.com/lobsters/lobsters/pull/1796/files#diff-86930c95a19b82f7e64a962a0053d44e855824813019b3698eae4917a90cdcac)
This Dockerfile.dev is responsible for building an image that can run the web server. Notice the CMD at the end. `CMD ["bundle", "exec", "rails", "server", "-b", "0.0.0.0"]` - this line will run when this container is started and the container will terminate when it's done running. This is a recipe for building a container - docker run or docker-compose are responsible for actually building it.

[.devcontainer/docker-compose.yaml](https://github.com/lobsters/lobsters/pull/1796/files#diff-e96a798889398aad623ee0fa8ef4d316b5a72a5383f4534acbe805619458f263)
Multiple containers need to run and be started in order to stand up a server. In this case - Maria DB and the custom web server container that was built by Dockerfile.dev. Ports, environment variables, network, and local volumes for storing data. Docker-compose is responsible for configuring all of that and running the server.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,2 +1,2 @@
-web: bundle exec rails server
+web: bundle exec rails server -b 0.0.0.0
 job: bin/jobs


Dockerfile.dev
```diff
@@ -5,6 +5,8 @@ FROM ruby:${RUBY_VERSION}
 # Install dependencies
 RUN apt-get update -qq && apt-get install -y \
   build-essential \
+  clang \
+  libclang-dev \
```
pushcx: Why does this need clang? Shouldn't the container be running a Linux environment inside?

.devcontainer/docker-compose.yaml
```diff
@@ -0,0 +1,31 @@
+
+services:
+  app:
+    stdin_open: true
+    tty: true
+    build:
+      context: ..
+      dockerfile: ../Dockerfile.dev
+      args:
+        RUBY_VERSION: 3.4.7
```
pushcx: Can this pull from `.ruby-version`? Or can there be some other way to force the running container to check against that file and give an informative error message if not? If we repeat the version we're never going to keep it in sync.

ChaelCodes: This is the core reason that I needed to create a separate compose file. The only reason the current docker setup works is that y'all are passing the correct ruby-version in the Makefile. If I run `docker compose up`, it grabs whatever RUBY_VERSION was most recently set in my environment variables. (Which did briefly trip me up when I was figuring this out.)

https://github.com/lobsters/lobsters/blob/6ca740c79ff0b0f4763f0c7bb0c0ebd867e5141b/Makefile#L10-L12

I did do some research into this, but I couldn't find any examples of pulling the version from .ruby-version or executing scripts to set arguments or environment variables.

This is really critical. Your base image is named [ruby:3.4.7](https://hub.docker.com/layers/library/ruby/3.4.7). That is a picture of the machine that your commands are running on. I don't have access to `.ruby-version` prior to that because I don't have access to an executable environment before then, so it has to be passed in through the config.

https://github.com/lobsters/lobsters/blob/794aeec66b4daa3911ff0ca905ed5d926a6338b4/Dockerfile.dev#L1-L3

I checked the docker documentation, docker compose documentation, Rails [default devcontainers and docker compose setup](https://github.com/rails/rails/tree/main/railties/lib/rails/generators/rails/devcontainer/templates/devcontainer), [Rails devcontainer documentation](https://guides.rubyonrails.org/getting_started_with_devcontainer.html), and Docker default ruby docker compose example.

All hard-code the ruby-version inside the Dockerfile. If we did this, the `make docker-serve` would be unnecessary, and both manual docker and the devcontainer could use the same docker-compose file.
https://github.com/rails/rails/blob/feca260a238b801010cf0473326603ae37701cd9/railties/lib/rails/generators/rails/devcontainer/templates/devcontainer/Dockerfile.tt#L1-L3

Another option is starting from a pure linux distro instead of a ruby image and installing Ruby and bundler in the docker container. Then we can copy in the .ruby-version file as one of the first commands.

Dockerfile.dev
ChaelCodes: Let me retest this. I'll try it without both, without clang, and without libclang-dev.

The short answer is that when I was first starting your docker environment `bundle install` failed, but I'll get you the exact error messages and double check this.

.devcontainer/docker-compose.yaml
ChaelCodes: I just want to highlight this option, because on second thought, I think you might like it.

> Another option is starting from a pure linux distro instead of a ruby image and installing Ruby and bundler in the docker container. Then we can copy in the .ruby-version file as one of the first commands.

pushcx: Maybe it is worth reiterating that I know very little about docker and can barely use it. I have the high-level goal of wanting it to be easy for developers to contribute and docker seems to have become very common. But I don't know how to do that, so I'm mostly asking questions with an eye to avoiding complexity and maintenance I won't know to do.

So I still have the overall question of what devcontainers adds to the existing docker setup? Or maybe another way to put it is: if you had a blank slate, is this how you'd add docker to the project? Because it's not a lot of code to replace, I have no sentimental attachment to it. If the existing setup didn't work reliably for you, we can bulldoze it to get the right out-of-the-box experience where setting up to develop is painless. (And then let's leave a clear path for the devs who have the existing thing running, though perhaps that's as simple as a command or two to rebuild their container?)

ChaelCodes:
Incomplete Updates:
- [x] Rebase to pull in docker changes
- [x] Merge two docker setups
- [x] Hard-code ruby-version
- [x] Explore ways to surface ruby-version mismatches (GH Action, RSpec, script, error messages -- It should reveal issues before merge) (Would an update to the ruby-version ever happen in a commit directly to main?? Outside a PR?? unlikely)

.gitignore
```diff
@@ -49,3 +49,5 @@ config/*.sphinx.conf
 config/unicorn.conf.rb
 lib/tasks/deploy.rake
 storage/
+
+/config/master.key
```
ChaelCodes, line 53: Ignoring this - it's produced by rails credentials.

Dockerfile.dev
ChaelCodes: Resolved by #1808

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,33 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-docker-compose
+{
+	"name": "Lobsters Dev Container",
```
ChaelCodes, line 4: This file configures our devcontainer - which configures VS Code and tells it how to open itself in Docker.

Gemfile.lock
```diff
@@ -464,6 +469,7 @@ GEM
     zeitwerk (2.7.3)
 
 PLATFORMS
+  aarch64-linux
```
ChaelCodes: We're using the [official Ruby image](https://hub.docker.com/layers/library/ruby/3.4.7) on dockerhub. It's multi-platform, so with some research, I might be able to fix it to a different platform and avoid adding a new platform to Gemfile.lock.

I committed it as-is, because maybe adding a new platform to the lockfile is fine? It doesn't hurt anything.

README.md
```diff
@@ -23,62 +23,6 @@ So our design philosophy is a little different than a typical commercial product
    We're trying to maximize the return on investment of testing rather than minimize errors.
  * We're willing to take downtime for big code changes rather than try to make them seamless.
 
-
-#### Development setup
```
ChaelCodes: Moved this to CONTRIBUTING.

config/database.yml.sample
```diff
@@ -2,7 +2,7 @@
 trilogy: &trilogy
   adapter: trilogy
   encoding: utf8mb4
-  host: 127.0.0.1
+  host: <%= ENV.fetch('DATABASE_HOST') { "127.0.0.1" } %>
```
ChaelCodes, line 5: Instead of manually editing the database.yml, we can now just copy the database.yml.sample. If DATABASE_HOST is set (and it is in Docker) then it uses that host, otherwise, it uses localhost. Local development isn't affected.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,33 @@
...
+	],
+	// The 'service' property is the name of the service for the container that VS Code should
+	// use. Update this value and .devcontainer/docker-compose.yml to the real service name.
+	"service": "app",
+
+	// The optional 'workspaceFolder' property is the path VS Code should open by default when
+	// connected. This is typically a file mount in .devcontainer/docker-compose.yml
+	"workspaceFolder": "/lobsters",
+  // Run `sleep infinity` instead of the default CMD
+  "overrideCommand": true,
```
ChaelCodes, line 16: This is a new to me! In the devcontainer, we can override the command that's running in Docker.

A Docker image ships with a default command - the container will keep running until the command completes.

Most rails docker images use `rails s -b 0.0.0.0` because that makes a lot of sense if your goal is to run the server.
Most devcontainers use `sleep infinity` because that makes a lot of sense if your goal is to run a development environment continuously. (Makes `bundle install` and restarting the server easier.)

This option will replace whatever command is on the docker image with `sleep infinity`! Handy!

ChaelCodes: This means we can start up the environment before it's fully set up and finish setup with access to Ruby and Rails and all of our files!

Makefile
```diff
@@ -8,7 +8,6 @@ lint:
 	bundle exec standardrb --fix-unsafely
 
 docker-serve:
-	export RUBY_VERSION=`cat .ruby-version` && \
```
ChaelCodes: No longer necessary - ruby-version is set in Dockerfile.

bin/setup
```diff
@@ -18,22 +18,22 @@ FileUtils.chdir APP_ROOT do
...
 
   puts "\n== Preparing database =="
-  system! "bin/rails db:setup"
-  system! "bin/rails db:reset" if ARGV.include?("--reset")
+  system! "rails db:setup"
+  system! "rails db:reset" if ARGV.include?("--reset")
 
   puts "\n== Restarting application server if running =="
-  system! "bin/rails restart"
+  system! "rails restart"
```
ChaelCodes, line 36: Docker couldn't find bin/rails, but it can call rails itself.

config/initializers/00_zeitwerk.rb
```diff
@@ -19,6 +19,6 @@
   if Rails.env.production?
     raise config + migrate
   else
-    raise config
+    puts config
```
ChaelCodes: We don't actually _need_ to set the credentials to start the Rails server. All the places we call the credentials can handle an empty value. We need the credentials to test certain features, but not to start the server. However, this here was preventing the Rails server from starting in development unless they were set.

Initially, I needed this because we were using `rails s` to run the container. However, I've now got it set up so we're using `sleep infinity` to run the container.

I can roll this back, but I think it's genuinely helpful.

.gitignore
ChaelCodes, line 53: This comes from Rails credentials.

bin/setup
```diff
@@ -18,22 +18,22 @@ FileUtils.chdir APP_ROOT do
   system("bundle check") || system!("bundle install")
 
   puts "\n== Removing any old logs and tempfiles =="
-  system! "bin/rails log:clear tmp:clear"
+  system! "rails log:clear tmp:clear"
```
ChaelCodes, line 21: Docker wasn't finding bin/rails but it can execute rails.

ChaelCodes:
This is finally ready for review!!

Procfile.dev
```diff
@@ -1,2 +1,2 @@
-web: bundle exec rails server
+web: bundle exec rails server -b 0.0.0.0
```
ChaelCodes, line 1: We need to bind this to 0.0.0.0 so your computer can access the dev server running inside the container.

ChaelCodes:
Rebased and tested with Ruby 4.0.0 - still working great!

- ran tests
- booted and viewed dev server

.devcontainer/docker-compose.yaml
pushcx: The sync issue I mentioned was resolved by #1843

pushcx:
Thanks for working through all my questions here, let me get this merged. :D

PR #1798, opened 2025-11-11
Fix #1797: comment score display

### LINKED ISSUE #1797: Some of the upvoters have text /login
I'm not sure but I think this is a bug. For example:
Image

This is while not logged in. (I don't have an account so I don't know if it matters.)

Thank you for lobste.rs!

toretore:
This fixes #1797 by making `upvoter_score` always return a value. `comment_score_for_user` now returns nil when the score shouldn't be displayed, and `upvoter_score` is now responsible for all formatting and returns "~" in that case.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -13,14 +13,15 @@
   can_flag = @user && @user.can_flag?(comment)
   score = comment_score_for_user(comment, @user)
   draw_tree_line = comment.reply_count > 0
-  collapsed = force_open ||
+  is_open = force_open ||
     @user.try(:is_moderator?) ||
     (comment.score > Comment::COLLAPSE_SCORE && !comment.current_flagged?)
+  collapsed = !is_open
 %>
 
 <input id="comment_folder_<%= comment.short_id %>"
   class="comment_folder_button" type="checkbox"
-  <% unless collapsed %>checked<% end %>
+  <% if collapsed %>checked<% end %>
 >
 <div class="<%= class_names(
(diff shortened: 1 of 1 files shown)


app/views/comments/_comment.html.erb
```diff
@@ -8,39 +8,49 @@
...
-      "></div>
 <div id="c_<%= comment.short_id %>"
   data-shortid="<%= comment.short_id if comment.persisted? %>"
-  class="comment
-    <%= comment.current_upvoted? ? "upvoted" : "" %>
-    <%= comment.current_flagged? ? "flagged" : "" %>
-    <%= comment.current_reply ? "replied" : "" %>
-    <%= comment.score < Comment::SCORE_RANGE_TO_HIDE.first ? "bad" : "" %>
-  ">
+  class="<%= class_names(
```
pushcx, line 34: How does this relate to the bug?

app/views/comments/_comment.html.erb
```diff
@@ -8,39 +8,49 @@
   # is_unread: show (unread) in byline
   # show_folder_control: whether to display the control to hide the comment
 %>
+
```
pushcx, line 11: How does this relate to the bug?

app/views/comments/_comment.html.erb
toretore: It doesn't. I can revert it if you want to keep the changes to what is strictly necessary to fix it.

toretore:
Closing as this has been fixed elsewhere.

PR #1799, opened 2025-11-11
Remove avatar links from tabbing order

fedemp:
Prevent avatars from focusing when tabbing through the page.

Hide the avatar link from screen readers.

Its functionality is duplicated by the link with the author name.


pushcx:
Seems totally reasonable, good catch!

PR #1800, opened 2025-11-12
Fix tags list on Tags page

fedemp:
Related links should use one list element to group them all.

"|" characters hidden from screen readers.

Add aria-label to "edit" links so they read what exactly they would edit.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1573,6 +1573,17 @@ input[type="submit"].link_post.pushover_button {
 	margin-right: 1rem;
 }
 
+/* Tags page */
+.tag-title {
+	display: flex;
+	gap: 0.5rem;
+	align-items: baseline;
+}
+
+.tag-title:first-of-type {
+	margin-top: -0.75rem;
+}
+
 @media only screen and (max-width: 480px) {
 	.radio-group-container legend {
 		float: none;
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -172,11 +172,13 @@ module ApplicationHelper
     super
   end
 
-  def tag_link(tag)
+  def tag_link(tag, options = {})
     link_to tag.tag,
       tag_path(tag),
-      class: [tag.css_class, filtered_tags.include?(tag) ? "filtered" : nil],
-      title: tag.description
+      options.merge({
+        class: [tag.css_class, filtered_tags.include?(tag) ? "filtered" : nil],
+        title: tag.description
+      })
   end
 
   def how_long_ago_label(time)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -175,7 +175,7 @@ module ApplicationHelper
   def tag_link(tag, options = {})
     link_to tag.tag,
       tag_path(tag),
-      options.merge({
+      options.reverse_merge({
         class: [tag.css_class, filtered_tags.include?(tag) ? "filtered" : nil],
         title: tag.description
       })
--- a/app/views/tags/index.html.erb
+++ b/app/views/tags/index.html.erb
@@ -5,33 +5,35 @@
 </p>
 
 <% @categories.each do |category| %>
-  <h2 id="<%= category.category %>">
-    <%= link_to category.category, category_path(category) %>
+  <div class="tag-title">
+    <h2 id="<%= category.category %>">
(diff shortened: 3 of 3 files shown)


app/assets/stylesheets/application.css
```diff
@@ -1518,6 +1518,17 @@ input[type="submit"].link_post.pushover_button {
 	padding: 2px;
 }
 
+/* For the "Tags" page. */
+.tag-title {
+ display: flex;
```
pushcx: Nitpick: we indent css with a single tab.

app/helpers/application_helper.rb
```diff
@@ -176,7 +176,8 @@ def tag_link(tag)
     link_to tag.tag,
       tag_path(tag),
       class: [tag.css_class, filtered_tags.include?(tag) ? "filtered" : nil],
-      title: tag.description
+      title: tag.description,
+      id: "tag-#{tag.tag}"
```
pushcx: This helper is used throughout the site, so adding an anchor here is going to put duplicate anchors to a lot of pages, like any list of stories.

pushcx (changes requested):
Thanks for this. I'm happy to have the ARIA improvements, I don't know a lot about it but it's great to improve things for screen readers. I'd be immediately merging if this was the only change, so if you want to separate a11y improvements from style changes in your PRs, that'd help move things quicker.

Please add a before and after screenshot of the changes you're proposing to the tag list. It's not just useful for reviewing, but for future maintenance when we have to figure out if an odd interaction was deliberate.

app/assets/stylesheets/application.css
fedemp: Fixed, my bad. :).

app/helpers/application_helper.rb
fedemp: Yeah, I noticed that and I kind of expected this comment.

Should I allow the `tag_link` to accept an options object then? Cause for the "Filter tags" page, I need to pass this function an `aria-label` to correctly label filtered tags.

fedemp:
Screenshots of the before and after. Notice that the `edit` link for the category has a normal size cause it's not part of the heading anymore.

**Before**

image

----

**After**
Screenshot From 2025-11-13 20-21-43

app/views/tags/index.html.erb
```diff
@@ -5,33 +5,35 @@
...
+
+  <ol class="category_tags" aria-labelledby="<%= category.category %>">
+    <% category.tags.each do |tag| %>
+      <li>
+        <% filtered = @filtered_tags.include?(tag.id) %>
+        <%= tag_link tag %>
+        <span class="<%= filtered || !tag.active ? 'inactive_tag' : '' %>"><%= tag.description %></span>
+        <span class="byline">
+          <% if !tag.active %>
+            <span aria-hidden="true">|</span> <em>inactive</em>
```
pushcx: I don't want to add a bunch of work to this PR, but if you wanted to create separate PR with a `divider_tag` helper for this, we could use it in a whole bunch of places across the templates.

app/helpers/application_helper.rb
pushcx: Yeah, taking an `options` arg like `link_post` and others would be great.

app/views/tags/index.html.erb
pushcx: I merged down #1806, so it's available for use here.

fedemp: Great. I used it here. :+1:

app/helpers/application_helper.rb
fedemp: Done. :+1:

pushcx:
Oops, that build failure is my fault. I let some partial work slip into `main`. I'll try to solve those in the morning.

app/helpers/application_helper.rb
```diff
@@ -172,11 +172,13 @@ def tag name = nil, options = nil, open = true, escape = true
     super
   end
 
-  def tag_link(tag)
+  def tag_link(tag, options = {})
     link_to tag.tag,
       tag_path(tag),
-      class: [tag.css_class, filtered_tags.include?(tag) ? "filtered" : nil],
-      title: tag.description
+      options.merge({
```
pushcx:
```suggestion
      options.reverse_merge({
```

Otherwise the caller can't override these two options.

pushcx:
I got `main` back to green, so I'm happy to get this merged.

Thanks for improving this page!

PR #1805, opened 2025-11-19
Backfill moderator data

luigieli:
Part of #1714 

This PR is a follow-up for https://github.com/lobsters/lobsters/pull/1773 based on [this suggestion in the issue](https://github.com/lobsters/lobsters/issues/1714#issuecomment-3416449239). It uses data from the `hats` table to determine if a story was created while the user was a moderator.

After the changes made [here](https://github.com/lobsters/lobsters/issues/1234), all images from stories that were allowed to have images were transformed into links:
image

We use the date to ensure the image renders correctly if the poster was a moderator when the story was created.

|Before rerendering|After rerendering|
|---|---|
|image|image|

I also added new tests to ensure correct behavior for the feature. Since the logic checking if a user was a moderator to allow images was altered, some tests had to be updated. I added `create(:hat, user: moderator, modlog_use: true)` to these tests to ensure the hat exists.

However, would that logic fit in a better place? Like on [`granted_moderatorship_by_user!`](https://github.com/lobsters/lobsters/blob/06418c0328ff400ddc82d8e3aff23f353c181c08/app/models/user.rb#L431)?

### DIFF
--- a/spec/features/submit_story_spec.rb
+++ b/spec/features/submit_story_spec.rb
@@ -64,7 +64,10 @@ RSpec.feature "Submitting Stories", type: :feature do
     end
 
     context "as a user who is a moderator" do
-      before { user.update(is_moderator: true) }
+      before {
+        user.update(is_moderator: true)
+        create(:hat, user: user, modlog_use: true)
+      }
 
       scenario "results in an image, not a link" do
         expect {
--- a/spec/requests/stories_spec.rb
+++ b/spec/requests/stories_spec.rb
@@ -276,6 +276,7 @@ describe "stories", type: :request do
     context "can have images" do
       it "puts https CSP header" do
         moderator = create(:user, :moderator)
+        create(:hat, user: moderator, modlog_use: true)
 
         story = create(:story, user: moderator, title: "test", url: "http://example.com/")
         expect(story.can_have_images?).to eq(true)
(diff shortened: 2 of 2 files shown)


pushcx:
If we add a field to the User model, someone can't stop and restart being a moderator. That hasn't happened to us yet, but seems inevitable. We do have a great place to stick this data: their moderator hat. The existing `modlog_use` boolean tracks whether this is a mod hat, and its `created_at` and `doffed_at` values are the data we want. (I doublechecked our current former mods and the data is correct.)

It's good that you recognized the need for the rerender method to change, but I'm puzzled by the use of `posted_as_moderator?`. There's already `Story#can_have_images?` method used to determine this, and I get nervous when I see two access control methods that are so similar. Please take a look at them both to figure out the right name and implementation for this functionality and unify it.

luigieli:
Thanks for the reply.
> If we add a field to the User model, someone can't stop and restart being a moderator. That hasn't happened to us yet, but seems inevitable. We do have a great place to stick this data: their moderator hat. The existing `modlog_use` boolean tracks whether this is a mod hat, and its `created_at` and `doffed_at` values are the data we want. (I doublechecked our current former mods and the data is correct.)

Thanks for the hint for these hats data, I'll definitely take a look at it then.
> It's good that you recognized the need for the rerender method to change, but I'm puzzled by the use of `posted_as_moderator?`. There's already `Story#can_have_images?` method used to determine this, and I get nervous when I see two access control methods that are so similar. Please take a look at them both to figure out the right name and implementation for this functionality and unify it.

I'll refactor this logic then and make it cleaner.

As soon as I address those points, I'll push it here and ping you.

luigieli:
Hey @pushcx, I've rebased the changes in this PR and updated the description. Is that okay? Or would you prefer I close this one and open a new PR?

pushcx:
This is great! Thank you for working through these revisions, this landed in a short, clear implementation with good tests.

As it's short, I'll squash it down.

luigieli:
Thanks @pushcx for reviewing and merging it. I really appreciate it and look forward to contributing again.

PR #1806, opened 2025-11-19
Use divider_tag function to create a separator

fedemp:
Create a function that renders a `|` that's hidden to screen readers.


pushcx:
Context: https://github.com/lobsters/lobsters/pull/1800#discussion_r2535475339

Thanks for this improvement, nice to have this little helper.

While I'm reticent to change core UI around reading, voting, and commenting because it's always annoying to have changes there, this seems like a solid improvement that's worth it.

PR #1808, opened 2025-11-20
Fix bundle install error for development

taisei-86:
## Error description

When "make docker-serve" was executed
```
87.20 thread 'main' (11522) panicked at
87.20 /usr/local/bundle/gems/commonmarker-2.5.0/ext/commonmarker/.rb-sys/stable/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bindgen-0.69.5/lib.rs:622:31:
87.20 Unable to find libclang: "couldn't find any valid shared libraries matching:
87.20 ['libclang.so', 'libclang-*.so', 'libclang.so.*', 'libclang-*.so.*'], set the
87.20 `LIBCLANG_PATH` environment variable to a path where one of these files can be
87.20 found (invalid: [])"
87.20   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
87.20 warning: build failed, waiting for other jobs to finish...
87.20 make: *** [Makefile:571: target/release/libcommonmarker.so] Error 101
```
## Possible causes

https://github.com/gjtorikian/commonmarker/issues/268

## Correction details

Added the Clang libraries installation to the Dockerfile.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -8,6 +8,7 @@ RUN apt-get update -qq && apt-get install -y \
   libjemalloc2 \
   libvips \
   mariadb-client \
+  clang \
   vim
 
 # Set working directory


taisei-86:
Please take a look when you are free.

pushcx:
I think @ChaelCodes hit this same dependency issue in #1796. She found she also had to add `libclang-dev`, was that not a problem for you?

This is probably going to be a recurring issue. We follow new Ruby versions pretty closely and this transitive dependency seems to prefer setting a conservative upper bound on Ruby versions.

ChaelCodes:
Hey @taisei-86 ! Thanks for catching this and fixing it! I needed to go through and check which lib was actually needed and grab the error message, but you already did it, so now I don't have to! 🎉

ChaelCodes:
@pushcx This is exactly the issue I ran into and I recognize the failing package is the same.

pushcx:
@ChaelCodes Thanks for taking a look. Sorry to give you a minor merge conflict, but I'm going to go ahead and merge this so we have this PR info linked from the vcs history.

@taisei-86 Thanks for running down this issue and figuring out what's happening upstream!

PR #1811, opened 2025-11-28
Fix #1660: Show comment tree lines between siblings using CSS

### LINKED ISSUE #1660: Comment tree lines are incorrect
Comments should have thin vertical lines connecting siblings to make it easier to follow conversations. Comments with no siblings shouldn't.

The current display is a known regression I accepted to merge some other comment layout fixes in b2965a6da73fbdc6785bf56cdb2adc7c7aba8a2d. It got cleaned up a little in #1639.

The best way to fix this would be to look back in archive.org/git at how comment tree lines were laid out prior to April 15 and restore that.

toretore:
This replaces and fixes the way the dotted lines in a comment tree are rendered from using explicit HTML elements to adding the lines using a CSS `::before` pseudo-element. The lines are added to `.comments_subtree` elements that have a sibling following it using a combination of `:has` and the `+` sibling selector.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -648,7 +648,7 @@ li.comments_subtree {
 	position: relative;
 }
 
-/*  Render a dotted line between second-level siblings
+/*  Render a dotted line between second+ level siblings
  *  This works by finding the top-level .comments_subtree, then starting 2
  *  levels below that. There is an intermediate level rendered by the views
  *  that render the comments/_threads partial that has to be accounted for.


pushcx:
One small fix: this should draw lines between top-level replies. So in [your image](https://github.com/lobsters/lobsters/issues/1660#issuecomment-3584329663) I think both the comment from "test" reading "Level 1 Comment" and partially-visible comment from "vandervort_mickey" are top-level replies (`parent_comment_id` is nil). They should not have a line between them.

Off the top of my head you could prepend `li.comment > ` to start from the second level?

toretore:
Added another commit that changes this. It was a little tricky to figure out how to reliably detect which elements were replies, but I think I got it.

Screenshot from 2025-12-02 10-57-08

pushcx:
Thanks for working through this. Though this ended up being a pretty complicated selector, it drops a lot of complexity from our god-object template for comments that has a heck of a lot of complexity. (I considered the possibility of introducing an attribute or class to indicate that something was a second+ level reply, but that would get complicated with `~user/threads` showing subtrees.)

It's especially nice that this rolls up the verrrrrry subtle distinction between the parent tree lines and sibling tree lines into one concept.

Issue #547, opened 2018-09-12
External Profile Image support?


Jaruzel:
Hiya, as a new Lobste.rs user, I notice that I can only use Gravatar for Avatars...

Any chance of an extra field in users profiles that allows them to enter an external https link to a png/jpg image of their choice?

I'd look into a pull request for this, but I'm just not that good at coding, sorry.

Thanks!

pushcx:
This has previously come up in #232. Since then, the site has added a caching proxy to avoid the privacy problem of linking directly to Gravatar. To preserve that privacy (which came up again [during our april fool's gag this year](https://lobste.rs/s/qfbwqz/lobste_rs_visitor_statistics)), we'd have to cache those images, so it may as well just be an upload form. The storage use is no longer a significant concern: in production right now that's 303 MiB.

The big thing that I'm glad is Gravatar's problem is that they filter and resize images when they manage the uploads, which are three security-sensitive tasks. Filtering is important because they'd be files living right alongside the Ruby source of the site and, if Rails's (very eager) loading system could be tricked into loading one. Resizing is important because ImageMagick and similar tools were not designed with possible-hostile input in mind and have had a number of vulnerabilities that allow an attacker's image file to execute code. And uploading has a bunch of corner cases like someone uploading a 10 GiB file; when I last looked at this a few years ago Rails did not support for recognizing and canceling over-large uploads.

I'm open to the idea of removing this third-party dependency (preferably keeping it as an option for the sites like barnacl.es that are on Heroku), but we'd need a rock-solid answer to the security concerns. We'd also need to generate our own placeholder images, but that's minor, maybe some volunteer might take it as a be a fun weekend project. There may be other issues I'm not thinking of off the top of my head, though.

pushcx:
Survey of placeholder images: https://barro.github.io/2018/02/avatars-identicons-and-hash-visualization/

pushcx:
Been a few years. Is there now an image processing tool that someone considers rock solid in the face of untrusted input?

RGBCube:
> Been a few years. Is there now an image processing tool that someone considers rock solid in the face of untrusted input?

How could a PNG or JPG be untrusted? They don't contain anything other than binary data to render as an image.

You could verify if something is a png or a jpg by checking the binary header in the file as well

PhilArmstrong:
> > Been a few years. Is there now an image processing tool that someone considers rock solid in the face of untrusted input?
> 
[...]

PR #1812, opened 2025-11-28
Add story image preview

luigieli:
Part of #1525 

This PR is a follow-up for https://github.com/lobsters/lobsters/issues/1525 based on the [suggestion in the issue](https://github.com/lobsters/lobsters/issues/1525#issue-2999750092):

>One improvement: the image for our stories is always the Lobsters logo (or missing). After discussion in https://github.com/lobsters/lobsters/issues/547 I'm comfortable introducing libvips. Let's extract the image for the story we're linking to and overlay the Lobsters logo in the lower-left corner. We'll have more variety in our feeds and hopefully a clear visual distinction to people who see both a direct link to a post and our link to our story of that post.

Using `libvips`, this PR adds a Lobsters logo overlay to the image extracted from the linked story by creating an async job to follow the same logic as avatar handling. Note that I am not updating the deployment scripts to [avatar caching](https://github.com/lobsters/lobsters/blob/14bc8599c91f6d3acdd1fd4ce7fca5b99fcc64bd/hatchbox/pre-build#L7-L8) caching in this PR.

The [opengraph.dev](https://opengraph.dev/) was used to make the comparison.
|Before|After|
|---|---
|image|image|

### DIFF
--- a/spec/controllers/story_image_controller_spec.rb
+++ b/spec/controllers/story_image_controller_spec.rb
@@ -1,0 +1,47 @@
+require "rails_helper"
+
+RSpec.describe StoryImageController, type: :controller do
+  describe "GET show" do
+    let(:story) { create(:story) }
+    let(:cache_dir) { Rails.public_path.join("story_image/") }
+    let(:cached_image) { cache_dir.join("#{story.short_id}.png") }
+
+    before do
+      FileUtils.mkdir_p(cache_dir)
+    end
+
+    after do
+      FileUtils.rm_f(cached_image)
+    end
+
+    context "when the image exists" do
+      before do
+        FileUtils.touch(cached_image)
+      end
+
+      it "returns the image inline" do
+        get :show, params: {short_id: story.short_id, format: :png}
+        expect(response).to have_http_status(:ok)
+        expect(response.headers["Content-Type"]).to eq("image/png")
+        expect(response.headers["Content-Disposition"]).to include("inline")
+      end
+    end
+
+    context "when the image does not exist" do
+      let(:logo_path) { Rails.public_path.join("touch-icon-144.png") }
+
+      it "does not enqueue CreateStoryCardJob" do
+        expect {
+          get :show, params: {short_id: story.short_id, format: :png}
+        }.not_to have_enqueued_job(CreateStoryCardJob)
+      end
+
+      it "serves the fallback logo" do
+        get :show, params: {short_id: story.short_id, format: :png}
+        expect(response).to have_http_status(:ok)
+        expect(response.headers["Content-Type"]).to eq("image/png")
+        expect(response.body.bytesize).to eq(File.size(logo_path))
+      end
+    end
+  end
+end
--- a/spec/jobs/create_story_card_job_spec.rb
+++ b/spec/jobs/create_story_card_job_spec.rb
@@ -1,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe CreateStoryCardJob, type: :job do
+  describe "#perform" do
+    let(:story) { create(:story, url: "http://example.com/story") }
+    let(:job) { CreateStoryCardJob.new }
+    let(:story_image) { instance_double(StoryImage) }
+
+    before do
+      allow(StoryImage).to receive(:new).with(story).and_return(story_image)
+    end
+
+    it "calls generate on StoryImage" do
+      expect(story_image).to receive(:generate).with(story.url)
+      job.perform(story)
+    end
+  end
+end
--- a/spec/models/story_image_spec.rb
+++ b/spec/models/story_image_spec.rb
@@ -1,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe StoryImage, type: :model do
+  describe "#generate" do
+    let(:story) { create(:story, url: "http://example.com/story") }
+    let(:story_image) { StoryImage.new(story) }
+
+    it "handles download failures" do
+      allow(story_image).to receive(:fetch_url).with(story.url).and_return(nil)
+
+      expect {
+        story_image.generate(story.url)
+      }.not_to change { story_image.exists? }
+    end
+
+    it "skips large images" do
+      html_response = double("Net::HTTPSuccess", body: '<html><meta property="og:image" content="http://example.com/huge.png"></html>')
+      allow(html_response).to receive(:[]).with("content-type").and_return("text/html")
+      allow(story_image).to receive(:fetch_url).with(story.url).and_return(html_response)
+
+      large_body = "0" * (8.megabytes + 1)
+      image_response = double("Net::HTTPSuccess", body: large_body)
+      allow(story_image).to receive(:fetch_url).with("http://example.com/huge.png").and_return(image_response)
+
+      expect {
+        story_image.generate(story.url)
+      }.not_to change { story_image.exists? }
+    end
+  end
+end


luigieli:
I'm going through these CI failures to see what I can find out.

pushcx:
Thanks for taking this on, it'll certainly make our bot feeds more interesting. And it's great to introduce libvips for long-desired features like #547 and #986.

I can see how the design of `Story.fetched_attributes` led you to put this in the Story model, but I don't want to extend the pattern of having a model making http requests. I'd rather this work by having the `StoryController#create` action enqueue a new `CreateStoryCard` job. Move this this code in there. Maybe that's even two jobs, one to fetch the existing attributes and image. Then, if it sees a value for the image, it can enqueue a second job to fetch their image and save it with our logo added.

A small design thing, please call this the 'card_image_url' rather than just the 'image_url'. I know OpenGraph has a more generic name, but I still hear everyone call it the 'card image' and I want the variable names to have more than just the type in there.

An even smaller design note, please make the Lobsters logo much smaller, like half the width and height it is now.

For CI, the brakeman failure is it warning about the potential to use user-supplied data in filenames. The risk there is that a malicious user could set `params[:short_id]` so the final path is `/home/deploy/lobsters/current/public/story_image/../../app/initializers/foo.rb` and then they get to upload code that'll be run the next time the server starts. So we always want to be very careful about where we get strings that can wind up in filenames. Run `brakeman -I` when you're sure it's secure and it'll walk you through the process of ignoring the warning and leaving a note to explain why the code is safe.

The test failure about libvips - I'm guessing that doesn't happen to your local dev? It's probably something in the [CI action](https://github.com/lobsters/lobsters/blob/main/.github/workflows/check.yml) that needs to include libvips in the container that runs the tests.

You're really working up and down every layer of our stack with this! No wonder it's gotten a little complicated. You're off to a great start on this feature.

luigieli:
Thanks for the reply.

> I can see how the design of `Story.fetched_attributes` led you to put this in the Story model, but I don't want to extend the pattern of having a model making http requests. I'd rather this work by having the `StoryController#create` action enqueue a new `CreateStoryCard` job. Move this this code in there. Maybe that's even two jobs, one to fetch the existing attributes and image. Then, if it sees a value for the image, it can enqueue a second job to fetch their image and save it with our logo added.

I see. I'll make fetching the image an async job then.

> A small design thing, please call this the 'card_image_url' rather than just the 'image_url'. I know OpenGraph has a more generic name, but I still hear everyone call it the 'card image' and I want the variable names to have more than just the type in there.

Sure, descriptive variable names are always desirable.

> An even smaller design note, please make the Lobsters logo much smaller, like half the width and height it is now.

Ok.

> For CI, the brakeman failure is it warning about the potential to use user-supplied data in filenames. The risk there is that a malicious user could set `params[:short_id]` so the final path is `/home/deploy/lobsters/current/public/story_image/../../app/initializers/foo.rb` and then they get to upload code that'll be run the next time the server starts. So we always want to be very careful about where we get strings that can wind up in filenames. Run `brakeman -I` when you're sure it's secure and it'll walk you through the process of ignoring the warning and leaving a note to explain why the code is safe.

Oh, got it. In this case, I'll try to make the filename handling as secure as possible before ignoring the warning.

> The test failure about libvips - I'm guessing that doesn't happen to your local dev? It's probably something in the [CI action](https://github.com/lobsters/lobsters/blob/main/.github/workflows/check.yml) that needs to include libvips in the container that runs the tests.

Yup, I use Docker to run locally. Checking the Dockerfile, [libvips](https://github.com/lobsters/lobsters/blob/f6f4d7ee873d6482af05b6c018787ebc482ae7ca/Dockerfile.dev#L9) is installed there. Probably just installing it in the CI action would solve this.

luigieli:
Hey @pushcx 
I've implemented the requested changes:
- Moved image fetching and logo overlay logic to an asynchronous job.
- Renamed `image_url` to `image_card_url`.
- Sanitized `params[:short_id]` and documented the exception in `brakeman.ignore`.
- Installed `libvips` in the GitHub workflow.
- Scaled the Lobsters logo to a maximum of 1/5 of the original image height.
- Updated the PR description.

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+    story_card_image = generated_card_image(card_image_url)
+
+    if story_card_image.present?
+      FileUtils.mkdir_p(CACHE_DIR)
+
+      File.open("#{CACHE_DIR}/.#{File.basename(story.short_id)}.png", "wb+") do |f|
+        f.write story_card_image
+      end
+
+      File.rename("#{CACHE_DIR}/.#{File.basename(story.short_id)}.png", "#{CACHE_DIR}/#{File.basename(story.short_id)}.png")
```
pushcx: This seems like a no-op?

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
+class CreateStoryCardJob < ApplicationJob
+  queue_as :default
+  limits_concurrency to: 1, key: ->(story) { story.short_id }, duration: 5.minutes
+
+  CACHE_DIR = Rails.public_path.join("story_image/").to_s.freeze
```
pushcx:
```suggestion
  CACHE_DIR = StoryImageController::CACHE_DIR
```

But there is probably enough logic + file handling to be worth wrapping up a `app/models/story_image.rb`:

```ruby
class StoryImage
  CACHE_DIR = Rails.public_path.join("story_image").to_s.freeze
  
  def initialize(short_id)
    @story_id = short_id.is_a?(Story) ? short_id.short_id : short_id
  end
  
  def exists?
    File.exists? path
  end
  
  def path
    File.join(CACHE_DIR, short_id, '.png')
  end

  def generate url
  # ...fill cache...
  end
end
```

app/controllers/story_image_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    s = Story.where(short_id: short_id).first!
+
+    cached_image = File.join(CACHE_DIR, "#{File.basename(s.short_id)}.png")
+
+    if File.exist?(cached_image)
+      response.headers["Expires"] = 1.hour.from_now.httpdate
+      send_file cached_image, type: "image/png", disposition: "inline"
+    else
+      CreateStoryCardJob.perform_later(s)
```
pushcx:
```suggestion
```

I understand the goal here, that you want to fill the cache, but it looks like the job doesn't generate an image if the story doesn't have one (or it's too large, etc).

So then every time someone tries to load the image from us we'd enqueue a job that's going to hit the story URL again, seems like a bad behavior for a bot. We should trust that your PR queues the job every time a story URL is edited, and we can fix any missing paths if a prod bug becomes apparent.

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  private
+
+  def find_card_image_url(url)
+    response = fetch_url(url)
+    return nil unless response
+
+    # If it's a PDF or too large, skip
+    return nil if response.body.length >= 5.megabytes
+    return nil if response["content-type"].to_s.match?(/pdf/)
```
pushcx: Did you catch someone doing this in prod? What a funny bit of real-world complexity.

Maybe this check should flip to an allowlist of png/jpg/svg? Maybe if we're very lucky libvips exposes a list of formats/mime types it supports?

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    converted = response.body.force_encoding("utf-8")
+    parsed = Nokogiri::HTML(converted.to_s)
+
+    # Try <meta property="og:image">
+    card_image_url = parsed.at_css("meta[property='og:image']")&.attributes&.[]("content")&.text
+
+    # Then try <meta name="image">
```
pushcx: I appreciate that you thought to check a few places, the `&.` for handling the complexities of often-invalid pages, and the prudent `rescue` at the end. Good design!

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+    Rails.logger.error "Vips error while generating image: #{e.message}"
+    nil
+  end
+
+  def fetch_url(url)
+    s = Sponge.new
+    s.timeout = 3
+    # User submitted URLs may have an incorrect https certificate, but we
+    # don't want to fail the retrieval for this. Security risk is minimal.
+    s.ssl_verify = false
```
pushcx:
```suggestion
```

I'd rather fail safe. Over the years I have seen a few devs who have Strong Opinions about TLS certs and done very strange things. I admit I can't articulate a scenario where this would affect generating a story image, but I'm skittish around security controls.

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+    headers = {
+      "User-agent" => "#{Rails.application.domain} for job",
```
pushcx:
```suggestion
```

Oof, you revealed an ugly old bug here. We should be sending a better `User-agent` header. I'll fix this on `main` now.

app/jobs/create_story_card_job.rb
```diff
@@ -0,0 +1,98 @@
...
+      "Referer" => Rails.application.domain
```
pushcx:
```suggestion
      "Referer" => Routes.story_short_id_url(@short_id)
```

Assumes this method is moved into the StoryImage class explained in another comment.

(God bless the HTTP typo that will never die.)

config/brakeman.ignore
```diff
@@ -212,7 +212,7 @@
       "check_name": "CrossSiteScripting",
       "message": "Unescaped model attribute",
       "file": "app/views/stats/index.html.erb",
-      "line": 22,
+      "line": 28,
```
pushcx, line 215: When brakeman is touched it updates these line numbers that will slowly drift as other edits get made. I just bumped this on `main` for unrelated work so you are likely to get spurious merge conflicts about this file. You can drop every edit except your new fingerprint.

.github/workflows/check.yml
```diff
@@ -36,6 +36,11 @@ jobs:
         with:
           bundler-cache: true
 
+      - name: Install libvips dependency
+        run: |
+          sudo apt update
+          sudo apt install -y libvips
```
pushcx, line 44: I bet the `Dockerfile` needs this package added as well.

luigieli: `Libvips` is already in the [`Dockerfile`](https://github.com/lobsters/lobsters/blob/7f087cc775ca554927af09779ace36e7a28f208d/Dockerfile.dev#L8).

app/jobs/create_story_card_job.rb
luigieli: In the official [docs](https://www.libvips.org/) we can find this list. I didn't add all of them to the allowlist, just the ones that I judge to be the most used and would not cause any headache. However, I'll happily add any image format to it if needed.

luigieli:
Hey @pushcx, thanks for the feedback. I've addressed what you pointed out:
- Added a `StoryImage` model to handle the logic of fetching and caching, the job and controller just delegate to the model.
- Enabled `SSL` verification and fixed the `Referer` header.
- Added a `content-type` allowlist based on [libvips docs](https://www.libvips.org/).
- Removed the redundant image fetching, it is only fetched when the story is created or the URL is edited.

app/models/story_image.rb
```diff
@@ -0,0 +1,120 @@
...
+  rescue => e
+    Rails.logger.error "Error finding image URL for story: #{e.message}"
+    nil
+  end
+
+  def generated_card_image(url)
+    res = fetch_url(url)
+    return nil if res.blank?
+
+    # Limit the image size according to Meta developer docs
```
pushcx, line 83: Great comment, thanks for explaining.

pushcx:
This looks great! Thanks for working through my refactorings and nitpickings, this is very well-organized and will make our social promotion a lot nicer.

luigieli:
> This looks great! Thanks for working through my refactorings and nitpickings, this is very well-organized and will make our social promotion a lot nicer.

Thanks @pushcx. I appreciate your patience in guiding me through the implementation. I'm glad I could help improve this project even more.

pushcx:
Looks like this has a small bug in prod and some of the embedded logos are too low. Maybe the positioning and resize logic have slipped out of sync?

https://mastodon.social/@lobsters/115928039967254886
https://mastodon.social/@lobsters/115927253470344851

luigieli:
> Looks like this has a small bug in prod and some of the embedded logos are too low. Maybe the positioning and resize logic have slipped out of sync?
> 
> https://mastodon.social/@lobsters/115928039967254886 https://mastodon.social/@lobsters/115927253470344851

That's odd. It seems like the resizing logic is interfering with the positioning. I'll investigate and post an update.

luigieli:
I tried opening the image URLs directly [here](https://lobste.rs/story_image/hrsaz6.png) and [here](https://lobste.rs/story_image/o08dqx.png).
|lobste.rs/s/o08dqx|lobste.rs/s/hrsaz6|
|---|---|
|image|image|

The Lobsters logo is fully visible on both images. Maybe Mastodon has a max card size or something and it's cropping bigger images. If so, we can resize the final image to prevent this from happening. I'll check on that.

pushcx:
Aha. And it's clearly cropping to the center of the image (perhaps even with clever 'center of interest' logic?) so we can't just put the logo in the upper-left and be done with it. Dang.

And of course other Fediverse clients might crop differently. Maybe we can add metadata to indicate the preferred cropping? I seem to recall that's in the UI for posting an image to Mastodon.

Whatever you find, please don't lose too much time to it. In the priority of things, this is a pretty minor visual glitch.

luigieli:
> Aha. And it's clearly cropping to the center of the image (perhaps even with clever 'center of interest' logic?) so we can't just put the logo in the upper-left and be done with it. Dang.
> 
> And of course other Fediverse clients might crop differently. Maybe we can add metadata to indicate the preferred cropping? I seem to recall that's in the UI for posting an image to Mastodon.
> 
> Whatever you find, please don't lose too much time to it. In the priority of things, this is a pretty minor visual glitch.

Hey @pushcx, I took a look at the Mastodon [preview card docs](https://docs.joinmastodon.org/entities/PreviewCard/), [media attachment docs](https://docs.joinmastodon.org/entities/MediaAttachment/#image), and this [Mastodon post](https://mastodon.online/@mastodonmigration/109440513681868009) about image cropping. As far as I could understand, focus point selection is a feature for media only, we cannot set it for preview card images.

One solution I thought of is to crop/resize the image ourselves based on the [Meta recommended size](https://developers.facebook.com/docs/sharing/webmasters/images/) to try to mitigate these problems. However, I suppose it will keep happening due to how different social media clients handle these images. Do you think it's worth it?

pushcx:
Thanks for looking into it. I think at this point the only thing to do would be to document it so that someone who looks into this in the future has a running start. Could you PR a comment to where the logo is positioned explaining "If it looks like the logo is mispositioned..." with the three Mastodon links you found? Given the unreliability, I don't think future work here is warranted for us.

If you (or a future reader) wants to get involved with Mastodon, it might be a nice project to try to add positioning metadata for preview card images similar to the media attachments.

luigieli:
> Thanks for looking into it. I think at this point the only thing to do would be to document it so that someone who looks into this in the future has a running start. Could you PR a comment to where the logo is positioned explaining "If it looks like the logo is mispositioned..." with the three Mastodon links you found? Given the unreliability, I don't think future work here is warranted for us.

[Done](https://github.com/lobsters/lobsters/pull/1887)!

PR #1815, opened 2025-12-06
Full opacity on Hidden section

fedemp:
When a story is hidden in the front page, reduce its opacity to 0.25.

When checking hidden stories in the "Hidden" section, restore its opacity so user can know what was hidden.

Show full opacity for a hidden story in its single page; if user landed there, let them see what the story was even if we show the "you have hidden this story" warning.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -730,14 +730,16 @@ li .byline a.story_has_suggestions {
 	margin: 1em 0 1em 32px; /* matching div.detail */
 }
 
-li.story.hidden *:not(div, .dropdown_parent, .dropdown_parent > div *, .comments_label, .link, .tags, img),
-li.story.hidden .score {
+:is(ol:not(.show_hidden) li.story.hidden:not(:only-child)) :is(
+	.dropdown_parent > .flagger, /* Flag button */
+	.dropdown_parent > label, /* Caches button */
+	.byline > :not(.dropdown_parent),
+	.details > :not(.byline),
+	.voters
+) {
 	opacity: 0.25;
 }
 
-ol.show_hidden li.story.hidden {
-	opacity: 1.0 !important;
-}
 li.story.saved a.saver {
 	color: var(--color-fg-affirmative);
 }


fedemp:
**Hidden story in the front page. No visual changes.**



---

**Current "Hidden" section**

image


**Proposed "Hidden" section**

image

---

**Current single post page**
image

**Proposed single post page**

image

app/assets/stylesheets/application.css
```diff
@@ -737,14 +737,16 @@ li .byline a.story_has_suggestions {
...
 }
 
-ol.show_hidden li.story.hidden {
-	opacity: 1.0 !important;
+li.story.hidden:only-child,
+ol.show_hidden li.story {
+	opacity: 1.0;
 }
+
+
```
fedemp: I assure you there are no two consecutive lines in my PR. Not sure why GH shows it this way.

pushcx:
Thanks for noticing this bug, I can see the hidden stories list isn't very useful as-is. Could you check on how the 'caches' and 'flag' menus appear on a hidden story? I'm pretty sure all the complexity in the old CSS is trying to make sure that those have normal visibility and usability. So I think this would reintroduce #1206 which has a note from me griping that this is an easy bug to reintroduce. :)

The build failure is me accidentally breaking main; if you rebase or merge main in, it'll disappear.

fedemp:
Hello, again.

You are right about the dropdowns. I have now fixed them.

Screenshot From 2025-12-31 00-10-25
Screenshot From 2025-12-31 00-10-04

pushcx:
Thanks for this thorough fix! Makes the hidden stories list a lot more usable.

Now that you know this code, could you think a sec about ways to improve this CSS to not include a stacking context to make it not so easy to keep reintroducing this bug? No worries if not, but we've been annoyed at it for over a decade now so I figured I should ask for fresh eyes on it. :)

fedemp:
I see your concern. Ideally we should just do a `.story.hidden {opacity: 0.25}` and call it a day, but unfortunately the dropdown, being inside the same tree of the story, seems to make this impossible.

Issue #1814, opened 2025-12-05
Very slow page loads recently


mk12:
Lobste.rs has always loaded instantly for me in the past, but in the past couple weeks or so performance has fallen off a cliff. Every page load takes multiple seconds. I’ve tried different browsers, networks, devices, logging out, it makes no difference. For example I just visited the homepage and it took 3 seconds, clicked on the Comments link at the top and it took 5 seconds. I’m in the NYC area on a fiber connection and not seeing problems with other sites. Let me know if there is any other information I can provide.

pushcx:
We're getting swamped by aggressive scrapers that use residential proxies to avoid rate-limiting. They spread their traffic out to hundreds or thousands of IPs so each only requests one page every ~10 minutes, but overall traffic is ~6-10x normal. When they finish a run we drop back into our normal snappy self.

Our full-page caching is still active and working. It's possible they're accidentally bypassing it by sending cookies, I'll have to add some logging to try to diagnose that.

They ignore robots.txt, of course.

#1781 is planned to be a bigger speedbump to crawlers. I hadn't wanted to have it start interdicting all traffic, but might have to with this kind of distributed load.

@technomancy suggested [iocaine](https://iocaine.madhouse-project.org/documentation/) to isolate the bots. It installs with a curlpipe which really doesn't fit our ubuntu server + hatchbox provisioning, so I don't see a clear path to installing it. It has a complicated configuration and scripting setup that's not clearly documented, but if someone wants to PR that and pinky-swear to maintain it, it seems worth a shot.

I'll keep this issue open in case anyone wants to offer their performance expertise. We've done the obvious stuff like db indexes, fixing 1 + ns (as much as Rails tries to reintroduce them), full-page and fragment caching, etc etc.

mk12:
Sad to hear you have to deal with that 😢 Thanks for quick explanation!

technomancy:
> It installs with a curlpipe which really doesn't fit our ubuntu server + hatchbox provisioning, so I don't see a clear path to installing it

The documentation is admittedly a little scattered, but you don't need curlpipe to install iocaine. I used this to install using apt: https://iocaine.madhouse-project.org/documentation/3/getting-started/debian/

Beyond that the only configuration I've needed to use is in the [`Caddyfile`](https://iocaine.madhouse-project.org/documentation/3/getting-started/#caddy-configuration):

```
blog.example.com {
  @read method GET HEAD
  reverse_proxy @read 127.0.0.1:42069 {
[...]

PR #1818, opened 2025-12-07
Apply iocaine as a layer before Puma.

hmadison:
Contextualized by: #1814

This, in theory, should be compatible with the rest of the Hatchbox generated configuration. Unfortunately, this is a good bit of "dictated but not read" due to the testing methodology with Hatchbox.

### DIFF
--- a/hatchbox/Caddyfile.pre
+++ b/hatchbox/Caddyfile.pre
@@ -63,3 +63,21 @@ rewrite @visitor /cache/{file_match.relative}
 
 # add CSP header when serving from cache; Rails adds when it renders pages
 header @visitor ?Content-Security-Policy "default-src 'none'; img-src 'self' data:; style-src 'self' 'unsafe-inline'"
+
+# Send un-cached requests for anonymous users through iocane.
+@iocaine_target {
+  host lobste.rs
+
+  not {
+    header_regexp Cookie `.*=.*`
+  }
+
+  not {
+    path cache/*
+  }
+}
+
+reverse_proxy @iocaine_target 127.0.0.1:42069 {
+  @fallback status 421
+  handle_response @fallback
+}
--- a/hatchbox/root-deploy
+++ b/hatchbox/root-deploy
@@ -27,8 +27,15 @@ APP_ETC=/home/deploy/lobsters/shared/etc
 # https://docs.digitalocean.com/products/networking/dns/how-to/manage-records/#ptr-rdns-records
 echo "lobste.rs" > /etc/hostname
 
+# Set up iocane package repository
+if [ ! -f /etc/apt/sources.list.d/iocaine.sources ]; then
+  curl -s https://iocaine.madhouse-project.org/_/debian/iocaine.sources \
+         -o /etc/apt/sources.list.d/iocaine.sources
+  apt update
+fi
+
 # server packages
-apt install iptables-persistent opendkim postfix restic
+apt install iptables-persistent opendkim postfix restic iocaine
 
 # configure opendkim for postfix
 cp opendkim.m4 /etc/mail/m4/
@@ -62,6 +69,7 @@ systemctl force-reload systemd-journald
 systemctl daemon-reload
 systemctl enable --now opendkim
 systemctl enable --now postfix
+systemctl enable --now iocaine
 
 # quiet logins
 cp sshd_config /etc/ssh/sshd_config


pushcx:
Hunter and I manually copied this setup into prod to try things out and revise.

It runs, but we don't feel confident that we've got a good configuration or that we can fix it if anything breaks. The docs are confused and incomplete*, there were odd failures like `systemctl enable --now iocaine` hanging, there's weird defaults like roto over lua, we can't add a footer to explain to humans what they're seeing, there's teenager humor like the default port being 42069, there's no debugging to know why iocaine passed or failed any particular request, and so on. Individually each of these are pretty small red flags, but each aspect of setting up iocaine raised another. The sum effect is that this feels like a clever hobby project that hasn't seen prod usage, and if we use it with real traffic it's going to break in some surprising, undocumented way that is very painful to debug.

 \* What is `wordlist`? What is `training-corpus`? Are they used at startup or runtime? Is the use of training corpus cached? What format do the files need to be in? etc etc

algernon:
> The sum effect is that this feels like a clever hobby project that hasn't seen prod usage, and if we use it with real traffic it's going to break in some surprising, undocumented way that is very painful to debug.

FYI, it's being used by [Codeberg](https://codeberg.org/Codeberg-Infrastructure/scripted-configuration/src/commit/425e602d1244393b75a47ea5d35459d88aff4c90/hosts/_reverseproxy/etc/caddy/forgejo-prod.site#L128-L141), and a client of mine with *quite* a bit of traffic, and a growing number of smaller deployments. It's definitely seen prod usage.

I'm also responsive to bug reports, and am happy to help configure, tune, or debug iocaine. I'm an email or toot away.

> The docs are confused

I'd love to hear which parts, in particular, so I can improve it! The current docs are relatively new, and were written by a person (me) who's way too familiar with iocaine. I'm sure I missed parts that need more explaining.

> What is `wordlist`? What is `training-corpus`? Are they used at startup or runtime? Is the use of training corpus cached? What format do the files need to be in? etc etc

Like this!

A wordlist, with the built-in script is used to construct URLs. Any text file (or files) will do, it will be split by whitespace, and unique words will be kept to make URLs out of them. A dictionary works wonderfully well for that.

The training corpus is a set of text files, used to train the Markov chain iocaine uses to generate garbage content. Unlike wordlists, it is more useful if this is some kind of comprehensible text. Public domain books of old are a great sources.

Both of these are loaded at startup, and are kept in memory until iocaine is running. There is no caching outside of this, so every time iocaine starts up, the wordlist & training corpus will be re-read, and re-trained. (I have experimented with dumping the trained data to disk & loading that back, but it yielded only marginally improved results so far.)

> there were odd failures like `systemctl enable --now iocaine` hanging

Hm, that's new. It *should* hang until iocaine is finished training, and is ready to serve. That might take a couple of seconds, depending on the size of the training data. How much text did you put under it? O:)

> there's weird defaults like roto over lua

That may seem weird to you, but Roto performs about an order of magnitude better than Lua.

One of the reasons of iocaine's existence is performance, so the more performant language was chosen as the default. Lua was added to ease developing custom request handlers, in a perhaps more familiar language, in situations where performance matters less. For a default, I believe performance is a better choice. The built-in script is available in both Roto and Lua, though, so if you're more familiar with the latter, the two are equal in functionality, you can switch between them freely.

> there's teenager humor like the default port being 42069

I like a little whimsy. Don't see why a project that aims to fight off corporate greed should use corporate-friendly "normal" port numbers.

> we can't add a footer to explain to humans what they're seeing

You can! You can [replace the template](https://git.madhouse-project.org/iocaine/iocaine/src/branch/main/data/defaults/README.md#template), and add any footer, or banner, or whatever you wish. The default template is [here](https://git.madhouse-project.org/iocaine/iocaine/src/branch/main/data/defaults/templates/garbage.html), and can also be extracted with `iocaine show embeds -c '/defaults/templates/garbage.html'` too.

Yes, this could be documented better.

> there's no debugging to know why iocaine passed or failed any particular request

If you enable logging, iocaine will log which ruleset was used for each and every request, along with the request method, path, and all headers. Granted, that doesn't tell you *why* a request failed (or passed), but I think it comes pretty close. If you have any suggestions how the log messages could be improved to make this clearer, I'm more than happy to adjust them.

Logging is disabled by default because it is comparatively expensive.

pushcx:
> The current docs are relatively new, and were written by a person (me) who's way too familiar with iocaine. I'm sure I missed parts that need more explaining.

FWIW, my usual advice to devs is to sit down next to someone who's never seen the project, ask them to think out loud, and watch them try to set up from the docs. Don't answer questions or help unless they've been stuck for at least 60 seconds. Take notes.

>>    there were odd failures like systemctl enable --now iocaine hanging

> Hm, that's new. It should hang until iocaine is finished training, and is ready to serve. That might take a couple of seconds, depending on the size of the training data. How much text did you put under it? O:)

None. I also did a `journalctl -f -u iocaine`, the entire contents were:

```
Dec 07 16:47:23 lobste.rs systemd[1]: Starting iocaine.service - iocaine, the deadliest poison known to AI...
Dec 07 16:47:23 lobste.rs iocaine[3671576]: 2025-12-07T16:47:23.357496Z  WARN iocaine::user: No ai-robots-txt-path configured, using default
Dec 07 16:47:23 lobste.rs systemd[1]: Started iocaine.service - iocaine, the deadliest poison known to AI.
```

I didn't see it in `htop` so I didn't get the impression it was doing anything after a minute or two. I figured it didn't fork the way the default systemd script expected, `ctrl-c`'d, and confirmed with `curl` that it was running.

> One of the reasons of iocaine's existence is performance

I called [the docs](https://iocaine.madhouse-project.org/documentation/3/scripting/) confused because of stuff like this.

> Lua and Fennel perform the same, and while Roto is considerably faster than both, in many cases, the request handler will spend most of its time generating garbage - and that is Rust code, shared between all three runtimes. The language you choose for the script handler will not be your bottleneck - choose one you’re familiar with, or one you like, or want to play with.

So the primary differentiation between languages is performance, but their performance difference doesn't matter so why mention it at all, but your comment says performance matters. I can't square these statements. And if it does matter, drop the other languages.

> Don't see why a project that aims to fight off corporate greed should use corporate-friendly "normal" port numbers.

No individual element was dispositive, and I probably wouldn't have cared about this, but in reading the docs and tinkering with it in prod, I was trying to answer the question: is this something I trust with prod traffic? Does this look like a clever toy or a reliable tool? We gave it a lot of leeway because it came highly recommended from an experienced community member, but after 30 minutes of a speedbump every minute or two, the cumulative effect was that I realized I would feel like a reckless fool if I found myself trying to debug a prod issue in a few weeks.

algernon:
> > The current docs are relatively new, and were written by a person (me) who's way too familiar with iocaine. I'm sure I missed parts that need more explaining.
> 
> FWIW, my usual advice to devs is to sit down next to someone who's never seen the project, ask them to think out loud, and watch them try to set up from the docs. Don't answer questions or help unless they've been stuck for at least 60 seconds. Take notes.

I'd love to do that, but for that, I'd need someone who's never seen the project, and who's willing to go through it together with me. That's surprisingly hard to find, unfortunately.

> > > there were odd failures like systemctl enable --now iocaine hanging
> 
> > Hm, that's new. It should hang until iocaine is finished training, and is ready to serve. That might take a couple of seconds, depending on the size of the training data. How much text did you put under it? O:)
> 
> None. I also did a `journalctl -f -u iocaine`, the entire contents were:
> 
> ```
> Dec 07 16:47:23 lobste.rs systemd[1]: Starting iocaine.service - iocaine, the deadliest poison known to AI...
> Dec 07 16:47:23 lobste.rs iocaine[3671576]: 2025-12-07T16:47:23.357496Z  WARN iocaine::user: No ai-robots-txt-path configured, using default
> Dec 07 16:47:23 lobste.rs systemd[1]: Started iocaine.service - iocaine, the deadliest poison known to AI.
> ```

Hrm, according to these logs, iocaine started, and systemd should've considered it as such: the first line is where it begins to start, the last is when it sent an `sd_notify` to systemd. Since the message is coming from systemd, it received it. Not quite sure why enable hung then. I'll look into it.
 
> > One of the reasons of iocaine's existence is performance
> 
> I called [the docs](https://iocaine.madhouse-project.org/documentation/3/scripting/) confused because of stuff like this.
> 
> > Lua and Fennel perform the same, and while Roto is considerably faster than both, in many cases, the request handler will spend most of its time generating garbage - and that is Rust code, shared between all three runtimes. The language you choose for the script handler will not be your bottleneck - choose one you’re familiar with, or one you like, or want to play with.
> 
> So the primary differentiation between languages is performance, but their performance difference doesn't matter so why mention it at all, but your comment says performance matters. I can't square these statements. And if it does matter, drop the other languages.

It's more complicated than that. *Sometimes* it matters, sometimes it doesn't. Sometimes a more familiar language is more useful than a performant one, sometimes it is not. That's not up to me to decide, hence the choice. For *me*, it matters, because I'm running it on a €4/month VPS that sometimes has to handle 50+ million requests a day. Every instruction counts.

In any case, thank you for documenting the issues you ran into, these notes will be helpful in improving iocaine.

PR #1820, opened 2025-12-08
fixup 32a118df

chamlis:
Just saw this one go by whilst catching up on commits, I started reading The Well-Grounded Rubyist so I finally understand more of what's going on in the backend code.


pushcx:
Oof, a little embarassing to get such a small patch wrong, but thank you. 

If you started with the beta of the 4th ed, I've to hear your review of it sometime.

PR #1827, opened 2025-12-17
database_consistency build step and fixes

### LINKED ISSUE #1817: Finish database_consistency, add to build
In 2019 I filed #745 where I added the `database_consistency` gem and a number of contributors helped work through the potential bugs it located. And then we forgot to add it to the build, so it's been quietly sitting unused while new inconsistencies accumulate.

 - [x] generate a [todo list](https://github.com/djezzzl/database_consistency/wiki/generate-todo)
 - [ ] add `bundle exec database_consistency` to the build (`.github/workflows/check.yml`)
 - [ ] close out all the todos and delete `.database_consistency.todo.yml`

Contributors: can work through the [todos here](https://github.com/lobsters/lobsters/blob/main/.database_consistency.todo.yml) to add this missing validations and delete those todo lines. You can do as many or as few as you'd like in a PR, don't feel like you have to commit to closing them all out.

fpsvogel:
Resolves #1817

- Adds the [database_consistency](https://github.com/djezzzl/database_consistency) gem as a build step.
- Updates the gem.
- Fixes the remaining issues flagged by database_consistency.

Each commit fixes one or two issues flagged by the gem, so it may be easiest to review this PR one commit at a time. In each commit message I included documentation from the database_consistency wiki, about the specific checker(s) addressed in the commit.

Some commits fix all instances of an issue, other commits leave them in the database_consistency config file as exceptions. In some commits, the issue is fixed in some places, and left as exceptions in other places.

Each time I left exceptions in the config file, I added a code comment at the top of the file explaining those exceptions. [Here's the final list of those comments.](https://github.com/lobsters/lobsters/pull/1827/changes#diff-9915bf99a834fa38d33945abfcc4330e5eea251ffa5ece66a59ae7a194ff185b)

## Migrations demo

Rolling back the migrations and then re-running them:

GIF showing the migrations for Lobsters #1827 being rolled back and re-applied

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -33,7 +33,7 @@ class Story < ApplicationRecord
     inverse_of: :story,
     dependent: :destroy
   has_many :tags, -> { order("tags.is_media desc, tags.tag") }, through: :taggings
-  has_many :votes, -> { where(comment_id: nil) }, inverse_of: :story, dependent: :restrict_with_exception
+  has_many :votes, -> { where(comment_id: nil) }, inverse_of: :story, dependent: :destroy
   has_many :voters, -> { where("votes.comment_id" => nil) },
     through: :votes,
     source: :user


.github/workflows/check.yml
```diff
@@ -100,3 +100,41 @@ jobs:
           gem install bundler-leak
           bundle leak update
           bundle leak check
+
+  database-consistency:
```
fpsvogel, line 107: This new job includes DB setup and steps copied from the `test` job [above](https://github.com/lobsters/lobsters/blob/5866c556629aebb7ad08f124a11da8dd59559597/.github/workflows/check.yml#L13-L43), because running `database_consistency` involves booting the app.

Along the way I tried to DRY up the DB setup in a new `x-database-setup` section at the top level (below), but [that caused GitHub Actions to fail](https://github.com/lobsters/lobsters/actions/runs/20311022551?pr=1827), so I ended up simply duplicating the setup from `test`.

```yaml
# ...

# Common setup for jobs that need database access.
# To use this setup, include the following in your job:
#
# jobs:
#   my-job:
#     <<: *database-setup
#     steps:
#       - uses: ...
#         ...
#       - name: Workaround for inability to templatize credentials (#1380)
#         run: cp .github/workflows/check/* config/
#       - name: Load database schema
#         run: ./bin/rails db:schema:load db:migrate
#       - name: <your step here>
#         ...
x-database-setup: &database-setup
  runs-on: ubuntu-latest
  timeout-minutes: 15
  strategy:
    fail-fast: false

  services:
    mariadb:
      image: mariadb:10.11.11
      env:
        MARIADB_ROOT_PASSWORD: root
        MARIADB_DATABASE: lobsters_dev
      ports:
        - 3306:3306

  env:
    DATABASE_URL: trilogy://root:root@127.0.0.1/lobsters_dev
    RAILS_ENV: test
    RUBY_YJIT_ENABLE: "1"

jobs:
# ...
```

app/models/origin.rb
```diff
@@ -5,11 +5,15 @@
 # foo.github.io and github.com/foo have two different Domains that both produce the Origin with
 # identifier github.com/foo. That Origin's domain is set to github.com.
 class Origin < ApplicationRecord
-  belongs_to :domain, optional: false
+  belongs_to :domain
```
fpsvogel, line 8: `optional: false` is the default, so unnecessary.

app/models/user.rb
```diff
@@ -112,29 +120,33 @@ class User < ApplicationRecord
...
+    length: {maximum: 255}
   validates :disabled_invite_reason,
     length: {maximum: 200}
 
   validates :show_email, :is_admin, :is_moderator, :pushover_mentions,
     inclusion: {in: [true, false]}
 
-  validates :session_token,
-    allow_blank: true,
-    presence: true
```
fpsvogel, line 136: [Moved up](https://github.com/lobsters/lobsters/pull/1827/changes#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR129-R132) so that there is only one `validates :session_token`.

db/schema.rb
```diff
@@ -228,13 +229,12 @@
     t.index ["token"], name: "index_messages_on_token", unique: true
   end
 
-  create_table "mod_activities", charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
+  create_table "mod_activities", charset: "utf8mb4", collation: "utf8mb4_uca1400_ai_ci", force: :cascade do |t|
```
fpsvogel: The collation changes happened during migrations. It seems that default collations in MariaDB have changed recently: https://jira.mariadb.org/browse/MDEV-25829

spec/models/moderation_spec.rb
```diff
@@ -25,20 +25,24 @@
     domain = create :domain
     story = create :story, user: user
     tag = create :tag
+    action = "edited"
```
fpsvogel, line 28: This spec needs `action` in order to pass now because of [a validation](https://github.com/lobsters/lobsters/pull/1827/changes#diff-ebcba190832fdd4df6e4bc3c1b70758c18dc18acd52c90af93202d0077007dd9R66) that I added to mirror a DB null constraint.

(Because of the null constraint that already existed, these records in `valid_moderations` were not actually valid, in the sense that trying to save any of them would have resulted in an error from the null constraint.)

.database_consistency.ignore.yml
fpsvogel, line 1: The issues listed in this file do not need to be fixed, as far as I know. They are exceptions to the rules, or false positives, which database_consistency will continue to ignore moving forward.

To explain each remaining flag, I made one big chunk of code comments at the top so that when more exceptions need to be made in the future, it'll be easier to copy over the one chunk of comments rather than diffing/copying individual comments throughout the file.

But let me know if individual interspersed comments would be preferable.

db/migrate/20251214043426_messages_short_id_non_nullable_and_default_value.rb
```diff
@@ -0,0 +1,7 @@
+class MessagesShortIdNonNullableAndDefaultValue < ActiveRecord::Migration[8.0]
+  def change
+    Message.where(short_id: nil).update_all(short_id: "")
```
pushcx: There aren't any in prod so this is a no-op, but as this is the user-visible key. Duplicates or missing values would be worth erroring out the migration over.

db/schema.rb
pushcx: I'd like this file to match prod, and prod reflects that tables have had different values over time.

I don't have a strong opinion on what those values are, but I want to use this new default via a migration rather than pushing `schema.rb` out of sync with prod. Could I ask you to do that in a small separate PR that I could merge before this?

app/models/category.rb
```diff
@@ -6,6 +6,7 @@ class Category < ApplicationRecord
     dependent: :restrict_with_error,
     inverse_of: :category
   has_many :stories, through: :tags
+  has_one :moderation, dependent: :destroy
```
pushcx: We should never delete `moderation` entries, please use `:restrict_with_error`.

app/models/domain.rb
```diff
@@ -6,7 +6,8 @@ class Domain < ApplicationRecord
     class_name: "User",
     inverse_of: false,
     optional: true
-  has_many :origins
+  has_many :origins, dependent: :destroy
+  has_one :moderation, dependent: :destroy
```
pushcx: We should never delete `moderation` entries, please use `:restrict_with_error`.

app/models/domain.rb
```diff
@@ -6,7 +6,8 @@ class Domain < ApplicationRecord
     class_name: "User",
     inverse_of: false,
     optional: true
-  has_many :origins
+  has_many :origins, dependent: :destroy
```
pushcx: We don't delete Domains or Origins, please use `:restrict_with_error`.

app/models/origin.rb
```diff
@@ -5,9 +5,13 @@
...
   has_many :stories
-  belongs_to :banned_by_user, class_name: "User", inverse_of: false, optional: true
+  belongs_to :banned_by_user,
+    class_name: "User",
+    inverse_of: false,
+    optional: true
+  has_one :moderation, dependent: :destroy
```
pushcx: `:restrict_with_error`

app/models/story.rb
```diff
@@ -33,7 +33,7 @@ class Story < ApplicationRecord
     inverse_of: :story,
     dependent: :destroy
   has_many :tags, -> { order("tags.is_media desc, tags.tag") }, through: :taggings
-  has_many :votes, -> { where(comment_id: nil) }, inverse_of: :story
+  has_many :votes, -> { where(comment_id: nil) }, inverse_of: :story, dependent: :destroy
```
pushcx, line 36: We don't delete Stories or Votes.

app/models/user.rb
pushcx: Good catch, this duplication is one of those odd places a codebase shows its age.

.database_consistency.ignore.yml
```diff
@@ -1,25 +1,29 @@
...
+#   and add a corresponding `has_many` with `dependent: :destroy`. (It might
+#   make sense to add `dependent: :nullify` on the `belongs_to` side, and/or
+#   make sure the foreign key has `on_delete: :nullify`.) Other exceptions:
+#   a `has_many` with `dependent: :nullify` (Story#merged_stories), and a case
+#   where child records should remain when parents are deleted (User#stories).
+# - MissingIndexChecker: False positives. The `moderations` table actually does
+#   have an index for each of Moderation's `belongs_to` associations.
 # - ForeignKeyChecker: StoryText#story uses `story_texts.id` in place of a
-#   foreign key.
+#   foreign key. Nonstandard but intentional.
```
pushcx, line 28: "weird but legit"

pushcx (changes requested):
Wow! I had figured this was one of those issues that was going to hang around for years, so I'm really pleasantly surprised that you picked it up and put so much time into it, thank you. I can see this is really thorough, thoughtful work and had few comments.

I stopped leaving comments on the 1bda4d02c7 commit because they would've become repetitive. We don't delete many things, and soft-delete almost all of them. (Off the top of my head, I think Messages are the exception, for privacy.) I don't want the possibility that a deletion that should never have happened could cascade through and delete more records, so please change that to use `:restrict_with_errors` throughout.

I think the only other significant change is about the collation.

db/schema.rb
fpsvogel: Ah yeah, that makes sense. I removed the collation changes from this PR. ([Latest diff for schema.rb.](https://github.com/lobsters/lobsters/pull/1827/changes#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3))

I haven't opened a separate PR because I saw you made some collation changes in https://github.com/lobsters/lobsters/commit/a4fb48fc35928b85ce998081f75c468a7a6353a9. Is that related?

Also, I saw that Lobsters may migrate to SQLite soon, so maybe it doesn't make sense to worry about the collations. But LMK if anything still needs to be done there.

app/models/category.rb
fpsvogel: In https://github.com/lobsters/lobsters/pull/1827/changes/3770966a3ee7fce8583e08ea0b03626ba3de35c2 I changed this and the other additions of `dependent: :destroy` to `:restrict_with_error`.

However, this change caused [test failures](https://github.com/lobsters/lobsters/actions/runs/21447675347/job/61767804319?pr=1827) that seem to be from uniqueness validation failures (records not being deleted between tests?). I'll do more research, but off the top of your head can you think of why this would cause these test failures?

db/migrate/20251214043426_messages_short_id_non_nullable_and_default_value.rb
fpsvogel: Done in https://github.com/lobsters/lobsters/pull/1827/changes/8bac6b0c59b79c3e9aef584d807e7213e7a5d884

db/schema.rb
fpsvogel: Also: I try to always make post-review changes in additional commits rather than by changing past commits, for ease of re-reviewing and so that I don't have to force push. But for some reason this time I removed the collation changes in an interactive rebase rather than in an additional commit. Sorry about that! All my other changes were in additional commits.

fpsvogel: Though, there were merge conflicts in this PR, so I would've had to rebase in any case.

app/models/category.rb
pushcx: `spec/rails_helper.rb` does create a Category named `category1` and two Tags (`tag1`, `tag2`). It's been long enough I'm not sure they're strictly needed. We started with Machinist, which was not as good at creating associated records in FactoryBot - so now we can say `create(:story)` and it can create an associated Tag.

I don't think there's anything else that should be carried between tests, but it's possible there's a misconfig. I've tried to write self-suffient tests that create all the data they need, so it's possible they've been quietly ignoring extra data.

I skimmed your build failure and nothing comes to mind, it doesn't look related to the test data I just mentioned. Maybe `:restrict_with_error` doesn't work with `DatabaseCleaner.strategy = :truncation`?

fpsvogel: It turns out that using `dependent: :destroy` for `Story#votes` prevents the test failures. See additional commit https://github.com/lobsters/lobsters/pull/1827/changes/c1c07cb75a9e99f226cf28222eeec8df9911028d

Also, I changed [one of the newer commits](https://github.com/lobsters/lobsters/pull/1827/changes/79d0c6859bc1c2df78566eef526953d3de5184c4) to use `dependent: :restrict_with_exception` instead of `:restrict_with_error`, because I noticed `:restrict_with_exception` is used in a number of other places, while `:restrict_with_error` occurs only once.

fpsvogel: There are also quite a few occurrences of `dependent: :destroy` for previously existing associations (none of which are for Messages), and changing most of those to `dependent: :destroy` does not cause test failures. The only associations that produce test failures (similar to the ones you saw in CI) are these:

- `Story#taggings`
- `Story#votes` (mentioned in my previous message above)
- `Story#links`
- `User#usernames`

I didn't make any changes to existing associations; I was just experimenting. But I can open a separate PR for this, if it would make sense to change existing associations to use `dependent: :restrict_with_exception` where possible.

pushcx: I hadn't noticed there were two options and had to look up [the difference](https://stackoverflow.com/questions/48071054/difference-between-restrict-with-exception-and-restrict-with-error) between `_with_exception` and `_with_error`. `exception` is definitely what we want to express, good call.

And yes, let's get this merged and then replace `destroy` with `_exception`, that's almost certainly preferred behavior everywhere.

pushcx:
I think with the association stuff done this is ready to merge? :crossed_fingers:

app/models/category.rb
fpsvogel: Yep, this is ready to merge 👍

pushcx:
Thank you working thoroughly through such a large project that'll keep the codebase safer (from me).

PR #1829, opened 2025-12-21
Set DATABASE_HOST in database.yml, defaults to local

ChaelCodes:
# Description

Right now, we have to manually set the host for the database in database.yml if we're using docker. In local development, we use localhost, but Docker needs to reference the db image.

This change sets an environment variable `DATABASE_HOST` in docker and the database.yml. No manual editing needed, and localhost continues to use localhost.


pushcx:
Looks like it makes Docker one bit easier, thanks!

PR #1830, opened 2025-12-21
Allow the server to start without credentials set in development

ChaelCodes:
# Description

We don't actually _need_ to set the credentials to start the Rails server. All the places we call the credentials in initialization can handle an empty value. We need the credentials to test certain features, but not to start the server. However, this raise in Zeitwork was preventing the Rails server from starting in development unless they were set.

Now the error is still put in the log, production still requires crednetials, and credentials are still included in the documentation, but the server can actually start without them.


pushcx:
After chatting with Chael about this, we decided we can just delete this file. It's been a year since the migration, and there are only two known sister sites so _hopefully_ they've updated already. And if not, it just means they have a less clear error rather than things breaking worse.

PR #1831, opened 2025-12-21
Move dev environment setup to CONTRIBUTING

ChaelCodes:
# Description

I think it's a good idea to move dev environment setup to CONTRIBUTING.md so when someone is contributing they have access to everything they need in one place. In addition, this means the README stays focused on using Lobsters and creating a sister site. It directs potential contributors to CONTRIBUTING, but doesn't contain the full dev setup.


pushcx:
Thanks, this makes for clearer instructions!

PR #1832, opened 2025-12-22
Remove credentials prompt

ChaelCodes:
References #1830 

Remove prompt for credentials from zeitwerk initializer. It was a temporary initializer for those migrating, but is no longer needed.

These credentials are referenced in contributing and sister site setup, so they're still documented.


pushcx:
Thanks!

PR #1834, opened 2025-12-24
Remove Unused Routes

hmaddocks:
Tighten the routes for the following resources 
`messages`
` hats`
` mod/origins`

Problems:
Hats `index` tries to create a link to `new_hat_request_path` but this action doesn’t exist on `HatsController`. I have partially commented this out but it needs to be fixed properly. Either remove the whole block or add new hat request.

`app/views/hats/index.html.erb`
```ruby
<% if @user %>
<h2><a href="<%# new_hat_request_path %>">Request Hat</a></h2>
<% end %>
```

There are unused routes on `hat_requests`
Hats Requests has a couple of restful forms in `app/views/hat_requests/index.html.erb`, but the edit and update actions don’t exist on the `HatRequestsController`. It’s probably easier to live with the unused routes that try to work around the forms ¯\\\_(ツ)_\/¯

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -219,7 +219,7 @@ Rails.application.routes.draw do
   post "/invitations/delete_request" => "invitations#delete_request",
     :as => "delete_invitation_request"
 
-  resources :hat_requests do
+  resources :hat_requests, except: [:edit] do
     member do
       post :approve
       post :reject


app/views/hats/index.html.erb
```diff
@@ -8,10 +8,10 @@ message. Doffed hats are not listed in this table, check a user's profile page.
 </p>
 
 <% if @user %>
-  <h2><a href="<%= new_hat_request_path %>">Request Hat</a></h2>
+<h2><a href="<%# new_hat_request_path %>">Request Hat</a></h2>
```
pushcx: This link [works in prod](https://lobste.rs/hats) though?

pushcx (commented):
If there's more unused templates, actions, or routes, I'd rather delete them entirely than leave them.

app/views/hats/index.html.erb
hmaddocks: My bad. I mis-read the spec failure and made several changes at once so I thought the `new_hat_request_path` was redundent. Reverted this change.

hmaddocks:
> If there's more unused templates, actions, or routes, I'd rather delete them entirely than leave them.

The `app/views/hat_requests/index.html.erb` is a bit unusual and I'm not sure what the intent is.
There are two forms. The first form doesn't have a submit button but even if it did submitting the form would go to the non-existent `update` action. So I don't know whether you want to remove the first form or combine the forms into one. Do you want to be able to edit the user submitted values (Hat and Link)? If not the fields could be disabled or removed and the values rendered separately.

Other minor things;
The `:link` label is duplicated.
There are two `submit`, "Approve" and "Reject" on the second form. Maybe refactor it to be something like this
```
<%= form_with model: hr, url: approve_hat_request_path(hr), method: :post do |f| %>
  <div class="labelled_grid">
    <%= f.label :reason %>
    <%= f.text_area :reason, required: true %>
  </div>
  <div class="row">
    <%= f.submit "Approve Hat Request" %>
    <%= f.submit "Reject Hat Request", class: "deletion", formaction: reject_hat_request_path(hr) %>
  </div>
<% end %>
```

But it depends on what you want this view to do.

When this is resolved `config/routes.rb:222` can be changed to 
```
  resources :hat_requests, only: %i[index new create] do
```

pushcx:
There's two forms because the user can create a hat request, then the mod can edit with the user submitted and more, then approve or reject the request to create the hat.

I have been slowly separating out mod functionality into controllers that live in `app/controllers/mod` because the old pattern of reusing one form for both users and mods caused a lot of bugs. You could do that move or just keep things working as-is. I'm happy to take partial cleanup here.

hmaddocks:
This is good to go. I will fix the hat_requests routes on another PR

pushcx:
Thanks for tidying these up!

PR #1835, opened 2025-12-24
Fix some dead links in chat page

hugoarnal:
When going on the https://lobste.rs/chat page, I noticed some dead / outdated links on there


pushcx:
Thanks for catching these and figuring out what they should be!

PR #1838, opened 2025-12-26
Ruby 4.0

caius:
## What?

- [x] [Upgrade standardrb in #1839](https://github.com/lobsters/lobsters/pull/1839)
- [x] Upgrade Bundler to 4.0.3
- [x] Add ruby platform back to `Gemfile.lock`
- [x] Let bundler upgrade a bunch of gems that need to be for Ruby 4.0 compatibility
- [x] Pin Ruby to 4.0.0

## Why?

Ruby 4 is our Christmas Present from Ruby core team.

The gems that got updated were forced upgrades by bundler/ruby 4.x running `bundle install`. Didn't have to do any magic to figure out which they were.

The ruby platform is added back so we correctly depend on non-specific-platform gems (which in turn pins dependencies better-er I think), and also means that anyone trying to `bundle install` on a platform that isn't Darwin/arm64 or Linux/amd64 is able to install without Gemfile.lock always being dirty in git.

We need the rubocop-ast upgrade from #1839 or we can't run linting on this branch. (This branch is stacked on that one, but GitHub doesn't support cross-fork stacked branches. Ugh.)


pushcx:
Thanks for contributing this, Merry Christmas to us.

Could you rebase or merge in `main`? In #1843 I merged a new build check to make sure I don't break the docker setup. I'd like to exercise that as long as we're doing one of our infrequent Ruby upgrades.

caius:
Pushed and check did indeed fail in https://github.com/lobsters/lobsters/actions/runs/20624466151/job/59232527496

Added the Dockerfile.dev change into the upgrade commit and pushed again, should be good now.

pushcx:
Thanks for proving that works and catches exactly the issue I was concerned about.

Happy new year, happy 4.0. :D

gjtorikian:
👋 I got pinged about a5c77ca644acfb91a40e5d8c293314033d161cfb and Commonmarker not supporting Ruby 4.0. Version 2.6.1 now supports Ruby 4.

PR #1839, opened 2025-12-27
Update standardrb gems

caius:
## What?

- [x] Update standardrb gems
- [x] Update rubocop-ast (indirect dependency)

## Why?

rubocop-ast needs updating for Ruby 4, and we probably want to update standardrb, et al anyway.

Also have `bundle exec standardrb --fix` fix all our new issues.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -190,7 +190,7 @@ class CommentsController < ApplicationController
 
       render partial: "comment", locals: {comment: comment, show_story: show_story, show_tree_lines: show_tree_lines}
     else
-      redirect_back fallback_location: root_path
+      redirect_back_or_to(root_path)
     end
   end
 


app/controllers/comments_controller.rb
```diff
@@ -190,7 +190,7 @@ def disown
 
       render partial: "comment", locals: {comment: comment, show_story: show_story, show_tree_lines: show_tree_lines}
     else
-      redirect_back fallback_location: root_path
+      redirect_back_or_to(root_path)
```
caius, line 193: Changed by `bundle exec standardrb --fix`

pushcx:
Thanks for bumping this. I've been meaning to but waiting on a release with the fix for https://github.com/standardrb/standard-rails/issues/112 - hopefully that'll be this one.

PR #1841, opened 2025-12-28
Lean on YJIT/Frozen strings everywhere in CI

caius:
## What?

- [x] Move `RUBYOPT` to apply to all CI jobs in check pipeline
- [x] Replace `RUBY_YJIT_ENABLE` with `RUBYOPT=--yjit`

## Why?

We weren't enabling frozen string literals everywhere in CI due to brakeman having an issue, but that's been fixed upstream so now we're good to use this everywhere. Keep the debugging enabled so we stand a chance of figuring out what broke if something does break in future.

We currently use `RUBY_YJIT_ENABLE` envariable to enable Ruby's YJIT in the test suite, given we're already defining `RUBYOPT` we might as well pass `--yjit` in there - if we want to tune the YJIT memory usage, etc then we'll need to do it from `RUBYOPT` in future anyway.

Also enable deprecation warnings in the test suite, `spec/spec_helper.rb` is basically the first "app file" loaded when we run `bundle exec rspec` so its a good place to enable these things and have warnings emitted from both the dependencies we load and the app itself.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  RUBYOPT: "--enable-frozen-string-literal --debug-frozen-string-literal"
+  RUBYOPT: "--yjit --enable-frozen-string-literal --debug-frozen-string-literal"
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks for catching this opportunity and updating us!

PR #1842, opened 2025-12-28
Improve page cache filename handling

### LINKED ISSUE #1826: rss filename generated for caching is too long for the filesystem
As discussed on IRC: my rss client reports that the rss lobste.rs feed broke 4 days ago. When I try the url with my browser, it returns a 500 Internal Server Error.
The URL has a lot of tags (30+). @pushcx said

> problem is the filename generated for caching is too long for the filesystem

Thanks!

caius:
## What?

- [x] Added specs for `ActionController::Caching::Pages::PageCache#cache_file` monkeypatches
- [x] Fix bug appending ".html" to non-HTML filenames in caching
- [x] Handle generated filename being > 255 characters long by rescuing write error

## Why?

Hopefully fixes #1826.

Prepend a module defining a `write` method that catches `Errno::ENAMETOOLONG` into the `PageCache` class. Calls the original code, but wraps it in a rescue block for this error.

### DIFF
--- a/config/initializers/page_caching_monkeypatch.rb
+++ b/config/initializers/page_caching_monkeypatch.rb
@@ -12,12 +12,28 @@
 # report or contribute a fix upstream.
 #
 # We only cache HTML pages, so this monkeypatch forces the '.html' extension.
+#
+# We also catch an error when caching files if the computed filename on disk is too long, due to the
+# filesystem used in production only supporting 254 character names.
 
 require "actionpack/page_caching"
 
 ActiveSupport.on_load(:action_controller) do
   module ActionController::Caching::Pages # standard:disable Lint/ConstantDefinitionInBlock
     class PageCache
+      module WritePatch
+        # Override: refuse to cache paths with long filenames
+        def write(...)
+          super
+        rescue Errno::ENAMETOOLONG => e
+          # #1826 - Handle write errors from filenames being longer than 255 bytes
+          Rails.logger.info "Failed to cache page #{e.inspect}"
+          nil
+        end
+      end
+
+      prepend WritePatch
+
       def cache_file(path, extension)
         name = if path.empty? || path =~ %r{\A/+\z}
           "/index"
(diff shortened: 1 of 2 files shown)


caius:
@pushcx just occurred to me writing the PR description, what serves these files from cache? Are they pulled from nginx without any ruby/rails in the mix in which case this will just break caching for those paths (as nginx doesn't know to look for the sha256 digest path on disk).

caius:
I know it's the old infrastructure, but it might give a hint as to what we have configured now. https://github.com/lobsters/lobsters-ansible/blob/master/roles/nginx/files/production/nginx/sites-available/lobste.rs#L107-L118 suggests we take the URL and read directly from disk for it, which means this PR is pointless (or rather, I should split out the tests and fix for the `.html` extension from it)

pushcx:
I replaced Ansible with Hatchbox, and that replaced nginx with Caddy. Very similar serving, though: https://github.com/lobsters/lobsters/blob/f1c190485df3d6f3c5e3dcd09d9350759d6a3839/hatchbox/Caddyfile.pre#L17-L30

There's no Rails involved in serving from the cache, that's why it's fast enough to stand up to scapers. So yeah, this will break.

Maybe the approach should be to figure out how to cleanly skip trying to write the cache if the filename is too long?

caius:
> Maybe the approach should be to figure out how to cleanly skip trying to write the cache if the filename is too long?

Thought about this some more and aside from trying to wrangle Caddy into searching a second computed path once its too long, preventing the write to the cache is probably the sensible option here for now.

caius:
Updated this so we patch the `write` method instead, and tested it through the `cache` method as that's public interface.

We just call the original code and then catch the `Errno::ENAMETOOLONG` error, doing nothing in that case. Should be enough to stop the 500s and doesn't write anything to the filesystem so we'll just carry on un-cached.

pushcx:
Looks like a safe approach, and thanks for the clear comments and tests. Always good to have plenty of explanation available when we're monkeypatching!

PR #1843, opened 2025-12-29
Set Dockerfile's Ruby version Manually

ChaelCodes:
# Description

Docker is great for occasional contributors who might drop in and out and need a reliable, easy setup, but it's often skipped by maintainers and regular contributors in favor of a local dev environment. It's difficult to keep a Dockerfile and .ruby-version file in sync if you're not using the docker environment. This PR adds a GitHub Action to ensure that both versions stay the same.

Also, we need to set a hard-coded ruby version, because the current option does not behave unless you call `make docker-serve` to set the RUBY_VERSION first which isn't going to be an option when we're using devcontainers (especially not in codespaces).

# Implementation

This PR sets the Dockerfile image to the `ruby:3.4.7` image instead of calculating it on the fly, and removes arguments that were added to support that.

This PR adds a new GitHub Action that only runs when there's a PR or push that changes `Dockerfile.dev` or `.ruby-version`. It uses grep to check if the `.ruby-version` and ruby image version in the Dockerfile match.

Here's an [example PR](https://github.com/ChaelCodes/lobsters/pull/4) demonstrating what this would look like.

[Unrelated PRs](https://github.com/ChaelCodes/lobsters/pull/5) do not trigger the workflow.

# Screenshots

## Passing Workflow
Screenshot 2025-12-29 at 1 59 23 PM

## Failing workflow
Screenshot 2025-12-29 at 2 26 37 PM

## Unrelated PR
Screenshot 2025-12-29 at 2 36 11 PM

### DIFF
--- a/.github/workflows/check-docker-ruby-version.yml
+++ b/.github/workflows/check-docker-ruby-version.yml
@@ -1,0 +1,22 @@
+name: "Check Dockerfile Ruby version"
+
+on:
+  # Catch if a mismatched version is pushed to the main branch
+  push:
+    branches: [main]
+    paths: 
+      - .ruby-version
+      - Dockerfile.dev
+  # Check PRs that modify .ruby-version or Dockerfile.dev
+  pull_request:
+    paths: 
+      - .ruby-version
+      - Dockerfile.dev
+
+jobs:
+  check-ruby-version:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


pushcx:
Thank you for protecting the codebase from me with this. Caius just opened #1838 for a version bump, so I'm going to get this merged down so we can see it work right away.

PR #1844, opened 2026-01-02
Add "aria-label" for "merged" icon, and "description present" icon.

fedemp:

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -609,11 +609,10 @@ li .domain {
 }
 
 .merge {
-	background: url(merge.svg) no-repeat;
+	background: url(merge.svg) no-repeat center left;
 	padding-left: 0.6875lh;
-	min-height: 1lh;
 	display: inline-block;
-	vertical-align: bottom;
+	color: var(--color-fg-contrast-4-5);
 }
 
 li input.comment_folder_button {


app/views/stories/_singledetail.html.erb
```diff
@@ -146,11 +146,11 @@ classes = [
 
   <% if merged_stories.count > 1 %>
     <div class="details byline">
-      <span class="merge"></span><%= merged_stories.count %> merged
+      <span class="merge"><%= merged_stories.count %> merged</span>
```
pushcx, line 149: Why not add the aria label to this one as well?

app/views/stories/_listdetail.html.erb
```diff
@@ -31,13 +31,16 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
       <% end %>
     </span>
     <% if story.stories_count > 0 %>
-      <span class="merge"></span><%= story.stories_count + 1 %>
+      <span class="merge" aria-label="<%= story.stories_count + 1 %> stories merged"><%= story.stories_count + 1 %></span>
```
pushcx, line 34: This isn't redundant, to have the same info in the `aria-label` as the tag body?

app/assets/stylesheets/application.css
```diff
@@ -606,10 +610,10 @@ li .domain {
 
 .merge {
 	background: url(merge.svg) no-repeat;
-	width: 0.6875lh;
-	height: 1lh;
+	padding-left: 0.6875lh;
+	min-height: 1lh;
 	display: inline-block;
-	vertical-align: middle;
+	vertical-align: bottom;
```
pushcx: I think this is the default and you could remove the line?

Could you include a before and after screenshot of the visual changes here?

pushcx (commented):
Thanks for continuing to make these improvements. I have a couple questions to make sure I understand how this'll work.

app/views/stories/_singledetail.html.erb
fedemp: I know it will read different than the one at the front page. I can add it if you prefer the section to read the same in both places. It will read "n merged" though, the "stories" word can be assumed. :)

app/views/stories/_listdetail.html.erb
fedemp: But the body of the `span` only reads a number; the `aria-label` will read "N stories merged".

pushcx: Thank you, I didn't catch that detail. I use the side-by-side diff and it makes reading longer/indented lines harder.

app/assets/stylesheets/application.css
fedemp: Actually the default is `baseline` :nerd_face: 

I added a new commit so the icon will look more centered. You'll have to squash my two commits cause jj didn't allow me modify the first one.

About the screenshots:

image

----

image

----

image

pushcx:
Thanks for this improvement. And fingers crossed we get a jj-native code forge this year. :D

Issue #1845, opened 2026-01-04
Can't link GitHub accounts because CGI.parse is gone


pushcx:
`extras/github.rb` attempts to uses [`CGI.parse`](https://ruby-doc.org/stdlib-2.5.3/libdoc/cgi/rdoc/CGI.html#method-c-parse), which [was removed](https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/#stdlib-compatibility-issues)

Following #1838, so ping @caius.

While we need to replace this call, it'd be good to doublecheck the rest of the (linked) release notes for other issues, as we do not have anything like 100% test coverage.

caius:
Ooh crumbs, missed that when I was looking through. I think there's still the `cgi` gem that provides everything which used to be in the stdlib, so we may be able to just depend on that and have it work.

And add a test for it, naturally 😬

caius:
PR up for the `CGI.parse` issue.

Looking at the others:
- we don't reference `SortedSet`
- Both places we use Net::HTTP we explicitly set a content-type header, so unaffected by that change
- the other `CGI.` calls in the codebase are all the ones provided by `cgi/escape` (although we'll have the full gem loaded anyway.)

pushcx:
Thanks for fixing this bug and checking on the rest of the notes. No worries about the small bug, these things happen.

PR #1846, opened 2026-01-05
`CGI.parse("oops")`

caius:
## What?

- [x] Adds test to prove `Github.token_and_user_from_code` currently fails in code
- [x] Depend on cgi gem to bring functionality back

## Why?

Fixes https://github.com/lobsters/lobsters/issues/1845 after we upgraded to Ruby 4 and the CGI stdlib is cut down to just `cgi/escape`.


pushcx:
Thanks for this fix and test!

PR #1847, opened 2026-01-06
Fix opacity of comments link for hidden stories in mobile

fedemp:
When hiding a story, the comments link should turn semi transparent instead of remaining opaque.


fedemp:
image

pushcx:
Ah, nice touch, thank you.

PR #1850, opened 2026-01-09
Hide << and >> of pagination to screen readers

fedemp:
Use CSS to show the << and >> characters in the pagination nav in a way that screen readers don't try to read them.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -780,19 +780,27 @@ ol.stories.list details.story_content:open summary span {
 	display: none;
 }
 
-div.morelink {
+nav.morelink {
 	float: left;
 }
 
-div.morelink,
+nav.morelink,
 div.page_link_buttons {
 	margin-top: 1.5em;
 }
-div.morelink a {
+nav.morelink a {
 	color: var(--color-fg-contrast-7-5);
 	font-weight: bold;
 	text-decoration: none;
 }
+
+nav.morelink a[rel="prev"]::before {
+	content: "<< "/"";
+}
+nav.morelink a[rel="next"]::after {
+	content: " >>"/"";
+}
+
 div.page_link_buttons {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/application.css
```diff
@@ -780,19 +780,27 @@ ol.stories.list details.story_content:open summary span {
...
 }
-div.morelink a {
+nav.morelink a {
 	color: var(--color-fg-contrast-7-5);
 	font-weight: bold;
 	text-decoration: none;
 }
+
+nav.morelink a[rel="prev"]::before {
+	content: "<< "/"";
```
fedemp, line 798: FYI the `/` does not escape any character. It's a divider for the two different contents that should be rendered for "normal" rendering and for screen readers respectively.

pushcx: Thanks for explaining, I didn't know it was possible.

app/views/layouts/application.html.erb
```diff
@@ -129,9 +129,7 @@
 
   <footer>
     <span id="office-hours-live">
-      <% if File.exist?('tmp/office_hours_live.txt') %>
```
pushcx, line 132: Why did you delete this conditional? I only stream Lobsters office hours a couple hours a week, if I have to stream 128 hours/week I'll die.

pushcx (changes requested):
Great to have, and I appreciate your use of the more-semantic `<nav>` element.

Please restore the office hours conditional and I look forward to getting this merged!

app/views/layouts/application.html.erb
fedemp: Reverted.

I was testing some alignment issues on the same commit, and this change was pushed alongside the fixes for the tests.

pushcx:
Thanks again, deploying now.

PR #1854, opened 2026-01-09
Use aria-label to tag banned users, new users, and post authors

fedemp:
If there's a visual indicator for this, we can add a non visual one too.


pushcx:
We definitely have never tried to audit the templates from the accessibility perspective that all visible information should also be available for screen readers, so thank you for catching this.

PR #1855, opened 2026-01-11
Add `media.ccc.de` to the automatic video tag

hugoarnal:
Out of all (69) of the submitted stories for the `media.ccc.de` domain, only [one](https://lobste.rs/s/fesju4/media_ccc_de_arch_conf_2020) was not a video but a link to an event containing videos.

I feel like it makes sense to suggest the video tag to `media.ccc.de` as that's what's mostly posted on lobsters from the site.

Also fixed a problem with the twitch.tv regex which matched any character


pushcx:
Great catch, thanks for improving this.

PR #1856, opened 2026-01-11
Document database error in docker

hugoarnal:
I came across this error on my machine. I think it's because I need sudo privileges to use docker on Arch, but I'm not sure.

Documenting it in case someone else encounters this error on their machine


pushcx:
Thanks for adding this reference. Odd bug, given how long-running it is.

PR #1857, opened 2026-01-11
Add open graph tags on users & home pages

hugoarnal:
I often share lobste.rs links on discord to friends, but only the comments page show Open Graph tags.

So I decided to add them on the home page and on the user pages.

Final result:

Final result

(I censored the tunnel link to my lobsters local dev)

The result seems pretty good, maybe we can add user description in the `og:description` tag

I don't know if this is truly something that interests you so feel free to close it :)

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -174,7 +174,7 @@ class StoriesController < ApplicationController
       format.html {
         @meta_tags = [
           {property: "og:type", content: "article"},
-          {property: "og:site_name", content: "Lobsters"},
+          {property: "og:site_name", content: Rails.application.name},
           {property: "og:title", content: @story.title},
           {property: "og:description", content: @story.comments_count.to_s + " " + "comment".pluralize(@story.comments_count)},
           {property: "og:image", content: Rails.application.root_url + "touch-icon-144.png"},


app/controllers/home_controller.rb
```diff
@@ -54,7 +54,14 @@ def index
     @root_path = true
 
     respond_to do |format|
-      format.html { render action: "index" }
+      format.html {
+        @meta_tags = [
+          {property: "og:type", content: "website"},
+          {property: "og:title", content: "Lobsters"},
```
fedemp: I think you should use `Rails.application.name` given that the Lobsters codebase is used by other projects.

hugoarnal: I agree. I copied this the existing `stories_controller.rb`: https://github.com/lobsters/lobsters/blob/5a5368de80f408e8b9695b637b7a90dc3f6de242/app/controllers/stories_controller.rb#L175-L181

Perhaps I should also change the `stories_controller.rb` file to use `Rails.application.name`?

fedemp: Probably 🙂

pushcx:
Looks like a real straightforward usability improvement, thanks for this.

Adding the user description sounds good. Doublecheck the length and format, we'll probably have to truncate to be the first couple hundred characters. The field is called `about`, and there's a `linkified_about` if you can send HTML.

PR #1858, opened 2026-01-12
Update minimum message length

winks:
Update minimum message length to 5, as per IRC discussion


fedemp:
Does this apply to direct messages between users? Cause we could update the front end to add a [minlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/minlength) validation there too.

pushcx:
We were talking about this because someone wanted to send a "Thanks!" DM and had to pad it out to 20 characters. 

We don't have too many of these frontend validations so I don't want to hold up this PR for it, but yes, I think that'd be a nice improvement.

PR #1860, opened 2026-01-13
Add database consistency check to CI

caius:
## What?

- [x] Update database_consistency gem to 2.1.1
- [x] Add default database consistency configuration
- [x] Update database consistency todo file
- [x] Run database consistency check in CI

## Why?

Helps progress https://github.com/lobsters/lobsters/issues/1817 by stopping us introducing new issues. Updating the todo file stops the check failing with newer violations. Adding the default file stops us warning about Rails' framework tables we don't need to alter

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -98,3 +98,38 @@ jobs:
           gem install bundler-leak
           bundle leak update
           bundle leak check
+
+  database-consistency-check:
+    runs-on: "ubuntu-latest"
+    timeout-minutes: 15
+    strategy:
+      fail-fast: true
+
+    services:
+      mariadb:
+        image: "mariadb:10.11.11"
+        env:
+          MARIADB_ROOT_PASSWORD: "root"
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -98,3 +98,38 @@ jobs:
...
+        image: "mariadb:10.11.11"
+        env:
+          MARIADB_ROOT_PASSWORD: "root"
+          MARIADB_DATABASE: "lobsters_dev"
+        ports:
+          - "3306:3306"
+
+    env:
+      DATABASE_URL: "trilogy://root:root@127.0.0.1/lobsters_dev"
+      RAILS_ENV: "test"
```
caius, line 119: Running as development environment causes an error, so flipped it to test environment. Guessing that'll make no difference for this purpose because the database schema is the database schema is the database schema.

caius:
Totally duplicates #1827, my bad.

PR #1862, opened 2026-01-19
Refactor "save story" for a11y and no JS

fedemp:
Change the "save" button in the UI so it renders as a native form instead of a link. This should allow non-JS users to submit the request, and also prevent  opening the link in a new tab.

Refactor "save story" in the BE so it returns different responses if the request used xhr. If not using xhr, BE will redirect the user back and will render success and error messages in the UI. If using xhr, BE returns a JSON with a success or error message. If the request succeded, the JSON includes the new action for the form, and the new "call to action" for the button.

For users with Javascript enabled: After clicking the "save" button, the button is marked as disabled to prevent double submissions. The success and error messages are appended to a hidden element with `role=alert` so screen readers can announce the result of the operation. In the single detail page for merged stories, change all the stories save button after success; this fixes a bug where after saving one of the merged stories, the others would still render a button with the outdated state.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -294,6 +294,19 @@ summary {
 	cursor: pointer;
 }
 
+/* Remove style from button to so it will look like a link */
+input.btn-link:hover,
+input.btn-link {
+	border: 0;
+	padding: 0;
+	background-color: var(--color-bg);
+	color: var(--color-fg-contrast-4-5);
+	line-height: 1;
+}
+input.btn-link:focus-visible {
+	outline: auto;
+}
+
 #qrsvg {
 	background-color: white;
 	display: inline-block;
@@ -306,6 +319,17 @@ summary {
 	margin: 0;
 }
 
+#toast {
+	position: absolute;
+	width: 1px;
+	height: 1px;
+	padding: 0;
+	margin: -1px;
+	overflow: hidden;
+	clip: rect(0, 0, 0, 0);
+	white-space: nowrap;
+	border-width: 0;
+}
 
 /* header */
 
@@ -711,6 +735,10 @@ li .byline a.story_has_suggestions {
 	color: var(--color-lobsters-fg-has-suggestions);
 }
 
+li .byline form {
+	display: inline-block;
+}
+
 .last_read_newest {
 	border-bottom: 2px solid var(--color-fg-shape);
 	border-top: 0;
@@ -730,7 +758,7 @@ li .byline a.story_has_suggestions {
 	opacity: 0.25;
 }
 
-li.story.saved a.saver {
+li.story.saved .saver .btn-link {
 	color: var(--color-fg-affirmative);
 }
 
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -318,26 +318,64 @@ class StoriesController < ApplicationController
 
   def save
     if !(story = find_story)
-      return render plain: "can't find story", status: 400
+      flash[:error] = "Can't find story"
+    elsif story.merged_into_story
+      flash[:error] = "Story has been merged"
     end
 
-    if story.merged_into_story
-      return render plain: "story has been merged", status: 400
+    if flash[:error]
+      respond_to do |format|
+        format.html { redirect_to "/" }
+        format.json {
+          render json: flash.to_hash, status: 400
+          flash.clear
+        }
+      end
+      return
     end
 
     SavedStory.save_story_for_user(story.id, @user.id)
 
-    render plain: "ok"
+    flash[:action] = story_unsave_path(story.short_id)
+    flash[:success] = "Story saved."
+    flash[:cta] = "unsave"
+
+    respond_to do |format|
+      format.html { redirect_back_or_to Routes.title_path(story) }
+      format.json {
+        render json: flash.to_hash, status: :ok
+        flash.clear
+      }
+    end
   end
 
   def unsave
     if !(story = find_story)
-      return render plain: "can't find story", status: 400
+      flash[:error] = "Can't find story."
+
+      respond_to do |format|
+        format.html { redirect_to "/" }
+        format.json {
+          render json: flash.to_hash, status: 400
+          flash.clear
+        }
+      end
+      return
     end
 
     SavedStory.where(user_id: @user.id, story_id: story.id).delete_all
 
-    render plain: "ok"
+    flash[:action] = story_save_path(story.short_id)
+    flash[:success] = "Story unsaved."
+    flash[:cta] = "save"
+
+    respond_to do |format|
+      format.html { redirect_back_or_to Routes.title_path(story) }
+      format.json {
+        render json: flash.to_hash, status: :ok
+        flash.clear
+      }
+    end
   end
 
   def check_url_dupe
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,3 +1,5 @@
+// @ts-check
+
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 
 // in console: app await import("application")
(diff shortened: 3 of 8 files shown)


fedemp:
Make sure to unmute the video.

[Kooha-2026-01-18-22-51-31.webm](https://github.com/user-attachments/assets/19ac6b47-26c2-4304-a1e8-6357d786b75d)

[Kooha-2026-01-18-22-54-55.webm](https://github.com/user-attachments/assets/ca061cdc-2a00-4a40-a661-69fb09baae89)

----

Notifications in the UI if not using JS.

Screenshot From 2026-01-17 14-03-12

app/javascript/application.js
```diff
@@ -1,3 +1,5 @@
+// @ts-check
```
fedemp, line 1: Needed for Typescript language server, to provide completion and linting, and also so it will pick JSDoc comments.

pushcx: I don't mind having it, sure.

app/assets/stylesheets/application.css
```diff
@@ -295,6 +295,19 @@ summary {
 	cursor: pointer;
 }
 
+/* Remove style from button to so it will look like a ink */
```
pushcx: typo, "like a link"

app/controllers/stories_controller.rb
```diff
@@ -311,27 +311,59 @@ def unhide
   end
 
   def save
-    if !(story = find_story)
-      return render plain: "can't find story", status: 400
-    end
+    if (!( story = find_story ) || story.merged_into_story)
+      flash[:error] = (story = find_story) ? "Can't find story." : "Story has been merged."
```
pushcx: I don't think combining these two guard clauses is an improvement because this next line of code needs to redo the test to figure out which error message to show.

app/javascript/application.js
```diff
@@ -25,33 +27,35 @@ export const on = (eventTypes, selector, callback) => {
...
 export const parentSelector = (target, selector) => {
   let parent = target.closest(selector);
-  if (parent === null) {
+  if (parent == null) {
     throw new Error(`Did not match a parent of ${target} with the selector ${selector}`);
   }
   return parent;
 };
 
-export const qS = (context, selector) => {
```
pushcx, line 40: Why remove the `export`s?

pushcx:
Thanks for opening this draft. It's nice to make JS optional in one more place, and much better to make this accessible.

I think the toast pattern is probably not the way to go for accessibility, I've seen a few real negative writeups: https://primer.style/accessibility/patterns/accessible-notifications-and-messages/#toasts

app/assets/stylesheets/application.css
fedemp: Fixed. Thanks.

app/controllers/stories_controller.rb
fedemp: I have refactored it so it will check if `flash[:error]` was set. Let me know if I should take another approach.

app/javascript/application.js
fedemp: None of these functions are `imported` from other files, so there's no need for the `export`.

fedemp:
> Thanks for opening this draft. It's nice to make JS optional in one more place, and much better to make this accessible.
> 
> I think the toast pattern is probably not the way to go for accessibility, I've seen a few real negative writeups: https://primer.style/accessibility/patterns/accessible-notifications-and-messages/#toasts

I'm using "toast" for a lack of better word actually. These alerts are not visible so there's no need to make them dismissable and can't make content jump. After the "save" action, there's no place to redirect the user if using XHR.  It's plain text without interactions so no need for focus. It's "disruptive to a user’s regular workflow or working context" because we need to alert ~~that the action was successful, and just the button changing its label looks unprofessional~~ if the request failed. 

Edit: I removed alerts on success and retained only for errors. On success, it should be enough with the button changing its label.

app/javascript/application.js
pushcx: I think it was there so that I could test in the browser console? I left myself a comment at the top of the file how to get the site code into the console namespace so I can run functions as I debug, but now that I just tested it and got a syntax error.

I'm not at all a js expert, can you help me out of this situation by updating that comment? Maybe the exports are indeed all redundant. I have no confidence in this area.

app/javascript/application.js
```diff
@@ -95,8 +99,107 @@ export const removeExtraInputs = () => {
...
+			method: "POST",
+			body: new FormData(form),
+			headers: {
+				Accept: "application/json",
+			},
+		});
+
+		/**
+		 * @type {unknown}
+		 */
```
```suggestion
		/**
		 * @type {unknown}
		 */
```

I'm narrowly OK with these annotations being added to functions, but I don't want any inside, that's really distracting.

app/javascript/application.js
```diff
@@ -310,25 +413,58 @@ export class _LobstersFunction {
...
-      act = "save";
-      li.classList.add("saved");
-      saverEl.innerHTML = "unsave";
+    const { target, submitter } = ev;
+
+    if (
+      !(
+        target instanceof HTMLFormElement && submitter instanceof HTMLInputElement
+      )
+    ) {
```
```suggestion
    if (!(target instanceof HTMLFormElement && submitter instanceof HTMLInputElement)) {
```

pushcx: What situation is this guard clause for? Why would the event fire without these set to valid elements?

app/javascript/application.js
```diff
@@ -310,25 +413,58 @@ export class _LobstersFunction {
...
+    const story = parentSelector(target, ".story");
+    const forms = Array.from(qSA(story, ".saver")).filter(
+      (f) => f instanceof HTMLFormElement,
+    );
+    const btns = Array.from(qSA(story, ".saver input[type='submit']")).filter(
+      (b) => b instanceof HTMLInputElement,
+    );
+
+    for (const b of btns) {
+      if (b !== submitter) {
```
pushcx, line 410: Why not disable the one the user clicked?

app/javascript/application.js
```diff
@@ -95,8 +99,107 @@ export const removeExtraInputs = () => {
...
+
+/**
+ * @param {string} msg
+ * @param {number} [timeout = 0]
+ */
+function notify(msg, timeout = 0) {
+	window.setTimeout(() => {
+		const toast = qS("#toast");
+		if (toast) {
+			toast.textContent = "";
```
```suggestion
			toast.textContent = "";
```

Replaced on next line.

app/javascript/application.js
```diff
@@ -549,6 +685,13 @@ onPageLoad(() => {
     }
   }
 
+  // Append an element where xhr fetches can publish their results. Used by screen readers.
+  const toast = document.createElement("div");
+  toast.role = "alert";
+  toast.ariaLive = "assertive";
```
```suggestion
  toast.ariaLive = "polite";
```

Let's not interrupt what the user was doing if the server is slow in responding; a notification that a story was saved is not particularly important.

Thanks for @chronoDave for noticing this on [office hours stream](https://push.cx/stream/2026-01-22).

pushcx (changes requested):
If it's not a toast, and I see how it's basically invisible - I think it'd be clearer to put the error message inline in the story byline where the error is. Setting the area-live on it will prompt the screenreader to pick it up, and it even works for non-screenreader users.

I left a couple questions and some minor cleanup.

app/javascript/application.js
fedemp: Actually you were right. I never thought of that trick because I'm used to work with large (React) bundles.

This has been reverted.

app/javascript/application.js
fedemp: I've removed this in particular. FWIW there's a type guard just after this that checks if we receive a JSON formatted error or some generic text error.

app/javascript/application.js
fedemp: This is Typescript being extra cautious. I refactored it to do a type casting using JSDoc instead of type guarding. I understand that you may be against inline type annotations, but this simple cast improves the developer experience.

app/javascript/application.js
fedemp: Changed. :+1:

app/javascript/application.js
fedemp: Because I chose to let the `asyncFormSubmit` handle disabling the submitter. But "save" and "hide" are different because there is more than one of each for merged stories, so when the user clicks "save" in one, we need, as a side effect, to disable all other.

app/javascript/application.js
fedemp: Let me investigate this a little more. The original concern was that, if were receive the same message twice, the new one would not be announced. BUT given how JS works, the browser may not announce the intermediate step of clearing the alert.

fedemp:
> If it's not a toast, and I see how it's basically invisible - I think it'd be clearer to put the error message inline in the story byline where the error is. Setting the area-live on it will prompt the screenreader to pick it up, and it even works for non-screenreader users.

If we have a live region for each post, screen readers would list 15 different live regions on the homepage, one for each "save" button on the home page. 30, if "hide" gets its own region. And so on. 

Aside from that, if you would like to show the error text next to the button, it can be done too.

app/javascript/application.js
fedemp: Tried locally with Orca and it's smart enough to announce the new but equal message.

pushcx:
OK, thanks for all your explanations, and of course for this improvement. I'm going to merge this down and then bikeshed the `#toast` name to `#aria-announce` because otherwise, when I see it a few years from now I'll be puzzled because we don't use the toast UI pattern.

Inline errors would probably break the byline layout. Which maybe nudges that maybe `#aria-announce` is a missed oportunity and we should call it something more like `#dynamic-flash-error` and put it in the `layouts/application.rb` right under where we print the Rails flash. That does mean it could appear above the user's scroll point (especially if the server responds slowly) but it least it would appear at all?

Having thought out loud, I'm going to merge this down as planned and let me know what you think of having it on every page.

PR #1865, opened 2026-01-19
Add min and max length FE validation for direct messages

fedemp:
Based on https://github.com/lobsters/lobsters/pull/1858


pushcx:
Thanks for adding this hint.

PR #1869, opened 2026-01-21
Fix #1859 - Improve text legibility

### LINKED ISSUE #1859: Improve typography of the Lobste.rs Web site
Hello.

This topic is highly subjective: in my opinion, the Lobste.rs Web site would benefit both from a larger and a nicer font.
I discussed this on IRC and the general opinion was that Lobste.rs needs to be read at 120%-150% browser magnification.
We could not agree on the font issue, and I am not a designer and don't know enough about Web fonts to suggest something specific. Still, Helvetica is a bit bland for my taste. I would prefer something beautiful, readable, with character, but not flashy.

These are a few examples of beautiful fonts; unfortuantely they are print fonts, so you probably can't use these specific ones on a Web site: [Font Share Pairings](https://fontshare.com/pairs).

Thank you.

dannycolin:
Remove all font size hardcoded values and default the body text to 1rem. The use of a relative size unit gives more control to the user since 1rem is respect whatever the value a user has configured in their browser as the root font size.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -9,6 +9,8 @@ But I make an unprincipled exception for :has because it solves so many big prob
 
 :root {
 	/* common color scheme */
+	--font-size-default: 1rem;
+
 	--color-bg: rgb(var(--base-bg));
 	--color-bg-50: rgb(var(--base-bg) / 50%);
 
@@ -50,7 +52,7 @@ html {
 
 body, textarea, input, button {
 	font-family: "helvetica neue", arial, sans-serif;
-	font-size: 10pt;
+	font-size: var(--font-size-default);
 	color: var(--color-fg);
 	line-height: 1.45em;
 }
@@ -71,13 +73,11 @@ li.story div.details span.link a:visited {
 }
 
 h1 {
-	font-size: 12pt;
 	font-weight: bold;
 	margin: 0.5em 0 1em 0;
 	padding: 0;
 }
 h2 {
-	font-size: 11pt;
 }
 
 .clear {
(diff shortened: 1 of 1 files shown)


pushcx:
Relates to #1859

pushcx:
Thanks for doing this. I played with it on a branch and, again, it's nicer than what we have now to just let the browser use all its default sizes. Reading this diff, I can see a lot of unprincipled tinkering that's accumulated over the years like inconsistent units and numbers that are picked by a dev nudging things up and down until going "eh good enough" rather than setting up a font ladder.

This one's going to be disruptive, it already broke my muscle memory for loading the site and immediately holding control to mouse-wheel-scroll the font size up a bit. I know many people have their browser set to impose a zoom on the site so I'd rather just break those settings once than slowly normalize all these.

And maybe this is more evidence in the bucket that we should have a competent designer take a comprehensive look at the site.

dannycolin:
> Reading this diff, I can see a lot of unprincipled tinkering that's accumulated over the years like inconsistent units and numbers that are picked by a dev nudging things up and down until going "eh good enough" rather than setting up a font ladder.

I had the same reaction. A bit of UX work to decide what elements (primary/secondary text, heading, etc) need to be more prominent might be a good first step. :)

madsmtm:
Yeah, this definitely tripped me up. As a bit of concrete feedback, the text on tag labels/buttons feels a bit too large IMO.

nightkr:
So *this* is why the whole site suddenly looked off.

I guess I can't discount that I'm just not used to it, but for some reason this new styling gives me eye strain to an almost painful degree.

In firefox, Zoom Text Only at 90% seems get it mostly back to normal, but annoyingly ZTO isn't tracked separately per-site.

avm99963:
I was also wondering why Lobsters looked so different from before. On mobile, everything looks too big, and I can't zoom out (I guess I'm very used to the previous font size).

PR #1705, opened 2025-08-18
Migrate to SQLite

thomasdziedzic:
Work to support https://github.com/lobsters/lobsters/issues/539

A high level todo:
- [x] Make tests pass using purely SQLite
- [x] Click around the site locally and make sure things are working
- [ ] Performance testing
- [ ] Migration script for data from MariaDB to SQLite

PR #1871, opened 2026-01-23
Migrate to SQLite again

thomasdziedzic:
This is a continuation of https://github.com/lobsters/lobsters/pull/1705 but since it has been marked as stale and I can't reopen it, I'm just going to open a new PR. The latest work that I completed was merging latest main into this branch and resolving all the issues resulting from the merge.

Work to support https://github.com/lobsters/lobsters/issues/539

A high level todo:
- [x] Make tests pass using purely SQLite
- [x] Click around the site locally and make sure things are working
- [x] Performance testing
- [x] Migration script for data from MariaDB to SQLite
- [x] Backup command for sqlite]

Open issues for the following post merge:
- [ ] remove strip_operators as [mentioned](https://github.com/lobsters/lobsters/pull/1871#discussion_r2729463819)
- [ ] remove mariadb specific keywords from [search_parser.rb](https://github.com/lobsters/lobsters/blob/1a7903c8577a870ceb1a1ff332d5a560c83e09b7/app/models/search_parser.rb#L6)
- [ ] remove the [`confidence_order_path` is a kludge](https://github.com/lobsters/lobsters/pull/1871#discussion_r2729373598)


thomasdziedzic:
Here's some initial performance test results:

I ran these under the development environment with mariadb on the main branch and sqlite on the td-sqlite-2 branch.

```
ab -n 1000 -c 2 http://localhost:3000/

mariadb
=======
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    31   36   5.2     34     102
Waiting:       31   36   5.2     34     102
Total:         31   36   5.3     34     102

sqlite
======
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    30   34   5.7     32      92
Waiting:       30   34   5.7     32      92
Total:         30   34   5.7     32      92

```

```
ab -n 1000 -c 2 http://127.0.0.1:3000/s/:a/:b

mariadb
=======
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    45   53   6.2     52     118
Waiting:       45   53   6.2     52     118
Total:         45   53   6.2     52     118

sqlite
======
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    31   35   6.4     33     109
Waiting:       30   35   6.4     33     109
Total:         31   35   6.4     34     109
```

Looks like sqlite has better performance in both of the tests though the home page seems like it's fairly close but there is a 18ms advantage for the individual story in favor of sqlite.

.github/workflows/check.yml
```diff
@@ -19,17 +19,8 @@ jobs:
     strategy:
       fail-fast: false
 
-    services:
```
pushcx, line 25: There's parallel setup in the Dockerfile and .devcontainer. ChaelCodes is the person who knows those best if you need a hand.

app/models/comment.rb
```diff
@@ -512,6 +524,20 @@ def update_associated_caches
     user.refresh_counts!
   end
 
+  def create_fts
```
pushcx: Optional: I expect this is duplicated for story title and storytext, with 3 examples it might be worth drying this up into a Concern.

app/models/comment.rb
```diff
@@ -559,65 +585,53 @@ def self.recent_threads(user)
...
+        from comments c join discussion on c.parent_comment_id = discussion.id
+      )
+      select comments.*
+      from comments
+      join discussion
+        on comments.id = discussion.id
+      order by comments.thread_id desc, discussion.confidence_order_path
+    SQL
+
+    Comment.select("comments.*").from("(#{query}) as comments")
```
pushcx: If `from` returns an `ActiveRecord::Relation` this is fine. Otherwise it'll need to return to the previous `.joins` strategy.

`confidence_order_path` is a kludge that uses strings to work around a limitation of mariadb, which doesn't have an array type available. SQLite does, in json if not elsewhere.

There's a [longer explanation of the problems with this kludge](https://github.com/lobsters/lobsters/issues/1236#issuecomment-1944031476) but in short, because it loses precision, sometimes comments are not sorted/parented correctly. A proper fix would give `confidence_order_path` the type `[[confidence, id]]`: sort by confidence, tiebreak by id, and specify the full id so that if there are two comments with the same confidence the children go under the correct parent.

app/models/search.rb
```diff
@@ -172,15 +169,11 @@ def perform_comment_search!
         val = strip_short_terms(strip_operators(value))
         # if punctuation is replaced with a space, this would generate a terms search
```
pushcx: Comment needs updating/deleting.

app/models/story.rb
```diff
@@ -693,9 +698,10 @@ def update_score_and_recalculate!(score_delta, flag_delta)
             )
         ),
         flags = (select count(*) from votes where story_id = stories.id and comment_id is null and vote = -1),
-        hotness = #{calculated_hotness}
-      WHERE id = #{id.to_i}
+        hotness = ?
+      WHERE id = ?
     SQL
+    Story.connection.exec_update(update_query, nil, [id, (created_at - FLAGGABLE_DAYS.days).utc.iso8601, calculated_hotness, id])
```
pushcx, line 717: Thanks for moving these out to parameters, that'll make brakeman happier. I think there's an option for named parameters, which would be easier to maintain than counting how many `?`s and lining up position in the array.

app/models/story_text.rb
```diff
@@ -32,4 +36,20 @@ def self.cached?(story, &blk)
 
     where(id: story).exists?
   end
+
+  def create_fts
+    ActiveRecord::Base.connection.exec_insert("INSERT INTO story_texts_fts (rowid, title, description, body) values (?, ?, ?, ?)", nil, [id, title, description, body])
+  end
+
+  def update_fts
+    # contentless-delete tables in sqlite require all the columns when updating them, see for more info:
```
pushcx: Useful comment, thanks.

db/migrate/20120701154453_create_invitations.rb
```diff
@@ -1,18 +0,0 @@
-# If you are reading this file because Rails complains that there isn't a
```
pushcx, line 1: Please shove these all into `db/old_migrations`, I look at them all the time to remember how things changed. I know they're all there in the jj history but I like having them to hand.

db/migrate/20250818182728_add_fts_tables.rb
```diff
@@ -0,0 +1,8 @@
+class AddFtsTables < ActiveRecord::Migration[8.0]
```
pushcx, line 7: Please copy in my grumpy "don't edit migrations" opinion from the original first migration.

db/migrate/20250818183847_add_nocase_collation_to_username.rb
```diff
@@ -0,0 +1,9 @@
+class AddNocaseCollationToUsername < ActiveRecord::Migration[8.0]
+  def up
+    change_column :users, :username, :string, limit: 50, collation: "NOCASE"
```
pushcx, line 3: This should almost certainly be on `categories.category` and `tags.tag` as well.

db/schema.rb
```diff
@@ -57,36 +57,35 @@
...
-    t.bigint "user_id", null: false, unsigned: true
-    t.bigint "parent_comment_id", unsigned: true
-    t.bigint "thread_id", unsigned: true
-    t.text "comment", size: :medium, null: false
+    t.bigint "user_id", null: false
+    t.bigint "parent_comment_id"
+    t.bigint "thread_id"
+    t.text "comment", null: false
     t.integer "score", default: 1, null: false
-    t.integer "flags", default: 0, null: false, unsigned: true
```
pushcx, line 77: The `_id`s don't need to it, but for this field please add a validation that it's a positive number. The voting code updates it by deltas and I want the error if we have a bug.

spec/models/search_spec.rb
```diff
@@ -318,16 +318,16 @@
...
-      expect(s.flatten_title({term: "as'df"})).to eq("as\\'df")
-      expect(s.flatten_title({term: "hj\"kl"})).to eq("hj\\\"kl")
-      expect(s.flatten_title({quoted: [{term: "cat'"}, {term: "scare"}]})).to eq("\"cat\\' scare\"")
+      expect(s.flatten_title({term: "as'df"})).to eq("\"as''df\"")
+      expect(s.flatten_title({term: "hj\"kl"})).to eq("\"hj\"\"kl\"")
+      expect(s.flatten_title({quoted: [{term: "cat'"}, {term: "scare"}]})).to eq("\"cat' scare\"")
     end
   end
 
   describe "#strip_operators" do
```
pushcx, line 327: You built your queries better using `?` to have ActiveRecord manage quoting. If possible, I'd like to entirely delete `strip_operators` and rely on AR's injection, as it's going to be better-tested (and battle-tested) than anything I could write.

Gemfile
```diff
@@ -7,7 +7,6 @@ gem "rails"
 # database and caching
 gem "solid_cache"
 gem "sqlite3"
-gem "trilogy"
```
pushcx, line 10: 🎉

pushcx (commented):
Wow! This is a lot, and you're hitting most of the hairy parts of the codebase along the way. I left some comments on [today's stream](https://push.cx/stream/2026-01-26) and I'm really curious to see the performance test on my machine with prod data, and then of course I'm excited to see this code in prod.

pushcx:
For your to-do list, we need backups. The [implementation in prod now for mariadb](https://github.com/lobsters/lobsters-ansible/blob/1782010287b589296542671917b7a2fd6f036635/roles/mariadb/tasks/main.yml#L49) is a cron job that restic stores in backblaze. The sqlite database is Just a file on disk so we could add it to the [current restic job](https://github.com/lobsters/lobsters/blob/maion/app/jobs/restic.job.rb), but I'd like to see a documentation somewhere that it'll be in a consistent state even as prod writes are happening. Maybe there's a proper tool or just `cp` to snapshot?

pushcx:
For migration, here's a command I've used to dump prod: `mariadb-dump -v -u lobsters --opt --skip-add-drop-table --allow-keywords --comments --compress --dump-date --force --routines --single-transaction --quick`

.github/workflows/check.yml
thomasdziedzic: Removed the mariadb config from dockerfile and devcontainer.

app/models/comment.rb
thomasdziedzic: refactored out into a concern

Gemfile
thomasdziedzic: :tada:

app/models/comment.rb
thomasdziedzic: Converted to using `joins`. This should help ensure the stability of the ordering.

~I'll take a look at fixing the confidence_order_path but that might be a good candidate for doing it in another PR after this potentially gets merged.~

I looked at this and if I understand the request, it's to create an array of arrays for each comment for it's confidence: `[[highest_confidence, id1], [next_highest_confidence, id2], ...]` and then sorting each comment by this array. I could try to use [json_array](https://sqlite.org/json1.html#jarray) by concatenating normalized (left padded) strings together and then sorting the resulting string, though I have a feeling like that might not be what you want. The reason why I'm mentioning string sorting is because afaik, sqlite represents json as strings and so an array of arrays in json would actually be represented with a single string and that's what we would have to work with for sorting.

app/models/story_text.rb
thomasdziedzic: :+1:

app/models/story.rb
thomasdziedzic: I tried using named params, but it seems like exec_update doesn't support named params, only the `?` placeholder:

```
[4] pry(main)> Story.connection.exec_update("select id from stories where id = :story_id", nil, {story_id: 1})
TypeError: can't cast Array (TypeError)

        else raise TypeError, "can't cast #{value.class.name}"
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
from /home/tom/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/activerecord-8.0.3/lib/active_record/connection_adapters/abstract/quoting.rb:105:in 'ActiveRecord::ConnectionAdapters::Quoting#type_cast'
```

app/models/search.rb
thomasdziedzic: deleted since sqlite doesn't use against

db/migrate/20120701154453_create_invitations.rb
thomasdziedzic: mv'ed

db/migrate/20250818182728_add_fts_tables.rb
thomasdziedzic: copy'd

db/migrate/20250818183847_add_nocase_collation_to_username.rb
thomasdziedzic: added NOCASE to categories.category and tags.tag

db/schema.rb
thomasdziedzic: added a validation.

thomasdziedzic:
> For your to-do list, we need backups. The [implementation in prod now for mariadb](https://github.com/lobsters/lobsters-ansible/blob/1782010287b589296542671917b7a2fd6f036635/roles/mariadb/tasks/main.yml#L49) is a cron job that restic stores in backblaze. The sqlite database is Just a file on disk so we could add it to the [current restic job](https://github.com/lobsters/lobsters/blob/maion/app/jobs/restic.job.rb), but I'd like to see a documentation somewhere that it'll be in a consistent state even as prod writes are happening. Maybe there's a proper tool or just `cp` to snapshot?

The command for backing up an online sqlite3 database is:

```
sqlite3 my_database.sqlite3 ".backup 'backup_file.sqlite3'"
```

There isn't much docs on the sqlite page, but here's a good guide I found online: https://oldmoe.blog/2024/04/30/backup-strategies-for-sqlite-in-production/

thomasdziedzic:
I was able to take a development mariadb snapshot and load it into sqlite using the following steps:

1. `mariadb-dump -v -u root -h 127.0.0.1 -p --compact --no-create-info --opt --skip-add-drop-table --allow-keywords --comments --compress --dump-date --force --routines --single-transaction --quick --ignore-table=lobsters_development.ar_internal_metadata lobsters_development > dump.sql`
2. `git clone https://github.com/dumblob/mysql2sqlite.git`
3. `./mysql2sqlite/mysql2sqlite dump.sql > converted-dump.sql`
4. `cat converted-dump.sql | sqlite3 db/development/primary.sqlite3`

spec/models/search_spec.rb
thomasdziedzic: Could we keep existing functionality in place since I don't think it hurts to keep strip_operators and do a separate search refactor later after this PR? There's still some mariadb logic that exists in the search parser which I don't want to touch because search works as is and search will probably need a post merge cleanup as a result.

app/models/story.rb
pushcx: That's a really unfortunate limitation. I spent a minute poking around Rails source but can't guess why it didn't work or if there's a slightly different function you could call to get that functionality.

We at least have some smoke tests around this function so if parameters get out of order or the `binds` array length doesn't match the number of `?`s, we'll get catch it before prod.

spec/models/search_spec.rb
pushcx: OK. Can I ask you to take notes as you go on what refactorings you'd like to see? And then open an issue for us to make sure we work through it. Or inline 'TODO' notes work, whatever's easier. I just want to make sure it's not only in your head. :)

pushcx:
The backup script is `app/jobs/restic_job.rb`.

I'm testing the conversion script now.

pushcx:
The conversion script has not progressed since the stall at 547538 lines of output observed on [yesterday's office hours stream](https://push.cx/stream/2026-01-29). So that's ~24 hours pinning one CPU with no additional records appearing in the `sqlite.sql`.

Complete output, not sure if this message coincides with the stall:

```
$ ./mysql2sqlite/mysql2sqlite mariadb.sql > sqlite.sql
awk: ./mysql2sqlite/mysql2sqlite:130: (FILENAME=mariadb.sql FNR=111578) warning: Invalid multibyte data detected. There may be a mismatch between your data and your locale
```

I had eyeballed it and guessed it was the start of the `comments` table, but `tail`ing the file it's comment 84424, which is an unexceptional single-sentence comment. The only thing that catches my eye as possibly invalid data is `confidence_order_path` - perhaps it's attempting to interpret that bytepacked nonsense as utf8? Those bytes are 0xff, 0xff, 0xc8.

spec/models/search_spec.rb
thomasdziedzic: opened a post merge todo list in the PR's description.

thomasdziedzic:
> The backup script is `app/jobs/restic_job.rb`.

Added a backup script for sqlite into the restic job

thomasdziedzic:
> The conversion script has not progressed since the stall at 547538 lines of output observed on [yesterday's office hours stream](https://push.cx/stream/2026-01-29). So that's ~24 hours pinning one CPU with no additional records appearing in the `sqlite.sql`.
> 
> Complete output, not sure if this message coincides with the stall:
> 
> ```
> $ ./mysql2sqlite/mysql2sqlite mariadb.sql > sqlite.sql
> awk: ./mysql2sqlite/mysql2sqlite:130: (FILENAME=mariadb.sql FNR=111578) warning: Invalid multibyte data detected. There may be a mismatch between your data and your locale
> ```
> 
> I had eyeballed it and guessed it was the start of the `comments` table, but `tail`ing the file it's comment 84424, which is an unexceptional single-sentence comment. The only thing that catches my eye as possibly invalid data is `confidence_order_path` - perhaps it's attempting to interpret that bytepacked nonsense as utf8? Those bytes are 0xff, 0xff, 0xc8.

I did try to repro the stall but I wasn't able to. Checking the [mysql2sqlite](https://github.com/mysql2sqlite/mysql2sqlite/issues) issues, there's a bunch of problems reported around hex values, binaries, blobs, etc that it makes me think that fixing this might be difficult especially since I don't have a reproducible test case to work off of. One option would be to just write a custom dump from mariadb to sqlite script in ruby which would probably be a lot cleaner.

thomasdziedzic:
I wrote a [rake task](https://gist.github.com/thomasdziedzic/bef5648f59dee89bb14b230b13b57951) (which you can put into `lib/tasks/migrate.rake`) which will generically dump all the data into a yaml file and load all the data into a database using these rake tasks:

1. `dump_db`: dumps all the data in a rails database into a file `dump.yml`
2. `load_db`: loads all the data in a file `dump.yml` into the rails database'

Here's how I tested it:

1. Dumping data: in the lobsters mariadb instance, run `rails dump_db`
2. Loading data: in the lobsters sqlite instance, run `rails load_db` after copying `dump.yml` from the mariadb rails instance. I used `rails db:drop db:schema:load load_db` as the test command for building out the script if you're wondering how to test it easily.

In addition, you need to manually populate the fts tables on initial load with the following 2 sql statements:
```sql
insert into comments_fts(rowid, comment) select id, comment from comments;
insert into story_texts_fts(rowid, title, description, body) select id, title, description, body from story_texts;
```

EDIT: I removed the old dump everything into memory task in favor of a streaming task which should allow you to dump the whole prod database. Old non-streaming task left [here](https://gist.github.com/thomasdziedzic/9a9d50abf872b01833e554cfa2378a52) for reference.

app/models/comment.rb
thomasdziedzic: moved this to the post merge todo list

pushcx:
@chamlis found the bug: https://github.com/rails/rails/issues/55373

One row `story_texts` and two `comments` have null bytes in them. The first one is a hex dump, the other two are pasting out of PDFs. I'll manually delete them from prod and modlog the action. (We ran down the other text entry fields and didn't spot any, but it's possible there's more I didn't immediately think of.)

@nogweii caught that We need to catch this error in the future. StoryText should probably just strip them, but how about comments and user-entered text? I'm leery of silently stripping them, but it's also not a particularly clear error message when a user's browser may not be print anything in the textarea for the null byte.

pushcx:
And one option occurred to me: Rails supports connecting to multiple databases. So rather than edit `config/database.yml` to switch `primary` from `trilogy` to `sqlite3`, we could _add_ the sqlite3 primary under another name, and have one process directly copy from mariadb to sqlite3, which would save the encoding risk of round-tripping out to a file.

My nightmare is still that there's some subtle encoding bug in 20 comments that we lose data to.

thomasdziedzic:
> @chamlis found the bug: [rails/rails#55373](https://github.com/rails/rails/issues/55373)

Wow, that's an impressive find @chamlis !
 
> One row `story_texts` and two `comments` have null bytes in them. The first one is a hex dump, the other two are pasting out of PDFs. I'll manually delete them from prod and modlog the action. (We ran down the other text entry fields and didn't spot any, but it's possible there's more I didn't immediately think of.)

Nice, so this will just turn out to be a data cleanup task. I think we don't even need to delete that data, we just need to delete the null byte since that's the main limitation I see with sqlite.

> @nogweii caught that We need to catch this error in the future. StoryText should probably just strip them, but how about comments and user-entered text? I'm leery of silently stripping them, but it's also not a particularly clear error message when a user's browser may not be print anything in the textarea for the null byte.

For now it'll throw unhandled exceptions on those user inputs after migrating to sqlite. A lost comment or story_text with a null byte isn't the end, and it honestly feel like it shouldn't be possible unless you're trying to do something malicious, even though it's technically valid utf8.

> And one option occurred to me: Rails supports connecting to multiple databases. So rather than edit `config/database.yml` to switch `primary` from `trilogy` to `sqlite3`, we could _add_ the sqlite3 primary under another name, and have one process directly copy from mariadb to sqlite3, which would save the encoding risk of round-tripping out to a file.
>
> My nightmare is still that there's some subtle encoding bug in 20 comments that we lose data to.

My understanding is that mariadb is completely utf8mb4 (utf8) encoded, yaml serialization is in utf8 and sqlite uses utf8. As far as I can tell this only showed up because SQLite [doesn't accept null bytes in strings](https://www.sqlite.org/invalidutf.html) even though [it should technically be possible](https://stackoverflow.com/questions/33060368/can-utf-8-string-contain-nulls-0x0-in-it). Otherwise I don't see an issue with the pipeline between mariadb -> yaml -> sqlite since it's all using utf8. We could do a dump of prod and compare it with a dump of mariadb loaded from sqlite.

chamlis:
> > @chamlis found the bug:
> > [rails/rails#55373](https://github.com/rails/rails/issues/55373)
> 
> Wow, that's an impressive find @chamlis !

Thank you, but full credit to you for putting in the hard work on the port! I just got lucky with a quick search after @pushcx mused about potential null-bytes.

pushcx:
New bug. I ran as `bundle exec rails load_db > db.log 2>&1`. It prints the entire batch, so the full file is >12kloc. The `head` and `tail` of that log is:

```
, [2026-02-17T13:33:34.450219 #1684966]  WARN -- : heinous_inline_partial initialized, found: {"app/views/comments/_threads.html.erb" => "app/views/comments/_comment.html.erb", "app/views/home/index.html.erb" => "app/views/stories/_listdetail.html.erb", "app/views/search/index.html.erb" => "app/views/stories/_listdetail.html.erb"}
bin/rails aborted!
ActiveRecord::StatementInvalid: SQLite3::SQLException: unrecognized token: "'Your comment on [What are you reading this month?](http://localhost:3000/s/zasxvy/what_are_you_reading_this_month) has been moderated: (ActiveRecord::StatementInvalid)

> The cleanroom methodology is based in two pillars:
>
>  * Formal design and requirements methods (verification)
> ":
INSERT INTO "messages" ("id","created_at","author_user_id","recipient_user_id","subject","body","short_id","deleted_by_author","deleted_by_recipient","hat_id","token") VALUES ([redacted for #1871, this is a real DM]
```

```
>
>  * Formal design and requirements methods (verification)
>
/home/pushcx/code/lobsters/lib/tasks/migrate.rake:28:in 'block (2 levels) in Migrate#load'
/home/pushcx/code/lobsters/lib/tasks/migrate.rake:27:in 'block in Migrate#load'
/home/pushcx/code/lobsters/lib/tasks/migrate.rake:26:in 'IO.open'
/home/pushcx/code/lobsters/lib/tasks/migrate.rake:26:in 'Migrate#load'
/home/pushcx/code/lobsters/lib/tasks/migrate.rake:41:in 'block in <top (required)>'
Tasks: TOP => load_db
(See full trace by running task with --trace)
```

I pulled up `dump.yml` and the message looks like any other. It's unlucky it's a DM so I can't paste it. vim says line `104262262/136655856` so on the positive side, the load _did_ get ~76% of the way. I don't immediately have an idea what's wrong here.

(I thought more about my idea of connected to both mariadb and sqlite at the same time to save the YAML round trip. It looks like [`connects_to`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connects_to) might permit it? In `ApplicationRecord`: `connects_to database: { reading: :mariadb, writing: :primary }` and then the rake task would `find_in_batches` and then loop the batch to save the records. Instantiating 9,095,807 ActiveRecord objects would not be fast, but might be less painful than debugging yaml.)

thomasdziedzic:
> New bug. I ran as `bundle exec rails load_db > db.log 2>&1`. It prints the entire batch, so the full file is >12kloc. The `head` and `tail` of that log is:
> 
> ```
> , [2026-02-17T13:33:34.450219 #1684966]  WARN -- : heinous_inline_partial initialized, found: {"app/views/comments/_threads.html.erb" => "app/views/comments/_comment.html.erb", "app/views/home/index.html.erb" => "app/views/stories/_listdetail.html.erb", "app/views/search/index.html.erb" => "app/views/stories/_listdetail.html.erb"}
> bin/rails aborted!
> ActiveRecord::StatementInvalid: SQLite3::SQLException: unrecognized token: "'Your comment on [What are you reading this month?](http://localhost:3000/s/zasxvy/what_are_you_reading_this_month) has been moderated: (ActiveRecord::StatementInvalid)
> 
> > The cleanroom methodology is based in two pillars:
> >
> >  * Formal design and requirements methods (verification)
> > ":
> INSERT INTO "messages" ("id","created_at","author_user_id","recipient_user_id","subject","body","short_id","deleted_by_author","deleted_by_recipient","hat_id","token") VALUES ([redacted for #1871, this is a real DM]
> ```
> 
> ```
> >
> >  * Formal design and requirements methods (verification)
> >
> /home/pushcx/code/lobsters/lib/tasks/migrate.rake:28:in 'block (2 levels) in Migrate#load'
> /home/pushcx/code/lobsters/lib/tasks/migrate.rake:27:in 'block in Migrate#load'
> /home/pushcx/code/lobsters/lib/tasks/migrate.rake:26:in 'IO.open'
> /home/pushcx/code/lobsters/lib/tasks/migrate.rake:26:in 'Migrate#load'
> /home/pushcx/code/lobsters/lib/tasks/migrate.rake:41:in 'block in <top (required)>'
> Tasks: TOP => load_db
> (See full trace by running task with --trace)
> ```
> 
> I pulled up `dump.yml` and the message looks like any other. It's unlucky it's a DM so I can't paste it. vim says line `104262262/136655856` so on the positive side, the load _did_ get ~76% of the way. I don't immediately have an idea what's wrong here.
> 
> (I thought more about my idea of connected to both mariadb and sqlite at the same time to save the YAML round trip. It looks like [`connects_to`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connects_to) might permit it? In `ApplicationRecord`: `connects_to database: { reading: :mariadb, writing: :primary }` and then the rake task would `find_in_batches` and then loop the batch to save the records. Instantiating 9,095,807 ActiveRecord objects would not be fast, but might be less painful than debugging yaml.)

I'll try writing an enhancement to the script tomorrow that will bisect the problematic record.

chamlis:
One aspect I didn't fully appreciate about the null-bytes is that SQLite itself has no problem storing them in `TEXT` columns. They [don't recommend](https://sqlite.org/nulinstr.html) doing so, though.
```
sqlite> CREATE TABLE foo (a TEXT) STRICT;
sqlite> INSERT INTO foo (a) VALUES (char(0));
sqlite> SELECT COUNT(*) FROM foo;
COUNT(*)
--------
1
```

The issue we were having before was that Rails generates an `INSERT` query with the values inline and doesn't treat the null-bytes specially, so they end up in the raw query between quotes. SQLite either than can't parse that, or treats that as the end of the query string, I'm not sure which.

I wrote a quick python script to see if SQLite returns the same error with any other raw bytes in the query, and it seems it's only null bytes that are an issue.

<details>
<summary>Script</summary>

```python
import sqlite3

db = sqlite3.connect(":memory:")

db.execute("CREATE TABLE foo (a TEXT) STRICT")

for c in range(256):
    if c == ord("'"): continue # rails replaces these

    try:
        db.execute(f"INSERT INTO foo (a) VALUES ('{chr(c)}')")
        db.commit()
    except Exception as e:
        print(f"Error on {c}")
        print(e)
        print()
```
</details>

My gut is telling me that a null-byte slipped in somewhere else. Would it be possible to grep the YAML file for null-bytes (looks like that escapes them as `\0`), which would cover all of the columns at once? I'm not sure how human-readable that file is.

pushcx:
Inconveniently, since the site topic is programming, we have people posting comments with code snippets that match 3,771 times according to `grep '\\0' dump.yml | wc -l`. That's more than I can manually review. Maybe I am getting the wrong level of quoting on that search?

I tried to produce a yaml file with a null byte and reference to a null byte to look at: `require 'yaml'; File.write('bytes.yml', YAML::safe_dump(["a\0b", "a\\0b"]))`. I got a file reading:

```yaml
--
- "a\0b"
- a\0b
```

So I tried to search for `\0` inside a quoted value with `grep '^ *- ".*[^\\]\\0' dump.yml  | head` and got no results.

This latest exception happened in `messages`. I did another check in mariadb, there aren't any null bytes in the subject or body of any messages (`select count(*) from messages where subject like '\0' or body like '\0';`). (The only other string columns there are `short_id` and `token`; no code path should generate null bytes in those but I also confirmed that they don't have them.)

I have learned from painful experience to not be confident when there's layers of quoting and character set issues, but having checked from several different layers I am pretty sure there are no longer null bytes in text columns in the dump.

pushcx:
I got it. It's so ridiculous.

On [Monday](https://push.cx/stream/2026-02-16) I wrote a migration in 3e729cbe099fcf3f32bebb0acaccf32e5c5a553b to delete the null bytes from the database for this PR. I had that migration create a Moderation to publicly log the edit to fulfill the site's commitment to transparency.

Creating a Moderation related to a Story or Comment generates a Message to the submitter/author explaining the edit. It includes a diff of the edit. The 'old' version would contain a null byte. So deleting the null bytes from the database created more null bytes in the database.

My query was wrong, I wrote `body like '\0'` when I should've written `body like '%\0%'`.

I'm currently cleaning up the null bytes in the messages, doublechecking the mod log, story texts, etc.

chamlis:
Thanks for running the greps, and good job tracking it down in the end. The kind of drily ironic issue only unthinking machines could come up with.

thomasdziedzic:
> I got it. It's so ridiculous.

Does that you mean figured out all the issues with loading the database dump into sqlite or are there more issues? I was planning on writing a bisecting algorithm to narrow down errors to the exact record the error is occurring on but if you already figured out everything then I don't want to spend time on something you wont use.

pushcx:
I think this might be everything, but I've run out of time today between other commitments. I'll try to finish running a dump/load cycle when I return home later this evening, to have more info before tomorrow morning's stream.

pushcx:
It loaded just fine. :smile: :tada:

So I'll roll into office hours in 2 hours with mariadb + sqlite having the exact same data, if we want to spend a minute thinking up how to make sure there aren't any character encoding or other sutble corruption issues I have anxiety about.

pushcx:
```
$ diff mariadb.dump.yml sqlite.dump.yml > diff                 
$ wc -l diff                                                     
0 diff                                            
```

I had to delete the contents of `comments.confidence`, but that's a calculated value.

thomasdziedzic:
That's good news!
I just resolved the merge conflicts so we should be good for tomorrow.
Looks like there's a test failure but it's also on [main](https://github.com/lobsters/lobsters/commits/main/).

pushcx:
Thanks for all this work, what a big project!

Any readers: we're on a call and going to migrate prod now.

Issue #1872, opened 2026-01-23
Further improve typography of the Lobste.rs Web site


earthboundkid:
This is a follow up to #1859. I don't know if you prefer to start a new issue or reopen an old one, so I'm opening a new one and copying [my comments](https://github.com/lobsters/lobsters/issues/1859#issuecomment-3790163279) from there.

- - - -

This is subjective, but I find the new settings extremely difficult to read. Also there is wild inconsistency between the size of text in the text boxes (too small) and article listings (very wide).

<img alt="Image" width="1575" height="1077" src="https://private-user-images.githubusercontent.com/222245/539763945-46603945-fe75-495e-92a9-5cf1877d2349.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjkxNzQ3MTIsIm5iZiI6MTc2OTE3NDQxMiwicGF0aCI6Ii8yMjIyNDUvNTM5NzYzOTQ1LTQ2NjAzOTQ1LWZlNzUtNDk1ZS05MmE5LTVjZjE4NzdkMjM0OS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwMTIzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDEyM1QxMzIwMTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0wNjE1ZmE0OTlkZjIzNmViMDJkYzIwNTIyNTRjN2NhZWQ1YzM2ZGZmODgxNDNhNDM4NzAzZjgyMjYxODM1NzdlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.GD8dF2JiGy2Id2WWU_6BDaFTdYqPtL14zr0xStzJOFg">

As you can see, the textbox text in my browser (Safari 26.2) is way too small.

[...]

PR #1873, opened 2026-01-23
Set textarea font-size to base font size

fedemp:
Font size is too small in Safari. Affects both textarea and text fields. See https://github.com/lobsters/lobsters/issues/1872#issuecomment-3791880878

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -42,7 +42,6 @@ But I make an unprincipled exception for :has because it solves so many big prob
 /* generics */
 
 html {
-	font-size: var(--font-size-default);
 	/* force a vertical scrollbar to avoid page-shifting */
 	overflow-y: scroll;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/application.css
```diff
@@ -42,7 +42,6 @@ But I make an unprincipled exception for :has because it solves so many big prob
 /* generics */
 
 html {
-	font-size: var(--font-size-default);
```
pushcx, line 45: This has to be here to set up the base size correctly. @dannycolin's original PR had it on the `body` tag and I saw very odd effects trying to create size variations. `rem` is explicitly "relative to the root element", which is `html`.

pushcx:
I committed 071efe1c416e38998219d56682e735962def6872 to start addressing this, I randomly saw on bsky that Safari has font changes on these elements and that `inherit` is a good, non-repetitive way to ensure consistent fonts.

I think this might address everything you identified in this PR? Does that sound right, or is there more to fix?

fedemp:
Textarea looks fine now in safari. Maybe those who reported the original issue may have some more comments, but this PR can be closed.

PR #1874, opened 2026-01-23
Split `StatsGraphsJob` into separate jobs.

### LINKED ISSUE #1863: StatsGraphJob gets OOMkilled
The StatsGraphJob is getting OOMkilled in prod. Probably a combination of there being more data and the server being under more load lately. Let's just split the job up and enqueue the 5 `Stats.fill_` jobs separately a couple minutes apart from each other.

ahmed-hany94:
## Description
This applies the suggestion made in [Issue: StatsGraphJob gets OOMkilled #1863.
](https://github.com/lobsters/lobsters/issues/1863)
Closes #1863.

### Changes
- Replaced the `app/jobs/stats_graphs_job.rb` with five _"job"_ for the five methods performed.


config/recurring.yml
```diff
@@ -17,9 +17,21 @@ production:
   sitemap:
     command: 'SitemapGenerator::Interpreter.run(config_file: ENV["CONFIG_FILE"], verbose: false)'
```
pushcx, line 18: You don't need to create 5 jobs, look at this `command:` line. You can give the single line of code for each.

The scheduling you picked is a little tight; these queries are not particularly fast, could you space them out to 3 minutes apart?

ahmed-hany94: So sorry for the mishap.
Should have checked better.
This is definitely something up Ruby's alleyway.
Thank you for your patience.

Issue #1801, opened 2025-11-14
Tags field on New Story form misaligned in Safari


caius:
Logged in, submitting a story on /stories/new. In Safari I see the tags field without a label and left-aligned with the labels with the other labels rather than aligned with the other fields. My memory might be faulty but I think the inputs used to line up.

I think this is specifically a Safari bug, as it shows the tag label in Chrome.

Safari screenshot:

<img width="817" height="496" alt="Image" src="https://github.com/user-attachments/assets/aac6082b-0236-4750-836d-b739c1ca64ca" />

Chrome screenshot:

<img width="765" height="508" alt="Image" src="https://github.com/user-attachments/assets/29a315ea-a959-4d4b-b87d-4ca4fcf70a01" />

fedemp:
Hello, @caius What version of Safari are you using? Maybe I can find a version of Chrome from around the same date that shows this behaviour.

Also, do you know how to use the developer tools of the browser? Because I would like to know if the `<label>` is there in the DOM tree, and if CSS has somehow misplaced it.

For reference, these are the only rules that should apply to the `<label>`.

<img width="632" height="897" alt="Image" src="https://github.com/user-attachments/assets/70b51eff-e4b3-47e0-9253-b91d3c7c98f6" />

pushcx:
This is also present on the new story form in iPadOS Safari.

And then it's also present on the tag suggest form, and the mod story editing form.

We've seen this bug before, I think. If I recall, the input is getting a little bit larger than the container and Webkit handles it differently than the other engines. I thought this would be solved long-term by the move to laying out these forms with CSS grid, but either it didn't or there's a new cause.

caius:
The HTML differs for that line looking at it:

<details><summary>Click for HTML from page</summary>
<p>

```
<div class="box">
    <div class="boxline">
      <label for="story_url">URL</label>
      <input type="text" value="" name="story[url]" id="story_url">
      <button name="button" type="button" id="story_fetch_title">Fetch&nbsp;Title</button>
    <p class="actions url-updated">
      The URL above has been updated to use the <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#canonical">canonical URL</a> given by the server.
    </p>
    </div>

  <div class="boxline">
    <label for="story_title">Title</label>
    <input required="required" maxlength="150" size="150" type="text" value="" name="story[title]" id="story_title">
    <p class="actions title-reminder">
[...]

PR #1876, opened 2026-01-24
Account for focused outline in input margins

### LINKED ISSUE #1867: Button overlapping field in 2fa disable form
Image

Current Firefox.

Looks like the enable form is the same, the button just isn't wide enough to overlap:

Image

chamlis:
The outline applied to focused elements doesn't affect layout. This is desirable so that focusing an input doesn't affect the positioning of other elements on the page. However, it may result in the outline overlapping adjacent elements. To prevent this, this commit gives the affected elements a margin of the same size of their potential outline, so that the outline can't protrude beyond the element's box.

This commit makes the outline size a variable so it can be reused, and moves the :focus selector next to the styles it's intended to override.

Closes #1867

Comparisions follow. **Note**: we're applying the margin to all four sides, meaning the submit button is indented slightly. If the outline "leaking" over the edge of the page isn't an issue, we can switch the margin to only apply to the top and bottom.

Before:
image

After:
image

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -37,6 +37,7 @@ But I make an unprincipled exception for :has because it solves so many big prob
 
 	/* dimensions */
 	--comment-controls-height: 1.2lh;
+	--input-outline-width: 2px;
 }
 
 /* generics */
@@ -174,6 +175,15 @@ textarea {
 	color: var(--color-fg-contrast-10);
 	background-color: var(--color-box-bg);
 	padding: 3px 5px;
+	margin: var(--input-outline-width);
+}
+input:focus,
+button:focus,
+select:focus,
+textarea:focus {
+	outline: var(--input-outline-width) solid var(--color-box-border-focus);
+	border-color: transparent;
+	color: var(--color-fg);
 }
 textarea {
 	resize: vertical;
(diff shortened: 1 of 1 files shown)


fedemp:
IMO this is not the best approach. There's already a rule to use grids for forms, that includes some margin between rows. I think it would be easier to just place the "submit" button in its own row, and if needed, increase the `row-gap`.

```css
.labelled_grid {
	display: grid;
	grid-template-columns: minmax(10rem, min-content) 1fr;
	word-break: break-word;
	column-gap: 1rem;
	row-gap: 12px;
}
```

chamlis:
I'm happy to defer to your prior experience, but I prefer the approach in this PR as it leans more on the CSS box model to ensure things don't collide. I see this as analogous to encapsulation in OOP, with elements responsible for managing their own space requirements, rather than having the outline of an element protrude past its box and accounting for this in the margins/gaps of other elements. Maybe best practice differs on this, though, I'm not sure.

fedemp:
Just for the record, when i said "There's already a rule" I meant a css rule in lobsters css that would help in this case, not some "community approved best practice". It was mentioned before the intention of using grids to avoid overlapping. https://github.com/lobsters/lobsters/issues/1801#issuecomment-3543787376

https://github.com/lobsters/lobsters/blob/7aa104d0d5dcab08fc6d596548fc8a2f5c7f4143/app/assets/stylesheets/application.css#L1261

pushcx:
It sounds like both things are worth doing. The inputs should have sufficient space to display their focus outline without overlapping nearby elements, and also the submit buttons for these (and other forms) should explicitly be on their own row.

I'm going to merge this one now, especially because it fixes more potential layout issues. I'll also merge fixes for these forms to put the submit buttons on their own explicit rows.

Issue #1879, opened 2026-01-27
Unread indicator appears over new top comments


fedemp:
When I entered into a story detail a second time, I got the usual unread comments indicator. Without reloading the page, I posted a new top comment. My comment appeared above the unreads indicator.

<img width="531" height="270" alt="Image" src="https://github.com/user-attachments/assets/eb959dee-bad8-425a-accb-a64bf52bf4ba" />

pushcx:
Ah, yeah, the result is a little odd. The js that posts the comment just replaces the comment form with the new rendered comment. I think it would be odd to put the "1 comment, 0 unread" line above the form, so let's change the js to put a new top-level comment below that.

The structure is in `app/views/stories/show.html.rb` for anyone who wants to take this on, look for the `commentbox` partial and `thread_summary`, but be aware this form and functionality are reused for the inline 'reply' forms for comments, so take care not to break that.

pushcx:
Fixed in #1879.

PR #1881, opened 2026-01-29
Fix Unread indicator appears over new top comments #1879

timmarinin:
Relates to #1879


pushcx:
Great, thanks for figuring this one out. I appreciate the comment, saves me a puzzle in a few months when I'm trying to figure out why this code exists. :)

PR #1887, opened 2026-01-31
Add note on preview card logo positioning

luigieli:
Part of #1525 

This PR is a follow-up for https://github.com/lobsters/lobsters/issues/1525 based on the [suggestion](https://github.com/lobsters/lobsters/pull/1812#issuecomment-3826121901):
>I think at this point the only thing to do would be to document it so that someone who looks into this in the future has a running start. Could you PR a comment to where the logo is positioned explaining "If it looks like the logo is mispositioned..." with the three Mastodon links you found? Given the unreliability, I don't think future work here is warranted for us.

If the logo looks mispositioned, it is likely due to a social media cropping.
|Mastodon preview|Story preview card|
|---|---|
|image|image|

Refs:
- https://docs.joinmastodon.org/entities/PreviewCard/
- https://docs.joinmastodon.org/entities/MediaAttachment/#image
- https://mastodon.online/@mastodonmigration/109440513681868009

### DIFF
--- a/app/models/story_image.rb
+++ b/app/models/story_image.rb
@@ -102,6 +102,13 @@ class StoryImage
       lobsters_logo = lobsters_logo.resize((image.height / 5.0) / lobsters_logo.height)
     end
 
+    # If it looks like the logo is mispositioned, it is likely due to social media
+    # cropping behaviors on preview cards.
+    #
+    # References:
+    # https://docs.joinmastodon.org/entities/PreviewCard/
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks, glad to have that documented, and maybe in the future mastodon + others will give us a better way to handle it. :)

luigieli:
> Thanks, glad to have that documented, and maybe in the future mastodon + others will give us a better way to handle it. :)

It's always an honor to contribute to this project. :smile:

PR #1890, opened 2026-02-02
Update sister_sites.md

aartoghrul:
Commons.news was moved from defunct sites list to the live sites as it's not defunct.


pushcx:
Glad to hear it's up, thanks for updating the list!

PR #1893, opened 2026-02-03
Fix JS typo in notify(). Breaks site if save fails.

fedemp:

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -118,7 +118,7 @@ function isObject(obj) {
 function notify(msg) {
 		const ariaAnnounce = qS("#aria-announce");
 		if (ariaAnnounce) {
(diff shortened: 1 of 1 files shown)


app/javascript/application.js
```diff
@@ -118,7 +118,7 @@ function isObject(obj) {
 function notify(msg) {
 		const ariaAnnounce = qS("#aria-announce");
 		if (ariaAnnounce) {
-			airaAnnounce.textContent = msg;
+			ariaAnnounce.textContent = msg;
```
fedemp, line 121: This is why Typescript checks are useful. 😉

pushcx: And type systems in general, looking at the null check wrapping it.

pushcx:
Thanks for catching and fixing this!

Issue #1884, opened 2026-01-30
Add local 'build' task for developers


pushcx:
1. We need to package up a `rails build` task to run the equivalent of our CI build: rspec, standardrb, brakeman, database_consistency. I think the current 'build' is mentioned in `CONTRIBUTING.md` but maybe more docs will need updating. Both the workflow and build task should have comments referring to the other so maintainers know to keep them in sync.

EDIT: the rest of this is handled, see comments for details.

2. With #1827 merged I ran the check on a fresh restore of prod data and got (of course ) a few more warnings:

```
$ be database_consistency --config .database_consistency.ignore.yml
W, [2026-01-30T14:37:41.754104 #1846154]  WARN -- : heinous_inline_partial initialized, found: {"app/views/comments/_threads.html.erb" => "app/views/comments/_comment.html.erb", "app/views/home/index.html.erb" => "app/views/stories/_listdetail.html.erb", "app/views/search/index.html.erb" => "app/views/stories/_listdetail.html.erb"}
Loaded configurations: .database_consistency.ignore.yml
RedundantIndexChecker fail ModActivity index_mod_activities_on_item index is redundant as index_mod_activities_on_item_type_and_item_id covers it
ColumnPresenceChecker fail Message short_id column should be required in the database
MissingUniqueIndexChecker fail Link url+from_story_id+from_comment_id model should have proper unique index in the database
MissingUniqueIndexChecker fail Link to_comment_id+from_story_id+from_comment_id model should have proper unique index in the database
MissingUniqueIndexChecker fail Link to_story_id+from_story_id+from_comment_id model should have proper unique index in the database
ForeignKeyChecker fail Username user should have foreign key in the database
MissingIndexChecker fail User usernames associated model should have proper index in the database
ForeignKeyCascadeChecker fail Story merged_stories should have foreign key with on_delete: :[:nullify] in the database
```

I belatedly (sorry @fpsvogel) re-dumped the schema in 1a7903c8. I guess rails doesn't do a full dump after each migration like I had assumed, it independently uses the migration to transform `schema.rb`.

fpsvogel:
Weird: after pulling the latest changes, when I run `bundle exec database_consistency --config .database_consistency.ignore.yml`, I don't see any warnings from the gem.

[...]

PR #1894, opened 2026-02-03
Silence WIP database_consistency issues to unblock CI

fpsvogel:
Silences `database_consistency` warnings that are causing CI to fail, in order to unblock PRs.

This change is temporary, until we resolve the issue causing those warnings (see https://github.com/lobsters/lobsters/issues/1884).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -145,4 +145,4 @@ jobs:
 
       - name: Check consistency of DB and models
         run: |
-          bundle exec database_consistency --config .database_consistency.ignore.yml
+          bundle exec database_consistency --config .database_consistency.temp.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -145,4 +145,4 @@ jobs:
 
       - name: Check consistency of DB and models
         run: |
-          bundle exec database_consistency --config .database_consistency.ignore.yml
+          bundle exec database_consistency --config .database_consistency.temp.yml
```
hugoarnal:
```suggestion
          bundle exec database_consistency --config .database_consistency.ignore.yml || true # This is temporary, check PR #1894
```

pushcx:
The details are over in #1884, but the summary is that this was my mistake and it's handled now.

PR #1895, opened 2026-02-03
Add open graph description for users

hugoarnal:
Continuation of #1857

According to the [Open Graph protocol](https://ogp.me/), the description is `A sequence of Unicode characters`. That essentially means that most websites won't treat it as HTML or Markdown but only as a simple string. To tackle this, I created a `Markdowner.to_raw` function which converts from Markdown to just raw text. 

I'm not a Ruby dev so make sure to tell me what I'm doing wrong :-)

Also according to [this article](https://ogtester.com/blog/what-is-maximum-length-of-og-title-and-og-description), most social medias automatically truncate the description with various lengths. Hence why the choice of not truncating the user description.

Unfortunately, the escaped symbols cannot be removed as it is directly in comrak:

image

Here's the result on Discord using pushcx's description:

image

<details>
<summary>Click here to a complicated example with lots of symbols:</summary>
image

Description used:
[Hello world](https://lobste.rs) ~~no~~ **yes** *hello!*

\\this is a !@*_ string\\
Example image
`backticks`

pre
    text

@test is my account

~test is also my account

aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

> yup1

pre
    text

> yup
</details>

### DIFF
--- a/extras/markdowner.rb
+++ b/extras/markdowner.rb
@@ -3,6 +3,20 @@
 require "commonmarker"
 
 class Markdowner
+  COMMONMARKER_OPTIONS = {
+    extension: {
+      tagfilter: true,
+      autolink: true,
+      strikethrough: true,
+      header_ids: nil,
+      shortcodes: nil
+    },
+    render: {
+      escape: true,
+      hardbreaks: false,
+      escaped_char_spans: false
+    }
+  }
   USERNAME_MENTION = /\B([@~]#{User::VALID_USERNAME})\b/o
 
   # Rerender all markdown (#1627) cached in db columns, a pattern that predates Rails frament
@@ -31,27 +45,12 @@ class Markdowner
 
     as_of ||= Time.current
 
-    commonmarker_options = {
-      extension: {
-        tagfilter: true,
-        autolink: true,
-        strikethrough: true,
-        header_ids: nil,
-        shortcodes: nil
-      },
-      render: {
-        escape: true,
-        hardbreaks: false,
-        escaped_char_spans: false
-      }
-    }
-
-    root = Commonmarker.parse(text.to_s, options: commonmarker_options)
+    root = Commonmarker.parse(text.to_s, options: COMMONMARKER_OPTIONS)
 
     walk_text_nodes(root) { |n| link_username_mentions(n, as_of:) }
 
     ng = Nokogiri::HTML(root.to_html(
-      options: commonmarker_options,
+      options: COMMONMARKER_OPTIONS,
       plugins: {syntax_highlighter: nil}
     ))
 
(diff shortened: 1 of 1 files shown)


extras/markdowner.rb
```diff
@@ -75,6 +75,42 @@ def self.to_html(text, allow_images: false, as_of: nil)
     end
   end
 
+  def self.to_raw(text)
+    if text.blank?
+      return ""
+    end
+
+    commonmarker_options = {
```
pushcx: Please lift this duplication up to a class-level constant.

pushcx (changes requested):
One small change to avoid duplication, but this looks great.

Thanks for doing the research on the spec and how clients will treat it, I really appreciate it when I don't know anything about these standards. :)

extras/markdowner.rb
hugoarnal: Should be good

pushcx:
Thanks for the feature, I'll get this deployed in a minute so we start having nicer links.

PR #1897, opened 2026-02-05
Add :build Rake task with lint, security, DB consistency, specs

### LINKED ISSUE #1884: Add local 'build' task for developers

fpsvogel:
Resolves #1884 

Adds a `build` Rake task containing the same commands as in CI (check.yml).

Screen Shot 2026-02-09 at 5 00 24 PM

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -89,24 +89,6 @@ jobs:
         run: |
           bundle exec brakeman --ensure-ignore-notes
 
-  leak-check:
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
-    strategy:
-      fail-fast: false
-
-    steps:
-      - uses: actions/checkout@v5
-      - uses: ruby/setup-ruby@v1
-        with:
-          bundler-cache: true
-
-      - name: Find leaky gems
-        run: |
-          gem install bundler-leak
-          bundle leak update
-          bundle leak check
-
   database-consistency:
     runs-on: ubuntu-latest
     timeout-minutes: 15


Gemfile.lock
```diff
@@ -740,7 +745,7 @@ CHECKSUMS
   zeitwerk (2.7.3) sha256=b2e86b4a9b57d26ba68a15230dcc7fe6f040f06831ce64417b0621ad96ba3e85
 
 RUBY VERSION
-  ruby 4.0.0
+   ruby 4.0.0
 
 BUNDLED WITH
-  4.0.3
+   2.7.2
```
fpsvogel: `bundler-leak` is incompatible with newer versions of Bundler, and the Bundler downgrade seems to break CI, but I left this in the PR for illustration purposes, in order to suggest that we:

- Leave `bundler-leak` out of `rake build` (I relegated it to [a separate commit](https://github.com/lobsters/lobsters/pull/1897/changes/382c27e13e086b3cb497b25a5cd092c46733ed6f))
- Remove `bundler-leak` [from check.yml](https://github.com/lobsters/lobsters/blob/5fb11d61c9841ea3d2420c9e881f513170fe7bd7/.github/workflows/check.yml#L89-L105). The [database that it draws from](https://github.com/rubymem/ruby-mem-advisory-db) seems to be out of date (not updated since 2021). So I suspect it's not adding much value by being run in CI.

lib/tasks/build.rake
```diff
@@ -0,0 +1,40 @@
...
+  puts "\n◾ Running database consistency checks (DatabaseConsistency)..."
+  DatabaseConsistency.run([".database_consistency.ignore.yml"])
+
+  puts "\n◾ Running gem memory leak checks (bundler-leak)..."
+  bundler_plumber_cli = Bundler::Plumber::CLI.new
+  bundler_plumber_cli.update
+  bundler_plumber_cli.check
+
+  puts "\n◾ Running tests (RSpec)..."
+  Rake::Task["spec"].invoke
```
fpsvogel: FYI, this doesn't run specs in `spec/slow`. Other than this difference, this Rake task is functionally the same as the commands in `check.yml` as far as I can tell.

Context:

`check.yml` runs two commands for specs:

```shell
bundle exec rspec
bundle exec rspec spec/slow/*_spec.rb
```

`spec_helper.rb` includes this line:

```ruby
config.exclude_pattern = "spec/slow/*_spec.rb"
```

thomasdziedzic:
The last commit for [bundler-leak](https://github.com/rubymem/bundler-leak) and the [advisory db](https://github.com/rubymem/ruby-mem-advisory-db) was 4 years ago. I would consider this a dead project due to the advisory db not having an update in such a long time.

lib/tasks/build.rake
pushcx: You got it, yep. The slow specs don't seem worth it for devs to run on the regular as they take quite a bit of time, fail infrequently, and only test code that's used in dev.

.github/workflows/check.yml
```diff
@@ -1,3 +1,6 @@
+# If any steps below are added or changed, make sure to keep build.rake in sync.
```
pushcx, line 1: Just what I was looking for here, thanks.

fpsvogel:
New changes after the stream:

- dropped the commit that added bundler-leak to the new Rake `build` task
- [removed](https://github.com/lobsters/lobsters/pull/1897/changes#diff-55f05888fd854d9962bdd4d8ccf23a07b610990f58174397b2586b58f8937428L89-L106) bundler-leak from check.yml
- reordered the task steps so that RSpec runs first; involves [an override to prevent RSpec from exiting the process](https://github.com/lobsters/lobsters/pull/1897/changes#diff-b00e5620ad8540b93af7375026ada9ceb4acd4ac355a3376c7b3dbf6af75aab1R26)

pushcx:
Thanks for taking this on, I sure didn't expect that there'd be so much hassle calling into these tools. I appreciate you working through it.

PR #1899, opened 2026-02-07
Display self promo guideline when option is checked

### LINKED ISSUE #1725: Display a warning about self-promotion when the user checks the "I am the author" checkbox

hugoarnal:
Show a dropdown text when the "I am the author of the story at this URL (or this text)" is checked.

Unchecked:
image

Checked:
image

Shows the text same way as the `Please remove extraneous components from titles such as the name of the site, blog, section, and author.` text is shown (dropdown).

Hopefully that reduces the amount of self-promo bans on the site.

Closes #1725

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -267,6 +267,14 @@ export class _LobstersFunction {
     }
   }
 
+  checkSelfPromo(input) {
+    if (input.checked) {
+      slideDownJS(qS('.self-promo-warning'));
+    } else if (qS('.self-promo-warning').classList.contains('slide-down')) {
+      qS('.self-promo-warning').classList.remove('slide-down');
+    }
+  }
+
   fetchURLTitle(button) {
     const url_field = qS('#story_url');
     const targetUrl = url_field.value;
@@ -687,7 +695,8 @@ document.addEventListener("DOMContentLoaded", () => {
 
   // Story
 
-  Lobster.checkStoryTitle()
+  Lobster.checkStoryTitle();
+  Lobster.checkSelfPromo(qS('#story_user_is_author'));
 
   Lobster.tomSelect();
 
@@ -699,6 +708,9 @@ document.addEventListener("DOMContentLoaded", () => {
   }
 
   on('change', '#story_title', Lobster.checkStoryTitle);
+  on('change', '#story_user_is_author', (event) => {
+    Lobster.checkSelfPromo(event.target);
+  });
 
   on('click', '.story #flag_dropdown a', (event) => {
     event.preventDefault();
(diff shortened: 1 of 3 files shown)


app/javascript/application.js
```diff
@@ -267,6 +267,14 @@ export class _LobstersFunction {
     }
   }
 
+  checkSelfPromo(input) {
+    if (input.checked) {
+      slideDownJS(qS('.self-promo-warning'));
+    } else if (qS('.self-promo-warning').classList.contains('slide-down')) {
+      qS('.self-promo-warning').classList.remove('slide-down');
```
fedemp, line 274: (I think) This can be made simpler using `classList.toggle(className, input.checked)`, the second param being a "force" value, and some CSS animations.

app/javascript/application.js
```diff
@@ -699,6 +708,9 @@ document.addEventListener("DOMContentLoaded", () => {
   }
 
   on('change', '#story_title', Lobster.checkStoryTitle);
+  on('change', '#story_user_is_author', (event) => {
+    Lobster.checkSelfPromo(event.target);
+  });
```
fedemp, line 713: Aside from my other other, can't you use `input:checked p {display: block}`? I know you would lose the fancy animation (that possibly can be refactored using CSS), but at least you won't depend on JS.

hugoarnal: I'm not sure. I think that the animation is pretty useful in making the warning more effective.
I'm going to wait for @pushcx opinion on this one.

---

On another note:
I agree with you that depending more JS is kind of annoying, but should this be another issue to just redesign the JS `slideDown` function to be made only using CSS tricks? 
I think this is possible? That would only leave a toggle in the `slideDown` function and the calculation could be separated (calc CSS function?). I'm no CSS pro though

app/javascript/application.js
hugoarnal: This works but also loses the animation, unless I factor out this part of the `slideDownJS` function and put it right before the toggle:

https://github.com/lobsters/lobsters/blob/99e023db447b9774dbcd5c3f21859c4cc61b1c2a/app/javascript/application.js#L66-L70

pushcx: I don't want to lose the animation, I like that it draws the eye on this and the suggestion about separators in titles.

pushcx:
Thanks Hugo! Fingers crossed this helps.

@fedemp I appreciate the cleanup with `:checked` but I don't want to wait for it, this is a daily issue. If you want to tidy it (keeping the animations), I'd be happy to have a followup PR.

PR #1900, opened 2026-02-09
Add `span` element to fix visual bug on the users page.

graefchen:
Adding an `span` element around the status of the user, so that it is correctly displayed on both rows, and not just in the first row on devices with a width under 480px.

Currently it looks like this:
image

With the fix it will look like this:
image


pushcx:
Thanks for spotting the odd layout and contributing a fix.

PR #1903, opened 2026-02-12
Configure Anubis

### LINKED ISSUE #1781: add Anubis with a conservative config
We haven't had a lot of problems from scrapers over the years because we have put a bunch of time into performance work, especially by using full-page caching for logged-out visitors to avoid even touching the app servers. That means we have the breathing room to take a very gradual, conservative approach to increasing mitigations like Anubis, which we should use to **only challenge the riskiest traffic**. An unusually high percentage of our users (unknown, but probably single-digit) cannot solve javascript proof-of-work challenges because they're using strict privacy/security controls, homebrewed browsers, or other unusual setups.

Saturday before last (2025-10-25) Lobsters was slow/mostly down because were were getting aggressively scraped by a bot that used many thousands of IPs to avoid rate-limiting.

In the immediate mitigation, I was able to block their IPs with `iptables` and a list of bad-actor IPs for ASINs owned by that network because cadey shared the list that's already in Anubis for them.

I also fixed the endpoint they overwhelmed in 5e29c8d60. It normally took ~8ms to look up a redirect, but the fact that it did 2-3 queries instead of serving a locally cached page allowed them to dogpile the database.

This probably isn't the only endpoint susceptible to abuse, nor the last aggressive scraper we'll see. So I'd like to add Anubis, **conservatively configured to only challenge ASINs from its bad actor lists**, which [are here](https://github.com/TecharoHQ/anubis/blob/main/data/bots/_deny-pathological.yaml) and [here](https://github.com/TecharoHQ/anubis/tree/main/data/crawlers).

hmadison:
Contextualized by: #1814
Closes: #1781

---------

This (attempts) to configure Anubis in a fashion that work with our Hatchbox hosting setup. I'm using the default configuration for now. I'm assuming that we will want to tweak things once we know it works.


hmadison:
Not stale

hatchbox/lobsters.anubis.env
```diff
@@ -0,0 +1,5 @@
+BIND=:9000
+DIFFICULTY=4
```
pushcx, line 2: Why does this specify a difficulty? It looks like the `thresholds` section of `losbters.botPolicies.yaml` sets them for the various conditions?

hatchbox/lobsters.botPolicies.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+status_codes:
+  CHALLENGE: 200
+  DENY: 200
+
+store:
+  backend: memory
+  parameters: {}
+
+thresholds:
```
pushcx, line 36: Can you add an [expression-based rule](https://anubis.techaro.lol/docs/admin/configuration/expressions) to only apply these tests on paths starting `/search`? That's our most expensive endpoint.

Then also it'd be real easy for me to comment that out in a time of high load to expand to the rest of the site. Or they document how to [use load](https://anubis.techaro.lol/docs/admin/configuration/expressions#using-the-system-load-average) to turn up the difficulty; we could do that above load of 4. (We have 4 CPUs and they suggest 2 * CPUs, but we're visibly struggling earlier than that.)

pushcx (commented):
Thanks for setting this up and defening it from stalebot. Maybe this weekend we could get this deployed.

hatchbox/lobsters.anubis.env
hmadison: From the [documentation](https://anubis.techaro.lol/docs/admin/installation#environment-variables):

> The difficulty of the challenge, or the number of leading zeroes that must be in successful responses.

pushcx: It's present on all the conditions in the policy, so why set a default that's never used? Or am I misunderstanding how they relate? Is it a safety measure you wanted in place for potential future conditions?

hmadison: Digging a bit more, this value is [the default](https://github.com/TecharoHQ/anubis/blob/7e037b65e8fd23b297e69391dddbe356dddeed40/run/openrc/anubis.initd#L13) difficulty which gets used if the lobsters.botPolicies.yaml does not override it with a specific value.

hmadison: To the best of my understanding, since we are explicitly setting the value, it's not going to do anything for us, but is needed to allow the application to boot.

hmadison:
@github-actions not stale

pushcx:
There is probably a clever joke to be made about anubis and bringing this PR back from the dead, but I am too tired to think of it, so I'll just say I'm sorry I've let this hang out so long.

PR #1907, opened 2026-02-16
Add tag combination filters

ibsusu:
## What this does

Lets users filter out stories that match a specific combination of tags. For example, hide stories tagged both "javascript" AND "culture" while still seeing stories with either tag alone.  Accessible from the existing /filters page via a link to /tag_filter_combinations.

## How

Two-phase bloom filter:

1. Each story gets a 64-bit bitmask (tag_hash) computed from its tag IDs
2. On query, a fast bitwise AND rejects most non-matching stories
3. A SQL verification pass catches bloom filter false positives

benchmark shows ~3.5% overhead vs traditional single-tag filters.

## Limits

- 2–15 tags per combination
- 15 combinations per user
- Enforced at the model layer

## Rollback

Single migration. Code guards against missing column/tables: the scope rescues ActiveRecord::StatementInvalid, callbacks check has_attribute?(:tag_hash). Rolling back the migration won't break the site.

Let me know if I've screwed something up.

### DIFF
--- a/app/controllers/tag_filter_combinations_controller.rb
+++ b/app/controllers/tag_filter_combinations_controller.rb
@@ -22,6 +22,17 @@ class TagFilterCombinationsController < ApplicationController
       return
     end
 
+    sorted_ids = tag_ids.sort
+    # tags association is ordered by id, so no sort needed on existing
+    existing = @user.tag_filter_combinations.includes(:tags).any? { |c|
+      c.tags.map(&:id) == sorted_ids
+    }
+    if existing
+      flash[:error] = "You already have this tag combination filtered"
+      redirect_to tag_filter_combinations_path
+      return
+    end
+
     @combination.tags = Tag.where(id: tag_ids)
 
     if @combination.save
--- a/app/models/tag_filter_combination.rb
+++ b/app/models/tag_filter_combination.rb
@@ -3,7 +3,7 @@
 class TagFilterCombination < ApplicationRecord
   belongs_to :user
   has_many :tag_filter_combination_tags, dependent: :destroy, inverse_of: :tag_filter_combination
-  has_many :tags, through: :tag_filter_combination_tags
+  has_many :tags, -> { order(:id) }, through: :tag_filter_combination_tags
 
   MAX_TAGS_PER_COMBINATION = 15
   MAX_COMBINATIONS_PER_USER = 15


pushcx:
Which coding agent did you use for this? Claude, Codex, something else?

ibsusu:
> Which coding agent did you use for this? Claude, Codex, something else?

https://github.com/lobsters/lobsters/blob/0e3d6ec70e7b78a03de4084799c783ff7003e73c/CONTRIBUTING.md?plain=1#L29

https://github.com/lobsters/lobsters/blob/0e3d6ec70e7b78a03de4084799c783ff7003e73c/AGENTS.md?plain=1#L2

All lines were written by me after consulting AI services in lieu of googling and stackoverflowing and then rewritten by me again from memory and benchmarking.  I did not submit any directly AI generated code.  I have no desire to get banned since you now know what my account is.

If it is good code please accept it.  If it is not then please tell me how I can fix or please just reject it.  I do my best to not submit shit code.  Nobody has time for that.

pushcx:
To take something off the table, I'm not going to ban you over this PR or the feature request story. I've blocked accounts that have dumped drive-by slop to prevent getting more of it, but I can tell this came from caring about the community and wanting to improve it. Bans are a tool for preventing future abuse; I don't think that's likely from you and if this is going to be the new normal in open source projects, banning individuals is not going to [stop the tide from coming in](https://en.wikipedia.org/wiki/King_Cnut_and_the_tide). This is a pretty useful example of how vibecoding tools are changing the norms for open source contributions, so I'd appreciate it if you could share more of your process and thoughts here. 

I asked which agent because I've been surprised by them agreeing to work on this repo again. Over the last few days I've cleaned up 3 obvious, low-quality contributions from LLM agents, the first we've seen since adding the agents.md 4 months ago. I was hoping to learn which tool is ignoring it so I can try to improve the prohibition.

I have a hard time believing LLMs were only used for research. For one example, you're obviously not a student or junior developer, but the PR doesn't define a constant for the maximum number of tags. It hardcodes `15` in a bunch of places that would all need to be edited in perfect sync. Every developer learns the value of constants in their first couple months, but agents are hit-or-miss, especially with values given by the prompt.

When I review code, even bad code, I know I'm helping someone learn the codebase and Rails. Here that's not going to happen. And the task of reviewing it is daunting; this PR is 4% the size of the current codebase. I'd put a ton of time into critique that would just be copied and pasted into a coding agent. Looking at this PR feels like an invitation to waste a lot of time.

ibsusu:
I've already benefited from you telling me about the hardcoding so even if you didn't mean to your light reviewal of the code was helpful, Rails is not my usual playground.  Even with that, I don't need you to teach me the codebase on this PR, rather what I would find more useful for the future is your philosophy on site and codebase maintenance.  I can learn what I need to about the codebase itself from the directory tree, references and schemas. All of these frameworks and languages are similar enough that it's possible to program by benchmarking and testing correctness.  Digging through commit and comment histories to find what works for you is more difficult. 

Regarding AI Prohibition:

Your prohibition technique is good for headless hit-and-run slop bots but not someone attempting to make precise, guided changes by using the LLM as a rubber duck to figure out what to write where, what will and won't work, what's performant, what's maintainable, what's redundant, etc.  They still lack 'taste' in the majority of circumstances unless prodded in certain directions so I find that unless I just don't care or the change is small they can't be let off the leash if I still want human maintainable code.

Even if you embedded a header at the top of each file, class, method, or line, that's a search and replace away from being modified wholesale.  That's one reason why that open source contribution vouching/trust system came into being.  It is currently impossible to tell who or what wrote the code, only who submitted it.  This is _my_ PR, not a bot's.  I tested just now, Gemini, Codex, Claude, Cline, and Grok will simply ignore the AGENTS.md and CLAUDE.md within a directory if you tell them to - not to mention you can just delete the AGENTS.md and say 'go'.  LLMs have become far more pliable since their initial release.  Many things that would previously lock them up or trip them up have been relaxed.

Again, I didn't let the AI write the code. I did crib heavily from examples and tune what I wrote but none of their output has directly made it into this PR.  Gimme a bit and I'll fix the hardcoding.

pushcx:
I was [reminded of this PR this morning](https://bsky.app/profile/byroot.bsky.social/post/3mgcjtlrfcc2t) and figured I'd close it out as there's no path forward for it.

ibsusu:
I don't really get your reasoning.  I actually spent hours on this.  I understand how it works, that it does work, that the code makes the site more featureful  if not better, and that if it didn't work it allowed you to rip it out of the site without feeling a thing.  There is no greater chance that the code is copyrighted or lifted from other locations than a standard for loop.

You're assuming incorrectly, which is your prerogative I guess, but I really want to note it.

PR #1912, opened 2026-02-17
Filter user settings from logs etc

chamlis:
This attribute contains secrets for Pushover, TOTP, GitHub, and Mastodon. In principle it should be possible to filter just those subfields with a filter like "settings.totp_secret", but due to [how Rails implements filtering on model attributes](https://github.com/rails/rails/blob/12900149d812792f4b7fdf38f591fb6d8800db9f/activerecord/lib/active_record/attribute_methods.rb#L531-L539), the filter only sees the settings hash in its stringified form.

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -17,6 +17,7 @@ Rails.application.config.filter_parameters += [
   :passw, # preventative
   :salt, # preventative
   :secret, # preventative
+  :settings, # contains secrets for pushover, totp, github, mastodon
   :session_token, # auth cookie value
   :token, # rss token - need to transition this to a typeid
   :totp_code # LoginController


chamlis:
I should mention that this could also be done with the following:
```diff
diff --git i/app/models/user.rb w/app/models/user.rb
index cecd1605..3527b46c 100644
--- i/app/models/user.rb
+++ w/app/models/user.rb
@@ -1,6 +1,8 @@
 # typed: false
 
 class User < ApplicationRecord
+  self.filter_attributes = [:password, :settings, :token]
+
   has_many :stories, -> { includes :user }, inverse_of: :user
   has_many :comments,
     inverse_of: :user,
```

The commit in the PR has the downside of mixing *attribute* filtering in with *parameter* filtering. This patch has the downside of spreading filtering across the codebase, along with the chance I missed something in recreating the filter list and there's a naked secret somewhere. I don't have a strong opinion either way.

pushcx:
Thanks for being proactive about preventing my footguns, I appreciate that you thought to do this.

I think your explanation of the tradeoffs leads to there being two things to do here:

1. Hybridize your strategies, in `ApplicationRecord` add `self.filter_attributes = Rails.application.config.filter_parameters`. (Little more complicated if ActiveRecord errors on nonexistent attributes: `self.attribute_names.map(&:to_sum) & :Rails.application.config.filter_parameters`.)

2. Split up `settings`. It's a junk drawer of actual settings which is... fine. I don't think it improves on using model attributes directly, but it doesn't hurt. But `settings` is **also** a junk drawer of sensitive tokens. The pain here is that `settings` mixing up two very different kinds of data that need to be treated differently. We should migrate these into their own `typed_store` junk drawer that can be filtered while `settings` is not. `mailing_list_token`, `session_token` (more?) could also go in there, too, as long as we can still do direct lookups like `User.where(session_token: ...)`. Maybe a name like 'auth_tokens' or 'secrets'?

chamlis:
No worries, thanks for the review!

>1. Hybridize your strategies, in `ApplicationRecord` add `self.filter_attributes = Rails.application.config.filter_parameters`. (Little more complicated if ActiveRecord errors on nonexistent attributes: `self.attribute_names.map(&:to_sum) & :Rails.application.config.filter_parameters`.)

I might be misunderstanding what you're suggesting here, but I think this is [what ActiveSupport does already](https://github.com/rails/rails/blob/12900149d812792f4b7fdf38f591fb6d8800db9f/activerecord/lib/active_record/railtie.rb#L340-#L344). Do you want to make this explicit our side instead?

>2. [...] We should migrate these into their own `typed_store` junk drawer that can be filtered while `settings` is not. `mailing_list_token`, `session_token` (more?) could also go in there, too, as long as we can still do direct lookups like `User.where(session_token: ...)`. Maybe a name like 'auth_tokens' or 'secrets'?

Agreed that moving tokens out of settings is a good idea. Is there a reason why we shouldn't promote them to model attributes/columns? I recall something about rows getting too wide at one point?

Would it be better to hold off on schema changes until after #1871?

pushcx:
> I might be misunderstanding what you're suggesting here, but I think this is what ActiveSupport does already. Do you want to make this explicit our side instead?

Nope, you got it exactly. I didn't realize it was already present.

So then yeah, that leaves splitting up `settings` as the right next step. While this is security-related, I agree that it's not so urgent that it's worth adding work to the already-large 1871. Hopefully I can land that on Saturday morning.

Issue #1913, opened 2026-02-18
UI: weird shape on caches boxes


lbellomo:
When the caches box is open, looks like is missing a bit of width and and the last letter of the longest named caches overflows to the next line.

All cases are logged out. This is on firefox on linux (and the same happend on chromium):

<img width="396" height="272" alt="Image of lofte.rs home with the caches box open that overflow the last character of 2 element to the next line" src="https://github.com/user-attachments/assets/169f1642-edd0-47a7-8cab-870d159e90f1" />

On firefox on android only one row overflow:

<img width="400" height="300" alt="Same, but on firefox on Android. Only one line overflow" src="https://github.com/user-attachments/assets/38bb74e0-7eb6-42cd-92b5-77c65de45733" />

pushcx:
Thanks for spotting this and the clear report.

PR #1916, opened 2026-02-18
fix(ui): reimplement archive dropdown with popover api

### LINKED ISSUE #1915: UI: caches box don't auto-close
If you open a cache box, it will not close if you click anywhere or if you open multiple cache boxes. And it can result in this strange state:

Image of lobters home with multiples caches boxes open and overlaping

This is logged out.

atlasgong:
Hi, this is my first PR to lobsters, and I'm new to Ruby (and Rails), so please let me know if anything looks off!

This PR fixes #1915. The current implementation for the archive dropdown uses a checkbox, but without JavaScript,  I can't think of a way to uncheck boxes when clicking outside or when another checkbox is checked. I found instead the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API), which enables a popover to close when clicking outside, or when another popover is opened.

This implementation doesn't support legacy browsers, unless we use JavaScript. The Popover API is classified as Baseline 2025: https://caniuse.com/?search=popover. I noted in your README that you don't want to serve JavaScript to logged-out visitors, so I think this is the best we can do.

To position the popover near the button without JavaScript, I used `position-area` from [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning). This is new and is classified for Baseline 2026, and in particular doesn't support versions under Safari 26: https://caniuse.com/?search=+CSS+anchor+positioning. This means on Safari 18 and lower, the popup will not be positioned aligned to the `caches` button. I'm not sure how major or minor of a concern this would be for lobsters.

I'm also happy to help with #1913 by extending the width of the dropdown, please let me know if you'd like it to be in a separate PR or if I should add it here for simplicity.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -970,7 +970,7 @@ div.comment_text code {
 
 .archive-dropdown {
 	position: fixed;
-	position-area: bottom span-left;
+	position-area: bottom span-right;
 	margin: 3px 0 0 0;
 	width: 100px;
 	border: 1px solid var(--color-box-border);


pushcx:
Hi! Thank you for volunteering to pick this up. I'm sorry to say that my oversight wasted your time.

At the top of our [at the top of `application.css`](https://github.com/lobsters/lobsters/blob/0e3d6ec70e7b78a03de4084799c783ff7003e73c/app/assets/stylesheets/application.css#L4-L5) I noted that we target the "widely available" baseline standard, but it's such an unusual restriction I should've also added it to the README alongside the javascript warning you noted. I've done that now, but if it was in place you wouldn't have spent time on this PR's design approach.

Now that you're familiar with this markup: can you see any path to fixing this UI within these two restrictions? If not, maybe we should just put cache links on a hidden second line of metadata and use the checkbox hack to reveal it. (`<details>` would probably be better for accessibility, but I don't think it could support this layout.)

atlasgong:
Hey, it's no worries! I should've read the top of the file haha, my bad for missing that. Good lesson to learn for next time. Appreciate you adding it to the README though!

I think `<details>`/`<summary>` has the same issue as using the checkbox. There's no way to click outside an open `<details>` to close it. It's possible to have mutual-exclusivity by assigning the same `name` attribute to group them, so opening one closes others, and this would also keep the CSS positioning simple, without needing anchor positioning, but this is Baseline 2024: https://caniuse.com/?search=Mutually+exclusive+%3Cdetails%3E+elements.

Not sure if there are any other ways around this for the time being, maybe somebody else could pitch in here. As an aside, I've been lurking on lobsters logged-out for a while, would be willing to send an invite my way? My email can be found at my personal website linked on my GitHub profile. Thanks!

pushcx:
If the caches link adds a line of text underneath the existing metadata, the rest of the page can move down. It won't need to have the modal-window behavior of clicking off it to hide it.

I sent you an invite.

atlasgong:
I think I understand what you mean; something like this?

image

Thanks for the invite!

pushcx:
Yes, that's exactly what I was thinking.

Maybe even, only in the open state, add a downward-pointing-triangle to the 'caches' link like the `<details>` element has by default. (It would be a distraction in the 'closed' state'.

PR #1917, opened 2026-02-19
ui: replace caches dropdown with inline reveal

### LINKED ISSUE #1913: UI: weird shape on caches boxes

### LINKED ISSUE #1915: UI: caches box don't auto-close

atlasgong:
Addressing the discussion in #1916, this PR replaces the caches dropdown with an inline reveal (keeping the checkbox logic). It does use the `:has` selector (Baseline 2023), but I see there is an exception for this. Please let me know if anything else should be changed.

Closes #1915 and closes #1913.

Thanks!

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -944,21 +944,21 @@ div.comment_text code {
 	background-color: var(--color-box-bg-shaded);
 }
 
-.archive-dropdown {
+.caches-dropdown {
 	width: fit-content;
 	margin: 0 auto;
 }
 
-.archive-dropdown a {
+.caches-dropdown a {
 	color: var(--color-fg-contrast-4-5);
 	text-decoration: none;
 }
 
-.archive-dropdown a:hover {
+.caches-dropdown a:hover {
 	text-decoration: underline;
 }
 
-.details:has(> .archive_button:checked) label[for^="archive_"]::before {
+.details:has(> .caches_button:checked) label.caches::before {
 	content: "▾ ";
 }
 
@@ -977,19 +977,16 @@ div.comment_text code {
 	z-index: 1;
 }
 
-/* archive caches */
-.archive_button {
+.caches_button {
 	display: none;
 }
-.archive_button:not(:checked) ~ .archive-dropdown {
+.caches_button:not(:checked) ~ .caches-dropdown {
 	display: none;
 }
-label[for^="archive_"] {
+label.caches {
 	cursor: pointer;
 	white-space: nowrap;
 }
-
-
 
 .controls details {
 	margin-left: auto;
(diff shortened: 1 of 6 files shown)


atlasgong:
Desktop logged out:
image

Desktop logged in:
image


Mobile logged out:
image

Mobile logged in:
image

pushcx:
Could you right-align it? That'll roughly put it under the `caches` control and we won't have to try to line it up with the indent following user avatar (which users can toggle off).

And please use the `divider_tag` helper rather than `:before`, it'll have the correct aria accessibility attributes.

atlasgong:
Right-align produces this on desktop:
image

How do you feel about the following approach with `width: fit-content` and `margin: 0 auto`:
image

Mobile is pretty much the same as before, both-ways.

fedemp:
Sorry in advance for being the party pooper. The "caches" and "flag" dropdowns both need some love but from an a11y side. If you would like to improve those, and specially that other issue (mentioned in the issue tracker) that allows every "caches" to be open at the same time, it would be better to use `details` and some CSS. I can not test your branch (cause I don't have my pc right now) but I can tell you that the current implementation in prod does not work right neither with Orca nor for keyboard navigation.

app/assets/stylesheets/application.css
```diff
@@ -958,15 +977,16 @@ div.comment_text code {
 	z-index: 1;
 }
 
-/* archive; dropdown styling is with #flag_dropdown to match it */
+/* archive caches */
```
pushcx: I know this is a chore that I failed to do, but similar to my comment in #1932 I would like the class names to match the user-visible copy.

I changed the link from "archive" because users thought it was a verb rather than a noun so didn't recognize what it was for. I changed it to "caches" and it's been much more discoverable, but that was a quick experiment and I never came back to tidy up the classes to match.

app/assets/stylesheets/application.css
```diff
@@ -958,15 +977,16 @@ div.comment_text code {
...
 .archive_button {
 	display: none;
 }
 .archive_button:not(:checked) ~ .archive-dropdown {
 	display: none;
 }
-.archive_button ~ label{
+label[for^="archive_"] {
```
pushcx: Add an explicit class rather than do partial string matching. It's just a little too clever, I like being able to lean on search for classnames.

```suggestion
label.caches {
```

pushcx:
This is great, thanks for fixing some recurring issues with this.

PR #1922, opened 2026-02-19
Fix typos

winks:


pushcx:
Well, we learned a word I can't type, apparently.

Thanks for spotting and fixing this.

PR #1924, opened 2026-02-21
Revert "Migrate to SQLite again"

pushcx:
Reverts lobsters/lobsters#1871

We [deployed](https://gist.github.com/pushcx/eb7cdf2dc9707dc3ab9e7173d197ddfc) and it was not performant. We ran out of debugging time, but would like to find an expert and try again later.


eapotapov:
@pushcx 
I've sent you an email (to peter at your nickname domain) with the investigation details. Hope this helps, if valid.

PR #1925, opened 2026-02-21
Fix username timeout validation

leameow:
Seems like the timeout validation for username change is inverted. The existing test is actually wrong.

Here's the error I got on lobste.rs: 

image

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -625,7 +625,7 @@ class User < ApplicationRecord
   def validate_username_timeouts
     return unless username_changed?
 
-    at = usernames.where("created_at <= ?", 1.year.ago).order(created_at: :desc).first&.created_at
+    at = usernames.where("created_at >= ?", 1.year.ago).order(created_at: :desc).first&.created_at
     errors.add(:username, "has already been changed in the last year (#{at.strftime("%Y-%m-%d")})") if at
 
     recently_used = Username.where(username: username).where("renamed_away_at >= ?", 5.years.ago).order(renamed_away_at: :desc).first&.renamed_away_at
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -30,15 +30,31 @@ describe User do
     expect { create(:user, username: "case-insensITive") }.to raise_error
   end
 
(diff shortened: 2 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -31,13 +31,21 @@
   end
 
   it "doesn't allow changing username in < 1.year" do
```
pushcx: This test has grown to check too many things, can you split it in 3?

```
  it "doesn't allow new users to change in < 1 year"
  it "doesn't allow changing usernames in < 1 year of last change"
  it "allows changing username with no creation/changing in < 1 year"
```

I struggled to read and understand what was being tested and which data was really being compared.

Let's also make the changes from "alice" to "betty", the adjective is too small. That's my mistake; I was trying to avoid distracting from this by using the usual "alice", "bob" progression, but that would pull in the question of whether the user is transitioning - which is a design consideration (explained on `/settings`) for username changes and invites.

app/models/user.rb
```diff
@@ -625,7 +625,7 @@ def inbox_count
   def validate_username_timeouts
     return unless username_changed?
 
-    at = usernames.where("created_at <= ?", 1.year.ago).order(created_at: :desc).first&.created_at
+    at = usernames.where("created_at >= ?", 1.year.ago).order(created_at: :desc).first&.created_at
```
pushcx, line 628: Argh date comparisons.

spec/models/user_spec.rb
leameow: > I was trying to avoid distracting from this by using the usual "alice", "bob" progression, but that would pull in the question of whether the user is transitioning - which is a design consideration (explained on /settings) for username changes and invites.

I'm not sure I follow 🤔 What's the design consideration and what is being explained on /settings?

In any case, I updated the tests as asked!

pushcx:
Great! I now understand the different scenarios being tested and will get clear feedback when I break this again. Thanks for finding and fixing this bug.

leameow:
> Great! I now understand the different scenarios being tested and will get clear feedback when I break this again. Thanks for finding and fixing this bug.

My pleasure, looking forward to change my name!

PR #1927, opened 2026-02-23
Migrate to SQLite (after the great Chicago fire of 1871)

thomasdziedzic:
Reverts lobsters/lobsters#1924

We had performance issues in production, so we reverted the original sqlite migration. This pr should do more extensive performance testing and work before merging it.


app/models/search.rb
```diff
@@ -213,7 +204,7 @@ def perform_comment_search!
     when :relevance
       # relevance is undefined without search terms so sort by score
       if terms.any?
-        query.order!(Arel.sql(terms_sql + " DESC"))
+        query.order!(Arel.sql("#{terms.first} DESC"))
```
pushcx: Bug: this is trying to sort by their search term instead of a relevance score.

app/models/search.rb
```diff
@@ -322,7 +307,7 @@ def perform_story_search!
     when :relevance
       # relevance is undefined without search terms so sort by score
       if terms.any?
-        query.order!(Arel.sql(terms_sql + " desc"))
+        query.order!(Arel.sql("#{terms.first} desc"))
```
pushcx: Bug: this is trying to sort by their search term instead of a relevance score.

app/models/search.rb
thomasdziedzic: wrote a test for this and fixed it by sorting by rank: https://www.sqlite.org/fts5.html#overview_of_fts5

app/models/search.rb
thomasdziedzic: Added test and fixed as well.

04116:
hi @thomasdziedzic any update on this one?
Just want to see how Sqlite perf in big public server-side production system like lobster

thomasdziedzic:
Leaving a comment to indicate that I'm still working on this ~(mainly waiting on someone to provide some patches)~

Edit: Looks like the patches I was waiting on are most likely not going to be done so I'll work on this myself.

chamlis:
Apologies if this isn't the best place for this.

Following the experimentation on today's office hours, I played around with the comment ordering query using a story with 1000+ comments. It looks like the performance of that query isn't much to do with actually doing confidence ordering. Putting a `return story.comments` in `story_threads` produces a simple `SELECT * FROM comments WHERE story_id=?` query that is within 20ms of the more involved one (both taking around 150ms in my setup). If I dropped the largest `comment` and `markeddown_comment` columns from the query, that cut its execution time down to ~30ms. This might not reproduce on the production DB, but if it does that suggests to me that the bottleneck is more to do with the volume of final result data, rather than how we're querying it.

pushcx:
That's a great idea for an experiment! Sounds like we're really not far off from the best we can do on arranging the tree.

We don't do anything with `comment` so we should leave that out of the `select`; maybe we can get (150-30)/2 = 60ms right there. :D

thomasdziedzic:
@pushcx I created a bulk creation script that I used to load ~ 160k stories and 300k comments. I was able to get the perf issues mostly solved for the home#index and stories#show actions by either adjusting the query or adding some additional index.

Let me know what you think about this. Maybe it would be worthwhile to do a test prod db that we can alias to sqlite.lobste.rs or something like that so that we would be able to have users test it out without affecting production?

thomasdziedzic:
Here are some performance metrics:

This is with 300k comments and 160k stories.

With PUMA_WORKERS=5

```
ab -n 1000 -c 5 http://localhost:3000/

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    48   60   8.5     58     150
Waiting:       48   60   8.5     58     150
Total:         48   60   8.5     58     150
```

```
ab -n 1000 -c 5 http://localhost:3000/s/e4egio/accusamus_magnam_velit

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       1
Processing:    34   45   9.8     43     141
Waiting:       33   45   9.8     43     141
Total:         34   45   9.8     43     141
```

thomasdziedzic:
Thought of a way to log any slow queries in prod if we encounter more perf issues: I added:
1. ENABLE_SLOW_QUERY_LOGS when set to true will log sql queries that exceed a certain duration.
2. SLOW_QUERY_THRESHOLD_MS which specifies the duration threshold in milliseconds, defaults to 100ms if unset.

pushcx:
I'd love to have both on. The app log dir is `/home/deploy/lobsters/shared/log`.

pushcx:
:tada: :tada: :tada: 

@thomasdziedzic and I migrated prod to sqlite this morning.

Thanks again for this big project.

PR #1931, opened 2026-02-26
Don't float "tags" label in "submit" form.

### LINKED ISSUE #1880: "Tags" label in "Submit" page is not clickable.
Visit https://lobste.rs/stories/new
Click on URL and the text input is focused.
Click on Tags* and nothing happens.

jamesnvc:
Fixes #1880

The "Tags" label wasn't clickable, because the div that the TomSelect library creates extends over top of the label (because it was floated to the left and hence didn't take up space). This removes the float and uses flexbox to get the correct layout.

As mentioned in the issue, it might be desirable to replace uses of `float` with flex (or something), but that is not completely mechanical (e.g. simply changing the styles in `.boxline` (adding `display: flex`) & `.box label` (removing `float: left`) *almost* works, but the "text" textarea is squished), and so probably better left for its own issue.

### DIFF
--- a/app/views/stories/_form.html.erb
+++ b/app/views/stories/_form.html.erb
@@ -42,9 +42,9 @@
     </div>
   <% end %>
 
-  <div class="boxline" style="margin-bottom: 2px;">
+  <div class="boxline" style="margin-bottom: 2px; display: flex;">
   <%= f.label :tags, "Tags",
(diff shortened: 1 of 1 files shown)


pushcx:
Great, thanks for fixing this one. I appreciate you sharing your reasoning about the scope of the fix.

Issue #1919, opened 2026-02-19
Highlight recent invitees


dzwdz:
@pushcx has been throwing around the idea of making users feel more responsible for the people they invite on stream. I suggested marking stories/comments submitted by people you've recently invited.

It would probably be best to just mark their usernames somehow. Here's a half-assed idea for how that could look like:

<img width="899" height="180" alt="Image" src="https://github.com/user-attachments/assets/b026b5c5-1de3-48b8-bbd3-2515baf67a83" />

The "recent invitee" color should also probably take precedence over "authored by"[^1]. This used `var(--color-fg-negative)`, which I suppose kinda suggests inviting users is something bad. Idk.

Maybe an icon could be nice too? ~~An emoji, to be more appealing for zoomers.~~ 👪, 👶, 🚼, 🐣 kinda make sense but are creepy (but the blue baby is kinda funny). 🌱 suggests the user is a plant.

[^1]: Tangent: perhaps there should be an icon for comments made by the author, instead of just relying on the color?

---

Since it's related, I'll also take the opportunity to criticize Peter's other idea for conveyng this responsiblity: including the invitee in the mod mail about new users attempting to use restricted tags (culture, vibecoding, ...).

I'd argue that in general, if an user tried to submit such a story and failed[^2], this is a signal they've honestly tagged the story. It doesn't really convey much information apart from "my invitee accidentally learnt one of the norms". You're not really going to get this warning more than once per invitee per tag. On the other part, from the invitee's perspective, knowing that the inviter got notified about this will just stress them out and amplify the "people frantically apologizing for nothing" brought up on stream.

You could make the argument that including them in the mod mail about trying to submit a new domain could be useful so you know when your invitee tried to self promote... but I don't think we should in any way "police" new users for things they didn't end up doing (even if it was the site itself that prevented them from doing so). Even if someone did want to self-promote:
- If they don't resubmit - they've learned that lesson. I (as the inviter) have no reason to care.
- If they do resubmit - then it's just up to the mods to ban them, since they deliberately violated a norm.

[^2]: I'm writing this under the assumption that they will be prevented from resubmitting the story.

pushcx:
I think this feature highlighting invitees would be a good improvement. As is the footnote about only using color to indicate that a commenter is the story's author or is a new user, that's inaccessible to colorblind users and we must fix it.

[...]

PR #1932, opened 2026-02-26
Show indication when a new user is your invitee

jamesnvc:
Addresses #1919.

This is just the visual component - I see that @pushcx added the "cc-ing inviter to the discussion" bit already. A pretty basic display, but perhaps a reasonable starting point for bike-shedding.

Screenshot 2026-02-25 at 22 31 42

I implemented this by adding to the `styled_user_link` helper, to ensure that it appears everywhere the user name does. However, because it needs to know the viewer, that helper is now assuming the viewing user is in `@user`. That works in practice, but I'm not sure if it's better to explicitly pass that in as another argument to the function (and then update all the call sites); it feels a bit dirty to me to rely on the instance variable being bound, happy to change this.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -63,7 +63,7 @@ module UsersHelper
           "(you invited)"
         end
       end
-    ], " ")
+    ].compact, " ")
   end
 
   def user_karma(user)
--- a/spec/helpers/users_helper_spec.rb
+++ b/spec/helpers/users_helper_spec.rb
@@ -1,0 +1,55 @@
+# typed: false
+
+require "rails_helper"
+
+describe UsersHelper do
+  describe "#styled_user_link" do
+    let(:viewing_user) { create(:user) }
+    let(:unrelated_user) { create(:user) }
+    let(:new_invited_user) {
+      create(:user, created_at: 1.days.ago,
+        invited_by_user: viewing_user)
+    }
+    let(:old_invited_user) {
+      create(:user, created_at: (User::MENTORSHIP_DAYS + 1).days.ago,
+        invited_by_user: viewing_user)
+    }
+    let(:new_unrelated_user) { create(:user, created_at: 1.days.ago) }
+
+    it "shows the user name as a link for normal users" do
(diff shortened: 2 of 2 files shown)


app/models/user.rb
```diff
@@ -640,4 +640,8 @@ def votes_for_others
                  "(votes.comment_id is null and stories.user_id <> votes.user_id)")
       .order(id: :desc)
   end
+
+  def recently_invited_by?(user)
+    invited_by_user == user && created_at.after?(NEW_USER_DAYS.days.ago)
```
pushcx:
```suggestion
    invited_by_user == user && created_at.after?(MENTORSHIP_DAYS.days.ago)
```

Please add a constant. For now just set it to `NEW_USER_DAYS * 2`, I'll tune later.

app/helpers/users_helper.rb
```diff
@@ -56,7 +56,18 @@ def styled_user_link user, content = nil, html_options = {}
...
+    safe_join([
+      link_to(user.username, user_path(user), html_options),
+      @user && newly_invited_notice(invitee: user, inviter: @user)
+    ], " ")
+  end
+
+  def newly_invited_notice(inviter:, invitee:)
+    if invitee.recently_invited_by?(inviter)
+      tag.span(class: "new-invitee") do
+        "👶 (your invitee)"
```
pushcx:
```suggestion
        "(you invited)"
```

I'm nitpicky about copy. In this case, I don't want to introduce a new noun, "invitee". And then sorry for the chore, but I'd like to use that in the css class as well, `.you_invited` rather than have separate terms in user-facing copy and code.

I don't think this needs to be a separate method, I'd rather it was included in `styled_user_link`. It'd be a great opportunity to add a test of `styled_user_link` - just this functionality is fine. We can expand it to cover the precedence later.

jamesnvc: Sounds good; I pushed the copy changes, I'll add the spec shortly.

pushcx:
Excellent, thanks for making edits so quickly. Let's get this live. :)

PR #1934, opened 2026-02-26
a11y: reimplement caches dropdown with details/summary

atlasgong:
Following up on the conversation in #1917. The current implementation for the caches dropdown uses a checkbox with toggled display visibility, which removes the checkbox from the tab order and accessibility tree. This is fixed by using  `<details>`/`<summary>` which handles everything natively (and we don't have to worry about multiple concurrent opened dropdowns since the archive links are inline now).

This also requires moving the `caches` button to the be last element in the byline, otherwise it gets stretched out to fit the summary: 
image

Here's it being last:
image

Keyboard navigation now works for Chromium and Gecko based browsers, tested with macOS VoiceOver on Sequoia 15.6.1, but there's a pre-existing issue for WebKit browsers where the only elements available in the tab order are the `caches` buttons. I don't have more time to look into this today, so I'm marking this PR as a draft.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -960,7 +960,7 @@ div.comment_text code {
 }
 
 .caches-details[open] > summary::before {
-	content: "▾ ";
+	content: "▾ " / ""; /* make '▾' invisible to screen readers by setting alt txt */
 }
 
 .caches-dropdown {


pushcx:
I'm glad you checked in on this, but we can't move the "6 comments" link out of the last position. There is way, way too much collective muscle memory on that one. And it'd probably break the mobile view, too.

atlasgong:
Thanks for the response! Mobile view is still pretty much the same as before; but I totally get the collective muscle memory thing haha. Should we close this PR then? I'm all out of ideas.

pushcx:
Thanks for taking a shot at it.

PR #1926, opened 2026-02-22
Improve visibility of dotted line between sibling comments

aashvikt:
I increased the thickness of the vertical dotted lines that indicate sibling comments. Previously, they were close to invisible on some devices and browsers.

PR #1935, opened 2026-02-27
Fix comment sibling indicator centering

aashvikt:
Created as #1926 doesn't center (relative to upvote button) the thickened vertical lines properly.


aashvikt:
On further inspection, it appears that the lines _are_ centered relative to their upvote buttons, just not the folder (`[-]`, `[+]`) icons. Closing PR.

pushcx:
Thanks for being willing to take a second look. I wondered if anyone would notice your tweak to the sibling lines, but this is the only mention of it in the last couple days.

PR #1938, opened 2026-03-01
Refactor "hide" so it can work without JS

fedemp:
Use native forms for the "hide" button. Use JS to do an async request.

Use flash messages only when not using JS. Also refactor "save" so flash is not used when sending a JSON response.

Modify "save" handler in JS since the logic is very similar to "hide".

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -93,20 +93,13 @@ export const removeExtraInputs = () => {
 }
 
 /**
- * @template {string} K
+ * @template T
  * @param {unknown} obj
- * @param {K[]} keys
- * @returns {obj is {[P in K]: string}}
+ * @param {Array<keyof T>} props
+ * @returns {obj is T}
  */
-function hasProperties(obj, keys) {
-	return (
-		isObject(obj) &&
-		keys.every(
-			(key) =>
-				key in obj &&
-				typeof (/** @type {{[key]: unknown}} */ (obj)[key]) === "string",
-		)
-	);
+function hasProperties(obj, props) {
+  return isObject(obj) && props.every((p) => p in obj);
 }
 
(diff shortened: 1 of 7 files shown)


pushcx:
This is an improvement, thank you. I like the semantics of making this a form.

I don't like replacing `hasProperties` with a version that throws. To say it in type terms, I see a lot more value in a function with the type `obj, [string] -> bool` than one that's `obj, [string] -> null` and works by side-effect. There's more info available for understanding the program, and using exceptions makes the control flow of the program a lot harder for me to read. Please revert that function, and then I suppose typescript will prompt the changes I'd like to see in the calling functions so that I can see better how it all fits together.

fedemp:
Thank you for your feedback; it's a pleasure to collaborate. I already have a child commit that migrates voting to this same format.

About the `try/catch` and `throw`, I have updated to PR to reduce its use and even prevent deep levels of indentation.

app/javascript/application.js
```diff
@@ -406,44 +376,78 @@ export class _LobstersFunction {
...
-      (b) => b instanceof HTMLInputElement,
-    );
-
-    for (const b of btns) {
-      if (b !== submitter) {
-        b.disabled = true;
+    const story = parentSelector(submitter, ".story");
+    const btns = Array.from(
+      qSA(story, `${config.selector} input[type=submit]`),
+    ).filter((el) => el instanceof HTMLInputElement);
```
pushcx, line 410: This is redundant, qSA is a thin enough wrapper around querySelectorAll that we can trust it gives back a list of HTMLInputElements given the selector on the previous line.

pushcx:
Thanks, this is a lot clearer to me now. I left one comment, but it's not particularly important, so I'm going to go ahead and merge to get this improvement live.

PR #1939, opened 2026-03-02
Fix user tree caching issue

### LINKED ISSUE #1937: /users incorrectly says I invited a user
I'm logged in as user "agwa"; https://lobste.rs/users incorrectly says that I invited user "oursin":

Image

Image

jamesnvc:
Add the ability to specify the "viewing" user for the `styled_user_link` helper, so we can not show the "you invited" indicator on the user "tree" view, as that gets cached and may be shown to a different user.

Fixes #1937

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -36,7 +36,7 @@ module UsersHelper
     end
   end
 
-  def styled_user_link user, content = nil, viewing_user = @user, html_options = {}
+  def styled_user_link user, content = nil, html_options = {}, viewing_user = @user
     html_options[:class] ||= []
     if content.is_a?(Story) && content.user_is_author?
       html_options[:class].push "user_is_author"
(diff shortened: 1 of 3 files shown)


pushcx:
This might pass the test, but isn't going to work in prod, unfortunately. I was puzzled by your PR's approach and had to dig in a little.

The root cause of #1937 is that UsersController#tree predates our use of Rails fragment caching (and probably the feature entirely). So the action implements its own caching, which doesn't account for the idea that a username can appear differently to different logged-in users. (We probably have accidentally shown the new "you invited" to logged-out visitors, but nobody has noticed and reported it yet.)

The right fix here would be to remove our action's calls to `Rails.cache`, add `caches_page :tree, if: CACHE_PAGE` (for visitors), and for users, in the view, wrap fragment caching around it with the current user as a key, `<% cache [@user, @newest_user] do %>` (`newest_user` exists in the action but will need to become an instance var to share it).

PR #1944, opened 2026-03-04
Fixes #1941 flag button does not update after action

### LINKED ISSUE #1941: Flag button not updated after flagging post
I've just flagged an entry, and even thought the request was successful, the UI did not update to show the change. Moreover, after clicking "flag" again, a request to `/unvote` is done. Only notable change is that vote score disappears after the first report.

fedemp:


pushcx:
Thanks, you're definitely making better use of the js API than I know to.

PR #1945, opened 2026-03-04
Fixes #1891 Reset textarea after comment

### LINKED ISSUE #1891: Comment textarea is not reset after posting and page reload
In the story details page, e.g. https://lobste.rs/s/lgnaoi/api_tooling_crisis_why_developers_are try posting a top comment. Make sure you have JS enabled.

Comment is published and the textarea disappears.

Reload the page.

The textarea is visible again, but prefilled with the original comment. The published comment is shown in the list as expected.

----

Second part: If you try to publish again without changing the comment, the duplicate comment is published at the top of the list. The HTTP request responds with a 200 ok.

----

Third part: You notice your error. Delete the top/newest comment with the delete button. Reload the page. Now there's only one comment of yours and it's deleted.

----

Fouth part: Undelete your original comment. Reload the page. Textarea is filled yet again. Post duplicate comment again. The new duplicate comment shows as published several minutes before (based on the original comment). I know this is a side effect of the system detecting the duplicate comment, but I though it was worth mentioning.

----

Expected behaviour: Aside from the textarea not being filled, it would be nice if the request failed with a "Duplicate comment detected". Also, whoever takes this should check that the fix works with JS disabled too.

fedemp:
When doing an async request to post a comment, clear the textarea before removing the whole form container. Prevents the textarea from being populated on page reload.

Make the textarea required to submit a comment.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -330,6 +330,7 @@ export class _LobstersFunction {
     qS('#modal_backdrop').remove();
   }
 
+  /** @param { HTMLFormElement } form */
   postComment(form) {
     const formData = new FormData(form);
     const action = form.getAttribute('action');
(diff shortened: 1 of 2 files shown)


pushcx:
Adding `required` is useful, I'd be happy to merge that.

I'm not going to merge this strategy of deleting comment text, though. This is too likely to lose user text, which is the whole reason the site exists. In #1891 I linked to a longer comment about why and an idea for a failsafe way to do so.

fedemp:
Understandable. I'll close this and create a new one only for the `required` since it's a good improvement.

PR #1946, opened 2026-03-04
Use fragment caching instead of Rails.cache for users tree view

### LINKED ISSUE #1937: /users incorrectly says I invited a user

jamesnvc:
Thereby avoiding showing "you invited" indications for other users, based on feedback in #1939 

Fixes #1937

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -52,6 +52,7 @@ class UsersController < ApplicationController
       @users = User.select(*attrs).order(karma: :desc, id: :asc).to_a
       @user_count = @users.length
       @title << " By Karma"
+      @for = :karma
       render action: "list"
     elsif params[:moderators]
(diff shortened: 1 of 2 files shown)


pushcx:
I suggested that cache key and I'm pretty sure I was wrong. In development mode, you can run `rails dev:cache` and restart your server to enable caching. (...you also have to remember to disable it, which I am very good at forgetting to do.) I think this key as written would share the cache between the different ways of sorting the users.

jamesnvc:
Ah, okay, I'll look at that. I had checked with the dev cache turned on to verify that different users didn't see the same thing, but not with different sort options.

pushcx:
Thanks for working through this, fingers crossed I don't break prod again today.

PR #1948, opened 2026-03-05
Account for clock drift in TOTP authentication

danlamanna:
When I logged into lobsters my autofilled TOTP code from bitwarden was invalid. I've never had that happen before and I thought the [spec](https://www.rfc-editor.org/rfc/rfc6238#section-6) had something to say about clock skew:

>  Because of possible clock drifts between a client and a validation
   server, we RECOMMEND that the validator be set with a specific limit
   to the number of time steps a prover can be "out of synch" before
   being rejected.
>
> This limit can be set both forward and backward from the calculated
   time step on receipt of the OTP value.  If the time step is
   30 seconds as recommended, and the validator is set to only accept
   two time steps backward, then the maximum elapsed time drift would be
   around 89 seconds, i.e., 29 seconds in the calculated time step and
   60 seconds for two backward time steps.
>
>   This would mean the validator could perform a validation against the
   current time and then two further validations for each backward step
   (for a total of 3 validations).  Upon successful validation, the
   validation server can record the detected clock drift for the token
   in terms of the number of time steps.  When a new OTP is received
   after this step, the validator can validate the OTP with the current
   timestamp adjusted with the recorded number of time-step clock drifts
   for the token.

ROTP supports checking for drift in [`ROTP::TOTP.verify`](https://www.rubydoc.info/github/mdp/rotp/main/ROTP/TOTP#verify-instance_method).

P.S. Thanks for the devcontainer configuration. It made getting a dev environment almost instant.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -237,7 +237,7 @@ class User < ApplicationRecord
 
   def authenticate_totp(code)
     totp = ROTP::TOTP.new(totp_secret)
-    totp.verify(code)
+    totp.verify(code, drift_behind: totp.interval, drift_ahead: totp.interval)
   end
 
   def avatar_path(size = 100)
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -153,6 +153,21 @@ describe User do
     expect(u.authenticate("hunteR2")).to be false
   end
 
+  it "allows totp codes from the surrounding time steps" do
+    u = create(:user)
+    u.totp_secret = ROTP::Base32.random
+
+    totp = ROTP::TOTP.new(u.totp_secret)
+
+    travel_to Time.current do
+      past_code = totp.at(Time.current - totp.interval)
+      expect(u.authenticate_totp(past_code)).to be_truthy
+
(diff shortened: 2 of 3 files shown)


pushcx:
Thanks for filing this!

I checked on prod, it really does look like our system clock is accurate and it's been correctly synchronized with ntp since its setup. So maybe doublecheck your desktop?

There doesn't seem to be a risk to taking an early TOTP code. If someone can generate a valid one, they have the secret.

Accepting a late one feels a little riskier, but after chatting with some people familiar during [today's office hours](https://push.cx/stream/2026-03-05) and reading section 5.2 (thanks for linking the spec!), it seems like an acceptable risk. I'm going to merge this.

Your PR raises another problem, maybe you'd like to make a second contribution?

> Note that a prover may send the same OTP inside a given time-step window multiple times to a verifier.  The verifier MUST NOT accept the second attempt of the OTP after the successful validation has been issued for the first OTP, which ensures one-time only use of an OTP.

We don't do this! We don't have a `last_authenticated_at` timestamp on the User model to record and pass as the `after` to `verify`. Would you like to fix this oversight?

PR #1952, opened 2026-03-07
Add an option to load avatar from GitHub

### LINKED ISSUE #986: An option to load avatars from GitHub

hugoarnal:
This is a work in progress but I need help because I'm not familiar with either Ruby or Rails but I wanted to work on this issue.

Also I'm unsure what this change will do on the DB, will Gravatar be the default or not? How can this be set?

Closes #986

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -281,7 +281,7 @@ class SettingsController < ApplicationController
 
   def user_params
     params.require(:user).permit(
-      :username, :email, :password, :password_confirmation, :homepage, :about,
+      :username, :email, :password, :password_confirmation, :homepage, :about, :avatar_preference,
       :email_replies, :email_messages, :email_mentions, :inbox_mentions,
       :pushover_replies, :pushover_messages, :pushover_mentions,
       :mailing_list_mode, :show_email, :show_avatars, :show_story_previews,
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -33,6 +33,12 @@ class SettingsController < ApplicationController
     redirect_to "/"
   end
 
+  def update_avatar_preference
+    flash[:success] = "Updated avatar preference. Your avatar isn't updated, press the Clear Cache button next to Gravatar."
+    @user.update(user_params)
+    redirect_to "/settings"
+  end
+
   def update
     @edit_user = @user.clone
 
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -33,7 +33,7 @@ class SettingsController < ApplicationController
     redirect_to "/"
   end
 
-  def update_avatar_preference
+  def update_avatar_source
     flash[:success] = "Updated avatar preference. Your avatar isn't updated, press the Clear Cache button next to Gravatar."
     @user.update(user_params)
     redirect_to "/settings"
@@ -287,7 +287,7 @@ class SettingsController < ApplicationController
 
   def user_params
     params.require(:user).permit(
-      :username, :email, :password, :password_confirmation, :homepage, :about, :avatar_preference,
+      :username, :email, :password, :password_confirmation, :homepage, :about, :avatar_source,
       :email_replies, :email_messages, :email_mentions, :inbox_mentions,
       :pushover_replies, :pushover_messages, :pushover_mentions,
       :mailing_list_mode, :show_email, :show_avatars, :show_story_previews,
--- a/app/models/avatar.rb
+++ b/app/models/avatar.rb
@@ -1,0 +1,3 @@
+class Avatar < ApplicationRecord
+  enum :type, Gravatar: 0, GitHub: 1
+end
--- a/app/models/avatar.rb
+++ b/app/models/avatar.rb
@@ -1,3 +1,0 @@
-class Avatar < ApplicationRecord
-  enum :type, Gravatar: 0, GitHub: 1
-end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -102,6 +102,7 @@ class User < ApplicationRecord
     s.string :mastodon_instance
     s.string :mastodon_oauth_token
     s.string :mastodon_username
+    s.string :avatar_preference
     s.string :homepage
   end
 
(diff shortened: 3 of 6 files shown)


app/models/user.rb
```diff
@@ -369,19 +370,22 @@ def comments_deleted_count
...
   def fetched_avatar(size = 100)
-    gravatar_url = "https://www.gravatar.com/avatar/" <<
-      Digest::MD5.hexdigest(email.strip.downcase) <<
-      "?r=pg&d=identicon&s=#{size}"
+    url = "https://www.gravatar.com/avatar/" <<
+        Digest::MD5.hexdigest(email.strip.downcase) <<
+        "?r=pg&d=identicon&s=#{size}"
+    if avatar_preference == "1"
+      url = "https://www.github.com/" << github_username << ".png"
+    end
```
hugoarnal, line 388: I wanted for Gravatar to stay the default, hence why it's the 0 in the selection array.

However I'm probably going to turn this random `"1"` into an enum.

app/views/settings/index.html.erb
```diff
@@ -49,6 +49,15 @@
...
         </div>
+
+        <%= f.label :avatar_preference, "Avatar" %>
+        <div>
+          <div>
+            <span>Choose your preferred method of avatar display</span>
+          </div>
+
+          <%= f.select :avatar_preference, [["Gravatar", 0], ["GitHub", 1]] %>
+        </div>
```
hugoarnal: This part has a few problems:

1. It looks ugly
2. It repeats with the Gravatar option down here https://github.com/lobsters/lobsters/blob/e4dfbc1bf8d13d409c983119ad7ccb2b1ff1f27b/app/views/settings/index.html.erb#L219-L224
I tried removing it and putting the `Expiry cache` button up here. Unfortunately I couldn't figure out how to make it work after moving it.
3. Definitely needs better phrasing

app/models/user.rb
pushcx: Yeah, I would much rather see an enum with values like `gravatar` and `github` than numerics or constants.

app/views/settings/index.html.erb
```diff
@@ -49,6 +49,15 @@
 
           <%= render :partial => "global/markdownhelp" %>
         </div>
+
+        <%= f.label :avatar_preference, "Avatar" %>
+        <div>
+          <div>
+            <span>Choose your preferred method of avatar display</span>
+          </div>
```
pushcx:
```suggestion
        <span class="hint">
          Lobsters fetches and serves avatars directly for visitor privacy
        </span>
```

app/views/settings/index.html.erb
pushcx: Yeah, there's multiple form tags on this page. Moving this into the settings section would've nested forms.

I think it would be better design to move the single 'avatar source' setting down here to the 'External Accounts' section. So it'll submit just the one value, which should be fine, that shouldn't blank any settings. Duplicate the form tag it's in now, but don't move the gravatar/github account settings into it.

pushcx (commented):
Well, you certainly picked an ambitious first project, and you'll get a good tour as you work up and down the stack to get this implemented.

Feel free to leave questions here or on zulip, I'm happy to give Rails pointers.

app/controllers/settings_controller.rb
```diff
@@ -33,6 +33,12 @@ def deactivate
     redirect_to "/"
   end
 
+  def update_avatar_preference
+    flash[:success] = "Updated avatar preference. Your avatar isn't updated, press the Clear Cache button next to Gravatar."
+    @user.update(user_params)
+    redirect_to "/settings"
```
hugoarnal, line 39: Need to still figure out how to clear the avatar cache

hugoarnal:
Unsure what to do about the failing CI.

I think it's telling me to create a new table because I created a new ApplicationRecord but it's only used for enums at the moment so it doesn't seem necessary.

hugoarnal:
It is open, so yes indeed.

app/controllers/settings_controller.rb
pushcx: It's the `expire` action in `app/controllers/avatars_controller.rb`. The clear happens inline there, so it should be extracted to its own function.

We're going to have avatar logic split in a clunky way, where the Gravatar is a single field on the User model so that has a bunch of code for gravatars, and your new Avatar class has parallel functions for working with uploaded avatars.

It's a bigger refactor, but it would be good to pull all the gravatar-related code out from User and the controllers to its own Gravatar model.

app/controllers/settings_controller.rb
```diff
@@ -281,7 +287,7 @@ def github_disconnect
 
   def user_params
     params.require(:user).permit(
-      :username, :email, :password, :password_confirmation, :homepage, :about,
+      :username, :email, :password, :password_confirmation, :homepage, :about, :avatar_preference,
```
pushcx, line 290: I know this is a chore, but how about `avatar_source`? All of these are "preferences" so it doesn't say much - what the user is preferring is where their avatar comes from.

app/models/avatar.rb
```diff
@@ -0,0 +1,3 @@
+class Avatar < ApplicationRecord
```
pushcx, line 1: The build is failing because you declared this model to inherit from ApplicationRecord, which is for models that are backed by SQL tables.

I don't think you need this class? Or at least not with this contents. This enum looks like it's info about the user preference, so it should be on the User model.

config/routes.rb
```diff
@@ -174,6 +174,7 @@
 
   get "/settings" => "settings#index"
   post "/settings" => "settings#update"
+  post "/settings/update_avatar_preference" => "settings#update_avatar_preference", :as => "update_avatar_preference"
```
pushcx, line 177: This doesn't need its own route if it's a user setting, it can just post to /settings like the `form_with model: @edit_user` on line 8.

pushcx:
Looking good. What work do you see remaining? I see the db_consistency gem complaining a little, that should be minor, but functionally I think this may be enough? Just grab from github and put it in the same dir. Does it handle multiple sizes the same?

hugoarnal:
> Looking good. What work do you see remaining? I see the db_consistency gem complaining a little, that should be minor, but functionally I think this may be enough? Just grab from github and put it in the same dir. Does it handle multiple sizes the same?

Yes, this is almost done. I just haven’t had the time to look into it much which is why I drafted it again. Just need to address some of your previous comments also. Will test handling sizes.

I’ll let you know once this is ready to go

config/routes.rb
hugoarnal: To be honest, I don't really know what to do here. The entire `External Accounts` part is part of no form_with which is why I created a route. Posting it to settings returns an error where the password needs to be set. I'm going to leave it as such for now.

hugoarnal:
Not stale.
However I'm probably not going to work much on it anymore, so if someone wants to take it over don't hesitate :)

config/routes.rb
pushcx: Oh! I hadn't realized, thanks. You made the right call here, I was wrong.

pushcx:
Thanks for being clear about your availability and willingness to hand off. You made a great start, let me see if I can get this over the line.

pushcx:
Thanks again, Hugo. I'm deploying this now.

hugoarnal:
> Thanks again, Hugo. I'm deploying this now.

Thank you, was quite cool discovering Ruby on Rails like this :)

PR #1962, opened 2026-03-14
UI improvements for "hide" action

### LINKED ISSUE #1959: [Bug] Can't hide merged stories
As a normal user, I can't hide merged stories even when the button is there waiting to be clicked.

e.g. in https://lobste.rs/s/jr3zym/relicensing_with_ai_assisted_rewrite I can't hide neither 'AI And The Ship of Theseus' nor 'Is legal the same as legitimate: AI reimplementation and the erosion of copyleft' (the two merged stories).

----

After reporting this, I loaded an older version of the site locally and noticed that the merged story calls the same endpoint as the main story, so current implementation should mimic that.

fedemp:
Use the main story endpoint for saving and hiding in merged stories. Fixes #1959.

Don't show hidden users count if others users count is 0.

In single story page, if the user has hidden the story and we are showing the "hidden by you and others", toggle the "by you" part is user unhides the story.

In the same page, toggle the visibility of the "you have hidden this story" paragraph after toggling the story hidden status.

Allow the "hide" cta in the "you have hidden this story" message to do async requests too.

Fix styling for "hide" button in the "you have hidden this story" so it looks like a link in the paragraph.

Fix background for save and hide buttons when an story is target link.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -102,9 +102,9 @@ function hasProperties(obj, props) {
   return isObject(obj) && props.every((p) => p in obj);
 }
 
-/** @param {unknown} obj @returns obj is object */
+/** @param {unknown} obj */
 function isObject(obj) {
-	return typeof obj === "object" && !!obj;
+	return !!obj && typeof obj === "object"
 }
 
 /** @param {string} msg */
@@ -118,16 +118,7 @@ function notify(msg) {
 /** @param {SubmitEvent} ev */
 export async function asyncFormSubmit(ev) {
   ev.preventDefault();
-
-  const submitter = /** @type {HTMLInputElement} */ (ev.submitter);
   const form = /** @type {HTMLFormElement} */ (ev.target);
-
-  // Prevent double submissions
-  if (submitter.disabled) {
-    return;
-  }
-
-  submitter.disabled = true;
 
   try {
     const response = await fetch(form.action, {
@@ -157,8 +148,6 @@ export async function asyncFormSubmit(ev) {
     const error = err instanceof Error ? err : new Error(JSON.stringify(err));
     notify(error.message);
     throw error;
-  } finally {
-    submitter.disabled = false;
   }
 }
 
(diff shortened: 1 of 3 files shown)


fedemp:
**Don't show hidden users count if others users count is 0.** Currently: 
image

**In single story page, if the user has hidden the story and we are showing the "hidden by you and others", toggle the "by you" part is user unhides the story.**

[Kooha-2026-03-14-00-16-59.webm](https://github.com/user-attachments/assets/1d1b04f7-c090-417c-8b2a-496da1fa2c77)

**In the same page, toggle the visibility of the "you have hidden this story" paragraph after toggling the story hidden status.**

[Kooha-2026-03-14-12-33-44.webm](https://github.com/user-attachments/assets/01a5e7b8-abe5-4c6f-8a4d-9fb02fe3d8fb)

**Allow the "hide" cta in the "you have hidden this story" message to do async requests too.**
[Kooha-2026-03-14-12-36-25.webm](https://github.com/user-attachments/assets/aa75e5fa-5550-4f2e-aafa-28f44313e9b4)

**Fix styling for "hide" button in the "you have hidden this story" so it looks like a link in the paragraph.** Currently:

image

**Fix background for save and hide buttons when an story is target link.** Currently:

image

app/javascript/application.js
```diff
@@ -378,76 +367,86 @@ export class _LobstersFunction {
 
   /** @param {SubmitEvent} ev */
   async hideStory(ev) {
-    await Lobster._handleStoryAction(ev, {
-      selector: ".hider",
-      stateProp: "hidden",
-      classToggle: "hidden",
-      activeText: "unhide",
-      inactiveText: "hide",
-    });
```
fedemp, line 395: I had to remove all this "automagic" because, as seen with "hide", many things may happen when an action is triggered, and the submit event may not always come from the expected place. e.g. the "You have hidden this story" must not change its button text, and it's not a child of `.story`.

Something similar happens when working on upvotes. we won't update a button text but we must remove the flags and replace the vote count which is outside the button.

app/javascript/application.js
```diff
@@ -378,76 +367,86 @@ export class _LobstersFunction {
...
+    const submitter = /** @type HTMLInputElement */ ( ev.submitter );
+    const story = submitter.closest('.story');
```
fedemp: Not using the `parentSelector` here because I don't want it to throw an error. I just need to know if the submitter was inside a `.story` (i.e. a "hide" button" in `.byline`) or not (i.e. the "unhide" in `#hide-alert`).

app/javascript/application.js
```diff
@@ -378,76 +367,88 @@ export class _LobstersFunction {
...
   /** @param {SubmitEvent} ev */
   async hideStory(ev) {
-    await Lobster._handleStoryAction(ev, {
-      selector: ".hider",
-      stateProp: "hidden",
-      classToggle: "hidden",
-      activeText: "unhide",
-      inactiveText: "hide",
-    });
+    const submitter = /** @type HTMLInputElement */ ( ev.submitter );
```
pushcx, line 370: This would be a good opportunity to rename this variable. "submitter" always sounded like "the person who submitted the story" and I've learned that it means "the DOM elements that was clicked to submit this event". It hasn't ever been worth the churn to rename it, but as long as you're moving it between functions, would you like to give it a better name?

app/javascript/application.js
```diff
@@ -378,76 +367,88 @@ export class _LobstersFunction {
...
+    const story = submitter.closest('.story');
+
+    // If no .story, action was triggered by #hide-alert.
+    const submitters = ( story ?
```
pushcx, line 374: It feels like these two code paths are the same? The first one searches in the context of `story`, the second (implicitly) in `document`. Then the same attempt to find the `hider` submit button. Then they have two slightly different ways of finding a nested element they care about - I'm not so certain about this last part being equivalent.

I think this would also let you tidy up the `story ?? qS('.story')` below.

app/javascript/application.js
pushcx: Thanks for this explanation. Yeah, it's gone from these UI elements having very similar workflow to not having it. You made a good call breaking things out instead of trying to force it to adapt by adding a callback arg or something.

pushcx:
Thanks, this all looks like good UI polish. I see what you're doing in the code. I see one opportunity to further improve it with a variable rename, if you want to do that. And then I had a question about maybe tidying a boolean.

app/javascript/application.js
fedemp: I think I understand your comment cause at first I was confused about why I did write those lines.

If `story` is defined, we are either on a details page or in the list page; can't know for sure. So we query *all* hide buttons in the *context of the story* cause for merged stories must have more than one. Plus, just in case, we also query the `#hide-alert`.

If `story` is not defined, `#hide-alert` was the trigger. We are in single details page. We can query all hide buttons in the document and get them all for merged stories and the `#hide-alert` too.

pushcx: Thanks, with the comment and your explanation it's clear to me what's happening now, let me get this merged.

PR #1963, opened 2026-03-14
Fixes #1942 can't edit own comment in "Newest comments".

### LINKED ISSUE #1709: Editing a comment orphans reply
When I edited a comment that had a reply, it moved to the top and the reply was orphaned. The reply looks like it's a reply to a different comment, and the dashed parent indicator is broken.

Image

### LINKED ISSUE #1942: Editing my own comment in "Newest comments" trigger JS error
Post a comment in any story. Go to https://lobste.rs/comments and your comment should be there at the top. Press "Edit".

First issue: the "by line" line disappears.

Press "Update". Your comment now dissapears and a JS error is thrown.

```
Uncaught (in promise) TypeError: can't access property "insertAdjacentHTML", comments.querySelector(...) is null
    postComment http://localhost:3000/assets/application-530421f0.js:364
    promise callback*postComment/< http://localhost:3000/assets/application-530421f0.js:343
    promise callback*postComment http://localhost:3000/assets/application-530421f0.js:342
    <anonymous> http://localhost:3000/assets/application-530421f0.js:897
    on http://localhost:3000/assets/application-530421f0.js:24
    on http://localhost:3000/assets/application-530421f0.js:22
    on http://localhost:3000/assets/application-530421f0.js:21
    <anonymous> http://localhost:3000/assets/application-530421f0.js:895
application-530421f0.js:364:24
```

fedemp:
Fixes #1942. Fixes #1709 too. editing a reply orphaned it from its parent comment (in the UI)

In "Newest comments", When editing a comment, don't replace the whole comment container but only the container of the text comment.

When submitting a comment, check if the form is inside a comment; this indicates that user was editing a comment.

### DIFF
--- a/app/views/comments/_commentbox.html.erb
+++ b/app/views/comments/_commentbox.html.erb
@@ -7,7 +7,7 @@
   </ol>
 <% end %>
 
-<div class="comment comment_form_container" data-shortid="<%= comment.short_id if comment.persisted? %>">
+<div class="comment_form_container" data-shortid="<%= comment.short_id if comment.persisted? %>">
 <%= form_with url: comment do |f| %>
   <% if comment.errors.any? %>
     <%= errors_for comment %>
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -323,18 +323,26 @@ export class _LobstersFunction {
           if (replyForm) {
             // user submitted from a temporary reply form, so this is a reply to an existing comment
             replace(replyForm, text)
-          } else {
-            // Iterating up the comments tree to the nearest parent. If there isn't one, we are creating
-            // a top-level comment, so find the top of the comments tree.
-            const comments = form.closest('.comments') || qS('.comments')
-            parentSelector(form, '.comment_form_container').remove()
+            return;
+          }
 
-            // if comments is .comments1, it is top-level comment: insert it deeper
-            if (comments.classList.contains('comments1')) {
-              comments.querySelector('#story_comments .comments').insertAdjacentHTML("afterbegin", text)
-            } else {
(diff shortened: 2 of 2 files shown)


app/views/comments/_commentbox.html.erb
```diff
@@ -7,7 +7,7 @@
   </ol>
 <% end %>
 
-<div class="comment comment_form_container" data-shortid="<%= comment.short_id if comment.persisted? %>">
+<div class="comment_form_container" data-shortid="<%= comment.short_id if comment.persisted? %>">
```
pushcx, line 10: Whew, good catch here. I had to go check all 76 instances of `.comment` in the css to confirm this didn't break anything, but I see how this is a good little fix, thank you for being thorough.

pushcx:
This is a great cleanup, thank you.

You've contributed a lot to the frontend the last few months. The comment form code has gotten fairly hairy over the site's 14 years because it's our god object. The template reuse is nice, but then the js has sprouted all of these fiddly little paths to deal with them working slightly differently in different contexts. Do you see a way to clean up all these special cases and prevent this class of bugs where the matrix of pages * edit vs reply * top-level vs inline means that fixing one case breaks some other case that we didn't think about? You've got the most context on this right now and I've appreciated the care you've put into these PRs. Maybe you see a way out?

fedemp:
I get what you say; this is like working with jQuery and even then, I don't remember it being so fragile.

If we were to continue like this, maybe we should be more rigid and add some error throwing when an querySelector does not match. I was even thinking if it would be possible to add some form of end to end testing and capture snapshots; at work we use Playwright for a React app.

Or we could go the really exciting way and migrate to HTMX. I'd love to work on something like that.

Issue #1964, opened 2026-03-16
Followup comment by moderator to a moderator comment does not appear in modlog


gustafe:
On 2026-03-15 18:25, moderator Irene posted a comment <https://lobste.rs/s/dw0hx5#c_bzeqf7> using the Sysop hat.

On 2026-03-15 18:41:29, moderator 355E3B posted a direct reply (https://lobste.rs/c/8bl89s), also using the Sysop hat.

Both comments are marked with hats in the thread:

<img width="1071" height="998" alt="Image" src="https://github.com/user-attachments/assets/245fe1c3-0f5f-4892-a6dd-04f47a311420" />

However, in the modlog, the latest entry is the first one, from Irene:

<img width="1244" height="630" alt="Image" src="https://github.com/user-attachments/assets/8fccb99e-2dc2-46a5-9089-317e16558939" />

pushcx:
I think this was two bugs. First, useful history:

On 2017-10-25 I added the functionality to modlog hats in bfc8cc0c9d5d49597fd74d2a7c9f7127b405f389, [announcement](https://lobste.rs/s/kmpizq/deleting_comments_instead_threads).

On 2023-07-03, I added 355e3b and aleph- [as moderators](https://lobste.rs/s/hhoopi/lobsters_is_11_today).

First bug, when I granted those two hats in the rails or db console (I don't remember which), I didn't set `modlog_use`. I caught it while doublechecking data from the second bug:

```
MariaDB [lobsters]> select username, modlog_use from hats join users on hats.user_id = users.id where hat = 'Sysop';               
+----------+------------+                                                                                                          
| username | modlog_use |                                                                                                          
+----------+------------+                                                                                                          
| kyle     |          1 |                                                                                                          
| jcs      |          1 |                                                                                                          
| Irene    |          1 |                                                                                                          
| pushcx   |          1 |                                                                                                          
| alynpost |          1 |                                                                                                          
| 355E3B   |          0 |                                                                                                          
| aleph    |          0 |                                                                                                          
+----------+------------+                                                                                                          
7 rows in set (0.004 sec)
```

I have corrected these in the prod db. I've been reluctant to add an admin "grant a hat" form to the web UI to prevent the chore involved when a user requests a hat by email/DM instead of [the hat request form](https://lobste.rs/hat_requests/new). This bug is some evidence that was a mistake.

[...]

PR #1965, opened 2026-03-16
Log modlog hats when they're added to a comment in an edit

chamlis:
Moving this callback to run on `after_save` was the only way I could find to make `previously_new_record?` and `hat_previously_changed?` reliable. I believe this is because the model may get changed multiple times during the transaction, so it's unlikely the callback will run while the changes we're interested in are still the most recent ones.

This may be the cause of #1964.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -35,9 +35,12 @@ class Comment < ApplicationRecord
   validates :short_id, presence: true, uniqueness: {case_sensitive: false}
 
   before_validation :assign_initial_attributes, on: :create
+
+  # We log hats before the transaction ends so we can accurately check whether the model changed
+  after_save :log_hat_use
   # Calling :record_initial_upvote from after_commit instead of after_create minimizes how many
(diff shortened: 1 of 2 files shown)


pushcx:
This is the only way I can see #1964 happening, thanks for diagnosing this. Could you add a test? This seems really easy to break and violates our commitment to mod transparency.

I'm taking a look at the prod database now to figure out how many times this has happened.

chamlis:
There should already be three tests in here, and as ever I'm open to feedback both about the choice of tests and their composition:
1. Creating a comment with a modlog hat logs
2. Adding a modlog hat in an edit logs
3. Editing a hatted comment and retaining the modlog hat *doesn't* log

I'm crossing my fingers there haven't been that many instances of this requiring manual cleanup.

pushcx:
Whoops, I started reading the diff, went "oh no I have to check how many times this happened" and failed to scroll down to see the tests. Thanks for a thorough fix, I'll merge + deploy this now.

Looks like there were 11 comments affected.

Issue #1967, opened 2026-03-17
Flagging comments trigger JS error


fedemp:
Maybe related to #1966 but the JS error is different. I tried to flag a comment, and I got this JS error. Comment was correctly flagged but it seems that setting the reason triggers the error.

```
Uncaught TypeError: can't access property "toLowerCase", Lobster.commentFlagReasons[reason] is undefined
    voteComment https://lobste.rs/assets/application-707868f4.js:641
    <anonymous> https://lobste.rs/assets/application-707868f4.js:895
    on https://lobste.rs/assets/application-707868f4.js:24
    on https://lobste.rs/assets/application-707868f4.js:22
    on https://lobste.rs/assets/application-707868f4.js:21
    <anonymous> https://lobste.rs/assets/application-707868f4.js:892
application-707868f4.js:641:44
```

fanf2:
I'm still seeing this problem. I guess based on the cache busting hash the fix hasn't been deployed yet?

```
[Error] TypeError: undefined is not an object (evaluating 'Lobster.commentFlagReasons[reason].toLowerCase')
	voteComment (application-707868f4.js:641)
	(anonymous function) (application-707868f4.js:895)
	(anonymous function) (application-707868f4.js:24)
```

pushcx:
Thanks for reporting. This was fixed by #1967.

PR #1980, opened 2026-03-20
Fix comment flagging (#1967)

sknebel:
In #1944 the structure for the data about flag reasons was changed, but only the code for the story flag menu was adjusted. Adjust comment flag menu to match.

I've only tested this in browser devtools, not on a deployment of the site, so please double-check it's applied correctly.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -889,10 +889,10 @@ document.addEventListener("DOMContentLoaded", () => {
     Lobster.modalFlaggingDropDown("comment", event.target, reasons);
   });
 
-  on('click', '.comment #flag_dropdown a', (event) => {
+  on('click', '.comment #flag_dropdown a', (/** @type MouseEvent */ event) => {
     event.preventDefault();
-    if (event.target.getAttribute('data') != '') {
-      Lobster.voteComment(parentSelector(event.target, '.comment'), -1,  event.target.getAttribute('data'));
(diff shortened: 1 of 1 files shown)


fedemp:
I've checked your branch. It does not trigger a JS error anymore, which is great. But on the other side, the reason is not appended to the flag button. Even so, that can be tackled in another issue.

app/javascript/application.js
```diff
@@ -889,10 +889,10 @@ document.addEventListener("DOMContentLoaded", () => {
     Lobster.modalFlaggingDropDown("comment", event.target, reasons);
   });
 
-  on('click', '.comment #flag_dropdown a', (event) => {
+  on('click', '.comment #flag_dropdown a', (/** @type MouseEvent */ event) => {
```
pushcx, line 892: Hey @fedemp someone else is using your typescript annotations. 😄

pushcx:
I'm sorry I let such an important site feature stay broken so long. This landed just as I stepped away for some travel.

@sknebel Thanks for fixing this.

@fedemp Do you want to handle the 'reason' issue you just spotted? I'd love to get that in promptly, and I agree it's worth merging this first.

app/javascript/application.js
fedemp: They are trying to steal my spotlight! /s

PR #1982, opened 2026-03-20
Fixes #1972 "Caches" dropdown in merged stories

### LINKED ISSUE #1972: Caches dropdown does not work in merged stories
Check https://lobste.rs/s/n4kbuj/ai_agent_published_hit_piece_on_me#comments_story_01kh9dhrn1ewwbp39hyct7894z

It's the title of one of the merged stories that wraps its own comments. The "caches" dropdown is there, but clicking it does not work.

fedemp:
Fixes #1972. Refactor the "caches" dropdown to use `<details>` so it can be used with a keyboard.

Use anchor positioning so it won't overflow outside of viewport.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -948,22 +948,42 @@ div.comment_text code {
 	background-color: var(--color-box-bg-shaded);
 }
 
-.caches-dropdown {
-	width: fit-content;
-	margin: 0 auto;
+.caches {
+	display: inline-block;
+	position: relative;
 }
 
-.caches-dropdown a {
-	color: var(--color-fg-contrast-4-5);
+.caches summary {
+	list-style: none;
+}
+
+.caches ul {
+	position: absolute;
+	background: var(--color-bg);
+	border: 1px solid var(--color-box-border);
+	white-space: nowrap;
+	list-style: none;
+	padding: 0;
+	z-index: 1;
+}
+
+.caches li {
+	border-bottom: 1px solid var(--color-box-border);
+}
+
+.caches li:last-child {
+	border-bottom: 0;
+}
+
+.caches a {
+	color: var(--color-fg-contrast-7-5);
 	text-decoration: none;
+	display: block;
+	padding: 3px 7px;
 }
 
-.caches-dropdown a:hover {
+.caches a:hover {
 	text-decoration: underline;
-}
-
-.details:has(> .caches_button:checked) label.caches::before {
-	content: "▾ ";
 }
 
 #flag_dropdown a.cancel-reason {
@@ -979,17 +999,6 @@ div.comment_text code {
 	bottom: 0;
 	right: 0;
 	z-index: 1;
-}
-
-.caches_button {
-	display: none;
-}
-.caches_button:not(:checked) ~ .caches-dropdown {
-	display: none;
-}
-label.caches {
-	cursor: pointer;
-	white-space: nowrap;
 }
 
 .controls details {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/application.css
```diff
@@ -948,22 +948,53 @@ div.comment_text code {
...
+	list-style: none;
+	padding: 0;
+	z-index: 1;
+}
+
+@supports (position-anchor: --caches) {
+	.caches ul {
+		position: fixed;
+	}
+}
```
fedemp: I know I broke the site (at least) once when using popovers but this time it's different. Even thought every `anchor` property is still _newly available_, the dropdown works fine without them; they are only there so the dropdown will flip its side before overflowing outside of the viewport.

This is latest Firefox.

https://github.com/user-attachments/assets/ef8580c2-7dd2-43db-a296-a41e25a2af4d

And this is Firefox ESR

https://github.com/user-attachments/assets/9a7a0d3a-345f-47fd-9e42-80de1d9c87d5

If you are not fine with using this features, we could settle on what ESR does.

pushcx:
Sorry to leave this hanging out while I was traveling.

I can see that you're doing progressive enhancement, but I had an idea for building out a proper detector for the baseline standard. I'd like to land that so I don't have to manually keep hitting caniuse on frontend questions. I kicked this idea around on the [last office hours stream](https://push.cx/stream/2026-04-16-/stream/2026-04-16-give-a-mouse-a-cookie-banner) and the first pass I took at it over the weekend is roughly correct. I'll try to land that in the next day or two.

pushcx:
I was got this working in rspec in 8221ec52 and fixed up some places where we'd missed the mark on the baseline.

Thinking about your PRs like this one using anchor positioning and the previous one using popover, I added a [`WANTED_PROPERTIES`](https://github.com/lobsters/lobsters/blob/8221ec52ee8444240dc4777969283d381b4ba8e8/spec/css_widely_available_spec.rb#L21-L22) test so that we get alerted when CSS features we'd like to use become widely available, then there's a comment with notes for the planned CSS change. Would you like to PR those two into the spec? Maybe more that I can't think of off the top of my head?

fedemp:
I don't agree with everything that was removed, but at least we now have a more clear law of what's allowed.

About the PR, I have removed the references to `anchor`. Upside is that it now works for keyboard users; downside it will overflow to the right if the by line and the screen are the right size.

pushcx:
> I don't agree with everything that was removed

Do you mean they should have been given exceptions, that the workarounds I found are not good enough equivalents, or something else?

It occurred to me that I should probably have added all of these to `WANTED_PROPERTIES`, too.

As long as I'm writing - I saw you left reviews on a couple of the 6 LLM-authored PRs we've gotten in the last week. I don't know what's prompted this wave of spam, but I'm sorry you lost time to it. The common elements I'm seeing are it's the contributor's first PR, they have an empty profile or a ridiculous amount of activity, they don't write comments or produce a wall of bullet points and emojis.

app/views/stories/_listdetail.html.erb
```diff
@@ -118,7 +118,13 @@ class="story <%= story.current_upvoted? ? "upvoted" : "" %>
         <% end %>
         <% if story.url.present? && (!story.is_gone? || @user.try(:is_moderator?))  %>
           <%= divider_tag %>
-          <label for="caches_<%= story.short_id %>" class="caches" tabindex="0">caches</label>
+          <details class="caches" name="caches">
```
pushcx, line 121: I played with this locally and was pleasantly surprised to see that I could only have one menu open at once. I learned from this PR how `name` works on `details`, thanks!

pushcx (approved):
Looks great, thanks for working with our conservative baseline!

Issue #1840, opened 2025-12-27
Upvoted Comments page has incorrect pagination


hugoarnal:
When logged in, in the `/upvoted/comments`, if you have less than 20 comments upvoted, it still shows the "Page 2" button. When clicked, it goes to page 2 which is empty.

<img width="250" height="164" alt="Image" src="https://github.com/user-attachments/assets/dbc314d3-aa0b-4dbe-90e0-25bf238179bc" />

I assume this is a bug as Upvoted Stories (`/upvoted/stories`) doesn't show this button when you don't have enough stories for another page to show.

hugoarnal:
Just saw that other issues about pagination (#1823) also link to #394

hugoarnal:
After finally hitting the Page 2 requirement, I'm seeing that Upvoted Comments seem to always have a +1 (I have the button to go to Page 3 with only having 1 comment liked on Page 2)

pushcx:
Thanks for this clear description, and for even connecting the root issue.

Contributors: I'd happily take a fix just for this bug, or for the larger needed cleanup of pagination.

fedemp:
Should this be closed now that #1986 was merged?

hugoarnal:
> Should this be closed now that [#1986](https://github.com/lobsters/lobsters/pull/1986) was merged?

Indeed, seems to be fixed on my end, thanks!

PR #1986, opened 2026-03-21
fix pagination on Upvoted Comments page

jhuang97:
This just fixes the bug reported in #1840 by adding another database query to get the number of upvoted comments, and then using that to decide whether to show the next page button.  I have not cleaned up pagination in general.

While looking into this, I noticed that the main /comments page also has the same bug.  However, seeing as I have to go past the end of all the comments to see this issue with pagination (for me on the main lobste.rs website, this was page 32387), I think it is rare for people to run into the bug on /comments. I didn't fix this case, wasn't sure if we'd want to add a query that counts all the comments.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -298,6 +298,8 @@ class CommentsController < ApplicationController
       .limit(COMMENTS_PER_PAGE)
       .offset((@page - 1) * COMMENTS_PER_PAGE)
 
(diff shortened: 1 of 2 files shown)


pushcx:
Thanks for taking this bug! As I said, I'm very happy to merge a fix for this specific bug. You made the right call on not fixing `/comments`, it's really only bots that dig back that far.

Issue #1973, opened 2026-03-17
User should not be able to flag own comments


fedemp:
It seems I can flag my own comments, and the server will send an `ok` when I do the request, but after reloading the page, I noticed that my comment was back to normal.

The server should have returned an error, and the "flag" button should not be available at all.

Slightly related: I tried deleting my comment, and that disabled the "flag" button. Before reloading the page, I undeleted my comment, but the "flag" button stayed disabled.

rebecaaras:
Possible solution, something like:

[app/models/comment.rb -> L390](https://github.com/lobsters/lobsters/blob/bd8fa141b6e99fb2bdf6f4ad2950b5561973e1cb/app/models/comment.rb#L378)
```ruby
def is_flaggable?(user)
  if user && user.id == user_id
    return false
  end
  if is_gone?
    return false
  end
...
```
or ...
```ruby
def is_flaggable?(user)
  if is_editable_by_user?(user)
    return false
  end
  if is_gone?
    return false
  end
...
```
Decided to tackle this after (and if!) #1966 is merged because of possible conflicts, maybe shouldn't have...?

fedemp:
IMO if `is_flaggable` returns false, the flag button (or HTML link) should not be rendered at all. That way we tackle two issues with one solution.

rebecaaras:
@fedemp you're right! I refrained because adding an argument to the function means I should update it everywhere to receive `user`... I'll look into it soon tho.

Edit: Looking at it later, changing `can_flag?` function to receive `user`, once it uses `Comment#is_flagglable?` feels completely off :( Relatedly, in case this is implemented, should the `is_flaggable?` be changed for `Stories` to, for symmetry reasons?
[/user.rb](https://github.com/lobsters/lobsters/blob/bd8fa141b6e99fb2bdf6f4ad2950b5561973e1cb/app/models/user.rb#L312)
```ruby
def can_flag?(obj, user) # really?
  if is_new?
    return false
  elsif obj.is_a?(Story)
    if obj.is_flaggable?
      return true
    elsif obj.current_flagged?
      # user can unvote
      return true
    end
  elsif obj.is_a?(Comment) && obj.is_flaggable?(user) #<--
    return karma >= MIN_KARMA_TO_FLAG
  end

  false
end
``

pushcx:
`is_editable_by_user` depends on the time and things are flaggable longer than they're editable, so it wouldn't be a good reuse.

pushcx:
[...]

PR #1988, opened 2026-03-25
Fix #1966 and #1973 Extend unflaggable comments stories

### LINKED ISSUE #1966: Flagging a deleted comment triggers a JS error
Just recently, out of boredom, I tried to flag this comment https://lobste.rs/s/3sxwnf/silent_struggle_anxiety_software#c_oj5muv after it was deleted, and I got a JS error.

```
Uncaught TypeError: can't access property "innerHTML", scoreLink is null
    voteComment https://lobste.rs/assets/application-707868f4.js:588
    <anonymous> https://lobste.rs/assets/application-707868f4.js:895
    on https://lobste.rs/assets/application-707868f4.js:24
    on https://lobste.rs/assets/application-707868f4.js:22
    on https://lobste.rs/assets/application-707868f4.js:21
    <anonymous> https://lobste.rs/assets/application-707868f4.js:892
    EventListener.handleEvent* https://lobste.rs/assets/application-707868f4.js:652
application-707868f4.js:588:26
```

BTW, should I be able to flag deleted comments or should we remove the button?

rebecaaras:
Flag button is now not rendered for deleted comments, own comments and own stories.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -391,6 +391,9 @@ class Comment < ApplicationRecord
   end
 
   def is_flaggable?
+    if is_gone?
+      return false
+    end
     if created_at && self.score > FLAGGABLE_MIN_SCORE
       Time.current - created_at <= FLAGGABLE_DAYS.days
     else
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -391,12 +391,7 @@ class Comment < ApplicationRecord
   end
 
   def is_flaggable?(user)
-    if is_gone?
-      return false
-    end
-    if user && user.id == user_id
-      return false
-    end
+    return false if is_gone? || user&.id == user_id
     if created_at && self.score > FLAGGABLE_MIN_SCORE
       Time.current - created_at <= FLAGGABLE_DAYS.days
     else
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -390,8 +390,11 @@ class Comment < ApplicationRecord
     user && user.id == user_id && created_at && created_at < DELETEABLE_DAYS.days.ago
   end
 
-  def is_flaggable?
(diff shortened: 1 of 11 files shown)


rebecaaras:
This should address #1966 and #1973
flag-comments

flag-deleted-comment

flag-stories

app/models/user.rb
```diff
@@ -309,17 +309,17 @@ def banned_from_inviting?
     disabled_invite_at?
   end
 
-  def can_flag?(obj)
+  def can_flag?(obj, user)
```
pushcx: Why does this need a `user` argument rather than use `self`? From a skim, it looks like you're always calling it on and with `@user`? If I'm missing something, can you rename this variable for the role this variable is playing?

app/models/comment.rb
```diff
@@ -390,7 +387,13 @@ def is_disownable_by_user?(user)
     user && user.id == user_id && created_at && created_at < DELETEABLE_DAYS.days.ago
   end
 
-  def is_flaggable?
+  def is_flaggable?(user)
+    if is_gone?
+      return false
+    end
+    if user && user.id == user_id
```
```suggestion
    if user&.id == user_id
```

The code in the erb templates you reused is older than the nil propagation operator. :)

app/models/user.rb
rebecaaras: No, you're right. I wasn't aware that the currently logged in user could be captured through `self` in `User`. Newbie mistake. Fixed now.

app/models/comment.rb
rebecaaras: Thanks for this lovely syntax suggestion.

pushcx:
This is a clean implementation of these fixes, thanks for working through my comments to fix our bugs/design oddities!

PR #1993, opened 2026-03-30
Fix #1990 add to docs how to use default password for all users locally

rebecaaras:
Addresses #1990


pushcx:
Thanks for reporting this - I think it'd be better to just fix this annoyance.

Could you have `fake_data` set `password` directly to `"fake"`? That's even easier to explain. :D

Gemfile.lock
```diff
@@ -511,7 +511,7 @@ DEPENDENCIES
   puma
   rack-attack
   rack-mini-profiler
-  rails (> 8.0, < 8.1)
```
rebecaaras: [CI failed](https://github.com/lobsters/lobsters/actions/runs/24471358067/job/71512132479), and the error message suggested to update `Gemfile.lock`. I did it to check if CI would then pass, and it did, but as I my changes do not affect the `Gemfile` and this is obviously unrelated to this PR, I believe this commit should be deleted or moved somewhere else.

I'll be happy to make any solicited changes.

pushcx: This is me, I broke the build. Sorry for the distraction! It's fixed now, if you want to rebase onto `main` or merge in `main`. Do whatever's convenient for you, this is such a small change that I plan to squash + rebase.

rebecaaras: Obviously it was worth for the [commit message](https://github.com/lobsters/lobsters/commit/4801eb2349f6dcc5076258b8a2862964cb65c5e0) :D

pushcx:
Thanks, this a great quality-of-life improvement for contributors.

PR #2000, opened 2026-04-15
Automatic release tag for GitHub or Codeberg releases

hugoarnal:
I've recently been seeing more and more github.com/<...>/<...>/releases links not getting tagged with the release tag.

I'm proposing this change to bring attention to users to use the release tag.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -773,6 +773,7 @@ document.addEventListener("DOMContentLoaded", () => {
       "\.pdf($|\\?|#)": "pdf",
       "[\/\.](asciinema\.org|(youtube|vimeo)\.com|youtu\.be|twitch\.tv|media\.ccc\.de)\/": "video",
       "\.(mp4|avi|mkv|webm)($|\\?|#)": "video",
+      "(https:\/\/(github\.com|codeberg\.org)\/[^\/]+\/[^\/]+\/releases)": "release",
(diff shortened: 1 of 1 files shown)


app/javascript/application.js
```diff
@@ -773,6 +773,7 @@ document.addEventListener("DOMContentLoaded", () => {
       "\.pdf($|\\?|#)": "pdf",
       "[\/\.](asciinema\.org|(youtube|vimeo)\.com|youtu\.be|twitch\.tv|media\.ccc\.de)\/": "video",
       "\.(mp4|avi|mkv|webm)($|\\?|#)": "video",
+      "(https:\/\/(github\.com|codeberg\.org)\/(.*)\/(.*)\/releases)": "release",
```
```suggestion
      "(https:\/\/(github\.com|codeberg\.org)\/[^\/]+\/[^\/]+\/releases)": "release",
```

I have previously been bitten by using `.` instead of not-separator.

And it doesn't need the parens to capture, the domain only does to scope the `|`.

pushcx:
Thanks, I love this kind of polish.

Sorry for the failing build, I fixed it. I don't need to see the build pass for this one because a js change can't fail the build.

app/javascript/application.js
hugoarnal: Works well, fixed!

pushcx:
Great, thanks for polishing this.

PR #2004, opened 2026-04-26
fix comment counters on inline replies (#1977)

JadinAndrews:
When posting a comment the comment counters were not updated until the page was refreshed. Added changes to postedreply.html.erb partial to pass the short-id and comment count, and updated application.js to use these attributes to update the comment counters.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -344,7 +344,41 @@ export class _LobstersFunction {
           } else {
             comments.insertAdjacentHTML("afterbegin", text)
           }
+          // Update comment counters
+          const newCommentEl = qS(comments, '[data-story-short-id]')
 
+          if (newCommentEl) {
+            const storyShortId = newCommentEl.getAttribute('data-story-short-id')
+            const newCount = newCommentEl.getAttribute('data-new-comment-count')
+            const countLabel = parseInt(newCount) === 1 ? 'comment' : 'comments'
+            
+            // Update byline counter
+            const commentsLabels = qSA('.comments_label')
+            for (const label of commentsLabels) {
+              if (parseInt(newCount) === 0) {
+                label.textContent = 'no comments'
+              } else {
+                label.textContent = newCount + ' ' + countLabel
+              }
+            }
+            
+            // Update thread summary counter
+            const threadSummaries = qSA('.thread_summary')
+            for (const summary of threadSummaries) {
+              const text = summary.textContent
+              if (text.includes('comment')) {
+                if (parseInt(newCount) === 0) {
+                  summary.textContent = 'no comments,' + text.substring(text.indexOf(','))
+                } else {
+                  summary.textContent = newCount + ' ' + countLabel + ',' + text.substring(text.indexOf(','))
+                }
+              }
+            }
+            
+            // Remove the data attributes after processing
+            newCommentEl.removeAttribute('data-story-short-id')
+            newCommentEl.removeAttribute('data-new-comment-count')
+          }
         })
       })
   }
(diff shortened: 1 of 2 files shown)


app/javascript/application.js
```diff
@@ -344,7 +344,41 @@ export class _LobstersFunction {
           } else {
             comments.insertAdjacentHTML("afterbegin", text)
           }
-
+          // Update comment counters
+          const newCommentEl = qS(comments, '[data-story-short-id]')
+
+          if (newCommentEl) {
+            const storyShortId = newCommentEl.getAttribute('data-story-short-id')
+            const newCount = newCommentEl.getAttribute('data-new-comment-count')
```
fedemp, line 352: You could `parseInt` the count here and avoid repeating it at 358.

app/javascript/application.js
```diff
@@ -344,7 +344,41 @@ export class _LobstersFunction {
...
+            for (const label of commentsLabels) {
+              if (parseInt(newCount) === 0) {
+                label.textContent = 'no comments'
+              } else {
+                label.textContent = newCount + ' ' + countLabel
+              }
+            }
+            
+            // Update thread summary counter
+            const threadSummaries = qSA('.thread_summary')
```
fedemp, line 366: Just asking cause I don't know, does each one of the merged stories get its own unread counter? I'm trying to understand why you chose `qSA` instead of `qS`.

app/javascript/application.js
```diff
@@ -344,7 +344,41 @@ export class _LobstersFunction {
...
+            for (const summary of threadSummaries) {
+              const text = summary.textContent
+              if (text.includes('comment')) {
+                if (parseInt(newCount) === 0) {
+                  summary.textContent = 'no comments,' + text.substring(text.indexOf(','))
+                } else {
+                  summary.textContent = newCount + ' ' + countLabel + ',' + text.substring(text.indexOf(','))
+                }
+              }
+            }
```
fedemp, line 376: IMO you should add a `class` in the HTML to find the right container element you want to update instead of relying on a regex to modify the whole sentence.

app/views/comments/_postedreply.html.erb
```diff
@@ -1,5 +1,6 @@
-<%# locals: (comment:, show_tree_lines: true) -%>
-<li class="comments_subtree">
+<%# locals: (comment:, story: nil, show_tree_lines: true) -%>
+<% story ||= comment.story %>
+<li class="comments_subtree" data-story-short-id="<%= story.short_id %>" data-new-comment-count="<%= story.comments_count %>">
```
fedemp, line 3: A very personal comment. Using `data` is surely clever, but given how `create` already returns a JSON response for errors, wouldn't it be more clean to return a JSON with the comment as rendered by the server, and the comment in another key?

app/javascript/application.js
```diff
@@ -344,7 +344,41 @@ export class _LobstersFunction {
...
+            
+            // Update byline counter
+            const commentsLabels = qSA('.comments_label')
+            for (const label of commentsLabels) {
+              if (parseInt(newCount) === 0) {
+                label.textContent = 'no comments'
+              } else {
+                label.textContent = newCount + ' ' + countLabel
+              }
+            }
```
fedemp, line 363: You can use 

```js
const labels = {one: "comment", other: "comments"}
const pluralize = new Intl.PluralRules("en-US")
const content = labels[pluralize.select(newCount)]
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules#examples

pushcx:
This doesn't solve a practical problem and is too brittle to maintain. This is not an SPA and we don't want to start down that road.

Issue #1886, opened 2026-01-30
merged stories UI shows two comment boxes


jcs:
<img width="2064" height="1598" alt="Image" src="https://github.com/user-attachments/assets/45e7464d-b8cc-4229-9e0f-90bd63446007" />

pushcx:
Yep, it's part of #1456. There was a lot of confusion about which link a comment was replying to on merged stories.

PR #2021, opened 2026-05-20
clean up merged story view

jcs:
This UI is a total shit show but this at least removes the confusing multiple text boxes for input.

The first commit removes the duplicate text boxes.

The second commit is probably optional but the comment counts were wrong since merging a story adds the child story's `comments_count` to the parent, so there should probably just be one 'n comments' at the top and hide all the others.  This also removes the duplicate story headings which clutter up the UI and flattens all of the story threads.  If that confuses users as to which story the comments were made on, that's probably a good indication the stories shouldn't have been merged in the first place because they're not similar enough, but whatever.

before:

before

after:

after

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -296,7 +296,8 @@ class Story < ApplicationRecord
   def accepting_comments?
     !is_gone? &&
       !previewing &&
-      (new_record? || created_at.after?(COMMENTABLE_DAYS.days.ago))
+      (new_record? || created_at.after?(COMMENTABLE_DAYS.days.ago)) &&
+      !merged_story_id
   end
 
   def already_posted_recently?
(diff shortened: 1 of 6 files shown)


pushcx:
Thanks for taking a swing at this, but I don't want to revert to how story merging looked a few years ago. As mentioned on your [last issue on merging](https://github.com/lobsters/lobsters/issues/1886#issuecomment-3844393482), there's a bunch of in-progress work that'd be a good place for discussion and contributions.

jcs:
You mean that other issue I filed 4 months ago about this problem that is still not fixed by #1456 that you started over a year ago and is still not halfway completed?

pushcx:
Yep. We have 197 open issues, including some good ideas dating to 2014. If you'd like one of those done sooner, it would be more effective to contribute to those than open PRs to delete things you privately disagree with.

PR #2020, opened 2026-05-19
Update tag definitions, coalescing recent meta discussions.

hmadison:
Following on from the various discussions in IRC and on the site.

PR #2022, opened 2026-05-20
minor typo fixes in updated flag definitions

u1f408:
I noticed a couple typos in the updated flag definitions from #2020, so here's a patch. second commit is just removing some whitespace that my editor was complaining about while making the typo fixes. 

very much appreciate the work y'all continue to put in behind the scenes :)


pushcx:
Thanks for the polish, this is the best kind of bug report to get. :)

PR #2023, opened 2026-05-21
Update phrasing for story submissions and site guidelines

oridb:
Lobsters appears to have decided on an LLM stance, based on the tag changes, but has not made the change particularly visible to users. This is a first draft of the phrasing I would propose.


runxiyu (approved):

app/views/about/about.html.erb
```diff
@@ -67,6 +67,13 @@
     As a rule of thumb, self-promo should be less than a quarter of one's stories and comments.
   </p>
 
+  <p id="llms">
+    <strong>GenAI</strong>:
+    Lobsters is for humans. The stories you submit should be written by people, and should not contain
+    significant amounts of LLM-generated content. There are a few exceptions, such as urgent news about
+    new vulnerabilities, where a human written description is not available.
```
GrayHatter, line 74: > There are a few exceptions, such as urgent news about new vulnerabilities, where a human written description is not available.

Why is it better for lobsters to sacrifice accuracy to prioritize latency?

GrayHatter (changes requested):

pushcx:
Thanks for starting this.

To link some context, here's the [recent meta thread](https://lobste.rs/s/29pm2f/llm_generated_submissions_should_be#c_cnmlxv) this is building off.

Similarly, the recent thread on [the social contract of writing](https://lobste.rs/s/kgem4b/social_contract_writing) has a lot of good food for thought or policies we could adapt.

I've tried to avoid describing the site as "news" because there's a lot of complexity that comes in with that, and I wrote [part of the topicality guideline](https://lobste.rs/about#:~:text=Will%20it%20be%20more%20interesting%20in%20five%20or%20ten%20years%3F) to deliberately move away from breaking news. (And as a programmer, I'm required to nitpick definitions; I see Lobsters a [journalism but not news](https://lobste.rs/search?q=commenter%3Apushcx+journalism&what=comments&order=newest#:~:text=my%20views%20on%20speech%20are%20heavily%20informed%20by%20my%20experience%20as%20a%20journalist%20at%20the%20Washington%20Post%3A%20I%20see%20running%20this%20forum%20as%20a%20form%20of%20trade%20journalism)).

Probably we should have a meta thread to collaborate on writing a new guideline. Let me try to volunteer c355e3b or Irene to start one. We had some internal mod discussion that we can reuse, mostly about where we can articulate a line that's clear enough to submitters and commenters. So your draft says "significant amounts of LLM-generated content" and "significant" could get its own thousand-word definition. We need to write a policy that's clear to people who disagree on whether LLMs should even be used at all; promoters would lean towards a definition like "one-shot with no human polish" and detractors would define more like "an LLM had any use by any contributor, even asking questions". Even if people don't agree on what the tool use should be, they should be able to agree on where Lobsters is drawing the line.

app/views/about/about.html.erb
oridb: My thinking was that for certain bits of news, a user would want to install patches as soon as possible. I'm not wedded to this.

oridb:
Giving this a gentle nudge.

hmadison:
Hey @oridb,

Thanks again for taking the time to start this. To @pushcx's point, we may want a meta thread and I'll need to find a free weekend to moderate that thread. While I try to get that into my schedule, I have a few general comments around thiese kinds of changes:

1. I think it's critical that whatever guideline(s) we put in place are "testable". That is to say, whenever we need to take a moderation action under the new guidelines, the guidelines provide some kind of mechanical test that myself or another moderator can use to validate some piece content violates the guidelien. In this case, being overly specific is helpful. 
2. These new guideline(s) are going to be seen by new users, or users returning to the site after some amount of time. It's important to make sure that we include context as to _why_ we got here in addition to what we are trying to achieve for the content on the site. For many users, this page will be their first, and possibly only, interaction with any form of site policy and as shuch, should try to give the maximal amount of context to help steer their behavior.
3. It worth using this space to also call attention to the "self" and "group" moderation mechanisms we have on the site. Explaining (again) where and how to use flags and story suggestions will go a long way towards having folks positivly engage with those mechanisms.

oridb:
giving a monthly ping.

PR #2026, opened 2026-05-25
Serve JPG avatars with the right mime type

### LINKED ISSUE #2009: Wrong Content-Type in JPEG images
Some avatar images are not loading in Dillo:

Image<br>

This is caused because the `Content-Type` HTTP header for the images are always set to `image/png`, despite being JPEG:

```
% curl -sI https://lobste.rs/avatars/ohrv-16.png | grep -i content-type
content-type: image/png
% curl -s https://lobste.rs/avatars/ohrv-16.png | file -
/dev/stdin: JPEG image data, JFIF standard 1.01, resolution (DPI), density 96x96, segment length 16, comment: "CREATOR: gd-jpeg v1.0 (using IJG JPEG v62), quality = 90", baseline, precision 8, 16x16, components 3
```

Dillo uses the `Content-Type` to select which image decoder to use. As it doesn't recognize the JPEG format as a PNG, it complains and ignores the image:

> ** WARNING **: "https://lobste.rs/avatars/ohrv-16.png" is not a PNG file.

Could you change this behavior to match the `Content-Type` to the image format?

fedemp:
Even though all avatars are saved with the `.png` extension, check its magic bytes to see if it's actually a JPG image. If so, serve with the right mime type.

Fixes #2009

### DIFF
--- a/app/controllers/avatars_controller.rb
+++ b/app/controllers/avatars_controller.rb
@@ -24,6 +24,10 @@ class AvatarsController < ApplicationController
     redirect_to "/settings"
   end
 
+  def content_type_for(data)
+    data.start_with?("\xFF\xD8\xFF".b) ? "image/jpeg" : "image/png"
+  end
+
   def show
     username, size = params[:username_size].to_s.scan(/\A(.+)-(\d+)\z/).first
     size = size.to_i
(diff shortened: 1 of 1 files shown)


pushcx:
Thanks. It's an odd circumstance, but this is a well-contained fix. I'll leave a longer note in the issue; reviewing this turned up some prod oddities.

PR #2028, opened 2026-05-26
Add badge next to comment author is user is submitter/author

### LINKED ISSUE #1997: Add hover captions to usernames
In the interest of accessibility and improved intuitive usability, add hover captions to usernames. Optionally, include whether the user is a regular commenter, the submitter, or the author.

This came about when I asked the following question in the Lobsters IRC channel:

> \<Chronos\> I'm reading the comments on this page - https://lobste.rs/s/gns27z/what_are_your_programming_hunches_you - and I noticed that, while using the dark theme, most usernames are white, but some are medium blue. What does that distinction mean?

Replies were:

> \<jbox\> Chronos: I believe the blue username indicates the thread author
> \<slowstatic_76\> Chronos: how many usernames are blue? it should just be the submitter

Later on, user pushcx said:

> \<pushcx\> Chronos: Could you file an issue about this? We should add a hover caption. We recently added an aria label; it's an oversight not to have something for sighted users.

This issue is the result of that comment.

fedemp:
Add a small "[OP]" next to user name in those comment where the user is the one who submitted the original story.

Fixes #1997

image

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -542,6 +542,12 @@ ol.stories li.story div.story_liner {
 	color: var(--color-fg-contrast-7-5);
 }
 
+.comment .user_is_submitter::after {
+	content: " [OP]"/"";
+	font-size: var(--font-size-tags);
+	vertical-align: top;
+}
+
 ol.stories li:first-child div.story_liner {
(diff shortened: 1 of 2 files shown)


pushcx:
Thanks for adding this.

PR #2030, opened 2026-05-27
Remove invisible characters from entry titles

fedemp:
Fixes #1984


fedemp:
For the record, I'm not exactly proud of this one. I would prefer the regexp to be a constant that it's not parsed everytime the method is called.

And if the regexp could be globally shared, we could do the same for comments.

BTW, this was sourced both from https://regex101.com/r/42jDca/1 and https://invisible-characters.com/

pushcx:
I see the overall strategy here, this looks great. If you move the regexp up to a class constant it'll only be evaluated once, see `GRAPHICS_RE` in app/models/story.rb.

I really appreciate that you commented the invidual ranges, that'll help with maintenance. Please also add a comment with the URL(s) where you got the ranges.

(Though this PR is also affect by me breaking main, some of the build failure looks relevant to this branch. You can merge in main or rebase, whatever's convenient. I'll rebase + squash a small PR like this.)

fedemp:
Thanks for the tip on the constant. I've converted the regexp to one, and also fixed my syntax errors.

pushcx:
I know it is not exciting, but this is my favorite kind of code. It's clear, it's maintainable, and it's a great bit of polish for the users. Thank you.

PR #2035, opened 2026-05-28
Simplify title length validation reuse

PuercoPop:
Store the lower and upper bounds of the length of the title in constants. Then
use those constants when defining the title length validator in the model; and
the title length validator in the HTML view.

Using a constant as the source truth means we can avoid using reflection to
extract the information from the validator we defined.

I understand that what simpler code is a matter of taste and hence subjective,
so feel free to close the pull request if you prefer the current implementation

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -205,7 +205,10 @@ class Story < ApplicationRecord
 
   include Token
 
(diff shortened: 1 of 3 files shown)


pushcx:
Thanks, this seems worth merging. I have done a lot of reflection on AR validations lately and it is all very clunky.

PR #2038, opened 2026-05-30
Native validation for username, plus aria-describe in settings

fedemp:
For the username field in the settings page, add html native validation via the same pattern used in the backend.

For every field in the same page that has an associated hint, add a `aria-describedby` attribute so screen reader will read it when focusing the field.

Screenshot From 2026-05-30 11-25-33


pushcx:
Thanks for extending this pattern, it's great to give immediate feedback.

All your frontend fixes have me reorienting my design notes for a web framework from the rails-style horizontal separation of concerns by technology into one that's vertically separated by component, so this kind of polish is only separated by a couple of lines rather than happening at opposite ends of a stack.

PR #2044, opened 2026-05-31
Native validation for category and tags edit/create forms

fedemp:
For the name and description fields in the tag and category edition page, add a browser validation via the same pattern used in the backend.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -14,9 +14,12 @@ class Category < ApplicationRecord
 
   attr_accessor :edit_user_id
 
-  validates :category, length: {maximum: 25}, presence: true,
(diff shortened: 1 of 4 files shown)


pushcx:
This is real nitpicky, but could you rename `NAME_FORMAT` to `CATEGORY_FORMAT` etc so that the variable's name matches the name of the attribute?

This is a little complicated by the fact that `category.name` would have been a better name than `category.category`. I was extending the pattern from tag, but `name` would be better there, too. That rename would be a real slog, though.

fedemp:
I've changed both tag and category to TAG_FORMAT and CATEGORY_FORMAT. :+1:

pushcx:
Great, thanks for improving this UI.

PR #2047, opened 2026-05-31
JS: Fix first top comment not rendered after submit

fedemp:


pushcx:
Thanks for catching this.

PR #2048, opened 2026-06-01
Fix broken comment replies on search results

### LINKED ISSUE #2027: Can't reply comments in search page
The 'reply' button is there, but clicking on it throws a JS error. e.g. https://lobste.rs/search?q=javascript&what=comments&order=newest

```
Uncaught TypeError: can't access property "prepend", children is null
    <anonymous> https://lobste.rs/assets/application-252f2493.js:875
    on https://lobste.rs/assets/application-252f2493.js:24
6 application-252f2493.js:875:5
```

Zurtar:
Fixes #2027.

The search page was missing the comment container that holds the temporary reply form.

### DIFF
--- a/app/views/users/standing.html.erb
+++ b/app/views/users/standing.html.erb
@@ -65,6 +65,8 @@
 
 <ol class="comments comments1">
   <% @flagged_comments.each do |comment| %>
-    <li><%= render "comments/comment", :comment => comment, :show_story => true, force_open: true %></li>
+    <li><%= render "comments/comment", :comment => comment, :show_story => true, force_open: true %>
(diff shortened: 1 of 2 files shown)


fedemp:
Works on my machine. :100:

app/views/users/standing.html.erb
fedemp, line 1: What is the url for this view? Cause it map over `@flagged_comments` and user should not be able to reply to flagged comments, AFAIK.

Edit: My bad. I see that this only happens in https://lobste.rs/~fedemp/standing when logged with my own user. I should be able to reply my own comment.

pushcx:
Thanks for solving this!

Issue #2042, opened 2026-05-30
Bug: flash:success not shown after filtering tags for non logged in users


fedemp:
As a logged in user, after submitting the form to filter out tags, the page will show the typical green banner reading "Your filters have been updated.".

As an non logged in user, if I do the same, I do not get any success message even when my filters were saved.

pushcx:
This is an odd one that's worth fixing, thanks for reporting it.

By default Rails sets a session cookie (ours is named `lobster_trap`). The session includes custom info like logged-in user, but also some default data like the CSRF token and the `flash` messages. This bug exists because we overlooked that the flash messages are in the session.

Thanks for opening #2058 but it's working by coincidence. We keep the visitor tag filters in a separate cookie, not in the session, so it doesn't make sense to check the session for the name of the tag filter cookie. The right fix is going to be to adding the flash to `SESSION_DEFAULT_KEYS` so we don't clear the session when there's a flash message for the visitor. I'd guess it's got the obvious name `flash` but that's just a guess.

fedemp:
Your suggestion did not work. 😕 
Since flash messages are stored in the session cookie, I assumed that not deleting it would make flash work. I have updated the PR.

pushcx:
Your PR looks good, but I don't understand that CI failure, it seems totally unrelated.

pushcx:
I guess it was transient. I pulled up your branch locally and it passed, I re-ran the build on your PR and it passed. I'm happy to whistle pass the graveyard and ignore the transient failure.

PR #2053, opened 2026-06-02
Fix title shortword search handling

### LINKED ISSUE #1933: Parser gives confusing response for operator:shortword search
I searched for `title:xz` trying to find the story about the xz backdoor. The mariadb search engine doesn't index two-character words, which `app/models/search_parser.rb` refers to as `shortword`s (not `stopword` - similar context and term, so they're easy to confuse).

But this query is parsed incorrectly:

Image

The `:xz` is confusing here. Only "xz" is a shortword, and the `:` should be associated with `title`. But it's not well-defined to search for a title with an empty string. I'm not sure off the top of my head what the correct parse of this should be, I just know the current one is misleading to the user.

This search should be added to `spec/models/search_parser_spec.rb` in the "bugs I've seen in prod" `describe` block.

lonelyhty:
Fixes #1933.

This keeps `title:` queries with short words associated with the title operator instead of falling through into the confusing `:shortword` parse case. When a bare title term is too short for full-text search, the search now treats it as an unrecognized title search instead of issuing a title full-text query, and the parse explanation calls out the minimum title length.

Changes:
- parse `title:xz` as a title shortword
- avoid issuing title full-text searches for bare title shortwords
- show a clearer parse explanation for title shortwords
- add parser and search coverage for the short title case

Validation:
- `git diff --check`
- `docker compose run --rm --no-deps app bundle exec standardrb app/models/search.rb app/models/search_parser.rb spec/models/search_parser_spec.rb spec/models/search_spec.rb`
- `docker compose run --rm app bundle exec rspec spec/models/search_parser_spec.rb spec/models/search_spec.rb` (`109 examples, 0 failures`)

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -91,9 +91,15 @@ class Search
   def flatten_title tree
     if tree.keys.first == :term
       ActiveRecord::Base.connection.quote_string(tree.values.first.to_s)
+    elsif tree.keys.first == :shortword
+      ActiveRecord::Base.connection.quote_string(tree.values.first.to_s)
     elsif tree.keys.first == :quoted
       '"' + tree.values.first.map(&:values).flatten.join(" ").gsub("'", "\\\\'") + '"'
     end
+  end
+
+  def title_shortword? tree
+    tree.keys.first == :shortword
   end
 
   # security: must prevent sql injection
@@ -148,6 +154,7 @@ class Search
         tags.or!(Tag.where(tag: value.to_s))
       # TODO unknown tag
       when :title
+        next if title_shortword? value
         title = true
         value = flatten_title value
         query.where!(
@@ -257,6 +264,7 @@ class Search
         tags.or!(Tag.where(tag: value.to_s))
       # TODO unknown tag
       when :title
+        next if title_shortword? value
         title = true
         value = flatten_title value
         query.joins!(:story_text).where!(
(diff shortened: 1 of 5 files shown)


app/models/search.rb
```diff
@@ -91,11 +91,17 @@ def perform!
   def flatten_title tree
     if tree.keys.first == :term
       ActiveRecord::Base.connection.quote_string(tree.values.first.to_s)
+    elsif tree.keys.first == :shortword
```
pushcx, line 94: This could use the `title_shortword?` helper you define a few lines down.

app/models/search_parser.rb
```diff
@@ -33,7 +33,7 @@ class SearchParser < Parslet::Parser
   rule(:submitter) { str("submitter:") >> match("[@~]").repeat(0, 1) >> match("[A-Za-z0-9_\\-]").repeat(1, 24).as(:submitter) >> space? }
   # reproduce the 'validates :tag, format:' regexp from Tag
   rule(:tag) { str("tag:") >> match("[A-Za-z0-9\\-_+]").repeat(1).as(:tag) >> space? }
-  rule(:title) { str("title:") >> (term | quoted).as(:title) >> space? }
+  rule(:title) { str("title:") >> (term | shortword | quoted).as(:title) >> space? }
```
pushcx, line 36: Aha, no wonder it didn't parse correctly.

Maybe this suggests we should have an abstraction over `term | shortword |quoted` but I'd prefer to wait until we see the pattern more.

pushcx:
Thanks for taking this one on, it looks like you've solved the bug.

I don't recognize this build failure, maybe the some test data is inappropriately short? I _did_ touch this `update_score` function recently, maybe this will just vanish if you rebase or merge in `main` (do whatever is convenient for you, I squash + rebase small PR fixes like this).

pushcx:
I saw the same odd build failure over on #2042. I reran this build and it's green now, so that's definitely not your fault.

I'd like you to make the `title_shortword?` reuse I mentioned before I merge, sound good?

pushcx:
I'm doing another pass through the open PRs and my tweak is so minor it's not worth holding this up. I'd rather merge this fix than let it hang out and go stale.

Thanks for fixing this bug.

PR #2055, opened 2026-06-05
Fixes missing underline for Top page in navbar

### LINKED ISSUE #2049: "Top" link in navbar does not have red underline when active
When visiting http://localhost:3000/top/1w, the link "Top" should have a red underline like every other link in the nav.

Zurtar:
Fixes #2049. 

The link_to_different_page function compares the navigation link with the current path and optionally adds the red-underline style. On the Top page the current path defaults to /top/1w, while the nav link is /top/. Fixed by adjusting the regex to optionally account for the time suffix.


pushcx:
Thanks for the fix. This is a really clear extension, and I appreciate that you got all the functionality under test.

PR #2057, opened 2026-06-06
Dont show "Most often tagged" if not enough related tags

### LINKED ISSUE #2043: "Most often also tagged with" may show even if empty
Extremely improbable in prod right now, but on tag individual page, the text "Most often also tagged with" may appear even if there are no related tags.

Image

fedemp:
Fixes #2043


pushcx:
Nice little fix for new tags and the sister sites, thanks.

PR #2058, opened 2026-06-06
Fix flash messages not shown to non logged in users

### LINKED ISSUE #2042: Bug: flash:success not shown after filtering tags for non logged in users

fedemp:
Fixes #2042


fedemp:
To be honest, I have no idea what I did. I just noticed that the `lobsters_trap` cookie was reset after submitting the form. A grep took me to this file and then I just tried different combinations of the logical operator to see what stuck to the wall.

pushcx:
Thanks for puzzling this out!

PR #2059, opened 2026-06-06
Fix tags select cropped if JS is disabled

### LINKED ISSUE #2052: "Tags" field shows cropped text if JS is disabled
For the **ai** tag, I can't see the full description.

Image

fedemp:
Fixes #2052


fedemp:
The scrollbar is now there, even if too thin. :confused: 

image

pushcx:
Thanks, looks like this is the limits of where we can get with CSS without the big job of creating a custom component.

PR #2065, opened 2026-06-06
Remove alt text for avatars

fedemp:
* Screen reader users can't read it because it's hidden by `aria-hidden`
* Graphical browsers won't show it unless user disables images. And in this case, users get a horribly long text saying "random-user avatar"
* w3m shows the alt text, but it's redundant when the link to the user profile is the next link.
* eww ignores links if they only contain an image. Does not even show alt text!

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -11,7 +11,8 @@ module ApplicationHelper
       srcset: "#{user.avatar_path(size)} 1x, #{user.avatar_path(size * 2)} 2x",
       class: "avatar",
       size: "#{size}x#{size}",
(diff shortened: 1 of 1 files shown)


pushcx:
Could you add a comment that the `alt` text is unused because the element is `aria-hidden`? Otherwise in a year or two I'll be like "oh weird oversight that alt text is missing, I'll fix that!" and I'll put it back.

pushcx:
I'm making a pass through the PRs and wanted to ping on this one, it just needs a short comment to prevent regression.

fedemp:
Done. :+1:

pushcx:
Thanks for protecting this fix from me.

PR #2073, opened 2026-06-13
make comment upvoting work without js

adamdruppe:
applying the technique explained in that comment thread last night, as you can see it is not a lot of code.

tbh i don't remember editing that _thread file but git insists i did. oh well, it is a copy anyway so keeping it in sync.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -519,6 +519,7 @@ div.voters {
 	align-items: center;
 	justify-self: center;
 	text-decoration: none;
+	border: none;
 }
 .upvoter:focus { outline: none; } /* don't outline after upvote click */
 .upvoter:focus-visible { outline: 2px solid var(--color-fg-link); } /* do outline when tabbing */
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -239,7 +239,11 @@ class CommentsController < ApplicationController
       0, comment.story_id, comment.id, @user.id, nil
     )
 
-    render plain: "ok"
+    if params[:nojs]
+      redirect_to(Routes.comment_target_path(comment, true))
+    else
+      render plain: "ok"
+    end
   end
 
   def upvote
@@ -251,7 +255,11 @@ class CommentsController < ApplicationController
       1, comment.story_id, comment.id, @user.id, nil
     )
 
-    render plain: "ok"
+    if params[:nojs]
+      redirect_to(Routes.comment_target_path(comment, true))
+    else
+      render plain: "ok"
+    end
   end
 
   def flag
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -239,10 +239,10 @@ class CommentsController < ApplicationController
       0, comment.story_id, comment.id, @user.id, nil
     )
 
-    if params[:nojs]
-      redirect_to(Routes.comment_target_path(comment, true))
-    else
+    if request.xhr?
       render plain: "ok"
+    else
+      redirect_to(Routes.comment_target_path(comment, true))
     end
   end
 
@@ -255,10 +255,10 @@ class CommentsController < ApplicationController
       1, comment.story_id, comment.id, @user.id, nil
     )
 
-    if params[:nojs]
-      redirect_to(Routes.comment_target_path(comment, true))
-    else
+    if request.xhr?
       render plain: "ok"
+    else
+      redirect_to(Routes.comment_target_path(comment, true))
     end
   end
 
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -30,7 +30,14 @@
     <% end %>
     <% score = comment_score_for_user(comment, @user)%>
     <% if !comment.is_gone? and !comment.previewing %>
-      <%= link_to upvoter_score(score), (@user ? '': login_path), :class => "upvoter", :title => score %>
+      <% if @user %>
+        <%= form_tag("/comments/#{comment.short_id}/#{comment.current_upvoted? ? "unvote" : "upvote"}") do %>
+          <input type="hidden" name="nojs", value="true" />
+          <button type="submit" class="upvoter" title="<%= score %>"><%= score %></button>
+        <% end %>
+      <% else %>
+        <%= link_to upvoter_score(score), login_path, :class => "upvoter", :title => score %>
+      <% end %>
     <% elsif @user&.is_moderator? %>
       <span class="upvoter"><%= score %></span>
     <% end %>
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -32,8 +32,7 @@
     <% if !comment.is_gone? and !comment.previewing %>
       <% if @user %>
         <%= form_tag("/comments/#{comment.short_id}/#{comment.current_upvoted? ? "unvote" : "upvote"}") do %>
-          <input type="hidden" name="nojs", value="true" />
-          <button type="submit" class="upvoter" title="<%= score %>"><%= score %></button>
+          <button type="submit" class="upvoter" title="<%= upvoter_score(score) %>"><%= upvoter_score(score) %></button>
         <% end %>
       <% else %>
         <%= link_to upvoter_score(score), login_path, :class => "upvoter", :title => score %>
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/application.css
```diff
@@ -519,6 +519,7 @@ div.voters {
 	align-items: center;
 	justify-self: center;
 	text-decoration: none;
+	border: none;
```
pushcx, line 522: @fedemp You know accessibility better than me - this won't break keyboard navigation, right?

app/controllers/comments_controller.rb
```diff
@@ -239,7 +239,11 @@ def unvote
       0, comment.story_id, comment.id, @user.id, nil
     )
 
-    render plain: "ok"
+    if params[:nojs]
```
pushcx: The idiomatic way for this is `request.xhr?` (even though it's no longer xhr in the frontend js).

app/assets/stylesheets/application.css
adamdruppe: focus is normally indicated with an outline, not a border. This is what it looks like now when you tab to it:

image

app/views/comments/_comment.html.erb
```diff
@@ -30,7 +30,14 @@
     <% end %>
     <% score = comment_score_for_user(comment, @user)%>
     <% if !comment.is_gone? and !comment.previewing %>
-      <%= link_to upvoter_score(score), (@user ? '': login_path), :class => "upvoter", :title => score %>
+      <% if @user %>
```
pushcx, line 33: Why did you split this up? Why not show the form to logged-out visitors and lead into the same login flow?

app/views/comments/_comment.html.erb
```diff
@@ -30,7 +30,14 @@
...
+        <%= form_tag("/comments/#{comment.short_id}/#{comment.current_upvoted? ? "unvote" : "upvote"}") do %>
+          <input type="hidden" name="nojs", value="true" />
+          <button type="submit" class="upvoter" title="<%= score %>"><%= score %></button>
```
```suggestion
          <button type="submit" class="upvoter" title="<%= upvoter_score(score) %>"><%= upvoter_score(score) %></button>
```

app/assets/stylesheets/application.css
adamdruppe: the difference between an outline and a border btw is an outline is always inside the content box, so it will never resize the element.

app/views/comments/_comment.html.erb
adamdruppe: the original code had it split up too (in the ternary then the script) so i just kept that... either way could work though, the controller can be like if not logged in, redirect to login. that would probably be better in case someone logs in in another tab, then the buttons on the open tab work without needing refreshing.... so yeah i can do that.

pushcx:
Real solid start here. I've got a couple small tweaks and questions, but this feels very close to merging.

The duplication isn't you, it's from `heinous_inline_partial`, a grody performance hack of mine. Please do continue to include it in your PR.

app/views/comments/_comment.html.erb
adamdruppe: actually maybe not, trying it triggers this "render plain: "not logged in", status: 400" from controllers/concerns/authenticatable.rb so it doesn't even make it to the controller.

app/views/comments/_comment.html.erb
adamdruppe: ok in the new commit

app/controllers/comments_controller.rb
adamdruppe: yeah i vaguely remembered rails had one but i forgot what it was lol. applied now still seems to test out ok.

app/views/comments/_comment.html.erb
adamdruppe: the new commit does NOT include this change since the login link is nicer to the user than a 400 error.

ps you caught me at a lucky time for a quick reply but ima be offline for a while now so i'll get back to you later if there's more.

app/assets/stylesheets/application.css
fedemp: Out of curiosity, what border are you removing?

adamdruppe: I think it is the default one around the `<button>` element. A thin light grey.

app/views/comments/_comment.html.erb
```diff
@@ -30,7 +30,13 @@
     <% end %>
     <% score = comment_score_for_user(comment, @user)%>
     <% if !comment.is_gone? and !comment.previewing %>
-      <%= link_to upvoter_score(score), (@user ? '': login_path), :class => "upvoter", :title => score %>
+      <% if @user %>
+        <%= form_tag("/comments/#{comment.short_id}/#{comment.current_upvoted? ? "unvote" : "upvote"}") do %>
+          <button type="submit" class="upvoter" title="<%= upvoter_score(score) %>"><%= upvoter_score(score) %></button>
+        <% end %>
+      <% else %>
+        <%= link_to upvoter_score(score), login_path, :class => "upvoter", :title => score %>
```
fedemp, line 38: I can't test on my computer right now, but wouldn't this produce a link where the inner text is the same as the title? If so, please avoid adding the title cause screen readers may read both texts. IMO I would make the title read "Login to upvote" or an aria-label reading "Score N. Login to upvote".

app/views/comments/_threads.html.erb
```diff
@@ -74,7 +74,13 @@
...
     <% score = comment_score_for_user(comment, @user)%>
     <% if !comment.is_gone? and !comment.previewing %>
-      <%= link_to upvoter_score(score), (@user ? '': login_path), :class => "upvoter", :title => score %>
+      <% if @user %>
+        <%= form_tag("/comments/#{comment.short_id}/#{comment.current_upvoted? ? "unvote" : "upvote"}") do %>
+          <button type="submit" class="upvoter" title="<%= upvoter_score(score) %>"><%= upvoter_score(score) %></button>
+        <% end %>
+      <% else %>
+        <%= link_to upvoter_score(score), login_path, :class => "upvoter", :title => score %>
+      <% end %>
```
fedemp, line 83: This is extreme nitpick and Peter should agree with it, but wouldn't this be a good opportunity to move this to a partial and then call it from the different routes?

app/views/comments/_comment.html.erb
adamdruppe: idk i just copied what was already there lol. but what's the difference between upvoter_score and score? Looks like that function is responsible for the ~ thing. maybe the original idea of the title was if the number was truncated so you can see it on hover?

app/views/comments/_threads.html.erb
adamdruppe: i could swear i never edited that file and went back to the log when running the test and make certain i never edited it and:

```
app_1  | W, [2026-06-17T21:20:16.491146 #1]  WARN -- : heinous_inline_partial initialized, found: {"app/views/comments/_threads.html.erb" => "app/views/comments/_comment.html.erb", "app/views/home/index.html.erb" => "app/views/stories/_listdetail.html.erb", "app/views/search/index.html.erb" => "app/views/stories/_listdetail.html.erb"}
```

then git diff shows the other edit too. so it is already some kind of automatic generation process. it is certainly bizarre, i dare say even heinous, but still it is an existing process out of scope for this small change.

app/views/comments/_comment.html.erb
pushcx: Yeah, that's `render_logged_in_user_or_400`, which throws a short 400 back. It was designed for JS usage and is only used on these couple of (formerly!) js-only endpoints where it didn't make sense to return a full HTML page that would be thrown away.

Off the top of my head it can be removed when all of the js is optional, but it's a bigger change than should be included in this one PR.

app/views/comments/_comment.html.erb
pushcx: No, the title was simply my oversight. I added `upvoter_score` to deny the small mean pleasure of watching the score go down when flagging. (Flagging is for sending things to the mods, not disagreeing or punishing.)

app/views/comments/_threads.html.erb
pushcx: This is `heinous_inline_partial` again, it's expected duplication. That initializer has a comment/confession at the top with more info.

pushcx:
Thanks for being such a good sport about me [volunteering you](https://lobste.rs/c/c6sdn5) to update this.

In the failing build - one was on main and harmless; I fixed the other in the merge. Then landed a small commit 6aec11c2 to fix up alignment. (That also convinced me we should remove the link-based presentation entirely, it's a pain to debug two versions of this.)

I'd appreciate if you want to continue making the last of the js optional, but no worries if not.

PR #2075, opened 2026-06-15
Hide URL search results for non-logged in users.

### LINKED ISSUE #2071: Non logged user can see search results count even if search is invalid
While being unlogged I tried using the search form to search for a url. As expected, I got an error message reading that that's not possible if not logged in. Even so, I got to see how many results were found.

Image

Zurtar:
Fixes #2071. 

Two changes:
- Swapped out manual results override with the `invalid` method of `Search`.
- Uses the URL check logic implemented in Search rather than the naive check for "https".


pushcx:
Thanks a real clear fix, thank you.

Issue #2068, opened 2026-06-10
extend css validator for wrapping text in select tags


pushcx:
In `spec/css_widely_available_spec.rb` I hacked up a CSS parser to enforce that the project's CSS meets the widely-available baseline standard. In #2052 I thought it would be nice to wrap the `<option>` text rather than scroll horizontally, but that's not available yet. 

So I took a quick pass at adding it to our wishlist (patch below), but that didn't work out. The check isn't quite expressive enough - `select` has long had the `appearance` attribute, but it's the value `base-select` that's nowhere near widely available.

```
diff --git a/spec/css_widely_available_spec.rb b/spec/css_widely_available_spec.rb
index 07f3523a5d..c8c96454a7 100644
--- a/spec/css_widely_available_spec.rb
+++ b/spec/css_widely_available_spec.rb
@@ -18,8 +18,10 @@
 # Unprincipled exception because it solves so many big problems.
 ALLOWED_SELECTORS = Set.new(%w[has])

+# select { appearance: base-select; } option { white-space: normal; } would let long tag
+# descriptions wrap lines instead of horizontal scrolling #2052
 # textarea { field-sizing: content; resize: vertical; } would let us delete autosize.js
-WANTED_PROPERTIES = Set.new(%w[field-sizing resize])
+WANTED_PROPERTIES = Set.new(%w[field-sizing resize])

 BaselineAtRule = Data.define(:name)
 BaselineProperty = Data.define(:name)
```

And then I tested adding `appearance: base-select;` and the spec didn't catch that this is an unsupported value for the property.

So there's 3 things here:

 1. The validator should catch that `select { appearance: base-select; }` is not yet widely available. I think in this role `base-select` is called an `attribute`? So we'd need to start extracting that from the baseline data. Maybe it's about the combination of selector and attribute.
 2. Add this specific CSS to a `WANTED` list so we know when it's available.
 3. Once it correctly catches that, it will almost certainly turn up that we have accepted CSS that doesn't meet the widely available standard. Ideally we'll be able to rewrite to meet the standard, but maybe we'll decide to add more to the `ALLOWED` lists.

kmooventhan98:
PR https://github.com/lobsters/lobsters/pull/2079 merged 
why still issue is opened?

pushcx:
An oversight, Zurtar's PR implemented everything.

PR #2079, opened 2026-06-16
Extend CSS validation to include property values

Zurtar:
Addresses #2068. 

To-Do

- [X] The validator should catch  that `select { appearance: base-select; }` is not yet widely available.
- [x] Add this specific CSS to a WANTED list so we know when it's available.
- [x] Rewrite Existing Non-Conforming CSS to OR add to ALLOWED.

> It will almost certainly turn up that we have accepted CSS that doesn't meet the widely available standard. Ideally we'll be able to rewrite to meet the standard, but maybe we'll decide to add more to the ALLOWED lists.

As expected there is CSS that doesn't meet the baseline. which is why the test currently fails.
- [cursor: pointer](https://caniuse.com/mdn-css_properties_cursor_pointer)
- [word-break: break-word](https://caniuse.com/wf-word-break-break-word)
- [user-select: none](https://caniuse.com/user-select-none)

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -529,7 +529,7 @@ li.story {
 ol.stories li.story div.story_liner {
 	padding-top: 0.25em;
 	padding-bottom: 0.25em;
-	word-break: break-word;
+	overflow-wrap: break-word;
 }
 .comment {
 	clear: both;
@@ -1387,7 +1387,7 @@ label:has(+ *[type="hidden"] + *:required)::after {
 .labelled_grid {
 	display: grid;
 	grid-template-columns: minmax(10rem, min-content) 1fr;
-	word-break: break-word;
+	overflow-wrap: break-word;
 	column-gap: 1rem;
 	row-gap: 12px;
 }
--- a/spec/css_widely_available_spec.rb
+++ b/spec/css_widely_available_spec.rb
@@ -14,7 +14,8 @@ ALLOWED_AT_RULES = Set.new(%w[])
 # writing comments, that is the beating heart of the site.
 ALLOWED_PROPERTIES = Set.new(%w[user-select])
 
-ALLOWED_PROPERTY_VALUES = Set.new([])
+# See https://github.com/lobsters/lobsters/pull/2079#issuecomment-4730563102 for reasoning.
+ALLOWED_PROPERTY_VALUES = Set.new([["cursor", "pointer"], ["cursor", "default"], ["user-select", "none"]])
 
 # :has() https://caniuse.com/css-has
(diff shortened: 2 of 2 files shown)


Zurtar:
I'm not sure why `cursor: pointer` is flagged. Looking at the JSON it isn't baseline, but it seems like a feature that would be.

```json
    "css.properties.cursor.pointer": {
        "baseline": false,
        "support": {
            "chrome": "1",
            "chrome_android": "18",
            "edge": "12",
            "firefox": "1",
            "firefox_android": "95",
            "safari": "1.2"
        }
    }
```

pushcx:
Great extension to the test, thank you.

`cursor: pointer`: The baseline standard is [defined with a list of browsers](https://web-platform-dx.github.io/baseline/), one of which is iOS Safari. It probably will never support it. (I'd bet the group defining baseline has debated whether to ignore their choice and mark properties as widely available in situations like this.) Please add this to the ALLOWED.

`word-break: break-word` sounds entirely replaced by `overflow-wrap: break-word`, which is widely available, so let's replace our usage.

`user-select: none` I'm puzzled why this isn't considered widely available:

> "user-select":{"caniuse":["user-select-none"],"compat_features":["css.properties.user-select","css.properties.user-select.all","css.properties.user-select.auto","css.properties.user-select.none","css.properties.user-select.text"],"description":"The user-select CSS property controls which elements can be selected by the user.","description_html":"The <code>user-select</code> CSS property controls which elements can be selected by the user.","group":["css"],"kind":"feature","name":"user-select","spec":["https://drafts.csswg.org/css-ui-4/#content-selection"],"status":
  {"baseline":false,"by_compat_key":{"css.properties.user-select":{"baseline":false,"support":{"chrome":"54","chrome_android":"54","edge":"79","firefox":"69","firefox_android":"79"}},"css.properties.user-select.all":{"baseline":false,"support":{"chrome":"54", "chrome_android":"54","edge":"79","firefox":"69","firefox_android":"79"}},"css.properties.user-select.auto":{"baseline":false,"support":{"chrome":"54","chrome_android":"54","edge":"79","firefox":"69","firefox_android":"79"}},"css.properties.user-select.none":{"baseline":false,"support":{"chrome":"54","chrome_android":"54","edge":"79","firefox":"69","firefox_android":"79"}}

I'd say add it to ALLOWED and ask upstream if this might be a bug in the data?

Zurtar:
Thank you for the quick review.

I've made those changes, the current ALLOWED items are:
- `cursor: pointer`
- `cursor: default`
- `user-select: none`

The `word-break: break-word` fix was a quick replacement. Should be ready for review/merge.

pushcx:
This is great, thank you!

PR #2083, opened 2026-06-20
Add es-module-shims to support older browsers

### LINKED ISSUE #2070: Sailfish OS browser (Firefox ESR 91) can't load JS file
This is a draft. Not actionable items yet.

According to a comment by toastal, [the site does not work properly in Sailfish OS browser](https://lobste.rs/s/esvncd/how_building_html_first_site_doubled_our#c_zdp73n). The comment mentions that the browser is based on Firefox ESR91. According to [Sailfish releases](https://docs.sailfishos.org/Support/Releases/#sailfish-5---tampere-region), this is the most up to date browser they will be delivering.

It would be good to do a smoke test to see if there's something seriously broken. OP mentions the inability to upvote, ~but it's to be defined if it's because of JS being disabled~ even when JS is enabled, ~or if there's a compatibility issue~ so there may be a compatibility issue.

----

Confirming that JS won't load in Firefox v91. It can't load the JS file because of the way it's imported; it does not understand the `import "application"` syntax.

----

The app uses https://github.com/rails/importmap-rails to import the JS files, and it mentions a polyfill for browsers that do not support importmaps: https://github.com/guybedford/es-module-shims

IMO it's worth investigating. This version of Firefox, even if old, is not incapable of rendering a relatively modern site.

fedemp:
Fixes #2070, the default browser in Sailfish OS 5 does not support import maps. The readme from importsmap-rails[1] suggests using the es-module-shims[2] for compatibility with older browsers. A non minified copy[3] of the script is added to the public directory. An older version of es-module-shims is used because of Firefox not supporting dynamic imports.[4]

[1] https://github.com/rails/importmap-rails#supporting-legacy-browsers-such-as-safari-on-ios-15
[2] https://github.com/guybedford/es-module-shims
[3] https://npmx.dev/package-code/es-module-shims/v/1.10.1/dist%2Fes-module-shims.js
[4] https://github.com/guybedford/es-module-shims/issues/495

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -30,6 +30,7 @@
   <%= stylesheet_link_tag "TomSelect_remove_button", "data-turbo-track": "reload" %>
 
   <% if @user %>
+    <script async src="/es-module-shims@1.10.1.js" nonce="<%= content_security_policy_nonce %>"></script>
     <%= javascript_importmap_tags %>
   <% end %>
 
--- a/public/es-module-shims@1.10.1.js
+++ b/public/es-module-shims@1.10.1.js
@@ -1,0 +1,1067 @@
+/* ES Module Shims 1.10.1 */
+(function () {
+
+  const hasDocument = typeof document !== 'undefined';
+
+  const noop = () => {};
+
+  const optionsScript = hasDocument ? document.querySelector('script[type=esms-options]') : undefined;
+
+  const esmsInitOptions = optionsScript ? JSON.parse(optionsScript.innerHTML) : {};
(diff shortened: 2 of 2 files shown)


fedemp:
I do not expect this to be merged since we are not that far from a good experience with JS disabled; this is more about showing interest in this issue being fixed. I'm not eager to make the vast majority of the users to pay the extra toll of this script, but at the same time I feel (slightly) bad for the the Sailfish phone, being a fairly modern product and it getting a degraded experience.

pushcx:
Yeah, you predicted my response. I'm not willing to include this big a compatibility shim for a five-year old browser.

It also doesn't help my sympathy that it's to fix a migration the js ecosystem has handled so poorly and slowly.

PR #2088, opened 2026-06-21
fix: do not require the about section on sign up.

### LINKED ISSUE #2087: About field required in invite form
On the invite/sign-up form, there's a field for a brief bio. The label text says it's optional, but the field is marked as required. (see image)

Image

After creating an account, you're able to edit this field and leave it blank, which suggests the intended behavior is for it to not be required on sign-up either.

Zurtar:
Fixes #2087.

Removes the required flag on the about field.


pushcx:
Thanks for the quick fix!

Issue #1882, opened 2026-01-29
Unify rss routes with comprehensive tests


pushcx:
In fixing #1868 I realized we have multiple URLs to rss feeds. It would be nice to standardize them, but we have to take special care not to break RSS URLs because people copy them into their feed readers and forget about them for a decade. If we break those URLs, our site disappears from readers, more-or-less silently depending on their UI.

Specifically, we have many urls, like `/top/1w/rss`, that end in `/rss` instead of the more familiar `.rss`.

This work needs a number of routing specs to confirm that rss feeds work and the old `/rss` all redirect to their `.rss` counterparts.

One odd bit of behavior popped up in #1602 that we support `/page/2.rss` and `/top/1w/page/2.rss`. (Thanks @dzwdz for help locating these.) I'm not sure these were deliberately planned, the concept is kind of puzzling (a feed starting at page 2?), and I'll have to search logs to figure out if anyone ever uses them. If there _is_ a reason for these to exist, it would help if their specs included a comment with that reason so we don't forget in the future.

For debugging, our Routes module is handy for quick feedback on URLs: `Routes.top_url('1w', format: :rss)`.

pushcx:
Implemented by @Zurtar in #2090.

PR #2090, opened 2026-06-22
Adding redirects for /top rss endpoints to .rss versions

Zurtar:
Addresses #1882. 

This is a rough pass at switching theRSS routes from the trailing /rss to the standard .rss suffix.

A few spots that need a look or clarification:
- **Comment redirect**: I added a redirect for /comments/rss, but it's likely unnecessary since we didn't previously support that path.
- **Top routes**: I left the /top/:length/page/:page.rss behavior for now. The issue mentions checking the logs to see if the routes are actually used so I didn't want to remove them.
- **Test placement**: I used the existing tests as a guide but I am not 100% confident that I placed them in the right location. If they should be moved from requests to routing or somewhere else please let me know.

There are likely other issues that I'm missing and more tests that should be added as well.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -319,7 +319,7 @@ class HomeController < ApplicationController
 
     @rss_link ||= {
       title: "RSS 2.0 - " + @title,
-      href: "/top/#{length[:param]}/rss"
+      href: "/top/#{length[:param]}.rss"
     }
 
     respond_to do |format|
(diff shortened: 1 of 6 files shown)


pushcx:
I'd like all of these to be routing specs, yes. It's a minor speedup, but it does add up. `spec/routing/user_spec.rb` has examples of testing redirects. I'd like to have a spec for every URL being redirected because I think they're going to be around a very, very long time given that developers like to use feed readers with plaintext config files rather than store them in a database. It is a norm violation for an app to edit its own config files, but not its own database, so I expect we'll serve these redirects for a couple decades.

I'm still checking the logs but I'm confident enough that no one is using the per-page RSS feeds, please drop support for them. They don't need tests.

pushcx:
Great, thanks for this.

We'll have to maintain these redirects about 20 years before they roll out of people's feed readers, so mark your calendar. :D

PR #2091, opened 2026-06-22
Set a default --color-fg-accent

### LINKED ISSUE #2089: Late loading of css produces jump when current_page is triggered
Try loading https://lobste.rs/active and play special attention to the content below the main nav. It's moved some pixels to the bottom because the border for current page is rendered after the rest of the page, and this is because the color (accent/red) is defined in the system-system css. When the page loads, the rule for the border is not valid because the accent color is not defined.

https://github.com/user-attachments/assets/5a4e0081-b6b3-457b-89d3-aa92f072ca14

fedemp:
Border for current page needs a color value to create a valid css rule. Currently it's parsed as `2px solid`, not a valid css rule. Defaulting to whatever the default foreground color the browser reports prevent layout shuffling when colors are finally loaded.

Fixes #2089


pushcx:
Good catch, thank you.

PR #2095, opened 2026-06-23
Remove subnav for stories by user page

### LINKED ISSUE #2062: "Newest" and "Top" links in user profile
This is a feature request or a bug depending on the expected outcome.

Visit this random user profile: https://lobste.rs/~pushcx/stories

There are the two links "Newest" and "Top" to the top right. As a user, I would expect them to allow me to sort that profile stories, but I'm actually taken to a general list of entries.

fedemp:
Top stories by user is not implemented so there's no need to show the "newest" and "top" links.

Fixes #2062


pushcx:
Thanks for noticing and fixing this.

PR #2101, opened 2026-06-23
fix cancel comment button

### LINKED ISSUE #2096: Cancel reply button breaks comment functionality
As a logged in user, click "reply" on a comment to make the reply form appear.
Press "Cancel" to make it dissappear.
Press again on "reply".
Reply form does not appear again.

----

If you try to reply to a comment that has child comment, after pressing "cancel", all child comments are removed.

---

Post a comment.
Press "edit" to make form appear.
Press "cancel".
Nothing happens.

Zurtar:
Fixes #2096.

The cancel button was removing the `<ol class="comments">` element that holds the reply form. Seems like it was introduced in dcfedfbc29fd4bdc93d1c1a054e4b8a6524c476c. 

Fixed by removing the `reply_form_temporary` element instead of the parent.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -936,7 +936,7 @@ document.addEventListener("DOMContentLoaded", () => {
           response.text().then(text => replace(comment, text));
         });
     } else {
-      comment.parentElement.remove();
+      comment.remove();
     }
   });
 


pushcx:
Thanks for fixing this up.

Issue #1242, opened 2024-01-21
Expose comments/threads by user as RSS feed


i2:
I would like to propose the following for RSS feed:
expose RSS feed for threads/comments by a user like so: https://lobste.rs/~exampleuserabc/threads.rss

Currently, this is only possible for user "stories" and tags/topic.

[ I confirm I discussed this feature request in the chat room or a meta thread before posting. ]

PR #2102, opened 2026-06-23
RSS feed for user's comments

Zurtar:
Addresses #1242.

Adds a RSS endpoint for /~user/threads. Currently it only shows the comments by that user and not any parent/child comments.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -385,11 +385,21 @@ class CommentsController < ApplicationController
       @title = "Your Threads"
     end
 
+    @rss_link ||= {
+      title: "RSS 2.0 - #{@showing_user.username} Comments",
+      href: user_token_link(user_threads_url(format: :rss))
+    }
+
     @threads = Comment.recent_threads(@showing_user)
       .merge(Story.not_deleted(@user))
       .for_presentation
       .joins(:story)
     @threads = CommentVoteHydrator.new(@threads, @user)
+
+    respond_to do |format|
+      format.html { render action: "user_threads" }
+      format.rss { render action: "user_threads", layout: false }
+    end
   end
 
   private
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -384,21 +384,36 @@ class CommentsController < ApplicationController
       @showing_user = @user
       @title = "Your Threads"
     end
-
-    @rss_link ||= {
-      title: "RSS 2.0 - #{@showing_user.username} Comments",
-      href: user_token_link(user_threads_url(format: :rss))
-    }
-
     @threads = Comment.recent_threads(@showing_user)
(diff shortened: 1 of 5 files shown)


pushcx:
Thanks for picking this up. You highlight an important distinction between threads and comments, and it's implicitly in your code where does the work to query for the full thread but throws it away. I'm sorry for the chore, but could you move this route to ~username/comments.rss?

And then I guess we need to link this from a meta tag on the user page for discoverability.

People have wanted a paginated per-author page for a while. The improved search engine (`commenter:pushcx`) has worked, but we still get people puzzled by ~username/threads not being paginated. This is nice work in that direction. To be explicit: please don't add the HTML page to this PR, you can choose to do so in a separate PR if it sounds interesting.

app/controllers/comments_controller.rb
```diff
@@ -384,14 +384,39 @@ def user_threads
...
     @threads = CommentVoteHydrator.new(@threads, @user)
   end
 
+  def user_comments
+    if params[:user]
+      @showing_user = User.find_by!(username: params[:user])
+      @title = "Comments for #{@showing_user.username}"
+    elsif !@user
+      return redirect_to active_path
+    else
```
pushcx: I don't think this branch is reachable from this route now, you can delete it.

app/controllers/comments_controller.rb
```diff
@@ -384,14 +384,39 @@ def user_threads
...
       .for_presentation
       .joins(:story)
     @threads = CommentVoteHydrator.new(@threads, @user)
   end
 
+  def user_comments
+    if params[:user]
+      @showing_user = User.find_by!(username: params[:user])
+      @title = "Comments for #{@showing_user.username}"
+    elsif !@user
```
pushcx: You can remove this branch, `@user` is the logged-in user. We don't need to enforce that the user is logged-in, and many feed readers don't even support setting per-feed cookies/

app/controllers/comments_controller.rb
Zurtar: This was copied over from `user_threads`, I naively assumed it applied but should have checked.

I'm now seeing `user_threads` has a route at the top level (shown below) that redirects to `user_threads`, which is why it's needed there.
```ruby
  get "/threads" => "comments#user_threads"
```

I'll remove it.

app/controllers/comments_controller.rb
Zurtar: Same thing as the other comment, it was needed in `user_threads` for the redirect. Sorry about that, should've verified before copying it over.

pushcx: Ah, I get it. Yeah, there's a lot of complexity around some of these god objects. :)

pushcx:
This is now totally clear, thanks for putting up with my allergy to `if`s.

And of course thanks for adding this feature.

PR #2105, opened 2026-06-25
Remove :has from allowed css selectors

Zurtar:
`:has` is now [baseline](https://caniuse.com/css-has)! 🎉 This removes it from ALLOWED_SELECTORS to fix the failing test.

As an aside the test based alert worked very well.


fedemp:
Beware that `:has` support on Firefox started on v121, and v115 is the latest Firefox that (will) run in Windows 7 and 8. I would use cautiously.

https://support.mozilla.org/en-US/kb/firefox-users-windows-7-8-and-81-moving-extended-support

MDN still reports it as newly available. 

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has

Has any browser release happened recently that changed the status?

pushcx:
@Zurtar :clinking_glasses: Delightful to know I'm allowed to have principles again, thanks for the update and the report that my time bomb test worked correctly.

@fedemp It's not a new release changing the status, it's that the popular browsers have supported it for 30 months.

PR #2107, opened 2026-06-27
Refactor search pagination block

### LINKED ISSUE #2054: UI: chaotic layout in search pages
In [any search](https://lobste.rs/search?q=spam&what=comments&order=newest) with many results, the pagination of the results has a chaotic layout.

Normal, with a monitor (everything ok, just as a reference):

Image

Narrower browser window:

Image

And in a window of the size of a cellphone:

Image

The problem is that the next row overlaps with the previous row (for example: 12 takes a corner from both 4 and 5. And in other cases, it takes the entire bottom edge (such as 20 from 8 or 10 from 2):
Image

Also I'm not sure if there should be something below the `page:` Maybe the next rows should start at the same level as the first one.

All cases are logged out. This is on firefox on linux (and the same happend on chromium) when you switch to a small window (the size of a cell phone) or in firefox on android.

fedemp:
Wrap search pagination in a nav element, use a list for links, and add vertical spacing when the list wrap into two lines.

Fixes #2054

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -868,10 +868,12 @@ nav.morelink {
 	float: left;
 }
 
-nav.morelink,
-div.page_link_buttons {
+nav.morelink {
 	margin-top: 1.5em;
 }
+
+#page-link-buttons,
+#page-link-buttons a,
 nav.morelink a {
 	color: var(--color-fg-contrast-7-5);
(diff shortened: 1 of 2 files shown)


fedemp:
Before and after:

image

pushcx:
Thanks for fixing this, nice cleanup!

PR #2110, opened 2026-06-28
Repeat textarea placeholder as plain text for alternative browsers

fedemp:
Most (if not all) text browsers do not support the placeholder attribute. Even Netsuft, being a graphical support, does not support it.

By repeating it in the submit story page, we can show the same hint to those.

### DIFF
--- a/app/views/stories/_form.html.erb
+++ b/app/views/stories/_form.html.erb
@@ -1,4 +1,9 @@
 <%# locals: (story:, f:, suggesting: false) -%>
+<% guidelines = "Use this textarea for stories without a URL, to link additional context, or to paste abstracts from PDF papers.
+
+Please don't use this to promote the story, summarize the post, or explain why you posted it.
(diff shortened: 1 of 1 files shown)


pushcx:
Sure. Could you pull this text out to a variable in the template so we don't repeat it? I know it's only a couple lines away, but I know from bitter experience that repeats that close can still fall out of sync.

Please also add a comment that this use of `hidden` currently-true assumption that the same browsers that don't support placeholder also don't support `hidden`. I get it because I'm looking at this PR and it feels safe enough to rely on (and the stakes low enough for when it breaks!), but it's really not going to be obvious to someone reading the source without this context.

fedemp:
Done :+1:

fedemp:
"Bump" because github-bot asked for it.

pushcx:
Sorry for the delay, I'm glad to merge this.

PR #2112, opened 2026-06-30
Escape html in data-title attr of tag selector

### LINKED ISSUE #2108: Dillo cant submit new entries; tags fail
When trying to submit a story with Dillo 3.2.0, I'm always getting an error, seems like tags are not  included in the request.

> 1 error prohibited this story from being saved
> There were the problems with the following fields:
> Must have at least one non-media (PDF, video tag). If no tags apply to your content, it probably doesn't belong here.

fedemp:
For the tag selector in the submit story form, escape the html that is set in the data-title attribute. This used to break Dillo HTML parsing

Fixes #2108

### DIFF
--- a/app/views/stories/_form.html.erb
+++ b/app/views/stories/_form.html.erb
@@ -57,7 +57,7 @@
           (t.filtered_count == 1 ? "" : "s") << " filtering</em>"
       end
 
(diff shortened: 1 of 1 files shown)


pushcx:
I wonder if this was an xss vector. Thanks for removing the `raw` call.

PR #2113, opened 2026-06-30
Fix caches dropdown in netsurf

fedemp:
Fixes caches list overlapping other entries in Netsurf by hiding the styles using an attribute selector.

The display: inline-block is ignored by the rest of the browsers, and keeps the UI slightly more tidy in Netsurf

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1022,7 +1022,7 @@ div.comment_text code {
 	list-style: none;
 }
 
-.caches ul {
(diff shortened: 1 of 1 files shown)


fedemp:
Before and after:

image

image

pushcx:
Thanks for this fix.

I've really appreciated that you're thorough on this kind of work and acted responsibly when we've had the inevitable minor issues. I'm confident in your work on these kind of changes where I previously would probably have decided that it wasn't worth the risk for a marginal browser.

PR #2114, opened 2026-07-01
Use HTML list for tags

fedemp:
Tag links should be grouped to give them context. This is from my
understandind of "WCAG 2.1 SC 2.4.4 Link Purpose (In Context) (Level A)"

> The purpose of each link can be determined from the link text alone or
from the link text together with its programmatically determined link
context.

And then for "context", it reads

> [...] text that is in the same paragraph, list item, or table cell as
the link or in a table header cell [...]

So, I guess we can say there's context because once a screen reader
lands on the first items, it reads "Tags, list with n items, tag name,
link".

When listing links, the list context is lost, so an aria-label is added to prefix each link with "Tag"

https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context
https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context#dfn-programmatically-determined-link-context

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -664,6 +664,13 @@ li.story a.tag {
 
 li .tags {
 	margin-right: 0.25em;
+	display: inline-block;
+	vertical-align: middle;
+	padding: 0;
+}
+
+li .tags li {
+	display: inline-block;
 }
 
 li .domain {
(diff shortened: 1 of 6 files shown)


pushcx:
Thanks. This feels higher-risk because of the visibility of this, but I do like using semantic tags, and it's clear it'll be a better experience for screen readers. Thanks for including the docs.

Maybe at some point we want to consider a CSS reset to wipe default browser styles like the padding on `li`. They weren't very popular when the site started, but I've come to appreciate how much more less trivia you have to Just Know for the rest of your CSS.

PR #2117, opened 2026-07-02
Fix dotted line comment rendering

### LINKED ISSUE #2116: dotted line overlap visualization issue
This example can be found at https://lobste.rs/s/v2cbi5/stop_asking_writers_about_ai#c_i2hxi1

# Current behavior

## Uncollapsed

Image

## Collapsed

Image

Please note that the dotted line overlaps with the number "6".

# Desired behavior

## Uncollapsed

no desired change

## Collapsed

Image

GreenBeard:
Fixes #2116

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -533,7 +533,6 @@ ol.stories li.story div.story_liner {
 	overflow-wrap: break-word;
 }
 .comment {
-	clear: both;
 	padding-top: 0.4em;
 	padding-bottom: 0.4em;
 	position: relative;
@@ -712,6 +711,7 @@ li .comment_folder_button:checked ~ ol.comments li {
 }
 
 li.comments_subtree {
+	clear: both;
 	position: relative;
 }
 


pushcx:
Could you explain this fix a little? I'm not immediately getting it from the CSS.

GreenBeard:
> Could you explain this fix a little? I'm not immediately getting it from the CSS.

I'm not sure if you are asking me to explain the problem, how this fixes it, or why this solution over others?

# For the problem

I'll assume (hopefully correctly, if not I apologize) that the linked issue with the comment "Please note that the dotted line overlaps with the number '6'." is clear enough.

# For how this fixes it

Currently `.comments_subtree` effectively has unused padding (I use the term loosely and not in the real CSS `padding` definition) for the `.comment` as the `.comment` has a `clear: both;`. Unfortunately, this does not apply to other two children of `.comments_subtree`, the `.comment_folder_button` and `:before`. The `.comment_folder_button` is `display: none` so in effect it doesn't matter where it would display since it doesn't. The code assumes that it can determine the `:before`'s position relative to the `.comment` by placing itself relative the `.comments_subtree` using `top: 3.6lh;`. The `:before` is the dotted line which is why its position relative to the `.comment` is important. Since the `.comment` is offset vertically downward by the `clear: both` in a way that the `:before` and `.comments_subtree` are not, it creates this visual glitch.

If you want to understand why `clear: both` is causing the `.comment` to changes its vertical position that is because of the prior comment (i.e. `.comment` and `.comments_subtree` combo) containing a `div.voters` that is `float: left` which is not included in the height calculation of said prior `.comment` but is affecting the `clear: both` of this `.comment` (for the current code in `main`, changed to `.comments_subtree` in this proposed PR). This issue is exacerbated as `div.voters` is much larger than one may expect as it includes a `.upvoter` element which is `visibility: hidden` (not `display: none`) when collapsed. Even if `.upvoter` was `display: none` it would be relying upon the questionable assumption that the `.voters` will always be shorter than `.details` (and making it `display: none` is probably not desirable anyway).

# Why this solution over others

I'm not sure I like the `clear: both` on the `.comment` to begin with, but I was trying to make a minimal change since I'm not sure of the style that lobste.rs usually has for CSS and I lean towards minimal changes for projects I'm not used to. This was the least invasive fix that I could see which is why it is why I put in this PR. I'm also not a CSS expert by any means so I thought I would make a tiny proposal that you could rewrite if you didn't like it. I can outline some _alternative_ proposals below (other than the one already shown in this PR) in case that it is helpful or gives more context.

## Option 1 ( similar to https://www.w3schools.com/css/css_float_clear.asp#:~:text=The%20CSS%20Clearfix%20Hack ):

```
.comment::after {
    content: "";
    display: block;
    clear: both;
}
```

and probably remove the `clear: both` from `.comment`.

## Option 2:

```
.comment {
    display: flow-root;
}
```

and probably remove the `clear: both` from `.comment`. I'm least familiar with the pros and cons of this option as I just poke at CSS occasionally as a hobby.

> Formatting contexts affect layout because an element that establishes a new block formatting context will:
>
> * contain internal floats.
> * exclude external floats.
> * suppress margin collapsing.

https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Display/Block_formatting_context#:~:text=Formatting%20contexts%20affect%20layout and https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display#flow-root which seems to possibly imply that this could have undesirable behavior if you are used to "margin collapsing" or are worried about "exclude external floats".

pushcx:
Thanks for that thorough explanation, I get how it's working now. (I hadn't seen [display: flow-root](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display) before, thanks.)

The flow/clear-based layout is really complicated, I don't see that there's a simpler way to do this. You made the right call with a focused, minimal fix here.

PR #2119, opened 2026-07-02
Prefer rem over rlh in CSS

### LINKED ISSUE #2115: Servo  2026-06-30 renders some text fields too short
Not a big issue but worth investigating to see if other browsers are just magically fixing the CSS.

The username and password fields are rendered too short, in both profile and login pages. Other fields, like the search field, work fine.

Image

fedemp:
Servo does not seem to support rhl and default to some other unit that renders input too short. Replace with the equivalent size using rems.

Fixes #2115

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1402,7 +1402,7 @@ label:has(+ *[type="hidden"] + *:required)::after {
 }
 .labelled_grid input[type="text"],
 .labelled_grid input[type="password"] {
-	height: 1.5rlh;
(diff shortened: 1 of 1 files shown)


fedemp:
I saw that when `rlh` was added, there was a comment about fields breaking when the label wrap into more than one line, but this does not seem to happen even if we use `rem`.

image

pushcx:
Ah dang, I was so excited for `rlh`. Thank you for fixing it.

For future reference, here's the Servo issue about it: https://github.com/servo/servo/issues/24848

PR #2124, opened 2026-07-03
Dont let link be hidden by caches dropdown; remove line throght of user on active

fedemp:
There are two commits here.

* [Remove line through for inactive user on focus and hover.](https://github.com/lobsters/lobsters/commit/ea4f50cb2c053b44352c6ec6f12754ed27e619bf): like it says, the line through is removed for inactive users when the link is hovered of focused using a keyboard.

image

* [Don't let links be hidden by caches dropdown](https://github.com/lobsters/lobsters/commit/3ae6295bfccef09327e182afef89006bb9479666): Since the 'caches' dropdown does not close automatically, it may obscure other links near it. This change raises them when focused or hovered.

image

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -774,11 +774,14 @@ span.new_user, a.new_user,
 li .byline a.new_user {
 	color: var(--color-fg-affirmative);
 }
-span.inactive_user, a.inactive_user,
-li .byline a.inactive_user,
+a.inactive_user,
 .inactive_tag {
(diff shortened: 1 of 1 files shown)


pushcx:
I'm guessing you want to hide the strikethrough to make the name easier to read with a heavy box around it? Makes sense.

Thanks for the caches fix, too.

fedemp:
Yes. I was reading WCAG guidelines and it mentions one should make the best effort so focused links are not obscured.

PR #2125, opened 2026-07-06
Refactor user profile page

fedemp:
For those browsers that do not support CSS, the profile page was just one infinite line. Refactored it so use description lists, aka &lt;dt>

Inspired by https://benmyers.dev/blog/on-the-dl/


fedemp:
Before and after:

Screenshot From 2026-07-05 23-13-35

image

----

image


image

pushcx:
Really satisfying to get an improvement like this by better using HTML semantics, thank you.

PR #2127, opened 2026-07-08
Fix expiring avatars error

hugoarnal:
Recent merge of PR #1952 moves the logic of expiring the avatars from `avatars_controller.rb` to `user.rb`.

I wanted to try it in prod, however it causes errors due to some variables not being accessible outside of `avatars_controller`. Here's my proposed fix.

I have tested it in my dev environment.

(mentionning @pushcx as expiring your cache or setting your avatar source currently returns 500)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -378,11 +378,11 @@ class User < ApplicationRecord
   def expire_avatar!
     expired = 0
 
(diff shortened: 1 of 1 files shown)


pushcx:
Ah great, thank you. I moved that code over pretty quick, thanks for cleaning up my bug.

PR #2132, opened 2026-07-10
Disable Story Byline Buttons when Previewing

### LINKED ISSUE #2109: Story submit previews render broken save and hide buttons, and comments link
When previewing a story to submit, the buttons "save" and "hide" are enabled but, as expected, return a error 400 when submitted. A links to comment appears too but points to an invalid url.

Zurtar:
Fixes: #2109. 

Pretty straight forward, just disables the two forms or removes the link/href from the byline buttons when previewing.

### DIFF
--- a/app/views/stories/_singledetail.html.erb
+++ b/app/views/stories/_singledetail.html.erb
@@ -66,7 +66,7 @@ classes = [
           <a href="<%= edit_story_path(ms.short_id) %>">edit</a>
         <% end %>
 
-        <% if @user&.is_moderator? %>
+        <% if @user&.is_moderator? && !ms.previewing %>
           <%= divider_tag %>
           <% css = ms.has_suggestions? ? "story_has_suggestions" : "" %>
           <% if ms.previewing %>
--- a/app/views/stories/_singledetail.html.erb
+++ b/app/views/stories/_singledetail.html.erb
@@ -68,12 +68,7 @@ classes = [
 
         <% if @user&.is_moderator? && !ms.previewing %>
           <%= divider_tag %>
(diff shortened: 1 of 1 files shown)


app/views/stories/_singledetail.html.erb
```diff
@@ -68,7 +68,12 @@ classes = [
 
         <% if @user&.is_moderator? %>
```
pushcx:
```suggestion
        <% if @user&.is_moderator? && !ms.previewing %>
```

Yeah, there's no nice way to disable a link. Let's just hide it, mods won't notice.

pushcx (changes requested):
Thanks for picking this up, only one small simplification.

app/views/stories/_singledetail.html.erb
```diff
@@ -66,9 +66,14 @@ classes = [
           <a href="<%= edit_story_path(ms.short_id) %>">edit</a>
         <% end %>
 
-        <% if @user&.is_moderator? %>
+        <% if @user&.is_moderator? && !ms.previewing %>
           <%= divider_tag %>
-          <a href="<%= edit_mod_story_path(ms.short_id) %>" class="<%= ms.has_suggestions? ? "story_has_suggestions" : "" %>">mod edit</a>
+          <% css = ms.has_suggestions? ? "story_has_suggestions" : "" %>
```
pushcx: Sorry I didn't spell it out on the `if` above: you can revert this chunk.

Zurtar: Ah I see, missed that the suggestion didn't cover the whole revert. Should be good now, sorry about that.

pushcx:
Thanks for fixing this.

PR #2133, opened 2026-07-10
Hide reply form when collapsing a comment

### LINKED ISSUE #2121: Toggling a comment fold also toggle comment reply preview
Since folding a comment does not hide the reply form, I did not expect that previews would be hidden.

https://github.com/user-attachments/assets/1df28340-13b4-420f-b891-3e94d51c1481

Zurtar:
Fixes: #2121.

Adjusted the CSS so the collapse also applies to the reply form. This had the side effect of the reply form being hidden if you're replying to a collapsed comment, so I also added JS to un-collapse the comment when reply is clicked.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -719,8 +719,10 @@ li .comment_folder_button:checked ~ div.comment:not(:target) div.voters .upvoter
 	visibility: hidden;
 	margin-bottom: -15px;
 }
+
 li .comment_folder_button:checked ~ ol.comments ol,
 li .comment_folder_button:checked ~ ol.comments div.comment,
+li .comment_folder_button:checked ~ ol.comments div.reply_form_temporary,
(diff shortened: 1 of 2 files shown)


pushcx:
Can I add a little complexity and ask you to refactor the collapse data structure access into a couple of functions? Seeing it repeated like this tripped my personal sense of risk that things will slip out of sync.

Zurtar:
Ha! Funnily enough I had it factored out for that exact reason before opening the PR. Because there's very few functions in that file I ended up inlining it to match the style. I'll switch it back and update the branch

pushcx:
Heh, funny how that goes. I don't like to extract functions until I'm sure to reuse them or something about them gets especially complex.

d669578 ooks like you moved the functions and they got duplicated?

pushcx:
Great, thanks for fixing this and improving the code.

PR #2134, opened 2026-07-11
Peformance Improvement: Preload comments in /threads

thomasdziedzic:
Someone mentioned on IRC that the /threads endpoint is the slowest one for them. I looked for quick wins and this is what I found:

Local testing required comment queries of ~160ms twice for the comments which is by far the most expensive query on this page.

previously the #.empty? check along with an non-empty set of comments would lead to doubling the queries needed.

Preload this so we only need to query the comments once.

There might still be room for improvement to optimize the `@threads` query but for now this should cut the query time in half for the most expensive part of the request.


pushcx:
Aha, thanks for this fix.

PR #2137, opened 2026-07-12
add timestamp to slow query log

thomasdziedzic:
Slow query logs should now display a timestamp for easier debugging.

```
W, [2026-07-12T07:57:53.045335 #432526]  WARN -- : [SLOW QUERY] ...
```


pushcx:
Thanks, I'm going to deploy this now so we get feedback.

I'm a big fan of json logs and would like to use those. I dropped in `logger.formatter = Lograge::Formatters::Json.new` but got an odd error out of the `.warn`: `wrong number of arguments (given 4, expected 1)`. I'd rather not run that down this second.

PR #2140, opened 2026-07-12
Improve traffic queries

### LINKED ISSUE #2139: improve traffic-counting query
`config/recurring.yml` kicks off a recurring job `traffic_helper` to call `TrafficHelper.cache_traffic!`

This query buckets stories, comments, and votes into 15-minute increments to set the intensity of the logo as a quick at-a-glance indicator of how busy the site is. It's very cute! As the site has grown, this has gotten expensive.

The `TrafficHelper.traffic_range` query takes ~1800ms and each run does that work over again. Query plan:

```
QUERY PLAN
|--CO-ROUTINE v
|  |--SCAN votes
|  `--USE TEMP B-TREE FOR GROUP BY
|--MATERIALIZE c
|  |--SCAN comments
|  `--USE TEMP B-TREE FOR GROUP BY
|--MATERIALIZE s
|  |--SEARCH stories USING COVERING INDEX index_stories_on_created_at (created_at>?)
|  `--USE TEMP B-TREE FOR GROUP BY
|--SCAN v
|--BLOOM FILTER ON s (period=?)
|--SEARCH s USING AUTOMATIC COVERING INDEX (period=?)
|--BLOOM FILTER ON c (period=?)
`--SEARCH c USING AUTOMATIC COVERING INDEX (period=?)
```

The `TrafficHelper.current_activity!` query takes ~1300ms. Query plan:

```
select
        (SELECT count(1) AS n_votes   FROM votes    WHERE updated_at >= '2026-07-12 19:00:01 UTC') +
        (SELECT count(1) AS n_comment FROM comments WHERE created_at >= '2026-07-12 19:00:01 UTC') * 10 +
        (SELECT count(1) AS n_stories FROM stories  WHERE created_at >= '2026-07-12 19:00:01 UTC') * 20;
```

Maybe just improving the queries to use indexes better would do it. Off the top of my head, rather than filter by `created_at`, subquery to find the id of one row `created_at >= 90.days.ago` and then filter by id. I'm sure there are other, more clever query changes.

Maybe this should materialize the buckets to `Rails.cache` and just query the last 15 minutes to change the latest value in it, perhaps dropping the opening buckets older than 90 days.

Maybe you've got an even better idea? 😄 Lots of room for creativity on this one.

thomasdziedzic:
Adds expression indexes to the traffic_range query and additional indexes to the current_activity query to avoid full table scans.

traffic_range query before: 0.52 seconds
after: 0.08 seconds

current_activity! query before: 0.31 seconds
after: 0.00 seconds

Should fix #2139 

traffic range query plan after:

```
QUERY PLAN
|--CO-ROUTINE v
|  `--SCAN votes USING COVERING INDEX index_votes_on_period_and_updated_at
|--MATERIALIZE c
|  `--SCAN comments USING COVERING INDEX index_comments_on_period_and_created_at
|--MATERIALIZE s
|  `--SCAN stories USING COVERING INDEX index_stories_on_period_and_created_at
|--SCAN v
|--BLOOM FILTER ON s (period=?)
|--SEARCH s USING AUTOMATIC COVERING INDEX (period=?)
|--BLOOM FILTER ON c (period=?)
`--SEARCH c USING AUTOMATIC COVERING INDEX (period=?)
```

current_activity! query plan after:

```
QUERY PLAN
|--SCAN CONSTANT ROW
|--SCALAR SUBQUERY 1
|  `--SEARCH votes USING COVERING INDEX index_votes_on_updated_at (updated_at>?)
|--SCALAR SUBQUERY 2
|  `--SEARCH comments USING COVERING INDEX index_comments_on_created_at (created_at>?)
`--SCALAR SUBQUERY 3
   `--SEARCH stories USING COVERING INDEX index_stories_on_created_at (created_at>?)
```

### DIFF
--- a/db/migrate/20260712205915_add_indexes_to_traffic_related_columns.rb
+++ b/db/migrate/20260712205915_add_indexes_to_traffic_related_columns.rb
@@ -1,11 +1,5 @@
 class AddIndexesToTrafficRelatedColumns < ActiveRecord::Migration[8.0]
   def change
-    # Used by TrafficHelper.traffic_range
-    add_index :comments, "floor(unixepoch(created_at)/900), created_at", name: "index_comments_on_period_and_created_at"
-    add_index :stories, "floor(unixepoch(created_at)/900), created_at", name: "index_stories_on_period_and_created_at"
-    add_index :votes, "floor(unixepoch(updated_at)/900), updated_at", name: "index_votes_on_period_and_updated_at"
-
-    # Used by TrafficHelper.current_activity!
     add_index :comments, :created_at
     add_index :votes, :updated_at
   end
--- a/db/migrate/20260712205915_add_indexes_to_traffic_related_columns.rb
+++ b/db/migrate/20260712205915_add_indexes_to_traffic_related_columns.rb
@@ -1,0 +1,12 @@
+class AddIndexesToTrafficRelatedColumns < ActiveRecord::Migration[8.0]
+  def change
+    # Used by TrafficHelper.traffic_range
+    add_index :comments, "floor(unixepoch(created_at)/900), created_at", name: "index_comments_on_period_and_created_at"
+    add_index :stories, "floor(unixepoch(created_at)/900), created_at", name: "index_stories_on_period_and_created_at"
+    add_index :votes, "floor(unixepoch(updated_at)/900), updated_at", name: "index_votes_on_period_and_updated_at"
+
+    # Used by TrafficHelper.current_activity!
+    add_index :comments, :created_at
+    add_index :votes, :updated_at
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2026_06_13_004304) do
+ActiveRecord::Schema[8.0].define(version: 2026_07_12_205915) do
   create_table "action_mailbox_inbound_emails", force: :cascade do |t|
     t.integer "status", default: 0, null: false
     t.string "message_id", null: false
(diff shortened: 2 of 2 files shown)


pushcx:
I don't have great instincts for scale here, but and index feels like a heavyweight, nonlocal solution for a toy like this. All of our 3 biggest tables get wider and every insert slower so we can speed up one query every 5 minutes. Any idea how to measure that to see if it's a good tradeoff?

thomasdziedzic:
In terms of write overhead caused by the new indexes, I can only think of trying a large write to those tables and comparing the before and after times to see the impact on write performance. One proxy that doesn't involve that could be to see size of indexes. My local test db is about 2.2G so roughly half the size of prod and here are the sizes of all those new indexes:

```
sqlite> select name, pgsize AS total_bytes
from dbstat
where aggregate = true
and name in ('index_votes_on_period_and_updated_at', 'index_comments_on_period_and_created_at', 'index_stories_on_period_and_created_at', 'index_votes_on_updated_at', 'index_comments_on_created_at')
order by total_bytes desc;

index_votes_on_period_and_updated_at|18927616
index_votes_on_updated_at|16961536
index_comments_on_period_and_created_at|14192640
index_comments_on_created_at|12722176
index_stories_on_period_and_created_at|4702208
```

Here are the top 15 largest objects on my local db:

```
sqlite> select name, pgsize AS total_bytes
from dbstat
where aggregate = true
order by total_bytes desc
limit 15;

comments|1194774528
stories|409436160
comments_fts_data|141287424
links|101265408
index_links_on_url_and_from_story_id_and_from_comment_id|64258048
index_links_on_normalized_url|45305856
votes|22740992
idx_links_on_to_comment_id_and_from_story_id_and_from_comment_id|22683648
index_links_on_to_story_id_and_from_story_id_and_from_comment_id|22683648
index_votes_on_period_and_updated_at|18927616
index_links_on_from_comment_id|18448384
index_comments_on_token|17600512
index_votes_on_updated_at|16961536
index_links_on_from_story_id|15691776
index_comments_on_period_and_created_at|14192640
```

Note that index_votes_on_period_and_updated_at, index_votes_on_updated_at and index_comments_on_period_and_created_at show up in the top 15.

We could get rid of the indexes on expressions and just have the additional 2 indexes on comments(created_at) and votes(updated_at). Those 2 indexes will be used in the traffic range query as well, avoiding the table scans, but iirc when I tested that the performance on that query improved from 0.5 to 0.3 seconds (and current_activity was still at 0.00 seconds), an improvement but not as big as I would have liked which is why I added the indexes on expressions.

thomasdziedzic:
I removed the period specific indexes and kept the index_comments_on_created_at and index_votes_on_updated_at.

thomasdziedzic:
I'm going to think about the original issue some more before working on this some more.

PR #2143, opened 2026-07-13
Remove ref tracking query param from submitted URLs

### LINKED ISSUE #2111: Duplicate checker doesn't trim URL
When the url of a story you try to submit, contains a ref attribute, it counts as a new url and is not detected by the duplicate checker.

I've submittes this url: "https://overreacted.io/there-are-no-instances-in-atproto/?ref=selfh.st" which count as a new url because of "?ref=selfh.st". 

I think trimming or sanitizing urls before submitting would be a good way to prevent this.

itsliamegan:
Removes the `ref` parameter from submitted story URLs. Fixes #2111.

It's is not as ubiquitous as Google Analytics' `utm_` parameters, but [Plausible](https://plausible.io/docs/custom-query-params), [Fathom](https://usefathom.com/docs/start/dashboard#refs), and [SimpleAnalytics](https://docs.simpleanalytics.com/how-to-use-url-parameters#using-a-url-parameter) document support for tracking it. In addition, lobste.rs itself [has code](https://github.com/lobsters/lobsters/blob/main/app/models/story.rb#L1193) assuming it's used for tracking/attribution purposes.

Rather than going in the `Utils.normalize` method, I put this next to the other code for stripping tracking parameters. Because this parameter is checked to create a mod note, I also moved the stripping itself after the check for attribution.

I didn't add a DB migration as suggested in [#2111](https://github.com/lobsters/lobsters/issues/2111#issuecomment-4837426170), both because this change affects the actual `url` field as opposed to the `normalized_url` field and because none of the previous changes to strip tracking parameters migrated the data.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -1185,15 +1185,15 @@ class Story < ApplicationRecord
     # strip out tracking query params
     if (match = u.match(/\A([^?]+)\?(.+)\z/))
       params = match[2].split(/[&?]/)
-      # utm_ is google and many others; sk is medium; si/pp is youtube
-      params.reject! { |p|
-        p.match(/^utm_(source|medium|campaign|term|content|referrer)=|^sk=|^gclid=|^fbclid=|^linkId=|^pp=|^si=|^trk=/x)
-      }
       params.reject! { |p|
         if /^lobsters|^src=lobsters|^ref=lobsters/x.match?(p)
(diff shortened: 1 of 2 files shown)


pushcx:
Thanks for picking this up! You nailed it on the code fix and tests. Sorry I gave such confused advice in the issue, I guess I didn't recall that code as well as I thought.

Issue #1864, opened 2026-01-19
Export GitHub Issues and PRs with their comments


pushcx:
GitHub has gotten really flaky over the last year and I've grown [uneasy]() with it having the only copy of our issues, the comments there, PRs, and those reviews and comments. (Note that all of these can be edited and I'd like to include their edit history.)

I've seen [git-bug](https://lobste.rs/c/riyczk) for syncing issues, maybe there's that or another tool that can export the issues and PRs into our repo on an ongoing basis. It's not yet time to leave GitHub, but I don't want to be left without our data if GH becomes unusable, whether through becoming a totally unreliable SPA or [tying](https://en.wikipedia.org/wiki/Tying_(commerce)) CoPilot invasively. It's also nice to ready for if [something better comes along](https://lobste.rs/c/efqrmj) like jujutsu has replaced git.

I have no strong opinion on the format they're saved to as long as it's reasonably humanly readable (JSON, XML, TOML, etc.).

I expect that this will end up being some kind of script that I have to run periodically to commit changes to our repo, though it would be neat if it used GitHub Actions to update as issue and PRs are updated.

pushcx:
Maybe I've been too slow with this.

![Image](https://github.com/user-attachments/assets/7f699aee-9fc5-4892-bf79-2b1cbc570445)

[via](https://bsky.app/profile/swizec.com/post/3mehgxapfmk2l)

PR #2145, opened 2026-07-14
Fix case-sensitive username mentions

### LINKED ISSUE #2141: @username mentions shouldn't be case-sensitive
I mentioned `@355e3b` in [a comment](https://lobste.rs/c/xhgudp) and it wasn't hyperlinked, presumably because his username is `355E3B`.

Perhaps this is a collation bug caused by our recent migration to sqlite in #539?

Zurtar:
Fixes: #2141.

Makes username lookups for mentions case-insensitive.

### DIFF
--- a/extras/markdowner.rb
+++ b/extras/markdowner.rb
@@ -115,8 +115,8 @@ class Markdowner
 
       node.string_content = before
 
-      if Username.where("created_at < ? and (? < renamed_away_at or renamed_away_at is null)", as_of, as_of).exists?(username: user[1..])
-        link = Commonmarker::Node.new(:link, url: Rails.application.root_url + "~#{user[1..]}")
+      if Username.where("created_at < ? and (? < renamed_away_at or renamed_away_at is null)", as_of, as_of).where("LOWER(username) = ?", user[1..].downcase).exists?
+        link = Commonmarker::Node.new(:link, url: Rails.application.root_url + "~#{user[1..].downcase}")
         node.insert_after(link)
 
         link_text = Commonmarker::Node.new(:text)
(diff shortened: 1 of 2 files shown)


Zurtar:
While debugging I added a unit test to check if mention notifications worked with mixed case, they passed so I didn't include them here but I can add them if you feel it's needed.

(test for reference)
```ruby
#spec/jobs/notify_comment_job_spec.rb
    it "sends mention notification with mixed case" do
      recipient = build(:user)
      recipient.username = "alice"
      recipient.settings["email_notifications"] = true
      recipient.settings["email_mentions"] = true
      recipient.save!

      sender = create(:user)
      c = create(:comment, user: sender, comment: "@aLiCe")

      NotifyCommentJob.perform_now(c)

      expect(sent_emails.size).to eq(1)
      expect(sent_emails[0].subject).to match(/Mention from #{sender.username}/)
      expect(recipient.notifications.count).to eq(1)
      expect(recipient.notifications.first.notifiable).to eq(c)
    end
```

pushcx:
Welp, it finally happened, GitHub's flakiness caused data loss. I wrote a review comment on a PR, clicked over to the diff to see something, and flipped back to find my comment gone. Ugh. Knowing this was probably coming didn't take the sting out of it.

---

Thanks for fixing this.

This is definitely caused by #539 and Thomas [called this out as a limitation](https://lobste.rs/s/ko1ji1/lobste_rs_is_now_running_on_sqlite#:~:text=Collation%20in%20SQLite%20is%20rather%20weak%20compared%20to%20MariaDB).

I don't think we need to double up on the test that if someone is mentioned, they get a notification.

Zurtar:
> Welp, it finally happened, GitHub's flakiness caused data loss...

Might be a good time to bump #1864.

PR #2147, opened 2026-07-14
WIP: Adding username case-insensitivity tests

Zurtar:
Follow up to #2145. See #2141 for context

Adding tests to areas that take in username input to validate that they are case insensitive.

### DIFF
--- a/spec/models/message_spec.rb
+++ b/spec/models/message_spec.rb
@@ -14,6 +14,12 @@ describe Message do
     expect(m_valid_short_id).not_to be_valid
   end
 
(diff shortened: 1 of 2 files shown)


Zurtar:
Closing as a better solution has been discussed [here](https://github.com/lobsters/lobsters/issues/2141#issuecomment-4971395253).

Anecdotally I did manually test the private messages, mod mail and search engine for the case sensitivity issue. Everything semed to work.

Although I only wrote full test cases for private messages and the search engine before closing this PR.

PR #2153, opened 2026-07-16
Experimental recursive CTE optimisations

chamlis:
I've been messing around locally to try and optimise the recursive CTEs for `story_threads`.

This attempt is based on these things:
- SQLite looks to have to do a lot of shuffling of data around for everything inside a recursive CTE. Therefore, this makes the CTE only have a column for the comment ID and one for "priority" - as confidence is always between 0 and 1, we can just add it to the comment depth to get a priority level for traversing each comment.
- A covering index is added specifically for this recursive CTE
- A `CROSS JOIN` is used to join the CTE to `comments` - this [lets us force a join order](https://sqlite.org/optoverview.html#crossjoin). As we know any one story will only contain a fraction of all the site comments, we want to scan the CTE and then look up comments by their ID, rather than the other way around even including a bloom filter.

This looks to be an improvement locally, but I dare not make any predictions about how it will behave on the production database. I've thrown the kitchen-sink here in the hopes that at least some of it is helpful, even if some of it is better off reverted!

I get the following query and query plan for this locally:

```sql
WITH RECURSIVE "comment_tree" AS ( SELECT comments.id, comments.depth + comments.confidence AS priority FROM "comments" INNER JOIN "stories" ON "stories"."id" = "comments"."story_id" WHERE ((stories.id = 90 or stories.merged_story_id = 90) and parent_comment_id is null) UNION ALL SELECT "comments".id, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc ) SELECT "comments".* FROM comment_tree CROSS JOIN comments ON comment_tree.id = comments.id;
```

```
QUERY PLAN
|--CO-ROUTINE comment_tree
|  |--SETUP
|  |  |--SEARCH comments USING INDEX comments_parent_comment_id_fk (parent_comment_id=?)
|  |  `--SEARCH stories USING INTEGER PRIMARY KEY (rowid=?)
|  `--RECURSIVE STEP
|     |--SCAN comment_tree
|     `--SEARCH comments USING COVERING INDEX index_comment_tree (parent_comment_id=?)
|--SCAN comment_tree
`--SEARCH comments USING INTEGER PRIMARY KEY (rowid=?)
```

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -592,12 +592,12 @@ class Comment < ApplicationRecord
     return Comment.none unless story.id # unsaved Stories have no comments
 
     Comment.with_recursive(comment_tree: [
-      Comment.joins(:story).where("(stories.id = #{story.id} or stories.merged_story_id = #{story.id}) and parent_comment_id is null").select("comments.id, comments.depth + comments.confidence AS priority"),
+      Comment.joins(:story).where("(stories.id = #{story.id} or stories.merged_story_id = #{story.id}) and parent_comment_id is null").select("comments.*, comments.depth + comments.confidence AS priority"),
       # work around Rails bug from https://github.com/rails/rails/pull/51549 that generates invalid
       # SQLite-style SQL by wrapping parens around the recursive case if it has an 'order' clause.
       # Comment.joins("JOIN comment_tree ON comment_tree.id = comments.parent_comment_id").order("depth desc", "confidence desc")
-      Arel.sql('SELECT "comments".id, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc')
+      Arel.sql('SELECT "comments".*, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc')
     ])
-      .from("comment_tree").joins("CROSS JOIN comments ON comment_tree.id = comments.id")
+      .from("comment_tree as comments")
   end
 end
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -591,28 +591,13 @@ class Comment < ApplicationRecord
   def self.story_threads(story)
     return Comment.none unless story.id # unsaved Stories have no comments
 
-    story_ids = [story.id] + Story.where(merged_story_id: story.id).pluck(:id)
-
-    inner_join = <<~SQL
-      inner join (
-        with recursive confidence as (
-          select
-            c.id,
-            cast(confidence_order as blob) as confidence_order_path
-            from comments c
-            join stories on stories.id = c.story_id
-            where stories.id in (#{story_ids.join(", ")}) and parent_comment_id is null
-          union all
-          select
-            c.id,
-            cast(concat(substring(confidence.confidence_order_path, 1, 3 * (depth + 1)), c.confidence_order) as blob)
-          from comments c join confidence on c.parent_comment_id = confidence.id
-        )
-        select * from confidence
-      ) confidence
-      on comments.id = confidence.id
-    SQL
-
-    Comment.joins(inner_join).where(story_id: story_ids).order("confidence.confidence_order_path")
+    Comment.with_recursive(comment_tree: [
+      Comment.joins(:story).where("(stories.id = #{story.id} or stories.merged_story_id = #{story.id}) and parent_comment_id is null").select("comments.id, comments.depth + comments.confidence AS priority"),
+      # work around Rails bug from https://github.com/rails/rails/pull/51549 that generates invalid
+      # SQLite-style SQL by wrapping parens around the recursive case if it has an 'order' clause.
+      # Comment.joins("JOIN comment_tree ON comment_tree.id = comments.parent_comment_id").order("depth desc", "confidence desc")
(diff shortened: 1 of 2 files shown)


app/models/comment.rb
```diff
@@ -591,28 +591,13 @@ def self.recent_threads(user)
...
-
-    Comment.joins(inner_join).where(story_id: story_ids).order("confidence.confidence_order_path")
+    Comment.with_recursive(comment_tree: [
+      Comment.joins(:story).where("(stories.id = #{story.id} or stories.merged_story_id = #{story.id}) and parent_comment_id is null").select("comments.id, comments.depth + comments.confidence AS priority"),
+      # work around Rails bug from https://github.com/rails/rails/pull/51549 that generates invalid
+      # SQLite-style SQL by wrapping parens around the recursive case if it has an 'order' clause.
+      # Comment.joins("JOIN comment_tree ON comment_tree.id = comments.parent_comment_id").order("depth desc", "confidence desc")
+      Arel.sql('SELECT "comments".id, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc')
+    ])
+      .from("comment_tree").joins("CROSS JOIN comments ON comment_tree.id = comments.id")
```
```suggestion
      .from("comment_tree as comments")
```

Can save a join. Might need `.select(“*”)`

chamlis:
Thanks for the suggestion! I've pushed that as an additional commit for the alternative. My gut feeling is that having a narrower CTE is good for performance and using the `CROSS JOIN` should make the final join efficient as it's done by rowid, but I can't pretend to know for certain.

The query plan for this alternative is:
```
QUERY PLAN
|--CO-ROUTINE comment_tree
|  |--SETUP
|  |  |--SEARCH comments USING INDEX comments_parent_comment_id_fk (parent_comment_id=?)
|  |  `--SEARCH stories USING INTEGER PRIMARY KEY (rowid=?)
|  `--RECURSIVE STEP
|     |--SCAN comment_tree
|     `--SEARCH comments USING INDEX index_comment_tree (parent_comment_id=?)
`--SCAN comments
```

pushcx:
That's a really good point. I'll have to perf test it.

I'm probably not going to touch this over the weekend now because I'm burning my day on db repair and could use a break.

pushcx:
I'm sorry I didn't get back to this for so long.

So the real frustrating learning that followed the database work I mentioned above was that query plans are inaccurate if the `sqlite_stat1` or `sqlite_stat4` tables are missing. `1` is just about the indexes and is harmless. `4` includes table contents in its `sample` column so cannot be distributed.

So on production data (and speaking of frustrating learnings, it's the same sqlite version, and compile flags, see 4227239d), here's the query plan for the original cross join version:

```
sqlite> explain query plan WITH RECURSIVE "comment_tree" AS ( SELECT comments.id, comments.depth + comments.confidence AS priority FROM "comments" INNER JOIN "stories" ON "stories"."id" = "comments"."story_id" WHERE ((stories.id = 90 or stories.merged_story_id = 90) and parent_comment_id is null) UNION ALL SELECT "comments".id, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc ) SELECT "comments".* FROM comment_tree CROSS JOIN comments ON comment_tree.id = comments.id;;
QUERY PLAN
|--CO-ROUTINE comment_tree
|  |--SETUP
|  |  |--MULTI-INDEX OR
|  |  |  |--INDEX 1
|  |  |  |  `--SEARCH stories USING INTEGER PRIMARY KEY (rowid=?)
|  |  |  `--INDEX 2
|  |  |     `--SEARCH stories USING COVERING INDEX index_stories_on_merged_story_id_and_hotness (merged_story_id=?)
|  |  `--SEARCH comments USING INDEX story_id_short_id (story_id=?)
|  `--RECURSIVE STEP
|     |--SCAN comment_tree
|     `--SEARCH comments USING INDEX comments_parent_comment_id_fk (parent_comment_id=?)
|--SCAN comment_tree
`--SEARCH comments USING INTEGER PRIMARY KEY (rowid=?)
```

Running this on my local dev took 0.030 - 0.040 seconds.

And here's my suggestion

```
sqlite> explain query plan WITH RECURSIVE "comment_tree" AS ( SELECT comments.*, comments.depth + comments.confidence AS priority FROM "comments" INNER JOIN "stories" ON "stories"."id" = "comments"."story_id" WHERE ((stories.id = 129963 or stories.merged_story_id = 129963) and parent_comment_id is null) UNION ALL SELECT "comments".*, "comments".depth + "comments".confidence AS priority FROM "comments" JOIN comment_tree ON comment_tree.id = comments.parent_comment_id ORDER BY priority desc, comments.id asc ) SELECT "comments".* FROM comment_tree as comments /* loading for pp */ LIMIT 11;
QUERY PLAN
|--CO-ROUTINE comment_tree
|  |--SETUP
|  |  |--MULTI-INDEX OR
|  |  |  |--INDEX 1
|  |  |  |  `--SEARCH stories USING INTEGER PRIMARY KEY (rowid=?)
|  |  |  `--INDEX 2
|  |  |     `--SEARCH stories USING COVERING INDEX index_stories_on_merged_story_id_and_hotness (merged_story_id=?)
|  |  `--SEARCH comments USING INDEX story_id_short_id (story_id=?)
|  `--RECURSIVE STEP
|     |--SCAN comment_tree
|     `--SEARCH comments USING INDEX comments_parent_comment_id_fk (parent_comment_id=?)
`--SCAN comments
```

That `SCAN comments` sure looks like a full table scan, but this ran on my local dev in 0.003.

On both I swapped in a story id that had 1 merge and 504 total comments.

PR #2156, opened 2026-07-17
fix username moderation log

### LINKED ISSUE #2151: username changes no longer seem to appear in the moderation log
A user named deactivated8885175 recently deactivated their account, and that was correctly reported in the moderation log. The lack of any corresponding username change message made me wonder whether these were still being logged.

I paged back to the beginning of 2026 in the log, and didn't see any users changing their own usernames in the moderation log. That seems unusual.

(I'm opening this bug at the request of td123 in #lobsters on libera.)

thomasdziedzic:
fixes #2151

Use the right dirty methods.

Tested locally.


thomasdziedzic:
this is not stale

pushcx:
Thank you for solving this, sorry it took a while to review.

PR #2157, opened 2026-07-17
Generate cacheable image URL for text stories

chamlis:
StoryImageController would send this logo anyway, but using the direct URL to it means it can be served without hitting Rails.


pushcx:
Great to offload, thank you.

PR #2168, opened 2026-07-21
Make 2fa secret easier to copy during enrollment

jamesa:
Hello! I ran into this while setting up Two Factor Authentication on my account. I found that it was very slightly annoying to copy/paste the 2fa secret code. When double-clicking on the secret code, instead of only selecting the code, I also selected "Reveal Secret":

Before - summary is selected along with the details (2fa secret)

I think it'd be nicer if only the code was selected so that I could paste it directly into my password manager, like this:

After - only 2fa secret is selected

I also thought the monospace font was more fitting, but let me know if you disagree.

By the way, this was only a test secret, don't worry. :)

### DIFF
--- a/app/views/settings/twofa_enroll.html.erb
+++ b/app/views/settings/twofa_enroll.html.erb
@@ -16,7 +16,9 @@ target="_blank">TOTP</a> application of choice:
   Or to add to a device manually enter the secret:
   <details>
     <summary>Reveal Secret</summary>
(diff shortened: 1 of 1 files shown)


pushcx:
That certainly looks annoying, and I agree monospace seems appropriate. Thank you for reporting and fixing this, sorry for my delay reviewing.

PR #2174, opened 2026-07-23
Fix numeric truncation in flagged_commenters calculation

### LINKED ISSUE #2166: SQLite migration quietly broke flagged_commenters calculation
SQLite handles division differently to MariaDB, so while dividing columns and post-multiplying by 100 worked in MariaDB, in SQLite, the division always comes out zero ( Or 1, if 100% flagged ). 

Compare running this query in https://codapi.org/mariadb/ and  https://codapi.org/sqlite/ :
```
    SELECT 1 / 5 * 100.0 as current_impl, 1 * 100.0 / 5 as fix;
```

Will demonstrate the difference:

```
Mariadb ( pre-migration ):
+--------------+----------+
| current_impl | fix      |
+--------------+----------+
|     20.00000 | 20.00000 |
+--------------+----------+

SQLite ( post-migration )
┌──────────────┬──────┐
│ current_impl │ fix  │
├──────────────┼──────┤
│ 0.0          │ 20.0 │
└──────────────┴──────┘
```

This affects a few pages where flagging matters: Users aren't getting warnings on their profile, the mod page won't be showing flagged commenters, /standing will almost always show zero, but most importantly, the self-deletion trigger that escalates it to make the account more difficult to reactivate.

richardcocks:
SQLite has different behaviour of integer division ( integer truncation ) than MariaDB had ( decimal results ).

With multiplication by 100 after the division, `percent_flagged` always became 0% ( for almost everything )  or 100% .

This PR fixes that behaviour by multiplying by 100.0 in the numerator, therefore implicitly casting it to float earlier.

Fixes #2166

### DIFF
--- a/app/models/flagged_commenters.rb
+++ b/app/models/flagged_commenters.rb
@@ -67,9 +67,9 @@ class FlaggedCommenters
           sum(flags) as n_flags,
           sum(flags)/count(distinct comments.id) as average_flags,
           (
-            count(distinct if(flags > 0, comments.id, null)) /
+            count(distinct if(flags > 0, comments.id, null)) * 100.0 /
             count(distinct comments.id)
-          ) * 100 as percent_flagged")
+          ) as percent_flagged")
         .having("n_comments > 4 and n_stories > 1 and n_flags >= 10 and percent_flagged > 10")
         .order(sigma: :desc)
         .limit(30)


richardcocks:
This has had no activity for 30 days, so I am commenting that I am still willing to work on this, but there hasn't been maintainer activity. If this continues for another 30 days I'll let it go.

pushcx:
Sorry for the delay. Thanks for catching and fixing this bug.

PR #2176, opened 2026-07-23
Ignore public/cache but keep .gitkeep to hold the folder

richardcocks:
The `public/cache/` folder is deliberately kept live by way of `.gitkeep` in there.

However, it fills up with cached output which needs to be manually avoided when doing a `git add`.

This PR ignores `public/cache` with an override to not ignore `public/cache/.gitkeep`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,10 @@ public/application.js
 public/avatars
 public/sitemap*
 public/story_image
+public/cache
+
+## Override .gitkeep
+!public/cache/.gitkeep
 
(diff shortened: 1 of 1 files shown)


richardcocks:
I've not created an issue for this, since it's not a product-facing issue but development experience. If you have a rule that all PRs need an associated issue, I would be happy to raise one.

pushcx:
No, we definitely don't need an issue for all PRs. Thanks for this polish.

PR #2178, opened 2026-07-24
fix dead attribute check in token.rb

### LINKED ISSUE #2164: Broken check in `token.rb` causing large overhead to all ActiveRecord instantiation
https://github.com/lobsters/lobsters/blob/d5d54ec92b32331f8d458d3b03c4949ff9aa4c5c/app/models/concerns/token.rb#L6

This line is bugged, attributes returns a hash with string keys, so `attributes.include?(:token)` is always evaluated to false, but with a great cost: building the entire attributes hash, even for columns not in the select.

This line is causing large amounts of overhead in ActiveRecord Instantiation, removes any benefits of lazy attribute materialisation, affecting all models with Tokens, which is most of the important ones, including `User`, `Story`, `Comment`, etc.

I believe the line ought to have been:

    self.token ||= TypeID.new(self.class.to_s.parameterize) if new_record? || has_attribute?(:token)

So as it was written, instead of checking if it has the attribute, it forces instantiation of every attribute on each record, regardless of any previous narrowing.

Worst affected is the wide `User` table.

 On my local container environment, without this fix hitting `/Users` first post-boot request takes 2170ms, or ~1300 after.

After this change, it takes ~270ms and ~250ms respectively, with the main instantiation dropping from ~1200ms to ~140ms.

JoeJoeflyn:
Fixes #2164.

`attributes.include?(:token)` is always false because `attributes` uses string keys, not symbols, so the check is dead code. Replace with `has_attribute?(:token)` as suggested in the issue.

`bundle exec rake build` passes (842 examples, 0 failures).

### DIFF
--- a/app/models/concerns/token.rb
+++ b/app/models/concerns/token.rb
@@ -3,7 +3,7 @@ module Token
 
   included do
     after_initialize do
-      self.token ||= TypeID.new(self.class.to_s.parameterize) if new_record? || attributes.include?(:token)
+      self.token ||= TypeID.new(self.class.to_s.parameterize) if new_record? || has_attribute?(:token)
     end
 
     validates :token,


pushcx:
Thanks for doing this.

That failing build is unreleated (and only on CI, irritating) so I'm going ahead to merge this.

JoeJoeflyn:
Thanks a lot! Really appreciate it. Looking forward to contributing more

PR #2170, opened 2026-07-22
fix frozen string warning in StoryImage

thomasdziedzic:
Running `bundle exec rspec` produces a warning:

```
/home/tom/work/lobsters/app/models/story_image.rb:62: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```

PR #2181, opened 2026-07-24
Copy the response body before force_encoding in story_image

richardcocks:
`force_encoding` mutates its receiver, so this was modifying a string owned by response.

This error was revealed by d5d54ec9, which removed the rescue that had been swallowing it since the spec was added.


richardcocks:
Closed, I realised #2170 covers this.

PR #2182, opened 2026-07-25
Performance: Compare invited_by_user_id instead of full user object to avoid loading extra data

### LINKED ISSUE #2161: N+1 query against User(s) in story listings such as /Active, caused by `invited_by_user`
Each story in a story list such as `/Active` has `recently_invited_by`:

`invited_by_user == user && created_at.after?(MENTORSHIP_DAYS.days.ago)`

`invited_by_user` is not preloaded by `Story.for_presentation`, so this triggers a lazy load, which is then called in the partial template for each story, so loading `/active` triggers 20+ queries, as can be seen in the trace attached below.

Each query is <1ms, but there is significant overhead to doing so many individual queries.

All these queries can be avoided by comparing foreign keys, i.e. `invited_by_user_id == user.id` instead, because `invited_by_user_id` is already preloaded as part of the story query which `includes(:user)` 

Avoiding fetching the full `invited_by_user` row reduced the template render from ~70ms to ~20ms on my local container instance ( configured with production configs )  .

Relevant part of trace:
Image

richardcocks:
Fixes an issue on story list pages where loading author relationships causes n+1 queries.

The same underlying issue was previously worked around in story.rb by preloading all the data upfront, but was limited to fixing the `hottest` scope.

This was also affecting pages with comments.

This PR fixes both by comparing `invited_by_user` using the id directly. This avoids loading the rest of the object. This PR also eliminates the preload since it is no longer needed.

Fixes #2161

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -85,8 +85,6 @@ class Story < ApplicationRecord
       .filter_tags(exclude_tags || [])
       .positive_ranked
       .order(:hotness)
-      # solves n+1 user query in home#index
-      .includes(user: [:invited_by_user])
   }
 
   scope :tagged, ->(user, tags) {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -658,6 +658,6 @@ class User < ApplicationRecord
   end
 
   def recently_invited_by?(user)
-    invited_by_user == user && created_at.after?(MENTORSHIP_DAYS.days.ago)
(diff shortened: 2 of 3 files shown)


richardcocks:
This has had no activity for 30 days, so I am commenting that I am still willing to work on this, but there hasn't been maintainer activity. If this continues for another 30 days I'll let it go.

pushcx:
Great, thank you.

Issue #1585, opened 2025-05-23
Refactor away Keystore for Rails.cache


pushcx:
The [Keystore model](https://github.com/lobsters/lobsters/blob/1e7be9296adba828ba54ba6dbc05ef74b319dd21/app/models/keystore.rb) predates [Rails.cache](https://guides.rubyonrails.org/caching_with_rails.html) and [Solid Cache](https://guides.rubyonrails.org/caching_with_rails.html#solid-cache). It's most commonly used as a read-through cache for aggregations. Some of these, like the count of a user's deleted stories, are easy to express in ActiveRecord queries, not expensive to calculate, and are buggy (#1084).

The most complicated use is `comments.thread_id`, which is effectively an autoincrement integer and exclusively used in `Comment.recent_threads`. (`User.recent_threads` looks like dead code.) With AR having good support for RCTEs, this probably can be deleted. If not, let's replace with a TypeID.

#1548 is a blocker because several cron jobs depend on Keystore's durability, which `Rails.cache` doesn't guarantee.

PR #2183, opened 2026-07-25
Performance: Get traffic:intensity direct from Keystore without Rails.cache overhead

richardcocks:
Despite a stated goal in #1585 to migrate away from Keystore, the current traffic_helper cached value was still keystore but wrapped in a `rails.cache` that is on a short TTL, shorter than the job that populates the traffic:intensity value.

Without this PR:

On a cache hit, it is doing a SELECT to get the value from SolidCache.

On a cache miss, which is frequent, it is doing a SELECT to get the value, finding the cache expired, then getting the value from Keystore, then doing an UPSERT to write the value back to SolidCache.

Both of these operations are more expensive than reading the value from Keystore directly, so much so that on my local environment I ended up moving SolidCache to `tmpfs` to mitigate the performance impact.

This PR restores getting the value from Keystore directly. This should help toward #1814

### DIFF
--- a/app/helpers/traffic_helper.rb
+++ b/app/helpers/traffic_helper.rb
@@ -56,7 +56,7 @@ module TrafficHelper
   end
 
   def self.cached_current_intensity
-    Rails.cache.fetch("traffic:intensity", expires_in: 60) { Keystore.value_for("traffic:intensity") || 0.5 }
+    Keystore.value_for("traffic:intensity") || 0.5
   end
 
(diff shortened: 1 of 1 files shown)


richardcocks:
The merge base was from before the CI frozen-string fix, so I've rebased to latest main now to resolve that.

pushcx:
Yeah, this one's on every dynamic pageload. Thank you.

PR #2191, opened 2026-07-27
Allow development on remote access via http://hostname

supermarin:
We currently allow access from anywhere in development via 0.0.0.0, but Rails has built-in Host header security filtering that blocks connections like http://example:3000. This is useful if you're developing on a remote box with Tailscale or a similar approach.


supermarin:
An alternative to this would be to do something like `config.hosts << Socket.gethostname`.

pushcx:
Thank you, this let me delete my janky local workaround.

PR #2195, opened 2026-08-01
search: disable autocorrect in text field

mistydemeo:
When visiting lobsters on mobile, a papercut I run into frequently is that the phone's autocorrect ends up trying to "correct" domain names in `domain:` searches. I noticed that the search field disables autocapitalization, but not autocorrect. I feel like it'd be useful to disable autocorrect as well so this doesn't happen. I also disabled spellcheck, but can back that part out if you prefer.

### DIFF
--- a/app/views/login/forgot_password.html.erb
+++ b/app/views/login/forgot_password.html.erb
@@ -6,7 +6,7 @@
 <%= form_with url: reset_password_path do |f| %>
   <div class="labelled_grid">
     <%= f.label :email, "E-mail or Username" %>
-    <%= f.text_field :email, size: 40, autofocus: :autofocus, inputmode: :email, autocapitalize: :off %>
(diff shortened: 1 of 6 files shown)


pushcx:
This is a great idea, we should certainly expect a lot of searches for technical terms that won’t be in dictionaries. But yes, I think let’s leave spellcheck on to help out typos. 

Can you check our various email and username fields while you’re at this? (I would but am on my phone.)

mistydemeo:
> But yes, I think let’s leave spellcheck on to help out typos.

Backed that out.

> Can you check our various email and username fields while you’re at this? (I would but am on my phone.)

Done!

pushcx:
Sorry for my delay getting back to this, and thanks again.

mistydemeo:
No worries, thanks for merging!

PR #2197, opened 2026-08-05
Add missing titles for some pages

### LINKED ISSUE #2118: Site sections don't have specific <title> pages
If I shared a link to the site to a screen reader user, I would like them to be able to understand where they landed. This is also important so they could understand if activating a link took them to another subsection and not just the home page again.

So, the following section lack a useful title.

* Inbox: https://lobste.rs/inbox/all
* Unread: https://lobste.rs/inbox/unread
* ModMail: https://lobste.rs/mod_mails
* Public Profile Information: https://lobste.rs/settings

fedemp:
Fixes #2118

### DIFF
--- a/app/controllers/inbox_controller.rb
+++ b/app/controllers/inbox_controller.rb
@@ -14,6 +14,7 @@ class InboxController < ApplicationController
   end
 
   def all
+    @title = "Inbox"
     notifications_per_page = 25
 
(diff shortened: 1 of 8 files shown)


pushcx:
Thank you for finding and fixing these.

Sorry for my delay getting this merged.

PR #2200, opened 2026-08-08
Prefer await over promises. Prefer outerHTML over replace()

fedemp:
In order to simplify the logic flow, many chained promises were replaced with `await`.

Remove the `replace()` function and instead use [`outerHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML).


fedemp:
Working on something else so this is not important right now.

PR #2202, opened 2026-08-13
Use a string for the option to not use a hat

fedemp:


pushcx:
That change to the `label` seems like a small improvement, is that worth making?

PR #2199, opened 2026-08-07
Address HTML markup issues reported by Herb

marcoroth:
This pull request addresses (most of) the findings `bundle exec herb analyze` reported on Herb v0.10.3. 

There are still some things in  `app/views/comments/_threads.html.erb` and `app/views/users/tree.html.erb` that Herb doesn't fully understand yet which I left out for now.

I recommend reviewing the changes using [`?w=1`](https://github.com/lobsters/lobsters/pull/2199/changes?w=1).

If you are interested, I can do a follow up pull request to setup running `bundle exec herb analyze` in CI using the Ruby gem.

Follow up on #1633 and #1754

PR #2204, opened 2026-08-14
fix story #is_gone? condition

MatheusRich:
Story#is_gone? was using a local variable assignment instead of a comparison operator in the condition.

That made the whole right side of the or operator always evaluate to false. Unit tests were added to cover the different scenarios.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -774,7 +774,7 @@ class Story < ApplicationRecord
   end
 
   def is_gone?
-    is_deleted? || (user.is_banned? && score < 0 && is_moderated = false)
+    is_deleted? || (user.is_banned? && score < 0 && is_moderated == false)
   end
 
   def is_hidden_by_user?(user)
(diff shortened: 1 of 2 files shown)


pushcx:
Thanks for catching and fixing this bug. That explains some reports!

MatheusRich:
@pushcx I'm surprised this bug slipped in and no linter caught it.

pushcx:
`.erb` files are a bag of strings. It's trivial to output invalid markup, and it's far enough from Ruby code that existing linters didn't work. @marcoroth is developing [Herb Linter](https://herb-tools.dev/projects/linter) that can perhaps catch this and other issues. He's even been contributing fixes from the tool; mostly recently #2199 (and that links back to older ones).

Ultimately, this is one more in the infinite pile of examples of why generating html from string interpolation is a dead end.

MatheusRich:
I might be missing something, but isn't this one pure Ruby?

PR #2205, opened 2026-08-14
Use `exists?` instead of `find_by` for boolean checks

MatheusRich:
`exists?` is more efficient than `find_by` when checking for the existence of a record, as it does not instantiate an ActiveRecord object.


pushcx:
Makes sense, thanks!

PR #2210, opened 2026-08-24
Fix upvote arrow not showing as orange in /inbox/all (fixes #2128)

### LINKED ISSUE #2128: `lobste.rs/inbox/all` does not indicate that we've upvoted comments
The arrow is not orange

oyamairishitomi:
The vote arrow for already-voted comments in the comment list did not appear orange when /inbox/all was loaded. The issue was that CommentVoteHydrator was not being enumerated, even though it implements Enumerable. Added .to_a to CommentVoteHydrator.new(comment_notifications, @user) in the apply_current_vote method.

Fixes #2128

image


pushcx:
Thanks for taking this on.

The way this works by side effect is pretty subtle, could you remove the misleading `comment_notifications =` and add a comment to explain why it works?

PR #2211, opened 2026-08-24
Move Hatchbox build/deploy scripts

excid3:
Hatchbox now automatically detects scripts in .hatchbox so we don't have to configure them in Hatchbox.io anymore.

https://hatchbox.io/docs/articles/18-build-scripts-and


pushcx:
Neat! Very thoughtful of you to not just add this feature, but PR it into Lobsters, thank you.

Do they all run as `deploy`?

excid3:
Glad to contribute back something finally!

Yes, they work exactly the same. Just now can be nicely version controlled.

PR #2218, opened 2026-09-04
Fix force_open being applied too broadly

chamlis:
force_open used to be false in this location, by virtue of the default being false and a typo meaning this line originally read `force_option = true`. Having it be true here means that every comment is force-opened, even those the user has flagged or that have a sufficiently low score.

Fixes: 7a80cc1 (whose commit message was jadedly prophetic!)


pushcx:
Argh, yep, thanks for spotting and fixing.

Issue #1287, opened 2024-06-10
Manifest.json for web app installation


lucianmarin:
Can you a manifest.json to be able to install it on Android/iOS?

```html
<link rel="manifest" type="application/json" href="/static/manifest.json">
```

Below is the code I use on [Subreply](https://subreply.com/) and it works great. Maskable icon is the same as the favicon one, but zoomed out by 1.333x. Everything else is pretty obvious and straightforward.

```json
{
  "lang": "en",
  "name": "Subreply",
  "short_name": "Subreply",
  "description": "Tiny, but mighty social network.",
  "icons": [
    {
      "src": "/static/192.png",
      "sizes": "192x192",
      "type": "image/png",
      "purpose": "any"
    },
    {
      "src": "/static/256.png",
      "sizes": "256x256",
      "type": "image/png",
      "purpose": "maskable"
    }
  ],
  "id": "/feed",
  "start_url": "/feed",
  "display": "standalone",
  "orientation": "portrait",
  "status": "ok"
}
```

pushcx:
If it's that easy, sounds like a nice convenience to add. Would you like to open a PR?

Anything to be aware of for updates? I remember seeing a headline or two go by about bookmarking sites but I don't remember particulars.

lucianmarin:
I did, but I have no idea how to link PR to this issue.

pushcx:
Mention the idea of one from the other in the text of comment or commit message with the `#`, like:

#1288

kaycebasques:
followup from https://lobste.rs/s/k2tyg5/lobsters_pwa#c_zkksaw

![Screenshot_20240718-080647](https://github.com/user-attachments/assets/3171e484-c720-4d72-a081-c207e7a02aca)

pushcx:
I removed the `manifest.json` in e121f808, users [reported](https://lobste.rs/s/k2tyg5) intrusive popups on Android; more info in the site story.

mrexodia:
Would you be open to reconsider adding back PWA support? It makes the iOS experience much better. I’d be happy to put in some work also checking if Android still has popups. From what I remember Google kinda steered away from PWAs, so it might have improved since.

pushcx:
Yeah, I am still happy to offer the convenience if it's not going to cause a confusing UI or a maintenance burden.

pushcx:
This was reverted and closed in 2024 ago when we found Android had a very strange UI, but it's worth taking another look at.

[...]

PR #2228, opened 2026-09-18
add PWA support, fix #1287

mrexodia:
- Add back `/manifest.json` based on d86379c4
- Set purpose `any` for the `192x192` icon (see [DoesManifestContainRequiredIcon](https://source.chromium.org/chromium/chromium/src/+/main:components/webapps/browser/installable/installable_evaluator.cc;l=119;drc=3a77577699b19ed003bb500fc232b73f9fc9afaa) in the Chromium source code)
- A 512x512 icon is required according to [MDN](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#required_manifest_members), generated using `magick -background none -density 3072 app/assets/images/logo-color.svg -resize 512x512 -strip PNG24:public/touch-icon-512.png`
- Add `theme_color` so the Android top bar uses the Lobsters color (otherwise it uses white which makes the time/battery etc unreadable)
- Update CSP for manifest to fix Chrome error `Loading a manifest from 'https://<redacted>.trycloudflare.com/manifest.json' violates the following Content Security Policy directive: "default-src 'none'". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback. The action has been blocked.`
- Call `preventDefault()` on `beforeinstallprompt` as documented on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event) to avoid the automatic installation popup
  - Technically non-standard, but well-documented and harmless on browsers that do not support it
  - This will print `Banner not shown: beforeinstallpromptevent.preventDefault() called. The page must call beforeinstallpromptevent.prompt() to show the banner.` on the console and I confirmed on my Xiaomi phone with Chrome that no popup is shown
- Tested the PWA installs and works on iOS


mrexodia:
Also tested with Javascript disabled. Unfortunately this does show the popup 😔 

image


However, after dismissing it I cannot get it to show again unless I clear all browsing data.

